PR #1, opened 2015-09-02
Fix issue with author changing type

megamit:
When cycling through the attributes of the data json 'author' within 'message' which was previously a user object is overwritten with a dictionary. This causes an AttributeError to be thrown ( and silently swallowed... thanks except: pass) whenever any of its attributes are referenced in the form message.author.x. User data should change in this step and the user object should not be modified for any reason so its safe to skip updating it.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -244,7 +244,7 @@ class Client(object):
                 message = copy.deepcopy(older_message)
                 # update the new update
                 for attr in data:
-                    if attr == 'channel_id':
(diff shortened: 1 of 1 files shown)


Rapptz:
I had to rebase it since I was working on some stuff.

Rapptz:
Somewhat unrelated to this PR but I added an `on_error` to propagate the uncaught exceptions if you wanted to check them out since you mentioned `except: pass`. They should no longer be silently swallowed (though if you throw from `on_error` you're on your own).

Issue #5, opened 2015-10-04
Hitting issues when trying to create a bot


johncoxon:
I'm trying to write a bot using this API which will let people play a game but I'm coming across some weird behaviours and I can't tell whether I'm doing something wrong or whether there's actually a problem.

The below code, upon `talk_to_discord()` being called, attempts to start a client and then listens for a message to tell it when to start the game (I'm going with !start). When it hears that message, it starts a Game object, passing both the Discord client (`self.client`) and the user who started the game (`self.owner`). The idea is that the Game object then messages the owner to ask who's playing (generating a list of usernames from the input string) and also what roles are available to players (generating a list of roles from the second input string).

The module is called `werewolf`. If I open Python and then type `import werewolf` and `werewolf.talk_to_discord()`, the client starts and I can message the relevant user on Discord. 

The chat log in Discord between me and the user (also called 'werewolf') looks like this:

**me**: !start 
**werewolf**: Let's start a game!
**werewolf**: Who will be playing?
**me**: Player 1, Player 2, Player 3, Player 4, Player 5
_pause_
**me**: !start
**werewolf**: Let's start a game!
**werewolf**: Who will be playing?
**me**: Player 1, Player 2, Player 3, Player 4, Player 5
**werewolf**: What roles will be in effect?
**me**: Tanner, Werewolf, Werewolf, Hunter, Robber, Seer, Insomniac, Villager

The console output before _pause_ is:
```
Successfully logged in as werewolf with ID 100233992065814528.
A game is starting!
users input
[]
```

The output after _pause_ is:
```
A game is starting!
users input
[u'!start', u'Player 1, Player 2, Player 3, Player 4, Player 5']
roles input
[u'!start', u'Player 1, Player 2, Player 3, Player 4, Player 5']
test
```

The code I've written is:
```
import discord

def talk_to_discord():
	
	client = discord.Client()
	client.login('email', 'password')

	@client.event
	def on_message(message):
		if message.content.startswith("!start"):
			client.send_message(message.channel, "Let's start a game!")
			Game(client, message.author)

	@client.event
	def on_ready():
		print('Successfully logged in as {0} with ID {1}.'.format(client.user.name, client.user.id))
		
	client.run()

# --------------------------------------------------------------------------------------------------

class Game(object):
	"""Object for the entire game."""

	def __init__(self, client, owner):
		import random
		print('A game is starting!')
		
		self.client = client
		self.owner = owner

		# Work out who's playing.
[...]

PR #8, opened 2015-10-06
Exit Gracefully on Keyboard Interrupt.

megamit:
Inspired by the 3rd issue [here](https://github.com/Rapptz/discord.py/issues/5).
Thoughts?

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -220,6 +220,9 @@ class Client(object):
         try:
             log.info('attempting to invoke event {}'.format(event_name))
             self.events[event_name](*args, **kwargs)
+        except KeyboardInterrupt:
+            log.info("Keyboard Interupt, Stopping...")
+            self.logout()
         except Exception as e:
             log.error('an error ({}) occurred in event {} so on_error is invoked instead'.format(type(e).__name__, event_name))
             self.events['on_error'](event_name, *sys.exc_info())
@@ -373,7 +376,11 @@ class Client(object):
     def run(self):
         """Runs the client and allows it to receive messages and events."""
         log.info('Client is being run')
-        self.ws.run_forever()
+        try:
+            self.ws.run_forever()
+        except KeyboardInterrupt:
+            log.info("Keyboard Interupt, Stopping...")
+            self.logout()
 
     @property
     def is_logged_in(self):


Hornwitser:
The semantics of the Client class with regard to ``run()`` and it's call to ``ws.run_forever()`` is highly counter-intuitive, and your suggested patch does not make it in any way better.

Firsts things first.  Never *ever* swallow a KeyboardInterrupt unless this is what the user expects.  With your patch, doing CTRL+C will cause python to continue off after client.run() as if nothing happened, instead of continuing to unwind the stack with a KeyboardInterrupt exception.

Getting to the real issue here, ``client.run()`` doesn't do what you'd except it to.  It blocks the thread that calls it until the WebSocket dies.  Events are processed in it's own thread after ``client.login()`` regardless of whether ``client.run()`` has been called or not.  This creates a lot of problems with synchronization and race conditions.  Just to name a few:

- All of the examples contain a race condition, they call ``bot.login(...)`` before attaching the event handlers.  This in turn means that for example the on_ready event could be fired before the event handler is attached to it.
- Exceptions thrown by on_error or Client itself causes the WebSocket to close and a traceback to be printed to stderr. Then it's ignored, and the thread that called ``client.run()`` will continue as if nothing happened.
- Calling ``client.logout()`` does not imminently close the connection, only initiate a closing handshake.  The significance of this is that event handlers may be called *after* a call to ``client.logout()``.
- Calling the builtin ``exit()``  function from an event handler is ill-defined.  It will close sys.stdin, and then raise a SystemExit exception.  Since exceptions in the WebSocketClient thread is ultimately ignored, a script doing this would continue on from the ``run()`` call as if nothing happened without being able to read from standard input.
- Ctrl+C will most likely raise a KeyboardInterrupt in the thread that does ``client.run()``, but not always.  Trying to properly handle this non-deterministic behavior is at best a kludge.

If you take a look at my fork you will see that I avoided more or less all of these issues by removing the dedicated WebSocketClient thread.  This also has the additional benefit of the run method being guaranteed to have closed the WebSocket connection when it returns.

megamit:
I understand your point. I think it would be best if i just closed this fork. Your fork makes significant changes to client.py (nice ones, your event overhall is interesting), so i will wait patiently for your fork to be merged into the base branch.

PR #19, opened 2015-10-16
Add events for sniffing WebSocket data

Hornwitser:
These events were needed in order to realize my Discord Logger bot.  Although quite a special case, they might be useful for debugging.

I've tacked on a small improvement in the json encoding I spotted while coding this as well.


Rapptz:
Rebased~

Issue #31, opened 2015-10-31
member_join and member_remove lack server parameter


Boberro:
Events on new user joining and user leaving the server lack information on server parameter. As a result, it's hard to have meaningful use of those events.
See: handle_guild_member_add and handle_guild_member_remove

Boberro:
Fixed in pull request https://github.com/Rapptz/discord.py/pull/32

PR #32, opened 2015-10-31
fixed: on_member_join and on_member_remove lack useful data

Boberro:
Fix for issue https://github.com/Rapptz/discord.py/issues/31


Rapptz:
Er.. there's a `Member.server` property.

Boberro:
RIght... My bad - never realised member is not of User class.

PR #35, opened 2015-11-10
Audio completely working

izy521:
Obviously, I wouldn't recommend accepting the PR, but instead pulling what you need out of it. I didn't develop a method for sending audio files or anything, I just put that all in the join_voice_channel one for a quick test.

Zeta mentioned the opus library (opuslib), I used, doesn't support Python 3. So I dunno, good luck!

It was a fun experience~

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -42,6 +42,7 @@ import json, re, time, copy
 from collections import deque
 import threading
 from ws4py.client import WebSocketBaseClient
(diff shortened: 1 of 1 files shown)


Rapptz:
Implemented in https://github.com/Rapptz/discord.py/commit/a6d6d832ff7f6c49d33f4b6a98a80308b108fb58 (asyncio only)

PR #38, opened 2015-11-15
Fixed update_voice_state to handle changing between channels

tuckismad:


Rapptz:
Please make the commit from the develop branch rather than master.

PR #40, opened 2015-11-16
Add before and after params to logs_from.

khazhyk:
Currently if the the id is garbage the api acts as if it was not specified, so it's up to the user to provide a proper id.

Messages are still always returned newest first.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -817,10 +817,12 @@ class Client(object):
         self._is_logged_in = False
         log.debug(request_logging_format.format(response=response))
 
-    def logs_from(self, channel, limit=100):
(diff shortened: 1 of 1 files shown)


Rapptz:
While I appreciate the patch, the design of the library operates on data classes and not strings. So `discord.Message` would be a proper parameter. If someone wants a string ID they can use `discord.Object(id='some-string')`.

khazhyk:
boop

Rapptz:
Rebased into https://github.com/Rapptz/discord.py/commit/fcc540ec44717c700d8acf7bc5e99f934e08bb15

PR #43, opened 2015-11-17
fixed voice_members not updating when changing channels

tuckismad:


Rapptz:
Rebased into develop.

PR #44, opened 2015-11-19
Add send_typing command to allow sending typing updates.

khazhyk:
Apparently you can send typing updates even if you have no permission to write to the channel. Go figure. It will, however, raise 403 if you do not have the `read` permission.


Rapptz:
Rebased into develop.

PR #53, opened 2015-12-07
change server members to a dict

berler:
Considering that we frequently find members by their ID, it makes sense
to store this as a dict where the keys are the IDs so we can look them
up (and add and remove them) in O(1) time instead of O(N) time.

Keeps backwards compatibility so you can still do server.members and
get a list (it just stores it as a dict internally).

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -204,7 +204,7 @@ class ConnectionState(object):
             status = data.get('status')
             user = data['user']
             member_id = user['id']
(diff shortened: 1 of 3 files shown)


Rapptz:
This doesn't work in Python 3.

Python 3 dict views are not accessible by index. e.g. `server.members[0]` will not work.

berler:
I've updated it to use an OrderedDict to preserve order and wrapped the property in list() so that you can access it by index.  Does this work?

PR #57, opened 2015-12-12
avoid potential bug when creating new private msgs

berler:
It probably isn't good to rely on an item that was added to a list to
still be the last item, especially if we could have other async
coroutines modify the list. This may not be an actual issue, but having
the function explicitly return the object that it just added to the list
should guarantee that we don't accidentally pull the wrong item from the
end of the list later.


Rapptz:
Rebased.

PR #65, opened 2015-12-23
Add game data class, replace game_id.

khazhyk:
it works :+1:

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -27,6 +27,7 @@ DEALINGS IN THE SOFTWARE.
 from . import __version__ as library_version
 from . import endpoints
 from .user import User
+from .game import Game
 from .channel import Channel, PrivateChannel
 from .server import Server
 from .message import Message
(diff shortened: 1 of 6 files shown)


discord/member.py
```diff
@@ -75,7 +76,10 @@ def __init__(self, deaf, joined_at, user, roles, mute, **kwargs):
         self.joined_at = parse_time(joined_at)
         self.roles = roles
         self.status = Status.offline
-        self.game_id = kwargs.get('game_id', None)
+        game = kwargs.get('game', None)
+        if game is not None:
+            game = Game(**game)
+        self.game = game
```
SamSchlegel: this can just be `self.game = game and Game(**game)`

discord/client.py
```diff
@@ -1427,27 +1426,27 @@ def change_status(self, game_id=None, idle=False):
...
 
+        if game is not None and not isinstance(game, Game):
+            raise InvalidArgument('game must be of Game or None')
+
         idle_since = None if idle == False else int(time.time() * 1000)
-        try:
-            game_id = None if game_id is None else int(game_id)
-        except:
-            raise InvalidArgument('game_id must be convertible to an integer or None')
+        game = None if game is None else {'name': game.name}
```
SamSchlegel: Can just be `game = game and {'name': game.name}`

discord/member.py
jhgg: .get() returns None by default, no need to add it explicitly.

khazhyk:
Note: change_status will not send a respective presence update for yourself. As is, your own member object will not be updated with your game status. Not sure if this was the old behavior or not.

ericburns:
:+1:

PR #67, opened 2015-12-24
Merge develop into master

SamSchlegel:
Would be best to have master match the version that's on pypi. Since github PRs don't fast-forward this should probably be done by hand.


Rapptz:
I'm not going to do this since develop is the new legacy branch. Instead async should be the one that's going to get the fast forward merge while develop will be split from master.

SamSchlegel:
Sounds good to me +1

PR #97, opened 2016-01-31
Removed broken argument from send_message

TheConfuZzledDude:
With that "*" argument in, you become unable to send tts through the bot

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -895,7 +895,7 @@ class Client:
         return resp
 
     @asyncio.coroutine
-    def send_message(self, destination, content, *, tts=False):
+    def send_message(self, destination, content, tts=False):
         """|coro|
 
         Sends a message to the destination given with the content given.


Rapptz:
That * is meant to be there. It's a keyword only argument. Python 3+ only. I suggest you stop using python 2.7.

TheConfuZzledDude:
I am using Python 3.4.3. I just confirmed it through sys.version_info, and it still gives me crap about "giving 4 positional arguments when 3 are required"

Rapptz:
http://stackoverflow.com/q/1419046/1381108

TheConfuZzledDude:
ahh ok got it. not a very well documented feature :/ sorry bout that

Rapptz:
It is documented: http://discordpy.readthedocs.org/en/latest/migrating.html#forced-keyword-arguments

TheConfuZzledDude:
I meant in python's own documentation. Yours is rather good

PR #108, opened 2016-02-20
pip3 instead of pip

Yackback:
pip gives an error as Python 3.4.2 is needed. pip3 does not.


khazhyk:
This is dependent on your own personal platform. pip3 may not exist on certain users's systems (it does not exist on my system for example).

Rapptz:
I don't think I'll merge this. Sorry.

PR #110, opened 2016-02-25
Clean up zombies

khazhyk:
self.process.communicate(timeout=0.100) will block

zombies probably would be cleaned up anyways but in a non deterministic fashion by the garage collector


hg:
i too collect garages.

Rapptz:
Rebased into ef57873f80b68c32aea3cca635d05f9ff2aff29f

PR #122, opened 2016-03-05
Add util method to extract creation date from discord ids

khazhyk:
Some people pointed out this might be useful to have in libs.


Rapptz:
Rebased.

PR #124, opened 2016-03-07
Support extension loading of already imported modules

tekulvw:
Not everything is going to have an easily found relative path from the main script. This allows for the usage of `imp.find_module` and `imp.load_module` to grab modules from any path.


Rapptz:
I don't like this patch.

`import_module` already takes into consideration already imported modules. If you have a module instance you can just pass in `mod.__name__` to get its name.

PR #132, opened 2016-03-14
server.icon_url uses api endpoint instead of cdn directly.

khazhyk:
For consistency with user.avatar_url, and since using api is "right" and cdn is "wrong" i would assume


Rapptz:
Wrote it myself here: 0ad1cb95a3773cd80039e5ca846c1b19ac0e8b1a

PR #134, opened 2016-03-15
Changed some things to make state.py more stable to not crash Websocket when you Join a server via invite.

AraHaan:
Seems from Danny's last commit if you join a invite that it could crash the websocket I have seem to fixed that due to state.py.

I also added a function to client.py too that you feed your tokens in too for the new bot accounts in accordance to the recent API changes. Got a bot that is like that call

    client.run(None, None, token)

For older bots:

    client.run(email, password, None)

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -481,7 +481,7 @@ class Client:
         Raises
         ------
         LoginFailure
-            The wrong credentials are passed. Also is raised when bot is set to True and Token is set to None.
+            The wrong credentials are passed.
         HTTPException
             An unknown HTTP related error occurred,
             usually when it isn't 200 or the known incorrect credentials
@@ -513,7 +513,6 @@ class Client:
 
         body = yield from resp.json()
         self.token = body['token']
-
         self.headers['authorization'] = self.token
         self._is_logged_in.set()
 
--- a/discord/client.py
+++ b/discord/client.py
@@ -176,7 +176,7 @@ class Client:
             # redo the cache
         except OSError:
             log.info('a problem occurred while opening login cache')
-            pass # file not found et al
+            pass # file not found at all.
 
     def _update_cache(self, email, password):
         try:
@@ -442,6 +442,30 @@ class Client:
     # login state management
 
     @asyncio.coroutine
+    def Bot_Account_Token_Login(self, token):
+        self.token = token
+        if self.token is None:
+            raise LoginFailure('No Token Specified.')
+            yield from self.close()
+        else:
+            self.headers['authorization'] = token
+            self._is_logged_in.set()
+            try:
+                self.loop.run_until_complete(self.connect())
+            except KeyboardInterrupt:
+                self.loop.run_until_complete(self.logout())
+                pending = asyncio.Task.all_tasks()
+                gathered = asyncio.gather(*pending)
+                try:
+                    gathered.cancel()
+                    self.loop.run_forever()
+                    gathered.exception()
+                except:
+                    pass
+            finally:
+                self.loop.close()
+
+    @asyncio.coroutine
     def login(self, email, password):
         """|coro|
 
@@ -457,13 +481,12 @@ class Client:
         Raises
         ------
         LoginFailure
-            The wrong credentials are passed.
+            The wrong credentials are passed. Also is raised when bot is set to True and Token is set to None.
         HTTPException
             An unknown HTTP related error occurred,
             usually when it isn't 200 or the known incorrect credentials
             passing status code.
         """
-
         # attempt to read the token from cache
         if self.cache_auth:
             yield from self._login_via_cache(email, password)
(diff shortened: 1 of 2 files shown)


discord/state.py
```diff
@@ -281,7 +281,10 @@ def parse_guild_member_remove(self, data):
     def parse_guild_member_update(self, data):
         server = self._get_server(data.get('guild_id'))
         user_id = data['user']['id']
-        member = server.get_member(user_id)
+        try:
+            member = server.get_member(user_id)
+        except AttributeError:
```
berler: `server.get_member` does not ever throw any sort of error. It will return `None` if the user_id is not found.

AraHaan: I throse it for me when I use client.accept_invite with the url or code someone messages it when it is not supposed to crashign the bot. When I did this it works fine.

berler: That is literally impossible. If you look at the code for `server.get_member`, it is **impossible** for it to throw an error (unless you did something crazy and overwrote the server._members to not be a dict).  You may have been using a much older version of the library, or your error may have been due to another reason.  My best guess is that whatever you changed happened to fix it for other reasons.

discord/client.py
```diff
@@ -442,6 +442,30 @@ def _reconnect_ws(self):
...
+    def login_with_token(self, token):
+        self.token = token
+        if self.token is None:
+            raise LoginFailure('No Token Specified.')
+            yield from self.close()
+        else:
+            self.headers['authorization'] = token
+            self._is_logged_in.set()
+            try:
+                self.loop.run_until_complete(self.connect())
```
berler: It does not make much sense for this to call connect; it looks like you are duplicating the functionality of `run` and `start`.  IMO it would make a lot more sense to keep `login` as only doing a "log in" while use `run` / `start` to handle calling `connect` and `KeyboardInterrupts` etc.

discord/state.py
AraHaan: If you like here is the code I use to accept invites.
    if(message.content.startswith('::join')):
        if message.author.id in somedict['Users']:
            await client.send_message(message.channel, message.author.mention + " Due to Continuous abuse you have been Bot Banned.")
        else:
            code = message.content[len("::join "):].strip()
            url = code
            if code.rfind('https://discord.gg/') == -1:
                try:
                    invite = 'https://discord.gg/' + code
                    await client.accept_invite(invite)
                    await client.send_message(message.channel, 'Successfully Joined the server.')
                except discord.errors.NotFound:
                    await client.send_message(message.channel, 'Sorry, You did not specify a code or a Valid URL to Join a Server.')
            else:
                if url.rfind('https://discord.gg/') != -1:
                    try:
                        code = url[len("https://discord.gg/"):].strip()
                        invite = 'https://discord.gg/' + code
                        await client.accept_invite(invite)
                        await client.send_message(message.channel, 'Successfully Joined the server.')
                    except discord.errors.NotFound:
                        await client.send_message(message.channel, 'Sorry, You did not specify a code or a Valid URL to Join a Server.')

discord/client.py
```diff
@@ -442,6 +442,30 @@ def _reconnect_ws(self):
...
 
     @asyncio.coroutine
+    def login_with_token(self, token):
+        self.token = token
+        if self.token is None:
+            raise LoginFailure('No Token Specified.')
+            yield from self.close()
+        else:
+            self.headers['authorization'] = token
+            self._is_logged_in.set()
```
berler: Unlike the other `login` method, this does not appear to do any sort of check that the token is actually valid. This seems like a mistake.

discord/state.py
berler: So looking at this more, the correct thing to check would be to check `if server is None`.  It is impossible for any error to be thrown from the `get_member` method, which means the only other possibility is that the `server` variable was `None`.

AraHaan: Well I did PM the invite code so that is why I added the try on it so it dont crash if someone PM's the invite. It is odd though that it happens on PM's.

discord/client.py
```diff
@@ -573,7 +573,7 @@ def start(self, email, password):
         yield from self.login(email, password)
         yield from self.connect()
 
-    def run(self, email, password):
+    def run(self, email, password, token):
```
berler: It seems a bit ugly to have token as an additional arg when it actually should replace the email/password. Either it should probably use a different method, or you should abstract the credentials to a single variable which could either contain an email/password or a token.

AraHaan: I added token for those who did not update their account yet as the "Bot" account type does not have a UI and if they fuck up they would have to make a new account.

discord/client.py
AraHaan: the original ``run`` / ``start`` does not work because they require a email and a password, The New type of account does not HAVE a email and a password just a token.

AraHaan:
Looks like Everything is now working looks like All I need is a way to determine if the Token is Valid or not.

hlaaftana:
+1

hlaaftana:
+1 again

hlaaftana:
+1 i just really like this commit

hlaaftana:
+1

PR #135, opened 2016-03-17
Allow arguments before input in create_ffmpeg_player

Funky7Monkey:
Allows for arguments to be sent to ffmpeg in front of the input argument, using optBefore in the same way as options.

### DIFF
--- a/discord/voice_client.py
+++ b/discord/voice_client.py
@@ -341,7 +341,7 @@ class VoiceClient:
         struct.pack_into('>I', buff, 8, self.ssrc)
         return buff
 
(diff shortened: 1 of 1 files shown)


Funky7Monkey:
The point of my PR: ffmpeg options `-ss`, `-t` and `-stream_loop` would have use when placed before `-i`. While `-ss` and `-t` function as desired when called after `-i`, performance is improved slightly having them called before `-i`. Unless I am mistaken, `-stream_loop` can only be called before `-i` and would be very useful for easily looping audio.

Rapptz:
Did it myself here 4fa1bcadaad34b3581eb6bca482fec71183a2a70

PR #139, opened 2016-03-19
Encrypted audio working with PyNacl

izy521:
Volt mentioned the 'plain' method for sending and receiving audio may not be there in the future. We should all probably switch to the encrypted audio method soon.

It was a couple of lines of code here, I decided to go with PyNacl since the other Nacl lib required that the user had a pre-installed libsodium.so (or whatever on Windows) on their machine. I only tested this under Linux, so I don't know if it works under Windows / OSX, but I assume it'd work on the latter.

Edit: Tested the proposed changes under Windows, so it works there as well.

### DIFF
--- a/discord/voice_client.py
+++ b/discord/voice_client.py
@@ -336,20 +336,20 @@ class VoiceClient:
         nonce = bytearray(24)
         box = nacl.secret.SecretBox(bytes(self.secret_key))
         
-        #Formulate header
+        # Formulate header
         header[0] = 0x80
         header[1] = 0x78
         struct.pack_into('>H', header, 2, self.sequence)
         struct.pack_into('>I', header, 4, self.timestamp)
         struct.pack_into('>I', header, 8, self.ssrc)
         
-        #Copy header to nonce's first 12 bytes
+        # Copy header to nonce's first 12 bytes
         for i in range(0, len(header)):
             nonce[i] = header[i]
         
-        #Encrypt the Opus data with the nonce
-        #This weird lib also prepends the nonce to the data
-        #So we extract the data after the 24th spot
+        # Encrypt the Opus data with the nonce
+        # This weird lib also prepends the nonce to the data
+        # So we extract the data after the 24th spot
         encrypted = box.encrypt(bytes(data), bytes(nonce))
         encrypted_ba = bytearray(encrypted[24:])
         return header + encrypted_ba
--- a/discord/voice_client.py
+++ b/discord/voice_client.py
@@ -347,9 +347,9 @@ class VoiceClient:
         for i in range(0, len(header)):
             nonce[i] = header[i]
         
-		#Encrypt the Opus data with the nonce
-		#This weird lib also prepends the nonce to the data
-		#So we extract the data after the 24th spot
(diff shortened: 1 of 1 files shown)


uniquoooo:
@Rapptz hey

b1naryth1ef:
This looks great. Time to merge it @Rapptz

uniquoooo:
Yes sir @Rapptz I agree with Andrei

uniquoooo:
"its got wheels!"

discord/voice_client.py
```diff
@@ -329,17 +332,27 @@ def is_connected(self):
...
+        
+        # Formulate header
+        header[0] = 0x80
+        header[1] = 0x78
+        struct.pack_into('>H', header, 2, self.sequence)
+        struct.pack_into('>I', header, 4, self.timestamp)
+        struct.pack_into('>I', header, 8, self.ssrc)
+        
+        # Copy header to nonce's first 12 bytes
+        for i in range(0, len(header)):
```
b1naryth1ef: This should just be

```
nonce = header[:12]
```

discord/voice_client.py
```diff
@@ -329,17 +332,27 @@ def is_connected(self):
...
+            nonce[i] = header[i]
+        
+        # Encrypt the Opus data with the nonce
+        # This weird lib also prepends the nonce to the data
+        # So we extract the data after the 24th spot
+        encrypted = box.encrypt(bytes(data), bytes(nonce))
```
b1naryth1ef:
```
return header + bytearray(box.encrypt(bytes(data), bytes(nonce))[24:])
```

discord/voice_client.py
izy521: The libsodium libs require the nonce to be 24 bytes in length, wouldn't that make it 12 bytes?

b1naryth1ef:
```python
nonce[12:] = header
```

discord/voice_client.py
izy521: Ok, checks out :D

discord/voice_client.py
```diff
@@ -335,24 +335,20 @@ def _get_voice_packet(self, data):
...
         # Formulate header
         header[0] = 0x80
         header[1] = 0x78
         struct.pack_into('>H', header, 2, self.sequence)
         struct.pack_into('>I', header, 4, self.timestamp)
         struct.pack_into('>I', header, 8, self.ssrc)
-        
+
         # Copy header to nonce's first 12 bytes
         for i in range(0, len(header)):
```
uniquoooo: This can just be:
```python
nonce[:12] = header
```

discord/voice_client.py
```diff
@@ -344,8 +344,7 @@ def _get_voice_packet(self, data):
         struct.pack_into('>I', header, 8, self.ssrc)
 
         # Copy header to nonce's first 12 bytes
-        for i in range(0, len(header)):
-            nonce[i] = header[i]
+        nonce[:12] = header
```
b1naryth1ef: :sparkles: magic

Rapptz:
Rebased.

PR #140, opened 2016-03-23
Add support for token login (for bots)

maxpowa:
Allows you to log in with bot accounts, using 'token' as the email and the bot's token as the password.

Ex (obv invalid token):
```
client.run('token', 'SDMSDNZIQ123Dsna1243sdfk.DcGdhR.A3245ASFIOAI12ASDF_121')
```

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -464,6 +464,24 @@ class Client:
             passing status code.
         """
 
+        if email == "token":
+            log.info('logging in using static token')
+            self.token = password
+            self.headers['authorization'] = 'Bot {}'.format(self.token)
+            resp = yield from self.session.get(endpoints.ME, headers=self.headers)
+            log.debug(request_logging_format.format(method='GET', response=resp))
+
+            if resp.status != 200:
+                yield from resp.release()
+                if resp.status == 400:
+                    raise LoginFailure('Improper token has been passed.')
+                else:
+                    raise HTTPException(resp, None)
+
+            log.info('token auth returned status code {}'.format(resp.status))
+            self._is_logged_in.set()
+            return
+
         # attempt to read the token from cache
         if self.cache_auth:
             yield from self._login_via_cache(email, password)


b1naryth1ef:
TBH this should be a separate function altogether, and perhaps @Rapptz can throw the old user/pass auth in a separate lib :^)

AraHaan:
I did makke a commit for this however Danny turned it down sadly. https://github.com/Rapptz/discord.py/pull/134/files#diff-5bab608907a27a2180166115e6299cef
Be careful though if you change it in your latest copy of client.py to back it up Also all of your bots who are not on the new account type must have ``None`` on the token arg option for them to work.
But If it is the new account type the ``email`` and ``password`` args can be ``None`` and be ok.

maxpowa:
This change doesn't break any existing behavior. It simply adds support for token auth in a non-invasive manner. This does not do any of the conversion from normal user account to bot account, so it's on the user to convert an account on their own (and as such 100% their problem if they convert their primary account to a bot account).

Looking at your PR, it would have broken existing implementations.

AraHaan:
I edited my code since the PR to add in token checking (from what I hear from some ppl it is always a length of 27 characters so, I did this.

https://github.com/AraHaan/discord.py/blob/master/discord/client.py#L580-#L644

uniquoooo:
@AraHaan You can send a request to `GET /api/auth/login` with the `authorization` header set and if it returns 400 Bad Request the token is invalid. If it succeeds it returns the token.

maxpowa:
@AraHaan please keep the comments relative to this PR and not others to reduce confusion.

I'm not really sure if it's necessary but it wouldn't be very difficult to change the validitiy test in this PR to use `GET /api/auth/login` instead of `GET /api/users/@me`.

discord/client.py
```diff
@@ -464,6 +464,24 @@ def login(self, email, password):
...
 
+        if email == "token":
+            log.info('logging in using static token')
+            self.token = password
+            self.headers['authorization'] = 'Bot {}'.format(self.token)
+            resp = yield from self.session.get(endpoints.ME, headers=self.headers)
+            log.debug(request_logging_format.format(method='GET', response=resp))
+
+            if resp.status != 200:
+                yield from resp.release()
```
khazhyk: You never release the response if it is successful - my debug log has been complaining about this. :)

Rapptz:
I did this a bit separately but I merged your work anyway. Just do `client.login('token')`.

PR #141, opened 2016-03-25
Move process killing code to run so that it will run on the process t…

khazhyk:
…hread - fixes exceptions thrown due to communicate timing out


Rapptz:
Rebased f1736bc08f8f13d8b808d73e63a31f1a2d77d4a9

PR #143, opened 2016-03-30
[commands] Fix infinite recursion on subgroups without a command

tekulvw:
Just as an FYI, ctx.invoked_subcommand will end up being an instance of commands.Group!

Example of issue:

Calling `!group subgroup` will cause infinite recursion.

```
@commands.group()
async def group(self):
    pass

@group.group()
async def subgroup(self):
    pass

@subgroup.command()
async def do_something(self):
    pass
```


Rapptz:
Rebased.

PR #145, opened 2016-04-04
Handle servers without member_count (small servers don't have it)

khazhyk:


Rapptz:
Apparently it's not small servers but unavailable ones?

Danny - Today at 10:47 PM
hey @Stanislav there are cases where member_count is missing in READY, is that intentional?
Stanislav - Today at 10:49 PM
If the guild is unavailable it won't have a member doubt 
Count 
Danny - Today at 10:49 PM
ah.
will it have it in the subsequent GUILD_CREATE then?
Stanislav - Today at 10:50 PM
Yes it should 
Danny - Today at 10:50 PM
alright thanks a bunch

PR #146, opened 2016-04-05
Explain how to install libffi for redhat users

mastergberry:


ChandlerScheurkogel:
yum install libffi-devel

TickerOfTime:
Oh my gosh. Unless you want to document linux distro basic 101 for all the distros, why would you do this?

PR #147, opened 2016-04-05
Fix handling of message update in MESSAGE_UPDATE

ZetaHunter:
There is a bug where message doesn't actually get updated in local store, thus subsequent edits reveal contents of original message rather the one after the `(n-1)`th edit.

This PR fixes that.

Also `clean_content` caching had to be removed due to limitations of slots editing them after creation was impossible, instead method renamed to `_clean_content` and in `_update` `clean_content` is set.


Rapptz:
This PR is fine except for the clean_content bit.

Revert that part and change it so you do:

```py
# clear the cached slot cache
del self._raw_mentions
del self._raw_channel_mentions
del self._clean_content
```

at the bottom of `update`.

Rapptz:
Rebased.

PR #156, opened 2016-04-07
Fix Attribute Error when the client joins a server that server is ``None`` so then server.get_member causes that AttributeError.

AraHaan:
I caught this when my bot was joining a server. This is a fix I think.

### DIFF
--- a/discord/state.py
+++ b/discord/state.py
@@ -280,20 +280,21 @@ class ConnectionState:
     def parse_guild_member_update(self, data):
         server = self._get_server(data.get('guild_id'))
         user_id = data['user']['id']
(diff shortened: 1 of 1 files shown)


khazhyk:
I'm not sure that guild_member_update should ever be sent before the first guild_create. Do you have websocket logs of this behavior happening?

AraHaan:
It happens when I join servers on my bot that my bot was not a part of and sadly I could never get the logs from d.py itself to show. FeelsBadMan

cookkkie:
It happens for me too. But just ignoring it is bad.

Rapptz:
Fixed it differently I believe.

PR #161, opened 2016-04-09
Add edit_profile bot accounts support

cookkkie:
Successfully tested. :smile:

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -1514,7 +1514,7 @@ class Client:
             try:
                 password = fields['password']
             except KeyError:
-                raise("Missing password argument.")
+                raise(KeyError("Missing password argument."))
 
             payload = {
                 'password': fields.get('password'),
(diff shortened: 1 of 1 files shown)


discord/client.py
```diff
@@ -1498,13 +1505,24 @@ def edit_profile(self, password, **fields):
...
+        if self.is_bot_account:
+            payload = {
+                'username': fields.get('username', self.user.name),
+                'avatar': avatar
+            }
+        else:
+            try:
+                password = fields['password']
+            except KeyError:
+                raise("Missing password argument.")
```
SunDwarf: This will actually raise a TypeError. You should either let the error pass through or re-raise it with the custom text.

cookkkie: @SunDwarf Thanks for the report ;)

Rapptz:
I implemented it differently here https://github.com/Rapptz/discord.py/commit/29d3f5a886545067a18d6d5e7f540074b86d7959

PR #162, opened 2016-04-09
Provisional support for multiple voice streams.

SunDwarf:
This allows you to use multiple voice clients on a bot account. I haven't extensively tested it yet, however, so there's probably things that are broken.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -358,7 +358,7 @@ class Client:
             server_id = data.get('guild_id')
             if user_id == self.user.id:
                 if self.is_voice_connected(server_id=server_id):
-                    self.voice.channel = self.get_channel(data.get('channel_id'))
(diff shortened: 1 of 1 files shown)


discord/client.py
```diff
@@ -355,15 +355,15 @@ def received_message(self, msg):
 
         if event == 'VOICE_STATE_UPDATE':
             user_id = data.get('user_id')
+            server_id = data.get('guild_id')
             if user_id == self.user.id:
-                if self.is_voice_connected():
+                if self.is_voice_connected(server_id=server_id):
                     self.voice.channel = self.get_channel(data.get('channel_id'))
```
ingmferrer: self.voice[server_id].channel = self.get_channel(data.get('channel_id'))

PR #167, opened 2016-04-11
Fix chunk requests not actually being batched.

khazhyk:
Note, it might be prudent to also support presence_update member creation as this delays the chunking significantly - 10+ seconds on my account


Rapptz:
I merged one of these commits.

PR #181, opened 2016-04-19
Change message.clean_content to handle all @ signs

SexualRhinoceros:
This change should future proof it from all added functionalities such as role mentioning as well with the only downside being reasonably a few added characters to the total limit.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -198,7 +198,7 @@ class Message:
 
         pattern = re.compile('|'.join(transformations.keys()))
         result = pattern.sub(repl, self.content)
-        return result.replace('@everyone', '@\u200beveryone')
+        return result.replace('@', '@\u200b')
 
     def _handle_upgrades(self, channel_id):
         self.server = None


abalabahaha:
This will block accidental @ everyones tho... not cool

Hatsuney:
good

fennekid:
This will break things like emails and URLs though....

imayhaveborkedit:
`re.sub("@[^@]", "@\u200b", result)`

SexualRhinoceros:
if you want to maintain emails and URLs, you can easily use `message.content`

clean_content is meant to strip any and all mentions. I've seen users abuse this in any way possible while running RH1 and this is a change that simply has to be made if you want to say that it cleans all mentions indefinitely from the bot owners end.

Manage Roles not inheriting all permissions W H E N

Rapptz:
When role mentions come out it will probably not be as obvious as "escaping the @ sign". Should the time come for it, `clean_content` will update for it in a different matter.

PR #183, opened 2016-04-23
Discord.py Resume support.

ccubed:
Made changes so that the library supports resumes. Fixed the resume packet to send the token (It was the only thing missing). Shouldn't interfere with anything else as it blocks all other threads until the reconnect happens (Which is necessary because otherwise everything implodes as things try to write to a dead websocket). Also successfully gets the pending missed events. Works good. Tested on my side in a custom build. Working good so far. Left in the reconnect_times variable, can be removed, but might be useful to someone somewhere.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -612,6 +612,7 @@ class Client:
                     raise ClientException('Unexpected websocket closure received')
                 elif self.ws.close_code != 1000:
                     log.info("Websocket closed on us. Reconnecting.")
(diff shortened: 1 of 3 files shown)


ccubed:
I've updated it so that it now includes your most recent commits and pull accepts. It's still able to merge automatically.

japherwocky:
have been running this for the past 2 days, +1

Rapptz:
Done differently.

PR #187, opened 2016-04-27
Keep_Alive_Handler is a function, not a future/task

ccubed:
This breaks bots if you ever get a redirect request.


Rapptz:
No longer a thing.

PR #189, opened 2016-04-27
Guild sharding support

cookkkie:
Sharding is now live.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -414,7 +414,7 @@ class Client:
                 }
             }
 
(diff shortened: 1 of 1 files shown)


ccubed:
I tested this. Can confirm that it works fine.

SunDwarf:
It doesn't work if your shard_id is 0, because of the truthy check.

cookkkie:
@SunDwarf Yeah, i'm dumb...

PR #194, opened 2016-04-30
Make Client.delete_message use _rate_limit_helper.

mmirate:
In response to the new message-deletion rate limit.

This should prevent HTTP 429 errors from propagating to user code when deleting messages.

N.B. `fgrep self.session. discord/client.py` reveals several other places where the rate limit helper is not used, and I've no clue whether such non-usage is intentional.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -118,6 +118,7 @@ class Client:
         self.loop = asyncio.get_event_loop() if loop is None else loop
         self._listeners = []
         self.cache_auth = options.get('cache_auth', True)
+        self._semaphores = {}
 
         max_messages = options.get('max_messages')
         if max_messages is None or max_messages < 100:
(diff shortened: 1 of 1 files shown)


khazhyk:
This might be a bad idea since a large use case is people trying to delete 100 messages at a time, which will slam 95 429s, then 90, then 85, and so on.

mmirate:
I was going to put the above three commits in their own PR, but it seems they're useful here.

With them, the `_rate_limit_helper` creates, for each applicable rate limit defined as "`x`/`y`", a `BoundedSemaphore(value=x)`. Then each request, before being made, `acquire()`s all applicable semaphores, makes the call, then schedules to `release()` each one after `y` seconds.

The effect should be that the ratelimit-helped requests are sent only at a rate that is within the rate-limits (or at least very nearly so; admittedly, I didn't subtract any safety margins; also, some requests use `self.session` directly and therefore won't be tracked, let alone limited).

PR #196, opened 2016-05-01
Add move_role

khazhyk:
This is changing soon anyways, but I already made it.

~~One thing I foresee is someone creating a role, then trying to move it - since it won't be in server.roles it won't work.~~ This should actually work, since I filter the roles and insert the new one - it shouldn't be affected if the role wasn't already in the server.roles list


discord/client.py
```diff
@@ -1996,6 +1996,57 @@ def delete_invite(self, invite):
...
+        roles = []
+
+        # Roles may come in with multiple roles with the same position attribute - fix this here by recalculating.
+        for r in sorted(server.roles, key=lambda x: x.position):
+            if r != role:
+                roles.append(r.id)
+
+        roles.insert(position, role.id)
+
+        payload = [{"id": z[0], "position": z[1]} for z in zip(roles, range(len(roles)))]
```
cookkkie: `payload = [{"id": z[0], "position": z[1]} for z in enumerate(roles)]`
built-in func :+1:

PR #204, opened 2016-05-09
Fixed being Rate limited on Delete messages.

AraHaan:
When anyone prunes messages by mass because of the rate limit of 4~5 msgs per second being able to be deleted I decided to change it myself. This is working and I tested it myself. I would not be proposing this if it was not working at all.

It seems to take me like 15 sec or so now to delete 100 messages with this too.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -972,10 +972,10 @@ class Client:
         """
 
         url = '{}/{}/messages/{}'.format(endpoints.CHANNELS, message.channel.id, message.id)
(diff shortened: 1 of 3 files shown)


AraHaan:
Added a detection for when discord devs add more rate limits so they can easily be identified later. (sadly it would lack the knowledge of what was actually doing the latelimit then so might need to add where the function names get sent to ``_verify_successful_response`` would not be such a bad idea.

PR #216, opened 2016-05-17
logs_from improvements

khazhyk:
Allow datetime objects to be used in logs_from.
Add time_snowflake to convert a datetime to a "snowflake".
Fix logs_from returning incorrect results when retrieving more than 100 messages and only specifying "after".
Fix logs_from returning incorrect results when retrieving and specifying both "before" and "after".
Add option to return messages in oldest->newest order when specifying only "after" or specifying both "before" and "after". (Recommended when only specifying "after" and retrieving over 100 messages)

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -51,6 +51,7 @@ import logging, traceback
 import sys, re
 import tempfile, os, hashlib
 import itertools
+import datetime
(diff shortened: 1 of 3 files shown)


Rapptz:
Rebased.

PR #220, opened 2016-05-21
Allow datetime in limit parameters for purge_from

Harmon758:
Use logs_from instead of LogsFromIterator directly


Rapptz:
This is not compatible with 3.4.

Rapptz:
Rebased.

PR #236, opened 2016-06-04
Add __slots__

Harmon758:


Rapptz:
The current classes that are missing `__slots__` are purely intentional. The only things that have `__slots__` are things that benefit from it due to being in cache. `discord.Object` is meant to have dynamic attributes, along with the two different Clients. Exceptions should not have `__slots__` (as its parents do not).

Harmon758:
This was in part to fix an issue with `getattr` not being able to find certain attributes of objects in the library. Would it be better if `__dict__` was included in the `__slots__` for the two Clients as well?

PR #264, opened 2016-07-05
Get created Objects from cache

SexualRhinoceros:
Get created Objects from cache rather than use objects that are useless if you wish to check values from them.


Rapptz:
Not like this.

PR #290, opened 2016-08-05
Updated Decorator's code.

Diakarus:
Error on my half, Decorator had the proper code but I didn't input it properly. Decorator also has been blocked by Danny, So I am posting it on Decorator's behalf.

### DIFF
--- a/discord/voice_client.py
+++ b/discord/voice_client.py
@@ -146,11 +146,11 @@ class ProcessPlayer(StreamPlayer):
                          client._connected, client.play_audio, after, **kwargs)
         self.process = process
 
-    def run(self):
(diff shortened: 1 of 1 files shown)


khazhyk:
what is the point of this change

Diakarus:
Dunno, Decorator said my spacing was wrong so I fixed it. Didn't know if it was pointless or not, either way Decorator can't pull on their own because they're blocked apparently.

Diakarus:
"also tell him this change instead of writing thread exceptions to console it will write them in discord.py's logging module to" - Decorator

khazhyk:
decorater is blocked for a reason, i think

PR #292, opened 2016-08-05
[fix] Exception in Daemon threads. (that are logged to console instead of somewhere else) & more fixes.

someunknownperson:
I am new to using discord.py and I only use it for voice related things. However it does get exception in Daemon threads if the Voice Server/Region was to crash while the bot is playing stuff which would result in spam city.

Edit:
Added a fix for some reconnecting logic due to ``run_until_complete`` being a pain in the arse in some of my other code I wrote before. The benefit of using ``run_forever`` is more than the reason for ``run_until_complete`` to exist.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -469,9 +469,9 @@ class Client:
         Roughly Equivalent to: ::
 
             try:
-                loop.run_until_complete(start(*args, **kwargs))
+                loop.run_forever(start(*args, **kwargs))
             except KeyboardInterrupt:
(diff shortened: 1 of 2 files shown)


someunknownperson:
Added a fix for some reconnecting logic due to ``run_until_complete`` beign a pain in the arse in some of my other code I wrote before. The benefit of using ``run_forever`` is more than the reason for ``run_until_complete`` to exist.

PR #308, opened 2016-08-20
Allow coroutines for predicate in utils.find

jonnyli1125:
Quality of life change


leovoel:
this won't work
https://docs.python.org/3.6/reference/expressions.html#await

PR #314, opened 2016-08-27
Allow Object in start_private_message

gperinazzo:
**What**
Change the `Client.start_private_message` function to accept a `discord.Object` and improve it's use outside of `Client.send_message`.

**Why**
Because `Client.send_message` expects that a `discord.Object` passed as destination contains a channel (and not a user) id, sending a DM to an user when you only have the user id requires you to search that user first.
Since the  `Client.start_private_message` only requires the user id to create a private channel, this will allow users to bypass the search for the User object.

``` python
channel = await client.start_private_message(discord.Object(user.id))
await client.send_message(channel, content)
```

A currently possible workaround  is to construct an User object directly to pass to `Client.start_private_message` or `Client.send_message`, however the docs state that data classes should not be created manually.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -238,13 +238,8 @@ class Client:
         elif isinstance(destination, Server):
             return destination.id, destination.id
         elif isinstance(destination, User):
-            found = self.connection._get_private_channel_by_user(destination.id)
-            if found is None:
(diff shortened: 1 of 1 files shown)


SunDwarf:
`await client.send_message(user, content)`

PR #316, opened 2016-08-29
Implement rate limit header support

nopjmp:
We will keep track of the headers in a per bucket dict. None key dict
is the default rate limits. We don't increment or decrement anything and
rely on the server sending us the informaiton needed.

We use asyncio.Event and set the flag. We clear the flag when we hit the
global rate limit and set it again when we can continue making requests.
This will allow the wait() coroutine to return immediately.

This should satisfy hammerandchisel/discord-api-docs#108

I'm not sure if this is how you want to accomplish this. I haven't ran into the rate limit headers, but this works with no headers being returned. I'm more than happy to changes.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -31,6 +31,7 @@ import sys
 import logging
 import inspect
 import weakref
+import datetime
 from random import randint as random_integer
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


Rapptz:
Thanks for the PR.

However, this isn't really how I wanted to do it. This one has a couple of deficiencies:

1. It's using a regular dict object, which is memory consuming since there will be lingering buckets.
2. It's using `asyncio.Event` which is the wrong tool for the job -- `asyncio.Lock` would be better.
3. It completely disregards user accounts (which do not have those headers). A cursory look at the implementation shows that you rely on those headers completely to get the right information.
4. The bucket used is completely incorrect. The buckets I hardcoded only made sense with the old model, not the new one with the headers.

I think it's better to wait until either Jake or I go around to doing this.

nopjmp:
Sounds good. I'm just getting started with asyncio and didn't want to use Lock because it seems like the wrong concept for what is needed. Should have probably done a condition instead.

PR #324, opened 2016-09-06
Add the ability to reset a cooldown on a command

Phxntxm:
Currently the only way to do so is accessing private members, which is a "nono", so this should be added at the command level.


Rapptz:
Rebased: https://github.com/Rapptz/discord.py/commit/b7ffbca0c781ae88e2593c5ebb2a063e7d4e2253

PR #333, opened 2016-09-17
Add voice receive support

adongy:
Just the barebones for voice support, no decoding into PCM or anything like that.

User is free to implement `on_speaking` (which is always triggered) and `on_speak` (but he has to call `enable_voice_events` first).

### DIFF
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -483,7 +483,7 @@ class DiscordVoiceWebSocket(websockets.client.WebSocketClientProtocol):
     SESSION_DESCRIPTION
         Receive only. Gives you the secret key required for voice.
     SPEAKING
-        Send only. Notifies the client if you are currently speaking.
+        Send and receive. Notifies who is currently speaking (including you).
     """
 
     IDENTIFY            = 0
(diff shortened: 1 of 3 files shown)


discord/gateway.py
```diff
@@ -567,11 +567,16 @@ def received_message(self, msg):
...
+        elif op == self.SPEAKING:
+            self._connection.client_ssrc[data.get('ssrc')] = data.get('user_id')
+            self._connection.dispatch('speaking', data.get('speaking'), data.get('user_id'))
+
 
     @asyncio.coroutine
     def initial_connection(self, data):
         state = self._connection
         state.ssrc = data.get('ssrc')
+        stats.client_ssrc = {}
```
Rapptz: NameError

discord/voice_client.py
```diff
@@ -305,6 +319,31 @@ def is_connected(self):
...
+        Returns
+        --------
+        boolean
+            True if successful, False otherwise.
+        """
+        log.info('enabling voice packet events...')
+        if not self._connected.is_set():
+            return False
+        func = functools.partial(VoiceClientProtocol, self.secret_key, self._handle_voice_packet)
+        connect = self.loop.create_datagram_endpoint(func, sock=self.socket)
```
Rapptz: This stuff doesn't work on Windows (or rather, certain event loops)

OrangeChaten:
I've played with this, and the fix is quite simple
`stats.client_ssrc ` should be `state.client_ssrc`
`VoiceClientProtocol` should subclass `asyncio.Protocol`(`class VoiceClientProtocol(asyncio.Protocol)`)
doing the above things made this work for me

PR #350, opened 2016-10-02
Fix Paginator still allowing lines which are too long.

Hemerythrin:


Rapptz:
Prefix your commit with [commands].

PR #359, opened 2016-10-14
[commands] pass arguments from bot.load_extension to cog.setup

Socialery:
tl;dr:

``` diff
+    def load_extension(self, name, *, _package=None, **kwargs):
+        lib = importlib.import_module(name, _package)
+        lib.setup(self, **kwargs)
```

usecase:

``` python
somebot.load_extension('.cogs.something',
                       _package=__package__,
                       data_location=data_folder + '/cogs.something`
)
```

Cleaner than pinning random data to the bot itself (imo)
---

I'm willing to rm the `_package` special case, as it is a bit finnicky and lameo, I'm also willing to set up some introspection for the lib.setup stuff, so if the module is an older cog, it won't cause an exception from being fed random arguments from the main bot code

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -711,17 +711,17 @@ class Bot(GroupMixin, discord.Client):
 
     # extensions
 
-    def load_extension(self, name, *args, **kwargs):
+    def load_extension(self, name, *, _package=None, **kwargs):
         if name in self.extensions:
             return
 
-        lib = importlib.import_module(name)
+        lib = importlib.import_module(name, _package)
         if not hasattr(lib, 'setup'):
             del lib
             del sys.modules[name]
             raise discord.ClientException('extension does not have a setup function')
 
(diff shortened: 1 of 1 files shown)


khazhyk:
What's the point of _package ?

Socialery:
in situations where you are running the bot from outside of its cwd

* Project
    * Zwei
        * botty.py
        * Cogs
            * A_Cog


    python -m Zwei.botty

which then imports `.Cogs.A_Cog`, will bugger up, since importlib uses the location relative to the module its used in, aka `discord.ext.bot``.Cogs.A_Cog`, which is obviously unwanted, since the relative import should be in relation to Zwei

If absolute imports are used, the package won't be on path, since python says so.

Programs should not rely on the folder you run them from, and should be allowed to correctly use their package-relative files (imo)

- - -

The alternate method is `load_extension(__package__+'.Cogs.A_Cog') but that is lame, and considering `import_module` has an argument for the package to import relative to, might as well do things correctly n' junk

PR #372, opened 2016-10-27
Add support for reactions.

khazhyk:
Reactions can be be standard emojis, or custom server emojis.

(Still needs some cleanup + documentation but should be functional).


khazhyk:
The buckets for adding reactions seem to be global, and they seem to be 1/1. The current rate limiting scheme will result in rapid requests getting 429'd every other time. whee

Rapptz:
Merged/Rebased mixture.

PR #384, opened 2016-11-05
Make any integer parameters passed to http cast to str.

SunDwarf:
aiohttp is stupid and won't convert ints to str.


SunDwarf:
According to https://github.com/KeepSafe/aiohttp/issues/1364, this is a permenant change and client code has to pass strings only now.

khazhyk:
Looks like aiohttp may be reconsidering this breaking change, for now, just don't upgrade.

asvetlov:
Fixed by upstream (yarl==07.0) I guess.

PR #397, opened 2016-11-16
Convert message embeds to Embed objects

mikevb1:


Rapptz:
This is a breaking change so it was postponed for the rewrite (where I did it locally).

PR #407, opened 2016-12-01
Find the longest prefix match instead of the first

neonobjclash:
Simple fix for prefixes that start with other prefixes


Rapptz:
No thanks. This is on the user to return a sorted list themselves.

PR #408, opened 2016-12-01
added nick_mention function

Kyu:
The reason I added this is because a line that goes:
   
```py
if message.content == client.user.mention:
```   

doesn't detect nickname mentions


Rapptz:
`Member.mention` already does this.

So.. do something like `message.server.me.mention` or something.

PR #410, opened 2016-12-03
Run global checks in help formatter.

khazhyk:
Help formatter was not checking global bot-level checks, resulting in showing commands a user did not have permission for with show_check_failure disabled.


Rapptz:
Rebased.

PR #415, opened 2016-12-14
Allowing a function to execute right before a command is invoked

AileenLumina:
This adds an optional parameter `before_invoke` of the type `function` to the `process_commands` coroutine. If `before_invoke` is given as an argument to `process_commands`, it will execute that function before a command is invoked (if `message` contains a command). This is useful because listening to `on_command` doesn't guarantee that the code that should be executed `on_command` is actually executed before the command is invoked.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -821,7 +821,7 @@ class Bot(GroupMixin, discord.Client):
         if invoker in self.commands:
             command = self.commands[invoker]
             if before_invoke is not None:
-                if asyncio.iscoroutine(before_invoke):
+                if asyncio.iscoroutinefunction(before_invoke):
                     yield from before_invoke()
                 else:
                     before_invoke()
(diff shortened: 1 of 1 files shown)


NotSoSuper:
Well if you're overriding on_message, just put your function right before you call process_commands?

AileenLumina:
@NotSoSuper That would execute the function even if the message doesn't contain a command, which is not the goal of this PR.

SunDwarf:
Several things:

1) This won't work on awaitables, only coroutine functions. 
2) You can do this anyway with a decorator.

```py
def my_decorator(func: typing.Callable) -> typing.Callable:
    @functools.wraps(func)
    def inner(*args, **kwargs):
        # do whatever you want here
        await func(*args, **kwargs)
    
    return inner
```

Rapptz:
This was implemented in the rewrite in 2 different ways.

The first is with the built-in `before_invoke` hooks that can be defined in 3 different ways:

```py
@bot.before_invoke
async def global_one(ctx):
  # do stuff here

class Cog:
  async def __before_invoke(self, ctx):
    # before the commands inside cogs are called

  @commands.command()
  async def foo(self, ctx):
    pass

  @foo.before_invoke
  async def before_foo(self, ctx):
    # before foo is called
```

At the moment the order of invocation is Command Local > Cog Local > Bot Global.

PR #428, opened 2016-12-31
[commands] Bugfix on the teardown function call.

PapyrusThePlant:
And happy new year !


Rapptz:
Rebased into `async`.

PR #434, opened 2017-01-02
Add the ability to add an embed object to file uploads

Avaren:
Add the ability to add an embed object to file uploads - this allows you to link to the uploaded file in the embed using attachment://filename - json_payload is documented here: https://discordapp.com/developers/docs/resources/channel#create-message


Avaren:
(Should have mentioned this has been implemented already in the rewrite branch)

PR #435, opened 2017-01-02
Add ability to set opus encoder input signal type.

MWisBest:


Rapptz:
Rebased this into `async`.

MWisBest:
Awesome, thank you!

I have other changes I'd like to make to the encoder, but I see you're doing an overhaul of some sort. Would you prefer I wait a bit to submit them?

PR #449, opened 2017-01-08
Import missing error class

Avaren:


Rapptz:
https://github.com/Rapptz/discord.py/blob/rewrite/discord/guild.py#L36

Issue #425, opened 2016-12-29
No default notification level attribute for the server object


OrangeChaten:
The api provides the default notification level for a server, a setting in server settings, but discord.py does not expose this anywhere, it may not be useful, but might as well add it for the odd person who uses it

AL1L:
Can this be added anytime soon?

Harmon758:
This has now been resolved by https://github.com/Rapptz/discord.py/commit/1ac432d9f053812a1caab8330d14d4a9ba526289 (#1722) on the rewrite branch.

PR #450, opened 2017-01-09
Add default notification settings to guild object

Avaren:
Able to get and set this guild parameter - for #425

### DIFF
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -74,7 +74,7 @@ class VerificationLevel(Enum):
     def __str__(self):
         return self.name
 
-class DefaultMessageNotifications(Enum):
+class MsgNotifications(Enum):
     all_messages  = 0
     only_mentions = 1
 
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -74,7 +74,7 @@ class VerificationLevel(Enum):
     def __str__(self):
         return self.name
 
-class MsgNotifications(Enum):
+class MessageNotifications(Enum):
     all_messages  = 0
     only_mentions = 1
 
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -74,6 +74,13 @@ class VerificationLevel(Enum):
     def __str__(self):
         return self.name
 
+class DefaultMessageNotifications(Enum):
+    all_messages  = 0
+    only_mentions = 1
+
+    def __str__(self):
+        return self.name
+
 class Status(Enum):
     online = 'online'
     offline = 'offline'
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -35,7 +35,7 @@ from .game import Game
 from .permissions import PermissionOverwrite
 from .errors import InvalidArgument
 from .channel import *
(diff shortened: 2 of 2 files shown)


discord/enums.py
```diff
@@ -74,6 +74,13 @@ class VerificationLevel(Enum):
     def __str__(self):
         return self.name
 
+class DefaultMessageNotifications(Enum):
```
Rapptz: I am not entirely sold on this name.

It's too verbose.

`default_message_notifications=DefaultMessageNotifications.only_mentions` is quite a mouthful. I know there's `verification_level` but even that one is pushing it.

discord/guild.py
```diff
@@ -606,6 +613,13 @@ def edit(self, **fields):
             raise InvalidArgument('verification_level field must of type VerificationLevel')
 
         fields['verification_level'] = level.value
+
+        if 'default_message_notifications' in fields:
+            if not isinstance(fields['default_message_notifications'], DefaultMessageNotifications):
```
Rapptz: Instead of doing `'key' in fields` and then doing `fields['key']` I implore you to use `try...except` instead as it is much faster.

Avaren:
Requested changes completed

discord/guild.py
```diff
@@ -614,11 +614,13 @@ def edit(self, **fields):
 
         fields['verification_level'] = level.value
 
-        if 'default_message_notifications' in fields:
-            if not isinstance(fields['default_message_notifications'], DefaultMessageNotifications):
+        try:
+            if not isinstance(fields['default_message_notifications'], MsgNotifications):
```
Rapptz: This type of style might be better:

```py
notifications = fields.get('default_message_notifications', self.default_message_notifications)    
if not isinstance(notifications, MsgNotifications):
    raise InvalidArgument('default_message_notifications field must of type DefaultMessageNotifications')

fields['default_message_notifications'] = notifications.value
```

discord/enums.py
```diff
@@ -74,7 +74,7 @@ class VerificationLevel(Enum):
     def __str__(self):
         return self.name
 
-class DefaultMessageNotifications(Enum):
+class MsgNotifications(Enum):
```
Rapptz: Still not really sold on this name either (We don't use `Msg` shorthand anywhere). I think `MessageNotifications` might be a better intermediate but again, not really sold on it. Other alternatives might be `GuildMessageNotifications` etc.

I'm not convinced there's a good name for this.

Modelmat: How about NotificationLevel?

Modelmat:
How about NotificationLevel?

AL1L:
Can we get this merged soon?

Harmon758:
This PR is now stale since #1722 has been merged with https://github.com/Rapptz/discord.py/commit/1ac432d9f053812a1caab8330d14d4a9ba526289 on the rewrite branch.

siddrath:
async with aiohttp.get('http://random.cat/meow') as r:
    if r.status == 200:
        js = await r.json()
        await client.send_message(channel, js['file'])

this shows error  AttributeError: module 'aiohttp' has no attribute 'get'

Issue #380, opened 2016-11-03
[Commands] Allow checks to be coroutines


itslukej:
My reason for this is because I need to use aioredis inside my checks ❤️

ccubed:
1/10 - 1 point for effort.

itslukej:
@ccubed 😘

Rapptz:
Planned for the rewrite.

Rapptz:
Implemented in rewrite.

PR #452, opened 2017-01-09
Allow checks to be coroutines - for #380

Avaren:
Wasn't sure if you wanted to support both or drop non-coroutines as part of the rewrite.

The formatter stuff is a bit messy but not sure how else to run the checks there.


Rapptz:
I'm probably going to do this one myself since the change is a bit more complicated and involves a complete rewrite on how the HelpFormatter works.

PR #474, opened 2017-02-07
Check HEARTBEAT_ACK to detect half-open connection

khazhyk:
If we have not received a HEARTBEAT_ACK within 2 intervals, consider
the connection bad and open a new one.


Rapptz:
This does not work with voice keep alive threads.

Rapptz:
You can see how I implemented it here: https://github.com/Rapptz/discord.py/commit/ecab8399ff65c53861e9d44668fce2cb6784d442 for the rewrite. The idea should be the same.

PR #476, opened 2017-02-09
[commands] add `usage` argument to Command to allow overriding the ar…

mikevb1:
…gument list in the default help command

One use case being unordered arguments, but still "named" on the users' end.
Example:
```py
@bot.command(usage='[a] [b]')
async def com(ctx, *args):
    ...
```
Shows as `!com [a] [b]` in the help command instead of `!com [args...]`

### DIFF
--- a/discord/ext/commands/formatter.py
+++ b/discord/ext/commands/formatter.py
@@ -203,7 +203,7 @@ class HelpFormatter:
             result.append(name)
 
         params = cmd.clean_params
-        if cmd.usage is not None:
+        if cmd.usage:
             result.append(cmd.usage)
         elif len(params) > 0:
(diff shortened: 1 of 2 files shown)


discord/ext/commands/formatter.py
```diff
@@ -203,7 +203,9 @@ def get_command_signature(self):
             result.append(name)
 
         params = cmd.clean_params
-        if len(params) > 0:
+        if cmd.usage is not None:
```
Rapptz: Just `if cmd.usage` is enough.

mikevb1:
Requested changes have been made.

Rapptz:
I rebased it.

PR #495, opened 2017-02-25
Convert .jpg URLs to .png

martmists:


khazhyk:
It'd be better to allow a choice, and default to webp or jpg. If it's the same as avatars, I figure images can be retrieved as jpg, webp, jpg, or gif.

PR #503, opened 2017-03-06
Add timeouts for websocket initial connections.

khazhyk:
In DiscordWebSocket and DiscordVoiceWebsocket the from_client factory
methods can hang indefintely on all websocket ops. We set static
timeouts and attempt to reconnect if we time out.

Additionally, do not return from DiscordVoiceWebSocekt.from_client
until we recieve the initial READY payload, after which we have set
up the keep alive thread.

### DIFF
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -71,7 +71,7 @@ class KeepAliveHandler(threading.Thread):
         while not self._stop_ev.wait(self.interval):
             if self._last_ack + 2 * self.interval < time.time():
                 log.warn("We have stopped responding to the gateway.")
-                coro = self.ws.close(1006)
+                coro = self.ws.close(1001)
                 f = compat.run_coroutine_threadsafe(coro, loop=self.ws.loop)
 
                 try:
@@ -191,7 +191,13 @@ class DiscordWebSocket(websockets.client.WebSocketClientProtocol):
         This is for internal use only.
         """
         gateway = yield from client.http.get_gateway()
-        ws = yield from websockets.connect(gateway, loop=client.loop, klass=cls)
+        try:
+            ws = yield from asyncio.wait_for(
+                    websockets.connect(gateway, loop=client.loop, klass=cls),
+                    timeout=60, loop=client.loop)
+        except asyncio.TimeoutError:
+            log.warn('timed out waiting for client connect')
+            return (yield from cls.from_client(client, resume=resume))
 
         # dynamically add attributes needed
         ws.token = client.http.token
@@ -206,7 +212,12 @@ class DiscordWebSocket(websockets.client.WebSocketClientProtocol):
         log.info('Created websocket connected to {}'.format(gateway))
 
         # poll event for OP Hello
-        yield from ws.poll_event()
+        try:
+            yield from asyncio.wait_for(ws.poll_event(), timeout=60, loop=client.loop)
+        except asyncio.TimeoutError:
+            log.warn("timed out waiting for client HELLO")
+            yield from ws.close(1001)
+            return (yield from cls.from_client(client, resume=resume))
 
         if not resume:
             yield from ws.identify()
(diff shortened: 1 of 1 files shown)


discord/gateway.py
```diff
@@ -553,6 +571,15 @@ def from_client(cls, client):
         }
 
         yield from ws.send_as_json(identify)
+
+        try:
+            # Poll for useless HELLO event, then READY
+            yield from asyncio.wait_for(ws.poll_event(), timeout=60, loop=client.loop)
```
jhgg: this is hella error prone. ur first 2 messages are not guaranteed to be HELLO then READY. you can get a heartbeat request in-between them very feasibly.

discord/gateway.py
```diff
@@ -553,6 +571,15 @@ def from_client(cls, client):
...
+            yield from asyncio.wait_for(ws.poll_event(), timeout=60, loop=client.loop)
+        except asyncio.TimeoutError:
+            log.warn("timed out waiting for voice client READY")
```
jhgg: u should also attempt to close the `ws` at this point. otherwise u are leaking the fd.

discord/gateway.py
```diff
@@ -206,7 +212,11 @@ def from_client(cls, client, *, resume=False):
         log.info('Created websocket connected to {}'.format(gateway))
 
         # poll event for OP Hello
-        yield from ws.poll_event()
+        try:
+            yield from asyncio.wait_for(ws.poll_event(), timeout=60, loop=client.loop)
+        except asyncio.TimeoutError:
```
jhgg: same thing here, re: leaking the websocket

khazhyk: how do i mark this done in github

khazhyk:
ping

discord/gateway.py
```diff
@@ -206,7 +212,12 @@ def from_client(cls, client, *, resume=False):
...
+            log.warn("timed out waiting for client HELLO")
+            yield from ws.close(1006)
```
jhgg: use 1000 instead of 1006. if u are giving up - clean close and don't resume.

discord/gateway.py
```diff
@@ -538,7 +550,14 @@ def send_as_json(self, data):
     def from_client(cls, client):
         """Creates a voice websocket for the :class:`VoiceClient`."""
         gateway = 'wss://' + client.endpoint
-        ws = yield from websockets.connect(gateway, loop=client.loop, klass=cls)
+        try:
+            ws = yield from asyncio.wait_for(
+                    websockets.connect(gateway, loop=client.loop, klass=cls),
```
jhgg: what happens to the task if the timeout is reached? is it cancelled proper?

khazhyk:
Error code here isn't checked for resume  Used 1006 since this situation
means the other end stopped responding and we don't know why.

Asyncio handles canceling tasks.

discord/gateway.py
khazhyk: The task is cancelled. If it is not that should be considered a bug in asyncio.

khazhyk:
Looking again, 1006 seems wrong here and the other place we use it, as we still do end up sending a close frame. 1001 "going away" seems most appropriate, we're not done but we're closing the connection.

khazhyk:
hi

khazhyk:
ping, fwiw this happened to me again

Rapptz:
I finally rebased this for you spoops, you can stop crying now. It'll all be okay.

khazhyk:
it'll never be ok

Rapptz:
I reverted this PR (https://github.com/Rapptz/discord.py/commit/4835ca3404d1691c4bc5639d042fc64ab2aec8c6) due to it causing a TypeError with its implementation.

My mistake for not testing this.

I suggest a better implementation would be to timeout the `from_client` call instead of the individual aspects inside the function. That at least would not cause a TypeError and I figure would be cleaner.

khazhyk:
fwiw this was tested & running for weeks on 3.5.1 & 3.5.2 with no issue and working as expected. (Although I suspect it won't work on 3.4)

PR #522, opened 2017-03-29
Update basic_bot.py

VirusControl:


Pandentia:
A Python file is not a shell or batch file...

PR #566, opened 2017-05-14
Added installation note regarding macOS opus support to README

chdsbd:
The macOS installation for the opus codec is slightly different than Linux, so I added a macOS specific note.


Rapptz:
The README doesn't go into detail of how to install libopus. Just PyNaCl. 

Thanks for your PR but I don't think it really belongs here. Maybe in the new documentation, but that doesn't have instructions with voice yet.

PR #572, opened 2017-05-19
Allow any iterable to be passed to `command_prefix`.

SunDwarf:
Flattens any iterable in `get_prefix`. This allows custom iterators to be passed.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -25,6 +25,7 @@ DEALINGS IN THE SOFTWARE.
 """
 
 import asyncio
+import collections
 import discord
 import inspect
 import importlib
@@ -666,6 +667,8 @@ class BotBase(GroupMixin):
             if asyncio.iscoroutine(ret):
                 ret = yield from ret
             return ret
+        elif isinstance(prefix, collections.Iterable) and not isinstance(prefix, str):
+            return list(prefix)
         else:
             return prefix
 


discord/ext/commands/bot.py
```diff
@@ -666,6 +667,8 @@ def get_prefix(self, message):
             if asyncio.iscoroutine(ret):
                 ret = yield from ret
             return ret
+        elif isinstance(prefix, collections.Iterable) and not isinstance(prefix, str):
```
mgardne8: Collections.Iterable is not a sure fire way to check that objects will iterate safely in this use case, for example byte objects:

```python
In [27]: example = bytes([50,60,100,255,200])

In [28]: isinstance(example,collections.Iterable)
Out[28]: True

In [29]: isinstance(example,str)
Out[29]: False

In [30]: isinstance(example,collections.Iterable) and not isinstance(example,str)
Out[30]: True

In [31]: example
Out[31]: b'2<d\xff\xc8'

In [32]: for x in example:
    ...:     print(x)
    ...:
50
60
100
255
200
```
I think it would be better to use a set list of built in iterable container types, such as
```python
elif isinstance(prefix, (tuple, list, dict, set)):
# ...
```
But I do support this concept overall

discord/ext/commands/bot.py
```diff
@@ -666,6 +667,8 @@ def get_prefix(self, message):
...
+            return list(prefix)
```
mgardne8: I would also support returning something along the lines of
```python
return list(map(str,prefix))
```
giving us the assurance that everything was cast to a string for future uses
```python
In [43]: r = 'lul'

In [44]: s = '@@'

In [45]: triggers = (r, s, '~~', 3)

In [46]: list(map(str,triggers))
Out[46]: ['lul', '@@', '~~', '3']
```

PR #573, opened 2017-05-20
tell people how to install the development version properly

TheDevFreak:
change thing in readme so it points to correct urls


TheDevFreak:
maybe applies to async branch too?

PR #582, opened 2017-05-29
Added XKCD and Crayon Colour presets within discord.ext.colour

mgardne8:
This PR adds XKCD and Crayon colour classes to the ext package for explicit import and use within Discord.py projects


Rapptz:
You figured out how to do extension packages so I'm closing this.

PR #607, opened 2017-06-23
Update MissingRequiredArgument exception message

jacklin213:
Changed message from "is a required argument that is missing" to "has a required argument that is missing"

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -209,7 +209,7 @@ class Command:
             if param.kind == param.VAR_POSITIONAL:
                 raise RuntimeError() # break the loop
             if required:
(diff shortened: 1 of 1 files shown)


Rapptz:
This message is incorrect.

The current message says "<name of parameter> is a required..." yours says "<name of parameter> has a required...".

jacklin213:
Tested it earlier on a discord server when I was running a command
!test 
when the command expected 1 argument, the error message produced was
```test is a required argument that is missing```
which I believe is incorrect??

Phxntxm:
Then you named the parameter you are requiring "test". The parameter that is missing, is what is named there.

jacklin213:
@Phxntxm ah yes you are correct, My mistake. Went back and took a look now and my parameter was named "test" 
Sorry about that

Issue #426, opened 2016-12-31
Commands ext overwrites `instance` when adding a cog


Fuyukai:
Opening because I don't know if this is fixed in rewrite.

When having multiple bots running in the same process, the `instance` property of the cog object created is always set to the last bot that claims the cog, due to the Command instance staying between object creation, as they're members of the type.

Take this example code:

```py
class Example(object):
     def __init__(self, bot, a: int):
        self.bot = bot
        self.a = a

    @commands.command()
    async def d():
        await self.bot.say(self.a)

bot1 = Bot(command_prefix="!")
bot1.add_cog(Example(bot1, 1))

bot2 = Bot(command_prefix="?")
bot2.add_cog(Example(bot2, 2))
```

When `?d` is called, it will correctly respond with `2`. However, when `!d` is called, it will respond with `2` due to the second `add_cog` call overwriting `instance` on the Command object representing the `d` command.

This can also cause problems with multi-threaded bots where separate instances of loops running in threads return the wrong loop when `.say()` or similar is called, due to the loop being from another thread.

A workaround is to use `ctx.bot`, but this causes problems when you need to access other self attributes specific to that cog.

Rapptz:
This should be fixed I guess in the rewrite. Feels weird though.

Rapptz:
Reopening this as a note that it is not actually fixed 😒

Hornwitser:
I've toyed with a Descriptor based method of fixing this.  It works much the same as how class functions turn into bound methods, but I don't really like it  See Hornwitser@1e10508339d42b8feee9a9ce4dcba25af34ec401 for a possible implementation.

orlnub123:
An easier fix would be to create a new command instance using copy.copy or something similar and set the instance attribute on it. I could make a PR if you'd like to see an implementation of it.

Rapptz:
Should be fixed in https://github.com/Rapptz/discord.py/commit/caf3d17d4aa3ce45435bb96ea6d99317bfef8618

PR #609, opened 2017-06-23
Fix issue #426 commands ext overriding `instance` propery 

Joshua-Butt:
link: #426 

Replace `Command.instance` property with `Command.instances` list.
on call to `Command,__get__()` `Cogs` are appended to `Command.instances` rather than overriding the value.

### DIFF
--- a/discord/ext/commands/context.py
+++ b/discord/ext/commands/context.py
@@ -87,7 +87,7 @@ class Context(discord.abc.Messageable):
         self._state = self.message._state
 
     @asyncio.coroutine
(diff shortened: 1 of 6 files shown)


Rapptz:
Definitely not a fan of this fix.

Joshua-Butt:
Absolutely agree it's a somewhat kludgy solution.
Was really hoping to get the ball rolling, though definitely not an issue commonly experienced by users.

if you're not happy with it I can try and come up with a cleaner approach somewhere down the line.
Unless you or someone else had already thought up a more effective means.

PR #615, opened 2017-07-01
Close ClientSession before interpreter exits.

NamedKitten:


Rapptz:
Hello. Thanks for the PR.

It seems that aiohttp keeps reverting its stance on whether `ClientSession.close` is a coroutine. I've been using `aiohttp` in this library since it was in v0.20 or so.

During that time, from v0.20 to around v0.22 it was not a coroutine so this solution did indeed work. When v1.x series came out, they decided to make it a coroutine, except this time it was a dud coroutine since all it did was return a completed `asyncio.Future`.

Now we're in v2.x and it seems to me that they've reverted this decision and now it is no longer a coroutine, even giving a deprecation warning about it:

```py
    def close(self):
        """Close underlying connector.

        Release all acquired resources.
        """
        # actual code goes here -- omitted for brevity
        return deprecated_noop('ClientSession.close() is not coroutine')
```

Anywho, this is the history why historically the solution you propose has never been added.

I'm not sure if I should decide to add it at all. On one hand it would suppress the ClientSession is not closed message, however that message seems to have a dual purpose of informing me when people don't actually run their bots -- since during normal execution the client will have cleaned itself up naturally.

I'll leave this up for a bit while I collect my thoughts.

I recommend removing the docstring since it's kind of meaningless.

NamedKitten:
Fixed some things.
Have a check now.

discord/http.py
```diff
@@ -728,9 +728,9 @@ def get_user_profile(self, user_id):
...
-        Closes an aiohttp ClientSession instance any
-        time intepreter exits before it is closed.
-        """
         if not self._session.closed:
-            self._session.close()
+            if not asyncio.iscoroutinefunction(self._session.close)
+                self._session.close()
+            else:
+                # close must have been changed back to a coroutine or future then.
+                self.loop.create_task(self._session.close())
```
Rapptz: We can't do this -- by the time `__del__` is called the loop will have already been closed.

NamedKitten:
Reverted back to original implementation.

NamedKitten:
Ready to merge PR?

Rapptz:
Based on a PR on aiohttp itself: https://github.com/aio-libs/aiohttp/pull/2027 it seems like merging this might be a bad idea.

I don't quite know why they keep doing the back and forth and making it a coroutine. Probably a future proofing strategy.

Rapptz:
Due to aiohttp changes I mentioned above I'm going to close this.

PR #621, opened 2017-07-04
[commands] fix unloading incorrect cogs

khazhyk:
unload_extension would incorrectly unload cogs/listeners
of other extensions if the name of one was a prefix of
another.

Tested w/ prefixes and w/ cogs in submodules

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -88,6 +88,9 @@ _mentions_transforms = {
 
 _mention_pattern = re.compile('|'.join(_mentions_transforms.keys()))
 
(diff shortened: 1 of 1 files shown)


Rapptz:
I'm not sure how this fixes the issue, or rather what the issue is at all.

If you're talking about unloading `cogs.foo` leading to unloading of `cogs.foo.bar` then this is actually intended.

khazhyk:
unloading `cogs.foo` would unload `cogs.foobar`, which is invalid

Rapptz:
Oh I see this now.

PR #628, opened 2017-07-05
Hide class constructors in documentation

datmellow:
hiding constructor on docs for objects that should not be created by users.


Rapptz:
Rebased.

PR #631, opened 2017-07-07
Add basic rewrite voice example

imayhaveborkedit:
Note I technically haven't tested this but I've been over it so many times I can't imagine something's broken.


Rapptz:
Rebased.

PR #649, opened 2017-07-15
Add get_role for Guild object

palmtree5:
Takes a role ID, gets the role using a list comprehension and returns the found role (or None if the resulting list from the comprehension is not found to be True). Suggesting this because `Role` seems to be left out in terms of getting an object based on the ID. I've tested this too and it works as written

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -314,6 +314,11 @@ class Guild(Hashable):
         """Returns a :class:`Member` with the given ID. If not found, returns None."""
         return self._members.get(user_id)
 
+    def get_role(self, role_id):
+        """Returns a :class:`Role` with the given ID. If not found, returns None."""
+        found = [r for r in self.roles if r.id == role_id]
+        return found[0] if found else None
+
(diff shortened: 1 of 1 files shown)


Gorialis:
I don't think it has been documented on the repo but @Rapptz has provided reasons for not adding this method in the past, mainly because of the variable complexity/performance between the implementation of this in comparison to the implementation of the other `get_*` functions.

palmtree5:
that begs the question: why is `Guild.roles` a `list`, given that `Guild._members` and `Guild._channels` are both `dict`? It seems like it'd make sense for `Guild.roles` to also be a `dict`

Rapptz:
Because it's a trade-off in usability.

Making it a dict makes the following cases faster:

1. `get_role`

Making it a list makes the following cases faster:

1. `Member.roles`
2. `Guild.roles`

Since iterating over the roles is arguably more common than retrieving one by ID (both internally and externally) this is the reason why it's not a `dict`.

PR #650, opened 2017-07-16
Add no_bulk argument to TextChannel.purge

khazhyk:
no_bulk=True will never use bulk message delete. Useful e.g. for
using the purge interface for deleting the bot's own messages.


ivandardi:
Why not make it `bulk=True` instead? That way if you don't want bulk you can just make `bulk=False` and get rid of a double negative.

khazhyk:
The reasoning behind no_bulk is that if using bulk=True, we still would sometimes fall back to single message deletes (when messages are older than 2 weeks, or if we are a user account). I can change it, but this seemed more clear.

khazhyk:
github won't let me change the merge target for a pull request... thx https://github.com/Rapptz/discord.py/pull/666

PR #663, opened 2017-07-21
Add Union for multiple converters

BeatButton:
Use `commands.Union(*converters)` to try to convert, in order, to each converter passed. This supports instances of `Converter` as well as any other converter built into the library.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -35,7 +35,7 @@ from .view import StringView
 __all__ = [ 'Converter', 'MemberConverter', 'UserConverter',
             'TextChannelConverter', 'InviteConverter', 'RoleConverter',
             'GameConverter', 'ColourConverter', 'VoiceChannelConverter',
-            'EmojiConverter', 'IDConverter', 'clean_content' ]
+            'EmojiConverter', 'IDConverter', 'clean_content', 'Union']
 
 def _get_from_guilds(bot, getter, argument):
     result = None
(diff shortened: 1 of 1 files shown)


discord/ext/commands/converter.py
```diff
@@ -441,3 +441,24 @@ def repl2(obj):
...
         pattern = re.compile('|'.join(transformations.keys()))
         return pattern.sub(repl2, result)
+
+class Union(Converter):
+    """Tries to convert the argument to each of a list of
+    converters, in order.
+
+    Attributes
+    ------------
+    converters : tuple(:class:`Converter`,any)
```
Rapptz: This should probably point to the documentation on converters.

Unfortunately it doesn't have a ref yet so it's not possible atm.

discord/ext/commands/converter.py
```diff
@@ -441,3 +441,24 @@ def repl2(obj):
...
+        self.converters = converters
+
+    @asyncio.coroutine
+    def convert(self, ctx, argument):
+        for converter in self.converters:
+            try:
+                return (yield from ctx.command.do_conversion(ctx, converter, argument))
+            except BadArgument:
+                pass
+        raise BadArgument('Converter for Union{} failed'.format(self.converters))
```
Rapptz: I don't think this is a useful default message to have.

BeatButton:
I think I addressed the error message problem, but I realize failures to convert don't always raise BadArgument, and am not actually sure swallowing TypeError and ValueError is a good idea, even if it's necessary for this to work as intended.

PR #664, opened 2017-07-22
Make `Message.add_reaction()` return the added reaction

ivandardi:


discord/message.py
```diff
@@ -622,6 +627,7 @@ def add_reaction(self, emoji):
             raise InvalidArgument('emoji argument must be str, Emoji, or Reaction not {.__class__.__name__}.'.format(emoji))
 
         yield from self._state.http.add_reaction(self.id, self.channel.id, emoji)
+        return Reaction(message=self, data={'me': True}, emoji=emoji)
```
Rapptz: This isn't going to return a proper object in case some other type is passed.

For example, I convert between many different types such as `Emoji`, `PartialReactionEmoji`, `Reaction` itself, and finally `str`. I then convert them to `str` and call it `emoji`. In contrast, `Reaction.emoji` expects only two things: `Emoji` or `str`.

PR #665, opened 2017-07-23
add static_format parameter to avatar_url_as

khazhyk:
static_format will only apply to static (not animated) avatars. Makes
it easier to grab gif-or-'format' of an avatar. Defaults to 'webp'

### DIFF
--- a/discord/user.py
+++ b/discord/user.py
@@ -24,14 +24,17 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 DEALINGS IN THE SOFTWARE.
 """
 
-from .utils import snowflake_time, _bytes_to_base64_data, parse_time
+from .utils import snowflake_time, _bytes_to_base64_data, parse_time, valid_icon_size
 from .enums import DefaultAvatar, RelationshipType, UserFlags
-from .errors import ClientException
+from .errors import ClientException, InvalidArgument
 
 from collections import namedtuple
 
 import discord.abc
 import asyncio
+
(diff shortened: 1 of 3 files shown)


discord/user.py
```diff
@@ -110,8 +110,11 @@ def avatar_url_as(self, *, format, size=1024):
         format: Optional[str]
             The format to attempt to convert the avatar to.
             If the format is ``None``, then it is automatically
-            detected into either 'gif' or 'webp' depending on the
+            detected into either 'gif' or static_format depending on the
             avatar being animated or not.
+        static_format: Optional['str']
```
Rapptz: This isn't `Optional` (this is reserved for parameters that take in `None` as a valid option).

khazhyk:
Fixed comment.

Use case is similar to that of avatar_url or avatar_url_as(format=None), just allowing you to specify non animated format, instead of always using webp

discord/user.py
```diff
@@ -128,7 +131,7 @@ def avatar_url_as(self, *, format, size=1024):
             if self.avatar.startswith('a_'):
                 format = 'gif'
             else:
-                format = 'webp'
+                format = static_format
```
jhgg: png,webp,gif,jpeg are the only supported static formats. everything else will throw a 415 status code. you could save the caller a lot of trouble by validating this and perhaps throwing an exception.

khazhyk: Size seems to also only accept powers of two in range [16, 2048] ?

khazhyk: Done

Rapptz:
Rebased.

PR #666, opened 2017-07-23
Add no_bulk argument to TextChannel.purge

khazhyk:
no_bulk=True will never use bulk message delete. Useful e.g. for
using the purge interface for deleting the bot's own messages.

### DIFF
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -211,7 +211,7 @@ class TextChannel(discord.abc.Messageable, discord.abc.GuildChannel, Hashable):
         yield from self._state.http.delete_messages(self.id, message_ids, reason=reason)
 
     @asyncio.coroutine
-    def purge(self, *, limit=100, check=None, before=None, after=None, around=None, reverse=False, reason=None):
+    def purge(self, *, limit=100, check=None, before=None, after=None, around=None, reverse=False, reason=None, bulk=True):
         """|coro|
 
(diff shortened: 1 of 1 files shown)


discord/channel.py
```diff
@@ -211,7 +211,7 @@ def delete_messages(self, messages, *, reason=None):
         yield from self._state.http.delete_messages(self.id, message_ids, reason=reason)
 
     @asyncio.coroutine
-    def purge(self, *, limit=100, check=None, before=None, after=None, around=None, reverse=False, reason=None):
+    def purge(self, *, limit=100, check=None, before=None, after=None, around=None, reverse=False, reason=None, no_bulk=False):
```
Rapptz: Maybe use `bulk=True`.

discord/channel.py
```diff
@@ -245,6 +245,8 @@ def purge(self, *, limit=100, check=None, before=None, after=None, around=None,
             Same as ``reverse`` in :meth:`history`.
         reason: Optional[str]
             The reason for doing this action. Shows up on the audit log.
+        no_bulk: bool
+            If True, never use bulk delete.
```
Rapptz: Swap the `True` -> `False` with the `bulk` change and then probably state that pretty much the only use case is for deleting your own messages due to the way bulk delete works Discord side.

khazhyk:
Done

Rapptz:
Rebased.

PR #667, opened 2017-07-23
retry on 500 + raise once out of retries

khazhyk:
Discord official client retries on 500, so worst case scenario, we're
not any worse than the official client which seriously outnumbers us.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -193,8 +193,8 @@ class HTTPClient:
 
                         continue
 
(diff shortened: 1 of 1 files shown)


Rapptz:
Instead of changing the already working retry count, why not use the following construct:

```py
for ...:
   pass
else:
  raise HTTPException
```

khazhyk:
The retry count doesn't change...?

PR #670, opened 2017-07-25
allow trailing null bytes in uploaded jpegs

khazhyk:
Some jpegs from In The Wild™ can have trailing null bytes. Discord handles it fine.


Rapptz:
Rebased.

PR #684, opened 2017-07-31
Remove accept_invite & references

Gorialis:
Bot accounts can't accept invites anyway and the endpoint no longer works on user/selfbots.
Also amended `delete_invite` docstring to no longer rely on the invite param explanation in `accept_invite`.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -2701,40 +2701,13 @@ class Client:
         return result
 
     @asyncio.coroutine
-    def accept_invite(self, invite):
-        """|coro|
-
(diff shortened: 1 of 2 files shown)


Rapptz:
No breaking changes.

Gorialis:
Fair enough, although I'd argue that since the endpoint no longer works as intended anyway, raising AttributeError (or at least logging a deprecation error a la [change_status](http://discordpy.readthedocs.io/en/async/api.html#discord.Client.change_status)) is more favorable for users than letting the accounts unverify. Parallel created at [687](https://github.com/Rapptz/discord.py/pull/687).

PR #686, opened 2017-08-01
Fix volume command

Nekorooni:
Volume should be a value between 0-1.

### DIFF
--- a/examples/basic_voice.py
+++ b/examples/basic_voice.py
@@ -103,14 +103,14 @@ class Music:
         await ctx.send('Now playing: {}'.format(player.title))
 
     @commands.command()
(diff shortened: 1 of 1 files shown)


imayhaveborkedit:
I my other file I have
```py
@commands.command()
async def volume(self, ctx, vol: int=None):
    if vol == None:
        await ctx.send("Current volume is %s%%" % vol*100)
    elif 0 < vol <= 100:
        ctx.voice_client.source.volume = vol/100
        await ctx.send("Volume set to %s%%" % vol)
    else:
        await ctx.send("Invalid volume, must be between 1 and 100")
```

Rapptz:
Borked's approach is better (using a number 1 to 100).

PR #688, opened 2017-08-01
Change VerificationLevel, ContentFilter to be IntEnums

Gorialis:
This allows comparisons like `guild.verification_level >= discord.VerificationLevel.low` and `guild.explicit_content_filter > discord.ContentFilter.disabled` to be done rather than needing to use the `value` attribute.


Rapptz:
I think `IntEnum` is a big too loose, but it's probably the easiest way to do this. 

By loose, I mean you can do `guild.verification_level == 4` which isn't very type safe (along with the implicit conversion to integer). Maybe it doesn't matter, who knows.

PR #689, opened 2017-08-01
Examples for AsyncIterator, change_presence & some minor changes

Gorialis:
Examples as in title, changing example on `typing()` to use the more desirable `async with` and removing a mention of dropped default guild channel


Rapptz:
Rebased.

PR #693, opened 2017-08-03
API was changed

SourSpoon:
Default channel was removed in API, example now direct messages member instead.

### DIFF
--- a/examples/new_member.py
+++ b/examples/new_member.py
@@ -8,8 +8,7 @@ class MyClient(discord.Client):
         print('------')
 
     async def on_member_join(self, member):
-        guild = member.guild
-        await guild.default_channel.send('Welcome {0.mention} to {1.name}!'.format(member, guild))
+        await member.send('Welcome {0.mention} to {1.name}!'.format(member, guild))
 
 client = MyClient()
 client.run('token')


mgardne8:
Why would we direct message a member to welcome themselves to the server?

Bluscream:
@mgardne8 Maybe ask the person that initially added this. Don't blame @SourSpoon for fixing your shit...

examples/new_member.py
```diff
@@ -8,8 +8,7 @@ class MyClient(discord.Client):
         print('------')
 
     async def on_member_join(self, member):
-        guild = member.guild
-        await guild.default_channel.send('Welcome {0.mention} to {1.name}!'.format(member, guild))
+        await member.send('Welcome {0.mention} to {1.name}!'.format(member, guild))
```
mgardne8: You removed the assignment of the guild variable, and then continued to use it within the format statement, this will raise a NameError exception.

DiNitride:
@Bluscream Uhhh, can you explain how Matt is blaming Sour for anything, he simply asked a question

SourSpoon:
Direct messaging someone when they join could be useful for things like sending rules ect. The point was however to remove the default channel reference as it no longer exists in rewrite

mgardne8:
I can understand wanting to correct and example to align with the changes in the api, but I believe it would be more beneficial to, while updating, expand on the code to function closely to the way it currently does and not just bandaid over it. (A functioning bandaid would also be beneficial, please test your PRs)

PR #699, opened 2017-08-07
Remove GuildChannel.is_default

SinisterRectus:
To keep consistent with the previous default channel removals.

I wasn't sure how to document this in `migrating.rst` because it was originally in the `Channel` class, now it was in `GuildChannel`, but elsewhere it is documented as part of `TextChannel`:

```
The following are now methods instead of properties (requires parentheses):

- :meth:`TextChannel.is_default`
- :meth:`Role.is_default`
- :meth:`Client.is_ready`
- :meth:`Client.is_closed`
```

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -254,10 +254,6 @@ class GuildChannel:
             ret.append(role)
         return ret
 
-    def is_default(self):
-        """bool : Indicates if this is the default channel for the :class:`Guild` it belongs to."""
-        return self.guild.id == self.id
-
     @property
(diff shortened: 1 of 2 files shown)


Rapptz:
If you could also document the entries in `migrating.rst` I'd appreciate it.

SinisterRectus:
Like I said, I wasn't sure how to properly document it, but I guess this works.

Rapptz:
Yup that's good.

Rapptz:
Rebased.

PR #702, opened 2017-08-08
Fixed ExponentialBackoff not working when integral

mgardne8:
random.rand_range is not a thing, this causes issues when trying to use ExponentialBackoff with `integral=True`

```py
In [14]: from discord.backoff import  ExponentialBackoff

In [15]: nonInt = ExponentialBackoff(integral=False, base=20)

In [16]: nonInt.delay()
Out[16]: 15.93630551139845

In [17]: nonInt.delay()
Out[17]: 6.208550157892292

In [18]: Int = ExponentialBackoff(integral=True, base=20)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-18-94850ca45364> in <module>()
----> 1 Int = ExponentialBackoff(integral=True, base=20)

C:\Program Files\Anaconda3\lib\site-packages\discord\backoff.py in __init__(self, base, integral)
     62         rand.seed()
     63
---> 64         self._randfunc = rand.rand_range if integral else rand.uniform
     65
     66     def delay(self):

AttributeError: 'Random' object has no attribute 'rand_range'
```

### DIFF
--- a/discord/backoff.py
+++ b/discord/backoff.py
@@ -61,7 +61,7 @@ class ExponentialBackoff:
         rand = random.Random()
         rand.seed()
 
-        self._randfunc = rand.rand_range if integral else rand.uniform
+        self._randfunc = rand.randrange if integral else rand.uniform
 
     def delay(self):
         """Compute the next delay


Rapptz:
Rebased

PR #705, opened 2017-08-09
Disallow setting non-permission attributes

BeatButton:
This will avoid issues where someone makes a mistake typing the name of a permission, or thinks the name is something else.


Rapptz:
Squashed.

PR #717, opened 2017-08-14
Add delete_after to 'edit'.

MysterialPy:
Allow for edited messages to be deleted after [float] seconds with optional parameter delete_after.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -27,7 +27,7 @@ DEALINGS IN THE SOFTWARE.
 import asyncio
 import re
 
-from . import utils
+from . import utils, compat
 from .reaction import Reaction
 from .emoji import Emoji, PartialReactionEmoji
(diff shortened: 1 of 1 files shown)


discord/message.py
```diff
@@ -518,6 +518,9 @@ def edit(self, **fields):
             If provided, the number of seconds to wait in the background
             before deleting the message we just edited. If the deletion fails,
             then it is silently ignored.
+        reason: Optional[str]
```
Rapptz: This is kind of pointless. Would be better to revert your two commits.

I'm going to end up removing them from `send` as well since by design discord does not log bot deletes.

MysterialPy: I thought as much but was trying to keep it inline with your current design.

Rapptz:
I've rebased it.

PR #730, opened 2017-08-22
Add property top_hoist to Member.

MysterialPy:
Returns the members highest hoisted role. If not found, will return None.


Rapptz:
I don't think this is common enough to warrant inclusion in the library.

MysterialPy:
Partially agree. Though, I think it's something which has a use case, since it's representing a members visible role. I'll close it off for now.

PR #738, opened 2017-08-29
Add __str__ to Message

orlnub123:
Made \_\_str__ return Message.content


Rapptz:
Thanks for the PR.

However, `Message` purposefully doesn't have `__str__` implemented.

PR #740, opened 2017-08-29
Add voice receive capabilities.

Bottersnike:
The API is as follows:

1. Create your `VoiceClient` object (usually from a `.connect()`)
2. Register your voice events using `@client.event` or `@bot.event` (see the documentation for the individual events)
3. That's it. The voice data or user speaking events will be sent to those callbacks.

You can also use `await voice_client.get_voice_data` to listen for audio from a user, then return it all once the user stops talking. This has the added side effect of reordering all the packets for you before compiling them into a `bytearray` object and returning it for you.

If you're _really_ lazy, you can do `await voice_client.pipe_voice_into_file` which will listen to a user and pip all of their voice data into a file-like object for you all reordered and accounting for missed packets and silence.

See `examples/basic_voice_receive.py` to have a look at what it might look like in reality.

### DIFF
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -553,6 +553,7 @@ class DiscordVoiceWebSocket(websockets.client.WebSocketClientProtocol):
         super().__init__(*args, **kwargs)
         self.max_size = None
         self._keep_alive = None
+        self._dispatch = lambda *args: None
 
     @asyncio.coroutine
     def send_as_json(self, data):
@@ -595,6 +596,7 @@ class DiscordVoiceWebSocket(websockets.client.WebSocketClientProtocol):
         ws.gateway = gateway
         ws._connection = client
         ws._max_heartbeat_timeout = 60.0
+        ws._dispatch = client._state.dispatch
 
         if resume:
             yield from ws.resume()
@@ -649,6 +651,22 @@ class DiscordVoiceWebSocket(websockets.client.WebSocketClientProtocol):
             yield from self.identify()
         elif op == self.SESSION_DESCRIPTION:
             yield from self.load_secret_key(data)
+        elif op == self.SPEAKING:
+            user_id = int(data['user_id'])
+            ssrc = int(data['ssrc'])
+
+            if 'speaking' in data:
+                user = self._connection._state.get_user(user_id)
+
+                self._dispatch('voice_speaking_state', self._connection, user, data['speaking'])
+
+                # Used for continuous streams
+                for f in self._connection._speaking_listeners:
+                    f(user, data['speaking'])
+
+                if self._connection.channel.id not in self._connection._ssrc_lookup:
+                    self._connection._ssrc_lookup[self._connection.channel.id] = {}
+                self._connection._ssrc_lookup[self._connection.channel.id][ssrc] = user_id
 
     @asyncio.coroutine
     def initial_connection(self, data):
@@ -695,5 +713,3 @@ class DiscordVoiceWebSocket(websockets.client.WebSocketClientProtocol):
             self._keep_alive.stop()
 
         yield from super().close_connection(force=force)
-
-
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -661,12 +661,14 @@ class DiscordVoiceWebSocket(websockets.client.WebSocketClientProtocol):
                 self._dispatch('voice_speaking_state', self._connection, user, data['speaking'])
 
                 # Used for continuous streams
-                for f in self._connection._speaking_listeners:
-                    f(user, data['speaking'])
+                for uid in self._connection._speaking_listeners:
+                    if uid == user_id:
+                        for f in self._connection._speaking_listeners[uid]:
+                            f(user, data['speaking'])
 
                 if self._connection.channel.id not in self._connection._ssrc_lookup:
-                    self._connection._ssrc_lookup[self._connection.channel.id] = {}
-                self._connection._ssrc_lookup[self._connection.channel.id][ssrc] = user_id
+                    self._connection._ssrc_lookup[self._connection.guild.id] = {}
+                self._connection._ssrc_lookup[self._connection.guild.id][ssrc] = user_id
 
     @asyncio.coroutine
     def initial_connection(self, data):
--- a/discord/opus.py
+++ b/discord/opus.py
@@ -36,11 +36,17 @@ log = logging.getLogger(__name__)
 c_int_ptr = ctypes.POINTER(ctypes.c_int)
 c_int16_ptr = ctypes.POINTER(ctypes.c_int16)
 c_float_ptr = ctypes.POINTER(ctypes.c_float)
+c_ubyte_ptr = ctypes.POINTER(ctypes.c_ubyte)
 
 class EncoderStruct(ctypes.Structure):
     pass
 
+class DecoderStruct(ctypes.Structure):
+    pass
+
 EncoderStructPtr = ctypes.POINTER(EncoderStruct)
+
+DecoderStructPtr = ctypes.POINTER(DecoderStruct)
 
 # A list of exported functions.
 # The first argument is obviously the name.
@@ -52,7 +58,18 @@ exported_functions = [
     ('opus_encoder_create', [ctypes.c_int, ctypes.c_int, ctypes.c_int, c_int_ptr], EncoderStructPtr),
     ('opus_encode', [EncoderStructPtr, c_int16_ptr, ctypes.c_int, ctypes.c_char_p, ctypes.c_int32], ctypes.c_int32),
     ('opus_encoder_ctl', None, ctypes.c_int32),
-    ('opus_encoder_destroy', [EncoderStructPtr], None)
+    ('opus_encoder_destroy', [EncoderStructPtr], None),
+
+    ('opus_decoder_get_size', [ctypes.c_int], ctypes.c_int),
+    ('opus_decoder_create', [ctypes.c_int, ctypes.c_int, c_int_ptr], DecoderStructPtr),
+    ('opus_packet_get_bandwidth', [ctypes.c_char_p], ctypes.c_int),
+    ('opus_packet_get_nb_channels', [ctypes.c_char_p], ctypes.c_int),
+    ('opus_packet_get_nb_frames', [ctypes.c_char_p, ctypes.c_int], ctypes.c_int),
+    ('opus_packet_get_samples_per_frame', [ctypes.c_char_p, ctypes.c_int], ctypes.c_int),
+    ('opus_decoder_get_nb_samples', [DecoderStructPtr, ctypes.c_char_p, ctypes.c_int32], ctypes.c_int),
+    ('opus_decode', [DecoderStructPtr, ctypes.c_char_p, ctypes.c_int32, c_int16_ptr, ctypes.c_int, ctypes.c_int], ctypes.c_int),
+    ('opus_decode_float', [DecoderStructPtr, ctypes.c_char_p, ctypes.c_int32, c_float_ptr, ctypes.c_int, ctypes.c_int], ctypes.c_int),
+    ('opus_decoder_destroy', [DecoderStructPtr], None)
 ]
 
 def libopus_loader(name):
@@ -146,10 +163,11 @@ class OpusError(DiscordException):
         The error code returned.
     """
 
-    def __init__(self, code):
+    def __init__(self, code, verbose=True):
         self.code = code
         msg = _lib.opus_strerror(self.code).decode('utf-8')
-        log.info('"%s" has happened', msg)
+        if verbose:
+            log.info('"%s" has happened', msg)
         super().__init__(msg)
 
 class OpusNotLoaded(DiscordException):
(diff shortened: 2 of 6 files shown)


discord/voice_client.py
```diff
@@ -107,6 +110,14 @@ def __init__(self, state, timeout, channel):
         self._runner = None
         self._player = None
         self.encoder = opus.Encoder()
+        self.decoder = opus.Decoder()
```
Rapptz: I believe this to be incorrect.

There should be one decoder per SSRC.

discord/voice_client.py
```diff
@@ -291,8 +306,188 @@ def is_connected(self):
...
+            raise ClientException('no event found named {}'.format(coro.__name__))
+
+        self._events[coro.__name__] = coro
+        log.info('%s has successfully been registered as an event', coro.__name__)
+        return coro
+
     # audio related
 
+    @asyncio.coroutine
+    def get_voice_data(self, user, timeout=400):
```
Rapptz: `timeout` parameters are typically in *seconds* not *milliseconds*.

discord/voice_client.py
```diff
@@ -291,8 +306,188 @@ def is_connected(self):
...
+        """|coro|
+
+        Receive voice data from a user untill they are silent for
```
Rapptz: `untill` -> `until`

discord/voice_client.py
```diff
@@ -291,8 +306,188 @@ def is_connected(self):
...
+
+        Parameters
+        -----------
+        user: :class:`abc.User`
+            The user to listen for audio from.
+        timeout: Optional[int]
+            The time in ms of silence after which to stop listening.
+
+        Returns
+        -----------
```
Rapptz: The docstring for `Returns` requires a return type.

discord/voice_client.py
```diff
@@ -291,8 +306,188 @@ def is_connected(self):
...
+        The received voice data for the user passed in `bytes` form.
+
+        The data is stereo 48kHz audio in raw PCM. You will have to add your
+        own headers if you want to save it as a `.wav` file.
+        """
+        global buffer, last_packet
```
Rapptz: Why is this here?

discord/voice_client.py
```diff
@@ -291,8 +306,188 @@ def is_connected(self):
...
+        data = b''
```
Rapptz: This should probably be a `bytearray` instead of `bytes` since that is actually mutable.

discord/voice_client.py
```diff
@@ -291,8 +306,188 @@ def is_connected(self):
...
+            global buffer, last_packet
+            if s_user == user.id:
+                last_packet = time.time() * 1000
+
+                heapq.heappush(buffer, (timestamp, decoded_data))
+
+        self._pcm_listeners.append(listener)
+
+        while (time.time() * 1000) - last_packet < timeout or last_packet == 0:
+            yield from asyncio.sleep(0.1)  # Continue to fill up the buffer.
```
Rapptz: This entire code looks pretty racy.

Consider using a synchronisation primitive since you're expecting read and writing from two threads.

discord/voice_client.py
```diff
@@ -291,8 +306,188 @@ def is_connected(self):
...
+    def _voice_receive_loop(self):
+        """
+        Listener loop.
+        This loop will poll `self.socket` until there is some data,
+        then it will unpack the data, decode it, and send it to the
+        user-registered callback(s).
+        """
+
+        while self.socket is None:
+            time.sleep(0.1)
```
Rapptz: Consider using a `threading.Event` instead of a busy short sleep loop.

See my comment above about refactoring this design so it follows the one from `AudioPlayer`.

discord/voice_client.py
```diff
@@ -291,8 +306,188 @@ def is_connected(self):
...
+
+        while self.is_connected():
+            # Take the events from the WebSocket and pass them onto the callbacks
+            while self.ws._events_to_fire:
```
Rapptz: I do not like this at all.

discord/voice_client.py
```diff
@@ -291,8 +306,188 @@ def is_connected(self):
...
+                p = self.ws._events_to_fire.pop(0)
+                if p[0] in self._events and self._events[p[0]]:
+                    fut = asyncio.run_coroutine_threadsafe(self._events[p[0]](*p[1:]), self.loop)
+                    try:
+                        fut.result()
+                    except:
+                        import traceback
+                        traceback.print_exc()
```
Rapptz: Debug print statements shouldn't be here.

Consider using logging instead.

discord/gateway.py
```diff
@@ -553,6 +553,8 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.max_size = None
         self._keep_alive = None
+        self._ssrc_lookup = {}
+        self._events_to_fire = []
```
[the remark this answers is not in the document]
Bottersnike: The gateway doesn't have a ref. to the VoiceClient, so I stop the SSRCs on the Gateway because I have no way to move them to the VoiceClient. Would it be better to give the Gateway a ref. to VoiceClient?

Bottersnike:
I've changed the way it works. I added a `VoicePacket`, but should I `__slot__` it or would that not really do much? Oh, and is it safe to squash the commits down like I've done with the refactor commits without harming the review?

Rapptz:
Feel free to squash.

discord/voice_client.py
```diff
@@ -56,9 +58,89 @@
...
+        # Check the packet is valid
+        if check != cls._CHECK and check != cls._CHECK2:
+            fmt = 'packet has invalid check bytes: {}'
+            raise InvalidVoicePacket(fmt.format(packet))
+
+        # Decrypt data
+        nonce = bytearray(24)
+        nonce[:12] = header
+        box = nacl.secret.SecretBox(bytes(voice_client.secret_key))
+        buff = box.decrypt(bytes(buff), bytes(nonce))
```
jhgg: this can fail. you should handle this case (consider it a dropped packet, but maybe log a warning).

discord/voice_client.py
```diff
@@ -56,9 +58,89 @@
...
+
+    @classmethod
+    def unpack(cls, packet, voice_client):
+        if len(packet) < 13:
+            raise InvalidVoicePacket('packet is too small: {}'.format(packet))
+
+        # Unpack header
+        check, seq, ts, ssrc = struct.unpack_from(cls._FORMAT, packet)
+        header = packet[:12]
+        buff = packet[12:]
```
jhgg: the RTC packet can have the extension bits set - which may make the buff appear later in the packet. 

see: https://tools.ietf.org/html/rfc3550#section-5.3.1

discord/voice_client.py
```diff
@@ -56,9 +58,89 @@
...
+        The decoded PCM data.
+    """
+    # Struct format of the packet metadata
+    _FORMAT = '>HHII'
+    # Packets should start with b'\x80\x78' (32888 as a big endian ushort)
+    _CHECK = 32888
+    # Some packets will start with b'\x90\x78' (36984)
+    _CHECK2 = 36984
+
+    def __init__(self, sequence=None, timestamp=None, ssrc=None, buff=None, user=None):
```
jhgg: a lot of these will probably be constructed. maybe it'd be advantageous to make this a slotted class.

discord/voice_client.py
```diff
@@ -56,9 +58,89 @@
...
+        The id of the user (or source) of this voice packet.
+    buff : bytes
+        The Opus packet (as bytes).
+    user : Optional[:class:`abc.User`]
+        The user who the packet belongs to
+    pcm : Optional[bytes]
+        The decoded PCM data.
+    """
+    # Struct format of the packet metadata
+    _FORMAT = '>HHII'
```
jhgg: i'd store a  `struct.Struct` here instead. https://docs.python.org/3/library/struct.html#struct.Struct

discord/voice_client.py
```diff
@@ -56,9 +58,89 @@
...
+        buff = box.decrypt(bytes(buff), bytes(nonce))
+
+        # Packets starting with b'\x90' need the first 8 bytes ignored
+        if check == cls._CHECK2:
+            buff = buff[8:]
+
+        # Lookup the SSRC and then get the user
+        user_id = 0
+        if voice_client.channel.id in voice_client._ssrc_lookup:
+            if ssrc in voice_client._ssrc_lookup[voice_client.channel.id]:
```
jhgg: it may be worth holding onto packets that you can't get a valid user ID for until the speaking event arrives from the websocket.

discord/gateway.py
```diff
@@ -649,6 +651,22 @@ def received_message(self, msg):
...
+                user = self._connection._state.get_user(user_id)
+
+                self._dispatch('voice_speaking_state', self._connection, user, data['speaking'])
+
+                # Used for continuous streams
+                for f in self._connection._speaking_listeners:
+                    f(user, data['speaking'])
+
+                if self._connection.channel.id not in self._connection._ssrc_lookup:
+                    self._connection._ssrc_lookup[self._connection.channel.id] = {}
```
jhgg: you'd really want to key this by guild id - not channel id. ssrc's are shared across channels in a guild.

discord/opus.py
```diff
@@ -276,3 +294,93 @@ def encode(self, pcm, frame_size):
...
+            frame_size = frames * samples_per_frame
+            log.debug('detected frame_size {} ({} frames, {} samples per frame, {} channels)'.format(frame_size, frames, samples_per_frame, channels))
+
+        # note: python-opus also multiplies this value by
+        # ctypes.sizeof(decode_ctype) but that appears to be wrong
+        pcm_size = frame_size * self.channels
+        pcm = (decode_ctype * pcm_size)()
+        pcm_ptr = ctypes.cast(pcm, ctypes.POINTER(decode_ctype))
+
+        decode_fec = int(bool(decode_fec))
```
jhgg: int(bool(...)) seems a bit redundant here... 

```1 if decode_fec else 0```

File: discord/opus.py [hunk 19]
```diff
@@ -276,3 +294,93 @@ def encode(self, pcm, frame_size):
...
+    def _decode(self, data, frame_size, decode_fec, decode_func, decode_ctype, arr_type):
+        if frame_size is None:
+            frames = self._packet_get_nb_frames(data)
+            samples_per_frame = self._packet_get_samples_per_frame(data)
+            # note: channels could be different from self.channels
+            # this doesn't actually get used in frame_size, but we get
+            # the value for debugging
+            channels = self._packet_get_nb_channels(data)
+            frame_size = frames * samples_per_frame
+            log.debug('detected frame_size {} ({} frames, {} samples per frame, {} channels)'.format(frame_size, frames, samples_per_frame, channels))
```
jhgg: generally you want to leave formatting for the logger (that way if it's not log level debug, the formatting is skipped)

discord/voice_client.py
```diff
@@ -293,23 +393,180 @@ def is_connected(self):
...
+        The returned data is stereo 48kHz audio in raw PCM. You will have to
+        add your own headers if you want to save it as a `.wav` file.
+        """
+        buffer = []
+        last_packet = 0
+        user_id = user.id
+
+        def listener(vc, vp):
+            nonlocal buffer, user_id, last_packet
+            if user_id == (None if vp.user is None else vp.user.id):
```
jhgg: `if vp.user and user_id == vp.user.id`

discord/voice_client.py
```diff
@@ -107,6 +189,12 @@ def __init__(self, state, timeout, channel):
         self._runner = None
         self._player = None
         self.encoder = opus.Encoder()
+        self.decoders = {}
+        self._ssrc_lookup = {}
+        self._voice_receiver = None
+
+        self._pcm_listeners = []
+        self._speaking_listeners = []
```
jhgg: if you are going to split by userid/ssrc this should be a mapping of user id (or ssrc) to listener (or list of listeners). not a list of listeners that then filter (wasted cpu time)

jhgg: on an architectural level, you could have this be a mapping of user id (or ssrc) to an object that handles the decoding (it could also own the decoder as well). 

You should also add some logic such that if the ssrc for a user changes, the decoder is reset as well.

discord/voice_client.py
```diff
@@ -293,23 +393,180 @@ def is_connected(self):
...
+        SAMPLE_SIZE = 2
+
+        last_packet = 0
+        user_id = user.id
+        voice_buffer = []
+        last_speaking = 0
+
+        def listener(vc, vp):
+            nonlocal voice_buffer, user_id, last_packet
+            if user_id == (None if vp.user is None else vp.user.id):
```
jhgg: u should probably drop packets that are too old (ie part of segments that have already been decoded) - this handles late packets.

discord/voice_client.py
```diff
@@ -293,23 +393,180 @@ def is_connected(self):
...
+                vp = VoicePacket.unpack(packet, self)
+                log.debug('Decoded Voice Packet: {}'.format(vp))
+
+                if vp.ssrc not in self.decoders:
+                    self.decoders[vp.ssrc] = opus.Decoder()
+
+                # Dispatch event
+                self._state.dispatch('receive_opus', self, vp)
+
+                # TODO: re-order packets
```
jhgg: you're already doing this to some effect in the listener. you should just move this decode logic into there.

discord/voice_client.py
```diff
@@ -293,23 +393,180 @@ def is_connected(self):
...
+        """
+        buffer = []
+        last_packet = 0
+        user_id = user.id
+
+        def listener(vc, vp):
+            nonlocal buffer, user_id, last_packet
+            if user_id == (None if vp.user is None else vp.user.id):
+                last_packet = time.time()
+                heapq.heappush(buffer, (vp.timestamp, vp))
```
jhgg: if you're just going to `buffer.sort()` after all this, using a `heapq` is worthless and will just add more work.

Given how python's sorting is implemented, and that packets are probably (for the most part) ordered correctly - you should just call `buffer.sort()` at the end (which should be mostly linear thanks to timsort)

jhgg:
reviewing the code, i think this is kinda flawed. I think the proper pipeline should be something like:

```
socket loop => route packets into decoder based upon ssrc => decoder does buffering and outputs PCM audio after a delay (poor man's jitter buffer). 
```

We can use a static delay of a reasonable time since we don't need real time decoding.

```
## Socket loop:
decoders = defaultdict(Decoder)
while True:
    packet = receive_packet()
    rtp_packet = decode_rtp_packet(packet)
    # Packet isn't an rtp packet!
    if not rtp_packet:
        continue
    
    decoders[rtp_packet.ssrc].feed(rtp_packet)
```

Then something like:

```py
class Decoder:
    def __init__(self, window_delay=2):
        self.current_packets = []
        self.next_packets = []
        self.start_timestamp = None
        self.end_timestamp = None
        self.swap_timestamp = None
        self.window_delay = window_delay
        self.decoder = make_decoder()

    def feed(self, rtp_packet):
        cur_ts = rtp_packet.timestamp
        if self.end_timestamp is None:
            self.start_timestamp = cur_ts
            self.end_timestamp = cur_ts + self.window_delay
            call_in_the_future(cur_ts + (self.window_delay * 2), self.flush_packets)
            self.current_packets.append(rtp_packet)
            return

        # The packet is way too old. Drop it.
        if cur_ts < self.start_timestamp:
            return 

        # The packet is in the currrent window.
        if cur_ts <= self.end_timestamp:
            self.current_packets.append(rtp_packet)
        # The packet is in the next window.
        else:
            self.next_packets.append(rtp_packet)

    def flush_packets(self):
            # Process current PCM frames.
            self.emit_pcm_frames(self.process_packets(self.current_packets))
            # Advance us to the next window.
            self.start_timestamp = self.end_timestamp
            self.end_timestamp = self.start_timestamp + self.window_delay
            # we could probably just make this one loop. 
            self.current_packets = [c for c in self.next_packets if c.timestamp <= self.end_timestamp]
            self.next_packets = [c for c in self.next_packets if c.timestamp > self.end_timestamp]
            call_in_the_future(self.window_delay, self.flush_packets)

    def feed_decoder_packet_loss(self, pcm_frames, dropped_packets):
        while dropped_packets > 0:
            pcm_frames.append(self.decoder.decode(None))
            dropped_packets -= 1

    def process_chunks(self, chunks):
        # No audio received in this window. Emit empty frames.
        pcm_frames = []
        if not chunks:
            self.feed_decoder_packet_loss(pcm_frames, self.window_delay / FRAME_SIZE)
            return pcm_frames

        chunks.sort(key=attrgetter('timestamp'))
        # Emit silence between start of the window and the first audio packet.
        last_packet_timestamp = self.start_timestamp
        for packet in chunks:
            # Calculate the gap
            self.feed_decoder_packet_loss(pcm_frames, (packet.timestamp - last_packet_timestamp) / FRAME_SIZE)
            pcm_frames.append(self.decoder.decode(packet.data))
            last_packet_timestamp = packet.timestamp

        self.feed_decoder_packet_loss(pcm_frames, (self.end_timestamp - last_packet_timestamp) / FRAME_SIZE)
        return pcm_frames
```

I have no idea if this will actually work. A more "correct" approach may be to use the sequence number to calculate packet loss, and then feed silence into the decoder as packet loss as well. But I think that using the timestamp may be good enough (this might not be true in practice tho - you should test this). We can let opus do the silence for us, by tricking it into thinking there was packet loss when we're missing data. Which should be a lot more fool-proof than hopefully doing it ourselves.

The above code assumes that the packet's timestamp is in seconds. The packet's raw timestamp is definitely not in seconds though, and may need to be calculated based off of the sampling interval. We could probably kill a bunch of redundant calculation if we operated off of sampling interval instead of seconds for timestamp though. The delta between timestamps would be the number of samples missed - which we could feed to the decoder as packet loss. 

Anyways, the above code is more of a proof of concept of how I think the decoder should work, rather than a correct (in terms of runnable) implementation. There are probably some math errors in dealing with timestamps that need to be corrected. But the main concepts are there:
   
    - rolling fixed-duration window for packets.
    - dropping too old packets
    - ordering of packets for the decoder
    - using simulated packet loss to generate silence.

discord/voice_client.py
Bottersnike: This is handled by the piping function. The raw callbacks don't yet have a jitter buffer on them, but I'm planning on adding one.

discord/voice_client.py
Bottersnike: I believe I handle that when I chop off the first 8 bytes? (Hence the two check keys)

discord/voice_client.py
Bottersnike: I believe I handle that when I chop off the first 8 bytes? (Hence the two check keys)

discord/voice_client.py
Bottersnike: The listener and the events are separate. The listens are for when a user uses one of the pre-built functions to receive audio. The events send the packets as they arrive. This means the listeners have the luxury of being able to buffer the entire data before reordering it, whereas I will need a genuine jitter buffer for the events.

discord/voice_client.py
Bottersnike: (Line 510)

Crypti-x:
Works this code?

Bottersnike:
Closing because the packet structure has now changed and I'm not motivated enough to re-jig my packet handler to accommodate for the new(er) structure.

Issue #601, opened 2017-06-18
Is it possible to set name for Cog


eiva:
Can it be some method or property, not only class name?
Thanks!

Phxntxm:
No, it's set by the class name.

ButterCheezii:
I use this line inside the cog's `__init__` to change the name displayed in the help command:

```py
type(self).__name__ = 'Derpibooru'
```

Rapptz:
Keeping this one as a TODO.

ButterCheezii:
Be warned, my solution requires you to use `"quotes"` if you want to use the help command to list all the commands of a cog with spaces its name, since for all the bot knows that *is* the name of the cog

brandonsturgeon:
I'm trying to modify my bot's display name. Is this a relevant issue?

Vexs:
No.

But it's client.edit_profile on async and clientuser.edit on rewrite.

brtwrst:
> I use this line inside the cog's `__init__` to change the name displayed in the help command:
> 
> ```python
> type(self).__name__ = 'name here'
> ```

I have been using this to "rename" my cogs in the past but today i wanted to implement a `__local_check()` for one of my cogs and it turns out changing the `type(self).__name__` of a cog completely breaks `__local_check() and __global_check()` (probably some more).
When you have changed the name the checks just don't get applied at all.
Checks you write like this:
```python
def my_check():
    async def predicate(ctx):
        #do checks
        return True/False
    return commands.check(predicate)
```
and apply with
```python
@my_check()
```
work just fine - even if you changed the __name__.

Rapptz:
Should be fixed in https://github.com/Rapptz/discord.py/commit/caf3d17d4aa3ce45435bb96ea6d99317bfef8618

e.g. `class Cog(commands.Cog, name='My Name')`

PR #750, opened 2017-08-31
Add display_name kwarg to bot.add_cog

TheTrain2000:
This kwarg allows for the names of cogs to be changed upon addition of a cog in it's setup function.

This is a suggested resolution to #601

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -545,7 +545,7 @@ class BotBase(GroupMixin):
             If not specified, this is the class name.
         """
 
-        type(cog).__name__ = display_name if display_name is not None else type(cog).__name__
+        type(cog).__name__ = display_name if display_name else type(cog).__name__
         self.cogs[type(cog).__name__] = cog
 
         try:
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -521,7 +521,7 @@ class BotBase(GroupMixin):
 
     # cogs
 
-    def add_cog(self, cog):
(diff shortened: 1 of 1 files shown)


TheTrain2000:
I didn't add the typehint because I didn't see it used in other places within the cog and stuck with consistency.

That is not necessary. `if display_name` is completely valid to check whether or not `display_name` is `None`.

Anyways, Danny has stated that he intends to implement the naming of cogs in the form of meta classes. I'm leaving this PR open in the case Danny will use it to display information on his implementation of this.

MysterialPy:
Consistency? What if someone passes in an int for example? Or an empty list etc

And I don't see why you would check for None (especially since you didn't type hint).

`x if x` will fail on any Falsy, whereas `x if x is not None` will return True. Why would we want our name to be an empty string? Or an empty dict, set, list etc etc.

For example:
```
thing = ""

if thing:  # Will be False...

if thing is not None:  # Will be True
```

TheTrain2000:
If a user were to pass an empty string, it should be `False`. Cogs should not have empty strings for names.

Also, I've stated that this PR isn't going to be updated in my previous comment.

MysterialPy:
That's my point boi. Look at your code you check `x if x is not None`

TheTrain2000:
Oh. :blobthinkingfast: I'll push a commit in the case anyone uses this in the future, but this won't be the official implementation.

MysterialPy:
Haha 👍 Yeah I remember Danno talking about the meta class thing

PR #757, opened 2017-09-03
fix TypeError when passing invalid PermissionOverwrite to create_channel

sgtlaggy:
Just discovered this when someone asked for help in the Discord API channel. Basically shadowing builtins is bad.


h3kt1c0:
Looks good!

BluePhoenixGame:
This is from 2017 and is tying to merge into async. I even referenced it in a issue called `Close old pull requests?` which you should see above. I really doubt it would get merged, but yes. I agree with you, it looks pretty neat.

PR #762, opened 2017-09-06
Added member and user to Message and Context

orlnub123:
I know that you can get one of them from `author` but it would be nice to directly get them like you get `channel` or `guild`.

### DIFF
--- a/discord/ext/commands/context.py
+++ b/discord/ext/commands/context.py
@@ -202,13 +202,23 @@ class Context(discord.abc.Messageable):
 
     @discord.utils.cached_property
     def guild(self):
-        """Returns the guild associated with this context's command. None if not available."""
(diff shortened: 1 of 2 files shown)


Helehelehele:
`Context.author` (and `Message.author`) is already either a `Member` object or a `User` object.
This is directly from the [documentation](http://discordpy.readthedocs.io/en/rewrite/api.html#discord.Message.author): 

> **Message.author**
> A `Member` that sent the message. If channel is a private channel, then it is a `User` instead.

So you only need to do a type check if you only want to make sure you are only working with a `Member` object (or `User` object, but since `Member` already does what `User` can do I don't see the point in worrying about it).

Rapptz:
As Hele said, `Message.author` is already a `Member` or ` User`. The difference is pretty minor if you stick to the subset that is `User` operations.

PR #772, opened 2017-09-10
Add __eq__ to Message

BeatButton:
This makes it more intuitive to compare ```Message``` objects, which you may want to do when, for example, waiting for a reaction, or a message edit.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -185,6 +185,11 @@ class Message:
         self.reactions = [Reaction(message=self, data=d) for d in data.get('reactions', [])]
         self._update(channel, data)
 
+    def __eq__(self, other):
+        if not isinstance(other, Message):
+            return NotImplemented
+        return self.id == other.id
+
     def __repr__(self):
         return '<Message id={0.id} pinned={0.pinned} author={0.author!r}>'.format(self)
 


Rapptz:
The reason this was not implemented is because it raises the question on whether two messages are the same if their content is different. I'm not sure what side of the argument I'm on but based on the discussion it seems to be pretty divided.

BeatButton:
Oh, that makes sense as a contention, and I didn't even consider it. For my part I think whether the objects point to the same message is a better indication for equality, but you do you. It might be good to do something to avoid failing silently when people try to use it to compare things, though?

Rapptz:
It's not comparing "just" content. The argument is whether it should be `lhs.id == rhs.id` or `lhs.id == rhs.id and lhs.content == rhs.content`.

Gorialis:
Thanks for clearing up.

Personally, I still think this is the best way to implement this by default. If the argument is that a message being edited to have different content is not the same message, you could easily edit a message from content A to content B and then back to content A and it would be considered the "same" once again.

Additionally, doing this would refresh the embeds on the message (for example, if an OEmbed for a blog or news article changed) and it'd still be considered "the same message" even though the embed it points to could be entirely different.

neko404notfound:
+1 to this.

If two messages have the same snowflake, we assume them to be the same message. If they have content which is cached, and that is not up to date, then surely this is irrelevant, since requesting an up-to-date version of the object from the API will result in an equal result anyway?

Should we not identify the message by the snowflake alone? I can see that there is also information such as the reactions that have been added that are stored in the Message class. These are not automatically updated as far as I am aware. With this logic, if `reactions` are not the same between two `Message` objects, but the content and snowflake is the same, are they equal, or in-equal?

Personally, I think comparison by ID is a sensible thing to do. The user just then needs to use their common sense as to when to use the implementation. That is just my personal thoughts though.

I would be interested to hear any reasoning against this, or for this :)

PR #773, opened 2017-09-10
Separate arg parsing errors from conversion errors

### LINKED ISSUE #764: [Enhancement] Converters raise meaningful exceptions when they fail
Right now, when an argument conversion fails, `commands.BadArgument` is raised with a message as its first argument, something like "Object {arg} not found".

If I want a bot to tell the user they passed in a bad argument, my only option (more or less) is to send them that message. Unfortunately, this would mean I'd also send them messages like "Unexpected quote mark in non-quoted string" when argument parsing fails, since the parser raises the same exception.

I suggest that the parsing errors and conversion errors be separate classes (derived from `commands.BadArgument`), so that bots have the option to give more concise responses to users who pass an argument in which isn't converted (or parsed) correctly, whilst not making any breaking changes.

Tobotimus:
### Motivation
To allow more information about command argument errors to be obtained from their exceptions.

### Implementation
Two new Exception classes have been added, `commands.ConversionFailure` and `commands.ArgumentParseFailure`. Both derive from `commands.BadArgument`. Where `BadArgument` used to be raised when an argument conversion fails (i.e. in a converter function, method or `commands.Converter` instance), a `ConversionFailure` is raised. Likewise, where `BadArgument` was raised when syntax errors were encountered whilst parsing command arguments, an `ArgumentParseFailure` is raised.

`commands.ConversionFailure` has some new attributes which can help extrapolate information about where it occured, which are:
 - `argument`: the string value of the command argument whose conversion failed
 - `converter`: the *instance* of the object used to do the conversion.

The `commands.BadArgument` exception has been unchanged, the new classes have simply been derived from this class. In this way, we don't make any breaking changes.

Resolves #764


Tobotimus:
This is still ready for review, are there any plans for this to be resolved?

khazhyk:
How does this handle function style converters? Converters can also be any `str -> * ` function, which is pretty common. (e.g., `int`)

Tobotimus:
That's a good point. It looks as though I'll need to update the `bool` converter, since it raises `BadArgument`.

As for converters which are simply callables, their exceptions propogate out of `Command.do_conversion` and are caught in `Command.transform` to raise `BadArgument`. I'll need to update this too.

Because not all converters are subclasses of `commands.Converter`, I'm going to make a slight change so that `ConversionFailure` takes the converter instance instead of the converter class.

Tobotimus:
Gonna give this a bump, since it is ready for review.

Tobotimus:
This is pretty stale, especially after 2321ae8d9766779cef9baa7cc299e72a2ac88141

PR #779, opened 2017-09-14
Made the self.loop part look better

kyb3r:


khazhyk:
Hacktoberfest isn't until two weeks from now

verixx:
it is now

PR #803, opened 2017-09-24
Changed type(self.id) to int

gamingfail123:
Wtf, why is User.id a String? D: O.o

### DIFF
--- a/discord/user.py
+++ b/discord/user.py
@@ -48,7 +48,7 @@ class User:
     -----------
     name : str
         The user's username.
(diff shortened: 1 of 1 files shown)


Bottersnike:
Because that's how the whole of the async branch works. Switch to rewrite for ints.

Bottersnike:
Oh, and that PR would likely break most of the code because d.py expects a string at all times.

Rapptz:
Don't do garbage pull requests.

PR #805, opened 2017-09-24
Add optional param `invoke_with_checks` to invoke

MysterialPy:
Added an optional kwarg to `Context.invoke` which allows the command being invoked to verify checks. This represents a more natural feel to invoking a command like a standard user. Defaults to False.


Rapptz:
This probably won't happen.

Not like this anyway.

People also want the converters to be called. Doesn't end there either -- maybe they want the before and after hooks to be called too.

So `Context.invoke` bypasses everything and doesn't pick and choose what it thinks it's worth calling.

PR #812, opened 2017-09-27
Update `Cooldown` and `Command`

MysterialPy:
**`Cooldown`**
Split `is_rate_limited()` and added `update_rate_limit()`.

**`Command`**
Added `is_on_cooldown()`.

The addition of `is_on_cooldown` in line with the changes to `Cooldown` allow us to check the cooldown status of a command without invoking the command or updating the bucket in a way which would cause the command to be placed on cooldown.

### DIFF
--- a/discord/ext/commands/cooldowns.py
+++ b/discord/ext/commands/cooldowns.py
@@ -48,15 +48,13 @@ class Cooldown:
         if not isinstance(self.type, BucketType):
             raise TypeError('Cooldown type must be a BucketType')
 
-    def is_rate_limited(self, current=None):
-        if not current:
-            current = time.time()
+    def is_rate_limited(self):
+        current = time.time()
+        tokens = self._tokens
 
-        # It's fine to update here, as long as we never create a new window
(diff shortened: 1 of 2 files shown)


discord/ext/commands/cooldowns.py
```diff
@@ -48,21 +48,26 @@ def __init__(self, rate, per, type):
...
             raise TypeError('Cooldown type must be a BucketType')
 
-    def is_rate_limited(self):
+    def is_rate_limited(self, current=None):
+        if not current:
+            current = time.time()
+
+        # It's fine to update here, as long as we never create a new window
+        # so check if our window has passed and we can refresh our tokens
+        if current > self._window + self.per:
```
Rapptz: Since this a public API that apparently takes a `current` parameter, updating is actually not a good idea.

Consider: Someone passing a time.time() 3 minutes in the future.

It would probably be best to just move this updating in the `update_rate_limit` call and use a local `tokens` variable here that defaults to `self._token`.

Rapptz:
Rebased.

PR #828, opened 2017-10-03
Ran the docs through an online grammar checker

yourWaifu:
The commits are a bit misleading. That's my fault. I didn't use spell check, only checked for grammar.


Rapptz:
Rebased.

PR #837, opened 2017-10-05
Updated the link to Discord4J

LiamDevJS:
The one currently in the readme is no longer maintained


Rapptz:
Thanks but in the other version of the lib I just removed all the related projects since it's kind of annoying to maintain.

PR #839, opened 2017-10-07
Changed new_member.py - no default channel

NCPlayz:
This example now sends the message to the member from `on_member_join`, since `default_channel` was removed recently.


Chrozayis:
This defeats the purpose of having a bot  announce entrances
A purpose that I admit is redundant now that Discord can do that too

All the same I'd prefer the example be updated with perhaps more complex code that achieves the same purpose, not equally simple code that achieves a slightly different purpose

PR #848, opened 2017-10-11
Fix AttributeError when passing ClientUser to TextChannel.permissions_for

CptCalcium:
When using `discord.Client.permissions_in` or `commands.Bot.permissions_in` on a `TextChannel` an AttributeError is raised: `'ClientUser' object has no attribute 'roles'`.

This is due to the `ClientUser` being passed instead of a `Member` object if `discord.Client.permissions_in` or `commands.Bot.permissions_in` is called.

This fix checks whether the passed member is a `ClientUser` in `TextChannel.permissions_for` and uses the `Member` recieved from `TextChannel.guild.me` instead if necessary.

### DIFF
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -23,6 +23,7 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 DEALINGS IN THE SOFTWARE.
 """
 
+from .user import ClientUser
 from .permissions import Permissions
 from .enums import ChannelType, try_enum
 from .mixins import Hashable
@@ -105,7 +106,11 @@ class TextChannel(discord.abc.Messageable, discord.abc.GuildChannel, Hashable):
         return self
 
     def permissions_for(self, member):
-        base = super().permissions_for(member)
+        # use Member instead of ClientUser
+        if isinstance(member, ClientUser):
+            base = super().permissions_for(self.guild.me)
+        else:
+            base = super().permissions_for(member)
 
         # text channels do not have voice related permissions
         denied = Permissions.voice()


CptCalcium:
Just realized that adding a check to `ClientUser.permissions_in` may be the smarter solution.
Adding 
```python
if isinstance(channel, TextChannel):
    return channel.permissions_for(channel.guild.me)
else:
    return channel.permissions_for(channel)
```

to `ClientUser.permissions_in` in `users.py` would achieve the same result.

Also apologies if this is a mistake on my part somehow, but the documentation states that it's supposed to be supported: http://discordpy.readthedocs.io/en/rewrite/api.html#discord.ClientUser.permissions_in

hedgehog1029:
An interesting point about this is that ClientUser, and Users in general shouldn't *really* have that method at all - discord user entities have no concepts of permissions, but in some cases discord.py throws fake permission values into the mix for compatibilty (e.g. it returns full standard permissions in direct message channels) for when you use permissions_for as pre-checking.

However, here you're trying to get a user's permissions in a guild channel, and the user will have a corresponding member object, so fake values would be confusing - hence the error. 

The method you're calling (`ClientUser.permissions_in`) is designed to be called on direct message channels, where it would return a fake value. Generally when working with permissions, you should always use `Guild.me`, which will return a Member object local to the specific server. In the case of direct message channels, you don't have a `Channel.guild` attribute, so you should pre-check the type of channel and act accordingly.

SunDwarf:
A better implementation of this would be to lookup the `User.id` in the members dict, then operate on that member instead. Would solve for both ClientUser and regular User.

PR #849, opened 2017-10-13
Show sha1 for development versions.

khazhyk:
This is handy as you can more easily tell which recent commit you are on vs. just commit count.

git describe --tags might also work well for dev versions but requires more tagging


Rapptz:
Rebased.

PR #855, opened 2017-10-17
Implement custom help category names

Gorialis:
This PR allows cogs to have custom category names when they appear in help.

For example:
```py
class MyClassName:
    __category_name = "Cool Commands"
    ...
```
produces default help like the following:
```
Description

Cool Commands:
  smile   :)
​No Category:
  help    Shows this message.

Type ?help command for more info on a command.
You can also type ?help category for more info on a category.
```

~~Main issue with this impl:~~
- ~~`?help category` no longer works on renamed categories (they still work on non-renamed); you must know the cog's actual name and it isn't displayed~~
    - ~~To search this would require a cache or an O(n) lookup every time the help command is used~~
    - ~~Multiple cogs sharing a category name would present even more problems~~

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -109,8 +109,15 @@ def _default_help_command(ctx, *commands : str):
         command = None
         if name in bot.cogs:
             command = bot.cogs[name]
+        elif name in bot.all_commands:
+            command = bot.all_commands[name]
         else:
-            command = bot.all_commands.get(name)
+            for cog in bot.cogs.values():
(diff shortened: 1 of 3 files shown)


Gorialis:
I've added in an O(n) lookup for category names now (with the assembly of the command list being slow enough anyway this shouldn't make too much difference).
This will still break if multiple cogs use the exact same custom name, but this also breaks if you load two cogs with the same class name anyway, and attempting to prevent that is probably more effort than it's worth.

NCPlayz:
Nice! Was waiting for this to happen 😃

PR #856, opened 2017-10-17
Embed.set_image/author/thumbnail don't set None

lethargilistic:
Before this change Embed.set_image(None) would append the string 'None'
to the request, leading to an opaque 400 Bad Request error introduced by
discord.py, not the user. This change detects None entered for the URL
in set_image and set_thumbnail and does not append that string 'None'.
For consistency, it also changes set_author to ignore None as well,
although the Discord API accepts the string 'None' as the author in that
case and does not produce a 400 Error.

Example code:
```python
import discord

channel_id = 'DEFINE_ME'
token = 'DEFINE_ME'

client = discord.Client()
channel = discord.Object(id=channel_id)

def run():
    em = discord.Embed()

    #You can uncomment each of these to see the errors from None

    #em.set_image(url='https://placekitten.com/123/321/')
    em.set_image(url=None)
    #em.set_author(name=None)
    #em.set_thumbnail(url=None)

    async def send_em():
        await client.wait_until_ready()
        await client.send_message(channel, embed=em) 

    @client.event
    async def on_ready():
        print('online')

    client.loop.create_task(send_em())
    client.run(token)

run()
```

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -218,6 +218,8 @@ class Embed:
 
     def set_image(self, *, url):
         """Sets the image for the embed content.
+        
+        If you pass in None, your request's image will not change.
 
         This function returns the class instance to allow for fluent-style
         chaining.
@@ -228,9 +230,10 @@ class Embed:
             The source URL for the image. Only HTTP(S) is supported.
         """
 
-        self._image = {
-            'url': str(url)
-        }
+        if url:
+            self._image = {
+                'url': str(url)
+            }
 
         return self
 
@@ -252,6 +255,8 @@ class Embed:
     def set_thumbnail(self, *, url):
         """Sets the thumbnail for the embed content.
 
+        If you pass in None, your request's thumbnail will not change.
+
         This function returns the class instance to allow for fluent-style
         chaining.
 
@@ -261,9 +266,10 @@ class Embed:
             The source URL for the thumbnail. Only HTTP(S) is supported.
         """
 
-        self._thumbnail = {
-            'url': str(url)
-        }
+        if url:
+            self._thumbnail = {
+                'url': str(url)
+            }
 
         return self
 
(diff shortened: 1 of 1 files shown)


lethargilistic:
Of course, another option would be to raise an error if a user input None so that we could notify them of this mistake before the 400 happens, but that would be burdensome.

khazhyk:
Why not just not call set_*() ?

This makes these particular calls semantically different from the other set_* calls - e.g. set_footer(None) still sets the footer to "None"

lethargilistic:
If I missed some of the other set methods in my pull request, that's not a reason to disregard it. That just means my PR was incomplete and I could stand to add to it.

But, more on point, as a user, I was confused for a long time why I was getting bad request errors. The 400 from Discord does not provide enough information, and I would not expect a wrapper for an API to ever create a bad request, which is what it appears to do in that instance. Moreover, when this functionality does work (such as for author), it only works because of the coincidence that those Discord API fields accept any strings.

Whether Discord ignores these Nulls, as I would prefer, or if it raises errors, some change is necessary here.

khazhyk:
We could validate and require set_* only takes a str, as is the lib converts automatically which is convenient but does also result in some confusion when users are unaware their parameters are being converted. (All of these functions only accept str per their documentation, but there is a str() call underneath which silently accepts anything)

This still wouldn't solve the issue when a user passes an invalid URL, which I honestly believe is the same issue -> invalid input results in discord returning a 400. If the errors that are thrown now aren't sufficient, perhaps we can improve them.

As is the error seems sufficient to me, but perhaps not others. Although the mapping of Embed.set_image -> embed.image.url perhaps isn't clear enough?

lethargilistic:
Sorry, things took off for me.

The problem, as I see it, is that the user should never see a generic bad request error because the point of a wrapper is to abstract away the process of making a request with the Discord API. The fact that it even gets to that level is a mistake. The fact that some of these methods only work because of a coincidental cast to `str` is a mistake.

However, reflecting on this without the specific problem I was trying to solve in mind, I agree that a descriptive error would be better than silently changing nothing.

PR #857, opened 2017-10-18
Fix unquoted Route params leading to `add_reaction` failing on some emoji

Gorialis:
Usually raw should work anyway, but the hash emoji (#⃣) fails with a NotFound.
Since the official client quotes (and suffers no issues, although apparently this is an aiohttp issue with raws) we might as well do so as well, since it doesn't appear to break anything but fixes the hash emoji and any other potential emoji that suffer from this.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -53,7 +53,7 @@ class Route:
         self.method = method
         url = (self.BASE + self.path)
         if parameters:
-            self.url = url.format(**parameters)
+            self.url = url.format(**{k: _uriquote(v) for k, v in parameters.items()})
         else:
             self.url = url
 


Rapptz:
This should probably be done inside `HTTPClient.add_reaction` and friends?

Since you're missing the `remove_reaction` one as well and we don't know how the interaction is like with custom emoji.

Gorialis:
I put it in `add_reaction` since this is where if an emoji is custom or not is determined, and custom emotes shouldn't contain unicode (and therefore shouldn't suffer from this issue). I'll fix `remove_reaction` now, but I don't think moving it to HTTPClient provides any benefit unless something interfaces with it directly.

PR #860, opened 2017-10-19
is_avatar_animated should always return bool

khazhyk:


Rapptz:
Why not just cast it to `bool`?

khazhyk:
do you mean why not `return bool(avatar and startswith)`? idk

khazhyk:
changed

Rapptz:
✔️

PR #865, opened 2017-10-23
Update README.md

Modelmat:
Clarifying use of `async def`


Rapptz:
I don't think this clarifies anything.

If anything it's incorrect. Python 3.5 or higher can use `asyncio.coroutine` too.

Modelmat:
I just find it really confusing.

PR #870, opened 2017-10-26
[commands] allow expanded unicode numerals for id converter

khazhyk:
The other converters are discord specific formatting which
only accept ascii numerals


Rapptz:
muh performance

PR #872, opened 2017-10-27
on_member_join

ttlsx:
This example now sends the message to the member from on_member_join, since default_channel was removed recently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,14 @@ discord.py is an API wrapper for Discord written in Python.
 
 This was written to allow easier writing of bots or chat logs.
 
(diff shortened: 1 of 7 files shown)


NCPlayz:
You seem to be taking elements from legacy to rewrite, you might want to fix that, in case it wasn't intentional.

Pandentia:
@NCPlayz Honestly this just looks like someone trying to get a cheap hacktoberfest t-shirt. Should probably be marked invalid.

NCPlayz:
Yeah, you're probably right.

PR #882, opened 2017-10-31
Update documentation of joined_at (member.py)

Utrechter:
joined_at is their most recent join datetime, not their first one.


fourjr:
bump?

PR #886, opened 2017-11-02
Added Category Support.

ReedRGale:
I basically wanted category support, so I skimmed the API and changed what I think was needed. Humbly request someone look over this before its added anywhere, but so far it seems to do the trick.

Comprehensive changes below.

Edited ENUM to contain a category ChannelType; edited Channel to have a parent_id attribute; edited edit_channel in Client and in Http object to be able to change parent_id value, effectively giving the ability to add categories.

### DIFF
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -83,7 +83,7 @@ class Channel(Hashable):
 
     __slots__ = [ 'voice_members', 'name', 'id', 'server', 'topic', 'position',
                   'is_private', 'type', 'bitrate', 'user_limit',
-                  '_permission_overwrites' ]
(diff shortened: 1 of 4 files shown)


DoctorJew:
Doubt this will be given too much thought seeing as the rewrite version already supports categories. Don't think the async branch is being as actively updated.

ReedRGale:
Interesting. Wow that's a much more active branch! :0

I'll have to check that out.

DoctorJew:
If you plan on updating to the rewrite branch, I suggest reading about migrating [here](https://discordpy.readthedocs.io/en/rewrite/migrating.html).

PR #893, opened 2017-11-08
Edited to work with rewrite models.

davidschaeffer2:
I was rebuilding my music bot and noticed much of this code was using the old models. This is my current, working copy of the code that takes requests from users.

My queue-handling and next song playing is probably hacky, so any improvements I could make there would be appreciated.

### DIFF
--- a/examples/playlist.py
+++ b/examples/playlist.py
@@ -45,7 +45,7 @@ class YTDLSource(discord.PCMVolumeTransformer):
         return self(discord.FFmpegPCMAudio(filename, **ffmpeg_options), data=data)
 
 
-class YouTube:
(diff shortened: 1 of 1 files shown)


Painezor:
`class YouTube:` vs `bot.add_cog(Music(bot))`

NCPlayz:
I'm not sure if you've noticed but there is already an updated version of `playlist.py` under the name of `basic_voice.py` which can be found [here](https://github.com/Rapptz/discord.py/blob/rewrite/examples/basic_voice.py).

davidschaeffer2:
Hm, you're right. Thanks.

PR #894, opened 2017-11-10
New Game messages

Mippy:
"Listening to"
"Watching"


khazhyk:
worth nothing 1 is only valid with a streaming url, and 2 and 3 are only valid without

Mippy:
Could add a disclaimer

PR #897, opened 2017-11-11
Added ability to create categories and add channels under categories.

harigovindind:
Needed this feature for personal use, so thought I could share :)
1.Added "category" type to ChannelTypes in enums 
2.Edited create_channel() in http.py and client.py so that users can create channels under categories.

PS: First pull request on github, please feel free to point our mistakes/suggestions.

Thank you!


harigovindind:
Just noticed it was already proposed.

PR #908, opened 2017-11-17
Add PresenceType enum

Gorialis:
As Discord had added 2 new presences (and may add more in the future) I think it's now more justifiable to have an enum available for this so people don't have to remember numbers.

This uses IntEnum to avoid breaking changes when comparing with `int`s alone, and `try_enum` in case Discord adds any more presence types.


Rapptz:
Not the approach that's going to be taken.

In stash I have an implementation with type-level separations that are slotted to reduce memory usage as well.

PR #909, opened 2017-11-19
Prefix sha to fix leading zeros being stripped

khazhyk:
/dist.py:352: UserWarning: Normalizing '1.0.0a1402+0965847' to '1.0.0a1402+965847'


Rapptz:
Rebased

PR #922, opened 2017-11-24
playlist.py for rewrite.

EvieePy:
Since it's been asked for a bit lately. A slightly adapted version from async.

Tried to keep it as simple as I could. It's in 3.6, though if you need I'll do it in 3.5.

### DIFF
--- a/examples/playlist.py
+++ b/examples/playlist.py
@@ -146,6 +146,8 @@ class Music:
 
     async def __local_check(self, ctx):
         """A check which applies to all commands in Music."""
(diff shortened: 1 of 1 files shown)


zenrac:
Tested and approved except for some parts :

First you send typing during all your YTDLSource.from_url, it can easilly became an Infinite send typing. For example, when the entry found is a live, it will send typing for ever, it may should check if the entry isnt a stream before trying to DL/play it. 
Second, it's a small point but your async timeout returns an Exception each time the bot leaves the vc.  

Anyway good work !

PR #931, opened 2017-11-28
Update to 2017

kot83:


Rapptz:
There are many more files where this should be done (if at all). I believe the copyright notices are up to date in the other branch.

PR #933, opened 2017-11-28
Rewrite add Proxy support

ZamieltheHunter:
A simple oneline change in http.py will enable proxy support for most linux machines


SunDwarf:
You, uh, might wanna try that again. You're trying to merge rewrite into async.

PR #934, opened 2017-11-28
Added trust_env to the clientSession constructor to add support for m…

ZamieltheHunter:
…achines that set the http_proxy and https_proxy variables

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -224,7 +224,7 @@ class HTTPClient:
         yield from self.session.close()
 
     def recreate(self):
-        self.session = aiohttp.ClientSession(connector=self.connector, loop=self.loop)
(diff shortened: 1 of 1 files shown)


Gorialis:
This only exists in aiohttp 2.3+, which the async branch does not support.

khazhyk:
You can already pass the http_proxy parameters to the discord client constructor

http://discordpy.readthedocs.io/en/rewrite/api.html#discord.Client

Rapptz:
`HTTPClient.recreate` isn't even used (it's old cruft) so I'm not sure what your PR accomplished either.

Either way, both versions of the library have proxy support for their respective aiohttp versions.

PR #936, opened 2017-11-29
get_channel accepts strings not integers

toomuchio:
Misleading confused me for a tad, minor fix.

### DIFF
--- a/examples/background_task.py
+++ b/examples/background_task.py
@@ -17,7 +17,7 @@ class MyClient(discord.Client):
     async def my_background_task(self):
         await self.wait_until_ready()
         counter = 0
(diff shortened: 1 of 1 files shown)


Rapptz:
It is a coincidence at best.

Rapptz:
This is wrong. Rewrite uses `int` not `str`

toomuchio:
How long ago did this change I just tried this for the first time today and it wasn't working with ints.

Rapptz:
There are two major versions of the library in the wild at the moment.

- One is rewrite (aka v1.0.0) which uses `int` as IDs.
- The other is v0.16.x which uses `str`.

This PR is for the rewrite branch so it's incorrect.

PR #958, opened 2017-12-19
Add intersphinx

### LINKED ISSUE #577: Add intersphinx
Just a reminder to do intersphinx mapping for the new doc page.

Just add the extension and add:

```py
intersphinx_mapping = {'python': ('https://docs.python.org/3', None)}
```

Then cross reference the stuff in the stdlib.

Tobotimus:
Resolves #577 

Note that a lot of references in the docs *won't* be linked to the stdlib docs unless specifically given a role, e.g. `:obj:` or `:any:`. For some reason, this includes attribute types, but not parameter or return types.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -54,7 +54,7 @@ class Snowflake(metaclass=abc.ABCMeta):
 
     Attributes
     -----------
-    id: int
+    id: :obj:`int`
         The model's unique ID.
     """
     __slots__ = ()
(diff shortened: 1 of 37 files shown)


Rapptz:
> Then cross reference the stuff in the stdlib.

Tobotimus:
Most stdlib references I could find now link to the python docs. There are probably some I missed though.

This makes the PR's number of files changed a bit gross, but I guess it had to be done.

SunDwarf:
Why are a lot of these `:obj:` references? They're stdlib classes, they should be `:class:` references.

Tobotimus:
@SunDwarf `:obj:` is generic, so functionally it doesn't really make a difference, and I tend to use it for stdlib references since there are lots of them and it saves 2 characters. However, it won't be that hard to replace the existing `"obj:` references, so I'll go ahead and do that.

Tobotimus:
The requested changes have been made.

Rapptz:
Thanks a lot. I rebased this.

PR #959, opened 2017-12-20
Added Client.task decorator method and Client.tasks

AileenLumina:
This should make it a lot easier and more straightforward for people to create tasks that run in the background. Additionally, those tasks will be able to continue where they left off, as they receive a dict called `state` (is that a good name?) that contains the state those tasks had when the connection was interrupted. For that to work, the tasks written by users would just need to store variables in the `state` dict. The Client.task decorator takes care of possible connection issues; when there are any, it will wait for the Client to reconnect, and restart the tasks, passing them the state they had when the connection issue occured.

Tasks decorated with Client.task will be stored in the Client's `tasks` attribute.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -111,11 +111,16 @@ class Client:
         The websocket gateway the client is currently connected to. Could be None.
     loop
         The `event loop`_ that the client uses for HTTP requests and websocket operations.
+    tasks
+        A dict of `asyncio.Task`_ objects containing all tasks that were registered
+        using the :meth:`task` decorator.
     """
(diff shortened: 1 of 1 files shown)


Helehelehele:
Your `task` decorator is flawed. You can't use the named parameters; e.g. you can't do
```py
@client.task(name='foo')
async def bar(...):
    ...
```

I am not a fan of the design. Also in it's current state, PR does not have a convenient way to cancel tasks.

Rapptz:
Superseded by the `tasks` extension package.

PR #964, opened 2017-12-25
fetch_message: use history endpoint for user accts 

khazhyk:
- Try first to get message from local cache, save us a discord query
- If unable to use get_message, fallback to logs_from

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -930,12 +930,15 @@ class Messageable(metaclass=abc.ABCMeta):
         """
         return Typing(self)
 
+    @utils.deprecated('fetch_message_fast')
     async def fetch_message(self, id):
         """|coro|
 
         Retrieves a single :class:`~discord.Message` from the destination.
 
         This can only be used by bot accounts.
+
+        Prefer using :meth:`fetch_message_fast`.
 
         Parameters
         ------------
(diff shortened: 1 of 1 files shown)


Rapptz:
I actually don't like this because `get_message` could be used as a way to bypass the cache as of now when it could potentially be stale due to Discord dropping packets or some strange event.

khazhyk:
(Noticing also this doesn't filter specifically per channel.)

Perhaps re-introduce bot.messages or similar for trying to grab from local message cache, and users can have their own local-then-remote logic?

Rapptz:
If something like that were to be introduced it'd have to be something like `Client.get_cached_message` or something.

khazhyk:
updated this, find_message or whatever will be in separate PR

khazhyk:
Oh, I guess one issue with this is get_message raises NotFound, whereas logsfrom will return None... could fake that I guess

khazhyk:
None case is a bit weird
 - different exception raised for using history vs. not is strange
 - changing exception from NotFound to something else is breaking
 - splitting out to different method misses point of having single api bots & users can use

LookupError seems least bad to me?

discord/abc.py
```diff
@@ -844,8 +848,18 @@ async def fetch_message(self, id):
...
 
         channel = await self._get_channel()
-        data = await self._state.http.get_message(channel.id, id)
-        return self._state.create_message(channel=channel, data=data)
+
+        if use_history is None:
+            use_history = not self._state.is_bot
+
+        if use_history:
+            data = await self._state.http.logs_from(channel.id, limit=1, before=id + 1)
```
khazhyk, line 856: use around=id limit=1, since discord supports this now

khazhyk:
Updated to use around=id, this doesn't fix the LookupError vs. HTTPException disparity. Considering this is opt-in behavior anyways, I'm not that sad about the different error, but we could "fake" a httpexception if you think that's better. wdyt?

khazhyk:
Updated this:
 - Removed use_history=None handling, since we're defaulting all users to True
 - defaulted all users to True
 - added the .. versionadded:: 1.5 stuff?

discord/abc.py
```diff
@@ -937,6 +937,8 @@ async def fetch_message(self, id):
 
         This can only be used by bot accounts.
 
+        Prefer using fetch_message_fast.
```
Rapptz, line 940:
```suggestion
        Prefer using :meth:`fetch_message_fast`.
```

discord/abc.py
```diff
@@ -961,6 +963,40 @@ async def fetch_message(self, id):
         data = await self._state.http.get_message(channel.id, id)
         return self._state.create_message(channel=channel, data=data)
 
+    async def fetch_message_fast(self, message_id: int):
```
Rapptz, line 966:
```suggestion
    async def fetch_message_fast(self, id):
```

Rapptz (changes requested):
It might be worth using the rather rare `deprecated('fetch_message_fast')` decorator on top of regular `fetch_message`.

khazhyk:
Thanks, updated

Rapptz (approved):

PR #965, opened 2017-12-25
Add message.clean_substr for cleaning part of message.

khazhyk:
This is useful for any echoed user input where we do not want to
allow the user to trigger a ping by the bot. Simple example being
an echo command:

@command()
async def echo(ctx, *, msg):
  await ctx.send(ctx.message.clean_substr(msg))

clean_content in this case is not sufficient as there is otherwise no mapping from raw content to clean content for `msg`


Rapptz:
There is a `clean_content` converter already.

khazhyk:
Nice

PR #966, opened 2017-12-25
[commands] Fix MissingRequiredArgument param handling

khazhyk:
Documentation was incorrect, and we were throwing
away the param we were passing.

### DIFF
--- a/discord/ext/commands/errors.py
+++ b/discord/ext/commands/errors.py
@@ -72,11 +72,11 @@ class MissingRequiredArgument(UserInputError):
 
     Attributes
     -----------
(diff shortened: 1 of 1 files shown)


Rapptz:
I thought the plan was to convert this from `inspect.Parameter` -> `str` since people wanted to raise it themselves?

khazhyk:
I gave this some thought, and given that you can use `ctx.command.params['name']`, this seems not that hard to impl, and if anyone is doing completely custom parsing they may want to use their own exceptions anyways.

Rapptz:
I rebased this.

PR #967, opened 2017-12-25
[commands] add require_var_positional

khazhyk:
Allow command require_var_positional=True to require[commands] add require_var_positional

Add command require_var_positional=True to require at least one
argument to var_positional argument types.
Defaults to False, as previous behavior.


Rapptz:
How is this any different from `command(arg, *args)`?

khazhyk:
with require_var_positional=False args can be []. You can do something similar by doing `arg args` and then concatenating the list yourself, but this is more true to the intentions - i want a non-empty list of args

khazhyk:
RIP brutally closed

Rapptz:
Remake it then

PR #968, opened 2017-12-27
Allowed one argument to be used for custom prefixes (message)

JakeMakesStuff:
For instance, not all prefix calls need the bot so it is just a unused object in the definition.


Rapptz:
This is overhead that honestly serves little purpose. Unused parameters aren't going to kill you.

Issue #974, opened 2017-12-29
Aliases issue for when string in passed


Modelmat:
I made a mistake rather than parsing a list to `@commands.command(aliases=['name'])`, I sent `@commands.command(aliases='name').` This game me weird errors about commands like `n` already existing.

Looking at the [code](https://github.com/Rapptz/discord.py/blob/rewrite/discord/ext/commands/core.py#L745), you can see that it does not check to make sure that it is a list, so it instead iterates through the characters of `name`. 

See PR #975

Vexs:
Bit of an idiot fix isn't it? It's a trivial thing to avoid.

Modelmat:
Well, yes it is. It's still a good thing not to have errors like that making other errors. Even if it was made to raise another error (aliases are not a list), that would be better.

Rapptz:
Fairly trivial. 

Fixed.

PR #975, opened 2017-12-29
Aliases fix for when string in passed instead of a list.

Modelmat:
Fix Issue #974


Rapptz:
I don't like this implementation.

1. It's undocumented
2. It's in the wrong place.
3. The commit message doesn't follow the guidelines/precedent set by the previous commits.

PR #979, opened 2018-01-01
Change PartialReactionEmoji to PartialEmoji, add a PartialEmojiConverter

Gorialis:
This change does a couple of things:
- Renames `PartialReactionEmoji` to PartialEmoji` (breaking)
- Adds `animated` field to PartialEmoji
- Updates `__str__` to reflect animated-ness
- Adds `url` property to PartialEmoji, dynamic to `animated`, returns `None` for unicode emoji.
- Adds PartialEmojiConverter that converts a `<a?:name:id>` tag into a PartialEmoji

Tested on own install including docs, everything seems to work fine. This shouldn't be too severe of a breaking change since nothing generally directly references PartialReactionEmoji; the only places I've seen it referenced are within typehints of `on_raw_reaction_add` etc.

Weird Discord bug I discovered while writing this: messages Discord doesn't dispatch (obtained via `.history` or `.get_message` etc) always have the reactions with animated=False. Affects the returned payload directly, so not much I can do about this, although if Discord fixes this should automatically work as well.

### DIFF
--- a/discord/__init__.py
+++ b/discord/__init__.py
@@ -20,7 +20,7 @@ __version__ = '1.0.0a'
 from .client import Client, AppInfo
 from .user import User, ClientUser, Profile
 from .game import Game
-from .emoji import Emoji, PartialReactionEmoji
+from .emoji import Emoji, PartialEmoji
 from .channel import *
 from .guild import Guild
 from .relationship import Relationship
--- a/discord/emoji.py
+++ b/discord/emoji.py
@@ -30,8 +30,8 @@ from collections import namedtuple
 from . import utils
 from .mixins import Hashable
 
-class PartialReactionEmoji(namedtuple('PartialReactionEmoji', 'name id')):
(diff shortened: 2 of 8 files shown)


khazhyk:
```Weird Discord bug I discovered while writing this: messages Discord doesn't dispatch (obtained via .history or .get_message etc) always have the reactions with animated=False. Affects the returned payload directly, so not much I can do about this, although if Discord fixes this should automatically work as well.```

This isn't true in my testing, getting json using http.get_message, animated is set to true

Gorialis:
Testing further it seems that it only applies to some messages, and possibly only some reactions/emoji. Could be to do with emoji filesize, timespan, or Discord removing or reprocessing data after time. It isn't really clear, but for the purpose of this PR it shouldn't matter (if Discord decides to tell us a should-be animated reaction isn't animated, there's not much we can do about it)

Rapptz:
Rebased.

PR #981, opened 2018-01-02
Added __del__ method to client class.

mariohackandglitch:
Added ` __del__ `method to client class. This method calls self.close() to ensure the connection is properly closed if the client object is destroyed.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -163,6 +163,9 @@ class Client:
             VoiceClient.warn_nacl = False
             log.warning("PyNaCl is not installed, voice will NOT be supported")
 
+    def __del__(self):
+        self.close()
+
     # internals
 
     @asyncio.coroutine


NCPlayz:
What kind of scenario would you use this where you can't use Client.logout() or Client.close()

NCPlayz:
image
Also, you might've missed this piece of info.

mariohackandglitch:
This ensure the connection is closed in case the python interpreter suddenly exits. (However, looks like not always the interpreter calls all the destructors on exit tho 🤔).
I'm also not sure if you meant that I missed I can just call .close() myself, or that the function is a coroutine. If you meant the second case, then my first guess would be adding `await` to the close() method call, but I'm probably most likely wrong.

Gorialis:
`__del__` is not an coroutine and therefore cannot await `self.close`, nor should it, really. Please test your commits before making PRs.

mariohackandglitch:
Ok then...

Issue #946, opened 2017-12-08
Update README for rewrite branch.


deltamimikyu:
Can we update the README in the rewrite branch to reflect the correct installation instructions? Some new users will likely find this confusing otherwise, as it shows code snippets for the new library, but the installation will still lead the user to install version zero.

Vexs:
This is intentional, the rewrite docs are written as if that's the current branch.

deltamimikyu:
Does that not leave room for confusion?

Helehelehele:
In it's current state, rewrite branch is not meant for new users. It is still in development. The only reason it's published as a branch was so that core users could play around with it, and possibly "break it".

So yes, it does leave room for confusion. And it's meant to. "New users" is not the audience.

PR #983, opened 2018-01-02
Updated the install commands

NCPlayz:


Nekorooni:
It's intentional
#946

PR #989, opened 2018-01-04
Added an enum representing the valid activity types.

Espeonageon:
I have added this, as it does not seem to be defined anywhere else
in this API wrapper. I have grepped all the files to try and find
a match, but as yet can not find anything.

Currently, the existing values for the game activity types are not
very clear. There does not seem to be anything in the dpy
documentation regarding this (at least, not in the docstrings in
the source code).

The docstring in discord.Game seems to imply that there is only
the option of playing a game, or streaming a twitch URL. As of
4th Jan 2018, this is incorrect. There is also the ability to
be "listening to" something, or "watching" something. Both have
been tested and work running under a standard discord bot account.

I have tweaked the docstring in discord.Game to reflect this change.

### DIFF
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -212,6 +212,20 @@ class UserFlags(Enum):
     partner = 2
     hypesquad = 4
 
+
+class ActivityType(IntEnum):
+    """An enum that defines valid flags for the ``type`` parameter
+    used when creating an instance of ``discord.game.Game``.
+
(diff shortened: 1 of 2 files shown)


scragly:
something like this was attempted in #908

BlackSynder:
Danny (Rapptz) is planning on adding this, hence why it wasn't already implemented. You can use the new types using the `type` kwarg in discord.Game.

PR #996, opened 2018-01-06
Ensure gif avatar urls end in `.gif`

khazhyk:
This is a workaround for discord having trouble animating gifs if
the url does not end in exactly `.gif`. Since avatar_url is common
for thumbnails etc., adding this workaround here is handy, and
likely restores expected behavior (animated avatars animating).

### DIFF
--- a/discord/user.py
+++ b/discord/user.py
@@ -155,7 +155,10 @@ class BaseUser(_BaseUser):
             else:
                 format = static_format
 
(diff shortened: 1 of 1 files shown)


khazhyk:
Also tested with # param,  this seems to be a embed/image server issue, so needs to be passed in query

Rapptz:
I don't really like the idea of adding workarounds for something that the Discord client itself should fix. I'll merge this one regardless though.

PR #1009, opened 2018-01-12
add get_all_roles to client.py

Ivoah:
This adds the method 'get_all_roles' which behaves similarly to 'get_all_channels' and 'get_all_members'


kittenswolf:
When would someone need this?

PR #1014, opened 2018-01-14
Add some commenting to the example code

IanGolsby:
Makes it a teensy bit easier for newcomers to see what exactly is going on

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -61,16 +61,19 @@ Quick Example
     import asyncio
 
     class MyClient(discord.Client):
+        # Tells you when the bot is connected to Discord and is ready to receive commands
         async def on_ready(self):
             print('Logged in as')
             print(self.user.name)
             print(self.user.id)
             print('------')
-
+        
+        # Runs whenever a message is sent in a server the bot is in
         async def on_message(self, message):
-            # don't respond to ourselves
+            # Don't respond to ourselves
             if message.author == self.user:
                 return
+            # Tells us how many of the last 100 messages in the channel were sent by the command user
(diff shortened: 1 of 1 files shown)


NCPlayz:
Everything is in the docs, so I don't really understand this

IanGolsby:
It's not like the most useful thing ever done to this readme file but it makes it easier for people who have trouble with docs (cause some people just don't like reading them) or people who are too lazy to click on the docs link. Also some people like to not have to look at the docs when they're trying to decide to get something, so this makes it easier for someone looking at the readability of discord.py

Helehelehele:
> cause some people just don't like reading them

And why do we care about them?

Vexs:
This all feels like 

`if a is b #checks if a is an instance of b`

TheTrain2000:
If you don't like reading docs, I doubt you're going to like reading basic comments. I feel that these comments are extremely unnecessary, in that the code explains itself fairly well.

`people who are too lazy to click on the docs link`

Also, if you're too lazy to click on links for the documentation, then you're going to have a hard time using this library.

Rapptz:
Thanks for the PR.

However I don't think I like these excessive comments since the code is pretty clear as is and the comments aren't really adding much. If someone is confused they can just check the documentation rather than having trivial comments littered across the example code.

Rapptz:
Thanks for the PR.

However I don't think I like these excessive comments since the code is pretty clear as is and the comments aren't really adding much. If someone is confused they can just check the documentation rather than having trivial comments littered across the example code.

PR #1035, opened 2018-01-24
Added handling for unicode quotes

b-hodges:
I've only added some special quotes, but more can be added to the dictionary easily

### DIFF
--- a/discord/ext/commands/view.py
+++ b/discord/ext/commands/view.py
@@ -108,13 +108,42 @@ class StringView:
 
 # Parser
 
+# map from opening quotes to closing quotes
+quotes = {
+    '"': '"',
+    "'": "'",
+    "«": "»",
+    "‘": "’",
+    "‚": "‛",
+    "“": "”",
+    "„": "‟",
+    "‹": "›",
+    "⹂": "⹂",
+    "「": "」",
+    "『": "』",
+    "〝": "〞",
+    "﹁": "﹂",
+    "﹃": "﹄",
+    "＂": "＂",
+    "＇": "＇",
+    "｢": "｣",
+    "«": "»",
+    "‹": "›",
+    "《": "》",
+    "〈": "〉",
+}
+all_quotes = set(quotes.keys()) | set(quotes.values())
+
 def quoted_word(view):
     current = view.current
 
     if current is None:
         return None
 
-    is_quoted = current == '"'
+    is_quoted = current in quotes
+    if is_quoted:
+        open_quote = current
+        close_quote = quotes[current]
     result = [] if is_quoted else [current]
 
     while not view.eof:
@@ -122,7 +151,7 @@ def quoted_word(view):
         if not current:
             if is_quoted:
                 # unexpected EOF
-                raise BadArgument('Expected closing "')
+                raise BadArgument('Expected closing {}'.format(close_quote))
             return ''.join(result)
 
         # currently we accept strings in the format of "hello world"
@@ -133,32 +162,32 @@ def quoted_word(view):
                 # string ends with \ and no character after it
                 if is_quoted:
                     # if we're quoted then we're expecting a closing quote
-                    raise BadArgument('Expected closing "')
+                    raise BadArgument('Expected closing {}'.format(close_quote))
                 # if we aren't then we just let it through
                 return ''.join(result)
 
-            if next_char == '"':
+            if is_quoted and next_char in (open_quote, close_quote):
                 # escaped quote
-                result.append('"')
+                result.append(next_char)
             else:
                 # different escape character, ignore it
                 view.undo()
                 result.append(current)
             continue
 
+        if not is_quoted and current in all_quotes:
+            # we aren't quoted
+            raise BadArgument('Unexpected quote mark in non-quoted string')
+
         # closing quote
-        if current == '"':
+        if is_quoted and current == close_quote:
             next_char = view.get()
             valid_eof = not next_char or next_char.isspace()
-            if is_quoted:
-                if not valid_eof:
-                    raise BadArgument('Expected space after closing quotation')
+            if not valid_eof:
+                raise BadArgument('Expected space after closing quotation')
 
-                # we're quoted so it's okay
-                return ''.join(result)
-            else:
-                # we aren't quoted
-                raise BadArgument('Unexpected quote mark in non-quoted string')
+            # we're quoted so it's okay
+            return ''.join(result)
 
         if current.isspace() and not is_quoted:
             # end of word found
--- a/discord/ext/commands/view.py
+++ b/discord/ext/commands/view.py
@@ -114,26 +114,7 @@ def quoted_word(view):
     if current is None:
         return None
 
-    # map from opening quotes to closing quotes
-    quotes = {
-        '"': '"',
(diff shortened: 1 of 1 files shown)


discord/ext/commands/view.py
```diff
@@ -114,15 +114,34 @@ def quoted_word(view):
     if current is None:
         return None
 
-    is_quoted = current == '"'
+    # map from opening quotes to closing quotes
+    quotes = {
```
Rapptz: This should probably be a global to avoid re-creating this *every* time.

discord/ext/commands/view.py
```diff
@@ -114,15 +114,34 @@ def quoted_word(view):
...
+        '〝': '〞',
+        '＂': '＂',
+    }
+
+    is_quoted = current in quotes
+    if is_quoted:
+        open_quote = current
+        close_quote = quotes[open_quote]
+    else:
+        open_quote = close_quote = ''
```
Rapptz: This breaks if the input is not actually quoted. Despite its name `quoted_word` takes an *optionally* quoted word.

discord/ext/commands/view.py
```diff
@@ -114,15 +114,34 @@ def quoted_word(view):
...
+        '"': '"',
+        '«': '»',
+        '‹': '›',
+        '“': '”',
+        '‘': '’',
+        '❛': '❜',
+        '❝': '❞',
+        '❮': '❯',
+        '〝': '〞',
+        '＂': '＂',
```
Rapptz: This list is missing a lot of other quotes.

Despite the previous comment by @mylbp2ps3 I do not want to support "multiple ending quotes" since at that point the overhead added is too much for little benefit. I want to optimise for the common case not for the exotic one.

mylbp2ps3:
Not sure how to make changes to this request, but these changes should fulfill your requests:

Added these lines to line 28
```
# Dictionary of all unicode characters marked as "Quoatation Mark=Yes", and some more.
# Keys are opening quotes, and values are lists with matching closing quotes.
quotes = {
    '"': '"',
    "'": "'",
    "«": "»",
    "‘": "’",
    "‚": "‛",
    "“": "”",
    "„": "‟",
    "‹": "›",
    "⹂": "⹂",
    "「": "」",
    "『": "』",
    "〝": "〞",
    "﹁": "﹂",
    "﹃": "﹄",
    "＂": "＂",
    "＇": "＇",
    "｢": "｣",
    "«": "»",
    "‹": "›",
    "《": "》",
    "〈": "〉",
}
```

and `quoted_word` looks like this:
```
def quoted_word(view):
    current = view.current

    if current is None:
        return None

    is_quoted = current in quotes.keys()
    if is_quoted:
        open_quote = current
        close_quote = quotes[open_quote]
    else:
        open_quote = close_quote = '"'

    result = [] if is_quoted else [current]

    while not view.eof:
        current = view.get()
        if not current:
            if is_quoted:
                # unexpected EOF
                raise BadArgument('Expected closing {}'.format(close_quote))
            return ''.join(result)

        # currently we accept strings in the format of "hello world"
        # to embed a quote inside the string you must escape it: "a \"world\""
        if current == '\\':
            next_char = view.get()
            if not next_char:
                # string ends with \ and no character after it
                if is_quoted:
                    # if we're quoted then we're expecting a closing quote
                    raise BadArgument(
                        'Expected closing {}'.format(close_quote))
                # if we aren't then we just let it through
                return ''.join(result)

            if next_char in (open_quote, close_quote):
                # escaped quote
                result.append(next_char)
            else:
                # different escape character, ignore it
                view.undo()
                result.append(current)
            continue

        # closing quote
        if current == close_quote:
            next_char = view.get()
            valid_eof = not next_char or next_char.isspace()
            if is_quoted:
                if not valid_eof:
                    raise BadArgument('Expected space after closing quotation')

                # we're quoted so it's okay
                return ''.join(result)
            else:
                # we aren't quoted
                raise BadArgument('Unexpected quote mark in non-quoted string')

        if current.isspace() and not is_quoted:
            # end of word found
            return ''.join(result)

        result.append(current)
```

b-hodges:
I've added changes and run a more thorough test suite based on quotes being optional

discord/ext/commands/view.py
```diff
@@ -108,21 +108,50 @@ def __repr__(self):
 
 # Parser
 
+# map from opening quotes to closing quotes
+quotes = {
```
Rapptz: This needs to be a private variable, e.g. `_quotes`. There is no need to expose this.

discord/ext/commands/view.py
```diff
@@ -108,21 +108,50 @@ def __repr__(self):
...
+    "﹃": "﹄",
+    "＂": "＂",
+    "＇": "＇",
+    "｢": "｣",
+    "«": "»",
+    "‹": "›",
+    "《": "》",
+    "〈": "〉",
+}
+all_quotes = set(quotes.keys()) | set(quotes.values())
```
Rapptz: Same concept here. Not actually sure why this is used though.

discord/ext/commands/view.py
```diff
@@ -108,21 +108,50 @@ def __repr__(self):
...
     current = view.current
 
     if current is None:
         return None
 
-    is_quoted = current == '"'
+    is_quoted = current in quotes
+    if is_quoted:
+        open_quote = current
+        close_quote = quotes[current]
```
Rapptz: I don't like this implementation at all.

You're essentially doing two lookups in the dictionary (once for `in` and a second for `quotes[current]`). Instead you should opt for `quotes.get(current)` which will do both at a time (it defaults to `None` is none is found which is false-y).

discord/ext/commands/view.py
b-hodges: all_quotes is used to speed up lookups for escaping both opening and closing quotes since otherwise it would have to iterate through quotes.values()

discord/ext/commands/view.py
b-hodges: I'll implement that

b-hodges: So, I changed the implementation, but I decided to run a quick test and it looks like using `in` and a lookup is actually faster than using `.get()`. Here's the script I used to test:

```
from timeit import timeit

setup = '''
quotes = {'"': '"', "'": "'", "«": "»", "‘": "’", "‚": "‛", "“": "”", "„": "‟", "‹": "›", "⹂": "⹂", "「": "」", "『": "』", "〝": "〞", "﹁": "﹂", "﹃": "﹄", "＂": "＂", "＇": "＇", "｢": "｣", "«": "»", "‹": "›", "《": "》", "〈": "〉"}

def get():
    a = quotes.get('"')
    b = quotes.get('a')

def in_():
    a = quotes['"'] if '"' in quotes else None
    b = quotes['a'] if 'a' in quotes else None
'''

print('Using get():', timeit('get()', setup=setup))
print('Using in {}:', timeit('in_()', setup=setup))
```

raistlin7447:
Anyone still working on this?  It's become a problem for users of my bot as well.

b-hodges:
Still waiting on feedback from @Rapptz on the latest changes

Rapptz:
Rebased this one.

PR #1045, opened 2018-01-29
Add equality check for discord.Message

Modelmat:
Implements simple check of id that are equal. Also adds the container in the documentation. Useful when comparing messages from Reactions.


Rapptz:
https://github.com/Rapptz/discord.py/pull/772

PR #1061, opened 2018-02-08
allow embed content on send_file

ARivottiC:
There is no reason this option is not available on send_file. Also, as documented [here](https://discordapp.com/developers/docs/resources/channel#create-message-using-attachments-within-embeds) it's quite useful.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -1173,7 +1173,7 @@ class Client:
         yield from self.http.send_typing(channel_id)
 
     @asyncio.coroutine
-    def send_file(self, destination, fp, *, filename=None, content=None, tts=False):
+    def send_file(self, destination, fp, *, filename=None, content=None, tts=False, embed=None):
         """|coro|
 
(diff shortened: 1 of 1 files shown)


Dannyiy:
It's send_file not send_embed... 🤔

gogurtenjoyer:
This is possible in the rewrite branch of the library:

```python
f = discord.File('/path/to/image.png', filename='D.png')
e = discord.Embed(description="hello")
e.set_image(url="attachment://D.png")
await ctx.send(file=f, embed=e)
```

ARivottiC:
@Dannyiy both send_message and send_file use the same endpoint in the API. As documented you should be able to send embed while uploading a file.

ARivottiC:
@gogurtenjoyer yes, that is what I was expecting.

BluePhoenixGame:
Shouldn't this be closed as async is not supported anymore?

PR #1063, opened 2018-02-10
Implement voice receive (again)

Bottersnike:
I've done it again, but this time I've actually implemented things like the RTP headers properly. I've also tried to take into account feedback, from Danny, b1nzy and Jake.

The functions added are documented in the `docs/api.rst` file, but an example usage could look something like this:
```py
TARGET = 161508165672763392
f = open('f.pcm', 'wb')
@client.event
async def on_voice_receive(vc, user, data):
    if user.id == TARGET:
        f.write(data)
```
and that would handle packet reordering, silence padding, etc.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -947,7 +947,7 @@ class Connectable(metaclass=abc.ABCMeta):
         raise NotImplementedError
 
     @asyncio.coroutine
-    def connect(self, *, timeout=60.0, reconnect=True):
+    def connect(self, *, timeout=60.0, pad_silence=True, reconnect=True):
         """|coro|
 
         Connects to voice and creates a :class:`VoiceClient` to establish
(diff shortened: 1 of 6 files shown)


bmintz:
When pad_silence is set to true, does that mean that the voice data is realtime, and when it's false, the voice data is cut when there's silence?

bmintz:
Also what type is the `data` parameter to `on_voice_receive`?

Bottersnike:
When set to true, silence will appear as a large number of `0x00`s preceding the next packet. When set to false, those aren't there. The `data` parameter is raw PCM data, 16 bit stereo LE 48kHz.

stevemulligan:
Is there a way to know when there are not more packets to process for a voice client?  on_voice_speaking_state is called immediately when the user is done talking and data continues to flow in via on_voice_receive until all packets have been sent.

Bottersnike:
As soon as the user stops speaking, all buffered packets are ordered and flushed. With the aim to keep the speaking state event as close to the actual as I can, I might add another event that fires once the buffer's finished flushing.

jtcressy:
Would anyone here happen to have a short cog example for using this? I cant seem to get _any_ voice data at all. the `on_voice_receive` event never even fires off. `on_voice_speaking_state` is the only event that gets called, and I cant get any pcm or opus data through that.

PR #1064, opened 2018-02-11
Allow builtins' unbound methods to be used as command converters

eltrhn:
Currently when, for instance, `arg: str.upper` is attempted in a command definition, a conditional in ext/commands/core.py raises `AttributeError: 'method_descriptor' object has no attribute '__module__'` at `if converter.__module__.startswith(...)`.

PR adds a check for `hasattr(converter, '__module__')` before this so the conditional can short-circuit instead of erroring.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -201,8 +201,13 @@ class Command:
         if converter is bool:
             return _convert_to_bool(argument)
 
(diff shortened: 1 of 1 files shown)


eltrhn:
Ahh, my bad. Recommitted+squashed (before I saw the edit, too!).

Rapptz:
lgtm, thanks.

PR #1067, opened 2018-02-11
Compatibility with websockets 4.0

BeatButton:
Making the method signature use `*args` and `**kwargs` means that the library can change pretty much however it wants and it won't break. Since all we do in `close_connection` is stop the keep-alive task, I'm pretty confident that we don't really care what the specific arguments are.


Rapptz:
Missing voice websocket as well.

BeatButton:
Whoops, that didn't come up when I searched for `close_connection` using github. Guess I should just use grep like a normal human.

PR #1081, opened 2018-02-16
Add consume_args option to commands.group

MageJohn:
When invoke_without_command is set to False (default), the arguments to the
callback function are parsed before the subcommand is parsed. If the group callback has
arguments, they will be consumed from the command and the subcommand
will be looked for in whatever is left. 

This adds a consume_args option, which when True (the default), keeps this 
functionality. When set to False, however, the arguments are parsed, 
but not consumed. This means that the callback can perform 
whatever setup or validation it likes on the arguments, and then the 
subcommand will still be called properly.

Here's a rough example of how this could be used:
```python
bot = commands.Bot(command_prefix='?')

@bot.group(pass_context=True, consume_args=False)
async def newgame(ctx, game, *challanged):  # gets all the args from a command 
    # All the games have similar semantics
    if ctx.invoked_subcommand is None:
        return await bot.say("Unknown game!")
    players = [ctx.message.author]
    for c in challenged:
        player = ctx.message.server.get_member_name(c)
        if player:
            players.append(player)
        else:
            return await bot.say("invalid players!")
    ctx.players = players

@newgame.command(pass_context=True)
async def tic_tac_toe(ctx):
    # set up game data
    # checking for the correct number of players can be done here

@newgame.command(pass_context=True)
async def ludo(ctx):
    # ditto
```

If a user types `?newgame tic_tac_toe myfriend` in discord, the `newgame` callback will validate the players, and the group invoker can still find the subcommand  because no arguments have been consumed. Without `consume_args=False` the `*challanged` would consume the entire message, and the invoker won't find the subcommand.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -605,18 +605,35 @@ class Group(GroupMixin, Command):
         the group callback will always be invoked first. This means
         that the checks and the parsing dictated by its parameters
         will be executed. Defaults to ``False``.
+
+    consume_args : bool
+        If the callback is called before the subcommand, indicates if
+        the parsed arguments should be consumed, or left for the
+        subcommand. If this is ``True``, then the parameters of the
+        callback will consume arguments from the command, and
+        whatever's left will be interpreted as the subcommand and its
+        arguments. If ``False``, the same arguments will be passed to
+        the callback, but they won't be consumed from the command.
+        This is useful if all the subcommands require the same setup,
+        which can be performed in the callback. Defaults to ``True``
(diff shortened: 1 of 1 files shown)


MageJohn:
Hey! I managed to completely muck up this PR. Sorry about that. I accidentally pressed Ctrl-Enter, and sent it early, then I panicked slightly and clicked Close Pull Request. Now I can't undo anything, and I can't remove the request. I'm going to finish writing the description, and if you see this maybe you could open the request again. Thanks!

PR #1107, opened 2018-02-28
Add method to search for cogs required by load_extension.

daegontaven:


Rapptz:
Why

daegontaven:
@Rapptz Makes it easier to get a list of cogs that bot developers can simply plug into load_extension if their plugins directory is already a package. :)

NCPlayz:
You don't necessarily need to add them one by one, usually people put all the cogs in a new directory within the bot's, and use a loop to list the files in there.

Vexs:
They do? Personally I've never really seen a reason to go much further beyond a list in a json file somewhere of cogs. Less chance of mistakes.

daegontaven:
@NCPlayz Yeah that's what I used to do until I needed to load plugins from another directory. A very good example use case if when you need to disable experimental versions of plugins. So you do `find_cogs(experimental-cogs)` and then pass them to `unload_extension()`.

This is just a convenience method that I used and I figured it would be helpful for any future users who might necessarily not want to deal with the edge cases that `pkgutil` has already solved.

PR #1115, opened 2018-03-04
Default invoke_without_command to True instead of False

Vexs:
The current implementation is at best kind of unexpected for how one would expect subcommands to work- if the second arg of a command is a subcommand, it's natural to assume that it will invoke the subcommand instead of being passed as an arg to the root command. The default option also seems to be a more specialty option, with less total usage by the average user. 

Furthermore, the naming `invoke_without_command` is... confusing at best. I'm still not completely sure what this means. This change should eliminate some minor confusion for new users.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -874,10 +874,10 @@ class Group(GroupMixin, Command):
         in case no subcommand was found. If this is ``False``, then
         the group callback will always be invoked first. This means
         that the checks and the parsing dictated by its parameters
-        will be executed. Defaults to ``False``.
+        will be executed. Defaults to ``True``.
     """
     def __init__(self, **attrs):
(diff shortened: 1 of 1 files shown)


EvieePy:
Personally I like this change.

Though I feel like if it were defaulted to `True`, meaning it would need to be explicitly set to `False` to get the current default behavior, the name would become even more confusing for people.

Rapptz:
The premise of this pull request is wrong and rooted in personal preference.

A quick search shows that [the default behaviour is used](https://github.com/search?l=Python&q=invoked_subcommand+await&type=Code&utf8=%E2%9C%93) more than the [non-default behaviour](https://github.com/search?l=Python&q=invoke_without_command+True+await&type=Code&utf8=%E2%9C%93).

About the only thing I can agree with here is changing the name of the parameter to something else but since that's not in this PR I'm going to close it.

Issue #945, opened 2017-12-07
Low level hooks for voice connections


Rapptz:
Some people want to hook and intercept the `VOICE_STATE_UPDATE` and `VOICE_SERVER_UPDATE` events for use in e.g. [Lavalink](https://github.com/Frederikam/Lavalink/blob/master/IMPLEMENTATION.md) so maybe look into that.

### Notes

- Maybe allow dispatching `on_voice_server_update` with the necessary data of the event payload?
- Have a `cls` parameter in `Connectable.connect` to allow for custom `VoiceClient` implementations that would work as an IoC kind of deal.

PR #1117, opened 2018-03-04
Make the shard websocket publicly available

tekulvw:
In the standard Client, the websocket is publicly available at `Client.ws`. However when you use `AutoShardedClient` you would have to use `Client._connection._get_websocket(guild_id)` in order to retrieve the websocket.

This PR adds a `get_websocket()` method to both `Client` and `AutoShardedClient` that will retrieve the correct websocket.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -130,7 +130,7 @@ class Client:
         self._connection.shard_count = self.shard_count
         self._closed = asyncio.Event(loop=self.loop)
         self._ready = asyncio.Event(loop=self.loop)
-        self._connection._get_websocket = lambda g: self.ws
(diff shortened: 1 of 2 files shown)


Rapptz:
This seems more like an argument to remove `Client.ws` to me.

What reason do you want the websocket to be exposed?

tekulvw:
Using a third party audio player requires us to have access to the websocket to initiate our own `VOICE_SERVER_UPDATE` event.

Specifically, I'd like access to the `discord.gateway.DiscordWebSocket.voice_state()` method.

Rapptz:
This isn't really how I want to solve #945 so I'm going to close this.

PR #1123, opened 2018-03-07
Add Spotify intergration colour

EvieePy:
Added property `colour` and alias `color` which returns the Spotify integration colour (#1db954).

Technically Discord uses both (#1cb050 and #1db954) but it appears #1db954 is an official Spotify colour.

### DIFF
--- a/discord/activity.py
+++ b/discord/activity.py
@@ -186,7 +186,7 @@ class Activity(_ActivityTag):
         except KeyError:
             return None
         else:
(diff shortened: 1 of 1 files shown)


EvieePy:
Included #1124 in this.

Rapptz:
I don't quite like this implementation, if it's just a named constant just make it a property instead of wasting a slot for it.

Rapptz:
Rebased.

PR #1128, opened 2018-03-09
Add condition raises and warns to add_cog

Gorialis:
In PR #640, I proposed raising on adding classes as cogs instead of instances to prevent minor errors from creating larger, harder to source exceptions and tracebacks.

It was closed a month later with no extended explanation or alternative countermeasure implemented. 
**It has now been 8 months, and there is still nothing to help mitigate problems like this.**

This PR extends the idea of the original PR, but adds extra countermeasures, summarising to this:

- `add_cog` raises when passed a class instead of an instance.
  - This prevents instance methods being loaded as if they were static or class methods, causing argument confusion (`ctx` being sent as `self`, etc)

- `add_cog` raises when passed a discord.Client instance
  - This prevents rare situations where one might accidentally load their own client as a cog, causing the client to re-register its own handlers (and therefore causing every handler to fire twice)

- `add_cog` logs a warning when a cog is loaded with no commands or listeners
  - This prevents possible unforeseen similar issues to the above by logging what exactly was loaded and informing the user that nonfunctionality may be due to their listeners or commands not being recognized for whatever reason.

I've tested these changes on my own bots and, with good reason, it impairs none of them, as I expect it to with any sensibly implemented bot.
Every edge case I can think of is due to bad implementation on the user end and can therefore be worked around by the user improving said implementation.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -28,6 +28,7 @@ import asyncio
 import discord
 import inspect
 import importlib
+import logging
 import sys
 import traceback
 import re
@@ -37,6 +38,8 @@ from .view import StringView
 from .context import Context
 from .errors import CommandNotFound, CommandError
 from .formatter import HelpFormatter
+
+log = logging.getLogger(__name__)
 
 def when_mentioned(bot, msg):
     """A callable that implements a command prefix equivalent to being mentioned.
@@ -540,7 +543,17 @@ class BotBase(GroupMixin):
         -----------
         cog
             The cog to register to the bot.
+
+        Raises
+        -------
+        :exc:`.ClientException`
+            The cog passed was a class or incompatible instance.
         """
+
+        if inspect.isclass(cog):
+            raise discord.ClientException('cog must be an instance, not a class')
+        elif isinstance(cog, discord.Client):
+            raise discord.ClientException('cog cannot inherit from discord.Client')
 
         self.cogs[type(cog).__name__] = cog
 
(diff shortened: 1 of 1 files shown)


Rapptz:
This PR is way too opinionated to be merged.

1. Raising when a class is passed instead of an instance is typically okay, I have no strong opinion against this one.
2. Doing `self.add_cog(self)` has been done before and is not something I want to disallow just because some people find it weird or strange. Consider the following (currently valid) program that your PR will reject:

```py
class X(commands.Bot):
    def __init__(self):
        super().__init__(command_prefix='?')
        self.add_cog(self)

    @commands.command()
    def do(self, ctx, *, test: str):
        await ctx.send(test)

x = X()

@x.event
async def on_message(message):
    if message.author.bot:
        return
    await x.process_commands(message)
```
3. This noisy logging warning is way too opinionated. Consider the following valid program that might be a little weird but now raises a warning:

```py
class Baton:
    def __init__(self):
        self.money = 2000

    def take(self, x):
        if x >= self.money:
            ret = self.money
            self.money = 0
            return ret
        else:
            self.money -= x
            return x

    def put(self, x):
        self.money += x

class Bank:
    def __init__(self, bot):
        self.bot = bot

    @commands.command()
    async withdraw(self, ctx, *, amount: int):
        baton = self.bot.get_cog('Baton')
        taken = baton.take(amount)
        await ctx.send(f'withdrew {taken}')

bot.add_cog(Baton())
bot.add_cog(Bank(bot))
```

PR #1142, opened 2018-03-14
Fixed new_member.py example by switching server with member

ClementJ18:
It's not necessarily better but at least now it works.


BlackSynder:
I assume the intention wasn't to DM the user, but to send to the default channel. Seeing as that no longer exists, system_channel might be a better alternative.

ClementJ18:
It is more likely that the intention of the example was to show how the `on_member_join` event and events in general worked. Therefore, I prioritized an example that would work a 100% of the time over one that would be closer to the initial example since `system_channel` isn't always set.

PR #1154, opened 2018-03-20
Add Equality Check

Modelmat:
Probably might have some issues like that of #772
Also can be applied to `discord.Game`, `discord.Spotify`, and `discord.Streaming`
(Should Spotify be same song?)


Gorialis:
Loose equality check, this ignores everything that might be included as part of Rich Presence (current status, party size, etc)

Rapptz:
Most things have `__eq__` except `Activity` because there are way too many fields of interest to just pick one.

PR #1155, opened 2018-03-21
Force RTD to use python 3.6 and the latest environment build

tekulvw:
This fixes the "missing" sphinxcontrib library.


Rapptz:
Let's see how well this works I guess.

PR #1161, opened 2018-03-24
Improve basic_voice example

imayhaveborkedit:
Add stream option and command so we don't always have to download a file
Move repeated voice check code to before_invoke hook
Add typing indicators when making YTDLSources


NCPlayz:
I've noticed that all the commands that `ensure_voice` still have a thing in common:
```py
if ctx.voice_client.is_playing():
    ctx.voice_client.stop()
```
Surely you can put that in `ensure_voice` too, right?

Rapptz:
Rebased & Squashed.

PR #1164, opened 2018-03-25
Multi-worded Cog Name Support

NCPlayz:
This pull request is for multi-worded cogs, as the title says.

```py
class Misc:
    def __init__(self):
        self.cog_name = 'Not Actually Misc Stuff' # this is the variable that is used

    @commands.command()
    async def add(self, ctx, left: int, right: int):
        """Adds two numbers together."""
        await ctx.send(left + right)
```
Not only does it work for the `HelpFormatter`, but it works when doing `Bot.get_cog` since `cog_name` is used as the key when referring to the cog itself. If there is no custom cog name, it defaults to the class name.

```
An example bot to showcase the discord.ext.commands extension
module.
There are a number of utility commands being showcased here.

Not Actually Misc Stuff:
  add      Adds two numbers together.
Owner:
  eval     Evaluates code.
​No Category:
  help     Shows this message.

Type ?help command for more info on a command.
You can also type ?help category for more info on a category.
```

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -542,10 +542,10 @@ class BotBase(GroupMixin):
             The cog to register to the bot.
         """
 
-        if not hasattr(cog, '__cog_name'):
-            setattr(cog, '__cog_name', type(cog).__name__)
+        if not hasattr(cog, 'cog_name'):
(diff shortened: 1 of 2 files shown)


fourjr:
Might be better for `name ` to be a kwarg in bot.add_cog

PR #1177, opened 2018-03-30
added cooldown BucketType for members

ClementJ18:
as the title says, this adds a cooldown for member objects using an user id and a guild id. This means that the cooldown will only be specific to a member (so only on cooldown for that user in that guild). This means they are free to use the command in another guild. This can be useful for commands that change guild specific user data.

### DIFF
--- a/discord/ext/commands/cooldowns.py
+++ b/discord/ext/commands/cooldowns.py
@@ -31,8 +31,9 @@ __all__ = ['BucketType', 'Cooldown', 'CooldownMapping']
 class BucketType(enum.Enum):
     default = 0
     user    = 1
-    guild  = 2
+    guild   = 2
     channel = 3
+    member  = 4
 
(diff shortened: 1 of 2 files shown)


discord/ext/commands/cooldowns.py
```diff
@@ -111,6 +112,8 @@ def _bucket_key(self, msg):
             return (msg.guild or msg.author).id
         elif bucket_type is BucketType.channel:
             return msg.channel.id
+        elif bucket_type is BucketType.member:
+            return ((msg.guild or msg.author).id, msg.author.id)
```
Rapptz: I would rather it be `((msg.guild and msg.guild.id), msg.author.id)` i.e. the first key is `None` if there's no guild.

ClementJ18:
I've edited the PR to fit the review.

PR #1189, opened 2018-04-05
Clarified a detail about .subcommand_passed

anchpop:
Specifically, it doesn't work on a context retrieved through `bot.get_context(msg)`.


Rapptz:
`get_context` doesn't exist here so...

PR #1190, opened 2018-04-05
Edited to clarify detail about subcommand_passed

anchpop:
I already did this for the async branch, forgot to do it here


Rapptz:
I don't really like the way this was done since there is more to it than just this attribute that are set during `invoke`.

Vexs:
A little blurb on how the subcommand structure works internally might be nice, it's a little obtuse.

PR #1204, opened 2018-04-10
example: Fixed an issue where there is no default channel and causes a crash.

Nickdoge:
Made a PR before but noticed, it only works when a join channel is set.

Now, instead of ``guild.default_channel``, ``guilds.channels[1]`` would get the job done.

### DIFF
--- a/examples/new_member.py
+++ b/examples/new_member.py
@@ -9,7 +9,7 @@ class MyClient(discord.Client):
 
     async def on_member_join(self, member):
         guild = member.guild
-        await guild.default_channel.send('Welcome {0.mention} to {1.name}!'.format(member, guild))
+        await guild.system_channel.send('Welcome {0.mention} to {1.name}!'.format(member, guild))
 
 client = MyClient()
 client.run('token')
--- a/examples/new_member.py
+++ b/examples/new_member.py
@@ -9,7 +9,7 @@ class MyClient(discord.Client):
 
     async def on_member_join(self, member):
         guild = member.guild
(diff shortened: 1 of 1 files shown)


Vexs:
That's not guaranteed to be able to send a message, nor is it guaranteed to be a textchannel at all.

Really no good solution to this short of putting an entire bit in there to have a config command that can set a specific channel, which is well outside the scope of a basic example.

examples/new_member.py
```diff
@@ -9,7 +9,7 @@ class MyClient(discord.Client):
 
     async def on_member_join(self, member):
         guild = member.guild
-        await guild.system_channel.send('Welcome {0.mention} to {1.name}!'.format(member, guild))
+        await guild.channels[1].send('Welcome {0.mention} to {1.name}!'.format(member, guild))
```
Zenrac: It may return an IndexError, moreover it can also be a channel categorie or a voice channel.
You should check if there's an available TextChannel before, with something like 
`if sum(1 for channel in guild.channels if type(channel) == discord.TextChannel):`
And maybe check if the bot can send message
```
if sum(1 for channel in guild.channels if type(channel) == discord.TextChannel):
    for channel in guild.channels:
        if type(channel) == discord.TextChannel:
            if channel.permissions_for(guild.me).send_messages:
                return await channel.send('Welcome {0.mention} to {1.name}!'.format(member, guild))
```

Rapptz:
This example is probably better off being rewritten.

PR #1230, opened 2018-04-21
Forbid passing non-role objects to add/remove_roles

Modelmat:
This provides a useful error for when an id is passed instead, rather than an attribute error.

### DIFF
--- a/discord/member.py
+++ b/discord/member.py
@@ -501,6 +501,9 @@ class Member(discord.abc.Messageable, _BaseUser):
         HTTPException
             Adding roles failed.
         """
+        for role in roles:
+            if not isinstance(role, discord.Role):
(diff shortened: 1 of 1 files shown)


Vexs:
This ignores the fact that there are situations in which you might not want to pass an instance of discord.Role, not to mention that the lib raises similar errors in other locations.  It also fails to cover member.edit().

Modelmat:
How does this not cover member.edit()? Unless you mean this isn't covered by the member.edit() function.

Vexs:
??
Member.edit() allows you to replace roles entirely. It covers what replace_roles did.

Rapptz:
No thanks.

This disallows valid types that meet `abc.Snowflake`, which is the documented type accepted.

PR #1232, opened 2018-04-23
Added function for decorating commands to ignore

SimZor:
Adds a function being able to be decorated on functions.

This will remove the need to always use the if statement to check if subcommand is invoked.


Vexs:
This already exists as a [kwarg](http://discordpy.readthedocs.io/en/rewrite/ext/commands/api.html#discord.ext.commands.Group.invoke_without_command)

PR #1243, opened 2018-04-28
Makes bot.process_commands ignore commands from bots

Modelmat:
This follows recommended usage of discord bots, and removes the need to override
on_message for just this. If they don't want it to ignore the bot, just override it.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -917,6 +917,8 @@ class BotBase(GroupMixin):
 
     @asyncio.coroutine
     def on_message(self, message):
+        if message.author.bot:
+            return
         yield from self.process_commands(message)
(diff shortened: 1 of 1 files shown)


LyricLy:
I don't see a need for this. It's also misleading.

Modelmat:
Alternatively, we could add this to `bot.process_commands(message)`. This would completely remove this issue? (Although it would be a breaking change to those who have split it into sections to edit).
This could be controlled by the kwarg in the bot.

fourjr:
This should be a parameter in the `commands.Bot` init `ignore_bots` similar to `case_insensitive`, default to false

Rapptz:
I did this one myself since there were a lot of conflicts.

PR #1265, opened 2018-05-06
Removes redundant asyncio import

Modelmat:
Not much point importing asyncio when it's not being used, it was there for when `@asyncio.coroutine` was used previously.


ItWasAllIntended:
Does it reeeally matter? Maybe leave it in as a challenge for anyone with a sharp eye :-)

PR #1299, opened 2018-05-18
Made a role class mismatch show the appropriate error

JakeMakesStuff:
This shows a good error for lets say a nonetype or a user trying to put for example a guild into the role list.


Rapptz:
I do not do type checks. Instead, I opt for duck-typing approach to passing types. This e.g. is documented as `abc.Snowflake`.

PR #1313, opened 2018-05-27
Fix member.top_role not accounting for role hierarchy changes

CarlGroth:
Currently, changing the role hierarchy in server settings doesn't change the order of `member.roles`, leading to `member.top_role` being wrong.


Rapptz:
This should probably be fixed in another approach that I'll get to later.

PR #1336, opened 2018-06-07
[Rewrite] Help formatter: handle nick mentions

mikeshardmind:
Modifies the help formatter to handle nicknamed bot users for mentions in clean_prefix


Modelmat:
This doesn't actually use the nickname - `user.display_name` is always going to be the user's username not the nickname (as user's don't have nicknames).

PR #1338, opened 2018-06-08
Fix case insensitive command removal

### LINKED ISSUE #1217: case insensitive commands are not removed when unloading the cog
when using a bot with `case_insensitive=True` and unloading a cog that includes a command with a capital letter in it's name (for example setStatus), the command will not be removed when the cog is unloaded

if i rename the command function to `setstatus` the command is removed when unloading the cog and i can reload it again at runtime

Harmon758:
Fixes #1217


Rapptz:
Rebased.

PR #1343, opened 2018-06-11
Fix typo in Permissions.is_subset/is_superset

Hornwitser:
Fix the name for the other's type when raising TypeError being incorrectly written as `__class__name` instead of `__class__.__name__` in the is_subset and is_superset methods of the Permissions class.  This was introduced at the creation of these methods in 21c88cf.

Will rebase on request.


Rapptz:
Rebased.

PR #1357, opened 2018-06-18
Update background_task.py

datmellow:
A new example that does not use subclassing as it can lead to confusion for people. They think that subclassing is the only way to use a background task.


Rapptz:
All examples use subclassing to be consistent with the recommended approach of writing a bot.

PR #1364, opened 2018-06-20
Message.jump_to_url: correct the URL format

bmintz:
The current URL format in Discord Stable for Desktop and Android has replaced ?jump= with /,
and the old URL format no longer works in Discord Stable.
Updates Message.jump_to_url to reflect this.


Rapptz:
I'm going to do this one myself whenever it becomes more stable, which at the moment I'm not sure it is.

bmintz:
Then in my opinion, the current property should be removed, so as not to mislead people into using it.

Issue #712, opened 2017-08-12
`None` being present as a command_prefix raises an unhelpful error


Gorialis:
I'll make a PR for this later if it's not already addressed by the time I get around to it, but I'm putting this here as a reminder.

At the moment, if `None` is present when Bot attempts to parse commands, it produces this exception:
```py
  File "/usr/local/lib/python3.6/site-packages/discord/ext/commands/bot.py", line 902, in process_commands
    ctx = yield from self.get_context(message)
  File "/usr/local/lib/python3.6/site-packages/discord/ext/commands/bot.py", line 847, in get_context
    invoked_prefix = discord.utils.find(view.skip_string, prefix)
  File "/usr/local/lib/python3.6/site-packages/discord/utils.py", line 167, in find
    if predicate(element):
  File "/usr/local/lib/python3.6/site-packages/discord/ext/commands/view.py", line 62, in skip_string
    strlen = len(string)
TypeError: object of type 'NoneType' has no len()
```
This is undeniably an unhelpful exception. and proposed fixes are as follows:
1. When `None` is presented, simply ignore it. (this means that if `command_prefix` is `None` or any sequence of just `None`s (e.g. `[None]`) commands will not function at all.)
2. Raise whenever `None` is present (this is effectively the same as what is already happening, but can be used to have a more helpful error message)
3. Raise on `None` only if there are no alternatives present (e.g. `None` would raise but `[None, '?']` would not)
4. Ignore `None` as in 1 but use `warnings` to warn the user a la `utils.deprecated` (alternatively, you could use `logging` at WARN level)

Will update this issue if, and when, PR or fix is made.

Bluscream:
The third one looks best to me

henry232323:
I think it should just break? Does it need to have a check if I pass a socket object? Or a number? I think if your misusing it it should just fail when passing invalid things. 

If this really needs a fix the third option just ignores invalids silently which isn't good, it should just crash.

Gorialis:
Some people use databases or things like `dict.get` which are capable of returning `None`, meaning it's very potentially easy to make a callable prefix that returns `None` without users realizing it. This is the main reason this fix is proposed, at least if we choose to still raise on this, it should be a more useful error as in the second suggestion.

Rapptz:
Fixed.

PR #1370, opened 2018-06-22
Stop silently ignoring falsy command_prefixes

Hornwitser:
Changes the behaviour of handling iterable command_prefix types to not
silently ignore falsy prefixes and handle all iterable types in the same
manner.  Add special handling of a possible TypeError when the prefix is
a different type from what is expected.

Note that if Bot.get_prefix is overridden to return something other than
a list or string an unhelpful error message may still be produced.

---

This is in my opinion a much better way to handle #712, as silently ignoring falsy garbage passed to command_prefix while leaving the rest to still give the same unhelpful `TypeError: object of type '...' has no len()` is not good solution.  This change was originally motivated by me wanting to have an empty prefix in DM's only to find empty strings being silently ignored, but inconsistently so.

Performance wise this implementation is [faster than than the previous one in all cases](https://gist.github.com/Hornwitser/2fa2955b9ef7c1cc901f5a51fcb7f128).

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -25,6 +25,7 @@ DEALINGS IN THE SOFTWARE.
 """
 
 import asyncio
+import collections
 import discord
 import inspect
 import importlib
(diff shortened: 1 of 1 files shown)


Hornwitser:
I've updated the feature branch to use `message.content.startswith(tuple(prefix))` to quickly determine if the message matches any of the prefixes.  The benefit of this is that it's faster in determining that none of the prefixes match than using skip_string on each prefix, and it type checks the provided prefixes for more or less free.

PR #1382, opened 2018-06-30
Remove all asyncio.coroutine calls

bmintz:
One straggling @asyncio.coroutine was left behind from f25091efe1281aebe70189c61f9cac405b21a72f.

Also, @Client.async_event is no longer necessary, but I put its removal in a separate commit since it is a breaking change.


Rapptz:
Rebased.

PR #1403, opened 2018-07-09
Event for for WEBHOOKS_UPDATE

EndenDragon:
Similar to #451, I have a use to handle what happen when the server webhooks is changed.

### DIFF
--- a/discord/state.py
+++ b/discord/state.py
@@ -780,6 +780,10 @@ class ConnectionState:
         log.info('Processed a chunk for %s members in guild ID %s.', len(members), guild_id)
         self.process_listeners(ListenerType.chunk, guild, len(members))
 
+    def parse_webhooks_update(self, data):
+        channel = self.get_channel(int(data["channel_id"]))
+        self.dispatch("webhooks_update", channel)
+
     def parse_voice_state_update(self, data):
         guild = self._get_guild(utils._get_as_snowflake(data, 'guild_id'))
         channel_id = utils._get_as_snowflake(data, 'channel_id')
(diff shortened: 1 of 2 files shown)


docs/api.rst
```diff
@@ -380,6 +380,12 @@ to handle it, which defaults to print a traceback and ignoring the exception.
     :param last_pin: A ``datetime.datetime`` object representing when the latest message
                      was pinned or ``None`` if there are no pins.
 
+.. function:: on_webhooks_update(channel)
+
+    Called whenever a webhook is created, modified, or removed from a guild channel.
+
+    :param channel: The :class:`abc.GuildChannel` that had it's webhooks updated.
```
MaximusPX: Small typo, should be its.

discord/state.py
```diff
@@ -780,6 +780,10 @@ def parse_guild_members_chunk(self, data):
         log.info('Processed a chunk for %s members in guild ID %s.', len(members), guild_id)
         self.process_listeners(ListenerType.chunk, guild, len(members))
 
+    def parse_webhooks_update(self, data):
+        channel = self.get_channel(int(data["channel_id"]))
```
Rapptz: Two things:

1. I prefer single quotes over double quotes
2. Are you sure this can't be `None`? I tend to have an internal helper for those cases.

EndenDragon:
I've updated the code to reflect the recent code review.
In the case that channel can be `None`, the event would not be fired. Single quotes it is!
Thanks, @Rapptz!

bmintz:
I'm not sure how this would be useful if the only parameter to the event is `channel`. How can an event handler tell which webhook was modified? Was it removed, added, or edited?

EndenDragon:
@bmintz Discord Gateway only sends the channel of when the webhooks got updated. This event is implemented similar to Channel Pins Update event with the channel as the sole parameter.

PR #1404, opened 2018-07-09
add cog property to Command

MattIPv4:


Rapptz:
I get the point of this alias from a user perspective.

However, I don't think it should be done like this. If anything `Command.instance` could be renamed itself.

PR #1422, opened 2018-07-15
Allow cog name override

ed588:
I have added an additional parameter to bot.add_cog, allowing the name of the cog to be set to something other than the name of the cog class.

I think this will be useful in circumstances, for instance, where a developer wants to give all their cog classes names ending in `Cog`, but wants the "categories" in the help system to not have that suffix, without having to change any of the help functionality themselves.

Additionally, this adds extra configurability, which is generally useful in a library such as this.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -517,7 +517,7 @@ class BotBase(GroupMixin):
 
     # cogs
 
-    def add_cog(self, cog):
+    def add_cog(self, cog, name=None):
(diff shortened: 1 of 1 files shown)


uniflare:
This doesn't work since cog_name is not initialized outside the scope of the if statements.
Additionally, the cog_name in the cog_name property of commands are taken from the class type `__name__` rather than the dict key of the cog list.

I suggest:
https://github.com/Rapptz/discord.py/pull/1628

PR #1431, opened 2018-07-17
Added __int__ to colours

JakeMakesStuff:
This allows functions such as hash to work on the main colour.


Rapptz:
I don't see the benefit of this one over just doing `x.value`.

PR #1432, opened 2018-07-18
Fix behavior of everyone role mention

Vexs:
Prior to this change, attempting to do `everyone_role.mention` would result in the mention being formatted strangely when viewed in discord, appearing as `@@everyone`. 

This fixes this behavior, which appears as a kind of "gotya" to anyone trying to make a message filter ala #1429, or trying, for some reason, to list off all a member's or server's roles including everyone.

### DIFF
--- a/discord/role.py
+++ b/discord/role.py
@@ -160,7 +160,7 @@ class Role(Hashable):
     @property
     def mention(self):
         """Returns a string that allows you to mention a role."""
-        return '<@&%s>' % self.id
+        return '<@&%s>' % self.id if self.guild.id != self.id else '@everyone'
 
     @property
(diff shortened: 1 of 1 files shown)


khazhyk:
Technically the everyone role mention is different from the everyone mention. Since the everyone role is not mentionable (and cannot be), the former will never mention anyone. The `@@everyone` is actually `<@&default_role_id>`, same as the other role mentions. 

This creates a situation where role.mentionable is false, but role.mention will mention the role (everyone), which is perhaps odd.

Rapptz:
Don't like this type of special casing, I don't really think it helps much either.

PR #1436, opened 2018-07-19
Implement roles kwarg for guild.create_custom_emoji and emoji.edit

PikalaxALT:


diceroll123:
`roles: Optional[list[int]]`

Optional is unnecessary, having Optional means it can be None. You default to an empty list.

discord/emoji.py
```diff
@@ -243,7 +243,7 @@ def guild(self):
         name: str
             The new emoji name.
         roles: Optional[list[int]]
```
Rapptz: You probably meant ```Optional[List[:class:`Role`]]```

discord/emoji.py
```diff
@@ -243,7 +243,7 @@ def guild(self):
...
-            IDs of the roles for which this emoji is to be whitelisted.  Leave empty to make it unrestricted.
+            List of Roles for which this emoji is to be whitelisted.  Leave empty to make it unrestricted.
```
Rapptz: You probably want to reference the `Role` class here, and preface it with "A". I don't like the wording in general.

i.e.

```A :class:`list` or :class:`Roles` that can use this emoji. Leave empty to make it available to everyone.```

File: discord/guild.py [hunk 3]
```diff
@@ -1005,8 +1005,8 @@ def _create_channel(self, name, overwrites, channel_type, category=None, reason=
         image: bytes
             The *bytes-like* object representing the image data to use.
             Only JPG and PNG images are supported.
-        roles: Optional[list[int]]
-            IDs of the roles for which this emoji is to be whitelisted.  Leave empty to make it unrestricted.
+        roles: Optional[list[Role]]
```
Rapptz: Same comment as the above.

discord/guild.py
```diff
@@ -1005,8 +1005,8 @@ def _create_channel(self, name, overwrites, channel_type, category=None, reason=
...
+            List of Roles for which this emoji is to be whitelisted.  Leave empty to make it unrestricted.
```
Rapptz: Same nitpick.

PR #1437, opened 2018-07-19
Add support for splash(_url) to invites

DismissedGuy:
I noticed that you apparently couldn't get the splash from a server invite if you're not in it yourself

### DIFF
--- a/discord/invite.py
+++ b/discord/invite.py
@@ -109,8 +109,12 @@ class Invite(Hashable):
             guild = Object(id=guild_id)
             channel = Object(id=channel_id)
             guild.name = data['guild']['name']
+
             guild.splash = data['guild']['splash']
-            guild.splash_url = 'https://cdn.discordapp.com/splashes/{0.id}/{0.splash}.jpg?size=2048'.format(guild) if guild.splash else ''
+            guild.splash_url = ''
+            if guild.splash:
+              guild.splash_url = 'https://cdn.discordapp.com/splashes/{0.id}/{0.splash}.jpg?size=2048'.format(guild)
+
             channel.name = data['channel']['name']
 
         data['guild'] = guild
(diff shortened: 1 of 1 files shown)


ApfelTV:
will this thing here see activity soon?

ApfelTV:
> 2 lines of code
>> `The code is messy`
>>> makes sense

DismissedGuy:
My additions are set using the exact same way as `guild.name` is, so if this is apparently the wrong way to do it I don't see why guild.name is there.

ApfelTV:
What about an answer? I like the activity here.

Harmon758:
> Add support for splash(_url) to invites
> Undocumented change
> 
> 2 lines of code
> The code is messy
> 
> makes sense

It's undocumented as in this PR doesn't add these changes to the documentation, and code doesn't need to be numerous in lines to be untidy.

> My additions are set using the exact same way as guild.name is, so if this is apparently the wrong way to do it I don't see why guild.name is there.

`guild.name` isn't undocumented or set dynamically. Nor does it have any semblance to this PR's `guild.splash_url`.

ApfelTV:
dude are you kidding me

ApfelTV:
just accept this shit
2 friggn lines dude

DismissedGuy:
even though this might not be the correct way to do this, how else would you go about adding this feature? it's not a full guild object, so would it be worth to make a separate object just for this?

ApfelTV:
Hey if this code is bad and you say we could do this on another way, what about showing the code for the other way?

LittleEndu:
Showing how something is done completely misses the point of what PRs are used for. If something is not critical to the system no owner/collaborator/contributor is going to PR a fix for it. If you want to have something fixed, you are the one who needs to make code that actually works.

DismissedGuy:
Could you guys at least give me some suggestions on how else to achieve this? I feel like it isn't worth making a separate object for (didn't get any response on that comment, so just gonna assume it isn't). The 2nd line could be cleared up I guess, but I personally don't see another way.

DismissedGuy:
I've cleaned it up a bit, does it look okay?

PR #1441, opened 2018-07-20
update add_field

ed588:
`name` and `value` no longer need to be given as keyword args.

So, where before you may have had
```python
emb.add_field(name="Some name", value="A value")
```
you can now do:
```python
emb.add_field("Some name", "A value")
```
Reasons I think this is acceptable:
- It is still reasonably obvious from the code what it does (presumably the reason it is like this in the first place)
- Making required parameters keyword-only is inconvienient

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -342,7 +342,7 @@ class Embed:
         """
         return [EmbedProxy(d) for d in getattr(self, '_fields', [])]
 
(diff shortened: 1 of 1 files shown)


Rapptz:
I value consistency a lot more than this.

ed588:
Consistency with what? What if I made the same changes to all the embed functions?

PR #1445, opened 2018-07-22
Added message as a attribute

JakeMakesStuff:


Rapptz:
Either use `__str__` or `Exception.args` like all Python exceptions.

PR #1451, opened 2018-07-24
message: de-duplicate reaction type conversion

bmintz:
Removes some duplicate code in Message.{add,remove}_reaction.
The code in question converts the emoji object from Reaction, Emoji, str, or PartialEmoji
to a string form suitable for sending over the wire.


bmintz:
I don't know what happened to my changes, and why a former pull request's code is presented instead of the code I wrote.

PR #1464, opened 2018-07-28
Added option to hide all commands from a cog

ducky11423:
Added a keyword argument to `bot.add_cog` that allows all commands from a cog to be hidden. Additionally added a keyword argument to the command class which forces the command to be shown in help, even if the entire cog is hidden.

Both arguments default to False, keeping all behavior the same unless the arguments are supplied.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -538,8 +538,9 @@ class BotBase(GroupMixin):
             The cog to register to the bot.
         hidden: bool
             Whether or not all commands in this cog should be hidden
-            from the help command. Defaults to ``False``. Commands with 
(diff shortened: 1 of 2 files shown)


ClementJ18:
Is it not possible that instead of a new keyword to force commands to show, to have it that commands that contain `hidden=False` are shown even if the cog is hidden? That feels like the more natural way to make full use of the `hidden` kwarg.

ducky11423:
I did think about such an implementation, however I didn't think it was necessary. I've pushed commits which change it though.

Rapptz:
I feel like these types of things (bulk-adding properties, cog-specific names) are better off implemented as a cog metaclass that should be inherited.

PR #1465, opened 2018-07-29
Fixes #1462

### LINKED ISSUE #1462: Adding an optional 'user' field for Emoji
The API's [emoji object](https://discordapp.com/developers/docs/resources/emoji#emoji-object) contains a user field with user data of the emoji creator.

This field is only available if the `MANAGE_EMOJIS` permission is present. Both in [List Guild Emoji](https://discordapp.com/developers/docs/resources/emoji#emoji-object) and [Get Guild Emoji](/guilds/{guild.id}/emojis/{emoji.id})

It would be nice to have this in the python API as well.

```json
{
    "managed": false,
    "name": "whip",
    "roles": [],
    "user": {
        "username": "Nezune",
        "discriminator": "2269",
        "id": "162070610556616705",
        "avatar": "104be0527d9f897b34814b197b4b3140"
    },
    "require_colons": true,
    "animated": false,
    "id": "255459240037056529"
}
```

GenericError:
Adding an optional 'user' field for Emoji

### DIFF
--- a/discord/emoji.py
+++ b/discord/emoji.py
@@ -67,8 +67,10 @@ class Emoji(Hashable):
     roles : List[:class:`Role`]
         A list of :class:`Role` that is allowed to use this emoji. If roles is empty,
         the emoji is unrestricted.
+    user : class`User`
+        The user who created the emoji
     """
-    __slots__ = ["require_colons", "managed", "id", "name", "roles", 'server']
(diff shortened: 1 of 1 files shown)


Rapptz:
As far as I'm concerned, `Emoji.user` is only retrievable from an endpoint that isn't exposed in this library. Ergo, we fetch it from the websocket which does not have this information.

PR #1468, opened 2018-07-30
emoji.py: correct and deduplicate the url format

bmintz:
All emoji URLs now have ?v=1 at the end.


Rapptz:
This is just to force cache invalidation from an old issue that they have, it means nothing.

PR #1474, opened 2018-08-01
Add Permissions.priority_speaker

slice:
"Users with this permission have the ability to be more easily heard when talking. When activated,
the volume of others without this permission will be automatically lowered."

8th bit.

### DIFF
--- a/discord/permissions.py
+++ b/discord/permissions.py
@@ -133,7 +133,7 @@ class Permissions:
     def all(cls):
         """A factory method that creates a :class:`Permissions` with all
         permissions set to True."""
-        return cls(0b01111111111101111111110011111111)
+        return cls(0b01111111111101111111110111111111)
 
     @classmethod
     def all_channel(cls):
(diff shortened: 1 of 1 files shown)


Werseter:
Does such permission even exist in Discord's endpoint? Even if so, I'm curious as of, if at all, was it introduced.

jwshields:
@Werseter I just updated to the newest Canary build and I see it in my options.

Werseter:
Interesting addition it is then, I am checking the regular documentation of Discord now and still have not found a proper entry, but then I suppose I should be looking at a beta one.

Werseter:
Well, even though this correctly seems to implement the flag, as I've tested it about now, still things in Canary may change and may confuse potential users if introduced to the library before an official release of the feature. While not invalid, this should most probbably be suspended and introduced, with proper changes if needed, as this feature may get expanded somehow, later on.

Harmon758:
Relevant documentation:
https://github.com/discordapp/discord-api-docs/pull/638
https://github.com/discordapp/discord-api-docs/commit/5adf4cb1c68b86573b82b9dc8d473d7da0176c00

PR #1483, opened 2018-08-06
Fix for #1462 (Implement Emoji.user attr)

mental32:
Please review prior to merging.

This change implements a user attribute on the discord.Emoji object of a discord.User type.
(should it be a Member instance instead?)

### DIFF
--- a/discord/emoji.py
+++ b/discord/emoji.py
@@ -141,8 +141,10 @@ class Emoji(Hashable):
         If this emoji is managed by a Twitch integration.
     guild_id: :class:`int`
         The guild ID the emoji belongs to.
+    user Optional[:class:`User`]
+        The user that created the emoji, None if not avaliable.
     """
-    __slots__ = ('require_colons', 'animated', 'managed', 'id', 'name', '_roles', 'guild_id', '_state')
+    __slots__ = ('require_colons', 'animated', 'managed', 'id', 'name', '_roles', 'guild_id', '_state', 'user')
 
     def __init__(self, *, guild, state, data):
         self.guild_id = guild.id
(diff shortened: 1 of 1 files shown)


discord/emoji.py
```diff
@@ -141,8 +141,10 @@ class Emoji(Hashable):
         If this emoji is managed by a Twitch integration.
     guild_id: :class:`int`
         The guild ID the emoji belongs to.
+    user :class:`User`
```
Rapptz: Needs `Optional[...]`.

discord/emoji.py
```diff
@@ -141,8 +141,10 @@ class Emoji(Hashable):
...
+        The user that created the emoji.
```
Rapptz: Needs a wording update that mentions that it's not always available and the permission required for it to be received.

discord/emoji.py
```diff
@@ -156,6 +158,9 @@ def _from_data(self, emoji):
         self.name = emoji['name']
         self.animated = emoji.get('animated', False)
         self._roles = set(emoji.get('roles', []))
+        self.user = user = emoji.get('user', None)
+        if user is not None:
+            self.user = self._state.get_user(user['id']) or self._state.store_user(user)
```
Rapptz: There is no point in doing this.

It's already a `User` object so just convert it into a `User` object via `self._state.store_user` which does the retrieval anyway. I'm not sure if it's worth caching the object globally anyway.

If anything I feel like this should follow the general type promotion rules (`User` -> `Member`) so it's a bit easier to use but again I'm not really sold on the idea or against it.

mental32: Would you prefer Emoji.user to be a Member rather than User?

mental32:
@Rapptz will this see any activity?

mental32:
closing...see #1496

PR #1489, opened 2018-08-06
Futher Explanation.

DavidCaughell:
Hopefully this should help new users get started more quickly and without disrupting the discord help channel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,9 @@
 
 discord.py is an API wrapper for Discord written in Python.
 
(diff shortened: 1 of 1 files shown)


Rapptz:
Good intentions but I'd rather not do it in the README.

There's already a documentation page on how to create a bot in-house in the rewrite documentation. The tidbit about `Client.run` just feels a bit condescending to the reader when singled out considering how much more there is to the example.

PR #637, opened 2017-07-09
rewrite branch specific install instructions

Bluscream:

PR #1493, opened 2018-08-08
Update the installing documentation

Espequair:
The rewrite documentation has a problem currently, following it will install the old version, not the rewrite. This commit fixes this in part, but a second look would be appreciated


Modelmat:
See #637, #946 and #983.
Will be master branch once done developement,

Espequair:
Very well, thank you.

PR #1497, opened 2018-08-10
Add typings

bryanforbes:
I've added `pyi` files as siblings to all of the library files:

* I've tried to only type the public API. I may have missed something or provided too many methods. Feel free to indicate what should be added or removed.
* Any class that is typed as generic (iterators, Command, Group, etc.) has to inherit from `typing.Generic` in order to be subscriptable and not throw errors when using generic subscripts in python code.
* In the course of doing this work, I converted the entire core codebase over to typed Python and found a few bugs. I've included those bug fixes (calls.py, message.py, state.py, and user.py).
* While Python 3.5.x is still the oldest version the library uses, the typings require Python 3.6 for PEP-526.
* I've added an optional feature that will pull in `mypy_extensions` and `typing_extensions` for `TypedDict` and `Protocol` respectively. These will not be pulled in unless specified by the consumer.
* I have also tested these typings in a [library I wrote that uses discord.py](https://github.com/bryanforbes/botus_receptus) as well as a [bot](https://github.com/bryanforbes/Erasmus) I wrote.


discord/ext/commands/cooldowns.pyi
```diff
@@ -0,0 +1,41 @@
...
+
+from .context import Context
+
+from typing import Optional, TypeVar, Type
+
+class BucketType(enum.Enum):
+    default = 0
+    user = 0
+    guild = 2
+    channel = 2
```
slice: Typo here -- supposed to be `channel = 3`.

Rapptz:
I understand the usage and reasons for `.pyi` files existing.

However, I do not think they apply to this right? Shouldn't the goal be to apply these typings to the main functions inside the library instead of using stubs?

bryanforbes:
There are a few advantages to using `.pyi` files alongside modules:

* Keeping typings out of the main library gives a small startup performance gain.
* Typing the function signatures in the library will cause `mypy` to type check the function body when it is run on the library code base.
* Only exposing the public API of the library.

As I said, I initially started out typing the entire library (including the function bodies), but after discussing with you in PM on Discord you mentioned that you didn't feel strongly about having the library code typed. I thought this was a good compromise.

Hornwitser:
The bugfixes should probably be split up into their own commit (or maybe pull request).  They don't really have anything to do with adding typings to the lib.

bryanforbes:
> The bugfixes should probably be split up into their own commit (or maybe pull request).

Good call. I'll separate them out.

bryanforbes:
I've separated the fixes out into their own PRs and removed them from this one

bryanforbes:
> BotBase type hints could use the `Dict[str, Any]` for cogs, extra_events and extensions.

Added

> Also, any call to discord.ext.commands going through such syntax complains about ext not being in the `__init__.pyi`.

As long as you `import discord.ext.commands`, you can use `discord.ext.commands.Bot()`. Even without typings, this is the case. You cannot access `discord.ext.commands` by simply using `import discord`.

Werseter:
If you do `from discord.ext import commands` and call `discord.ext.commands` _for some reason_ instead of just `commands` it will prompt up as not setup up for typing.

I never implied it is a proper way to use it, but it exists, and typings should cover for it too, instead of pulling up a warn.

bryanforbes:
I just tried the following:

```python
import discord

Bot = discord.ext.commands.Bot
reveal_type(Bot)
```

I don't get a warning from `mypy` and it correctly reveals that it's a function that will return `discord.ext.commands.bot.Bot[CT]`. This seems to be the scenario you're describing and it seems to work.

Werseter:
Could it be it is the PyCharm's linter issue?
image

That would be silly tho, but if it's fine with mypy, then I guess it's fine, it's a niche case anyway.

Werseter:
It's been a while, as I still am in middle of transitioning my project to Type Hints, but I caught that `abc.GuildChannel`, `abc.Messageable`, `abc.PrivateChannel` and `abc.User`, could use Snowflake hints, as per _"This ABC must also implement :class:`abc.Snowflake`."_

Werseter:
Well, given that any object typed as GuildChannel - i.e list returned by `ctx.guild.channels`, will be said to not have an `id` property, as it must be implemented by a subclass, but not by the abstract itself. Since the assumption of this implementation is declared as imperative, shouldn't the typings have those two properties included?

OrangutanGaming:
Is this going to be merged or should someone change it to be in code (like using Sphinx doc strings allowing it to be explicit on the docs too)?

Werseter:
I myself have been using this on my own for a while, and I have to say it has been of great assist, but given that using the type hints in python, at least now, afaik isn't really _too mainstream_, and can be confusing as to the usecase. I'd nevertheless welcome the addition of this at least as an optional extension.

OrangutanGaming:
When can this get merged :(

LewdNeko:
keep up the good work bryan im sure danno will notice soon... hopefully. You the man

😖

OrangutanGaming:
You might want to squash all your commits into one, 38 is a lot for one thing

bryanforbes:
> You might want to squash all your commits into one, 38 is a lot for one thing

I eventually will. There are people using my branch and squashing each time there is an update would cause issues pulling from it, so I'm trying to keep headaches to a minimum.

BluePhoenixGame:
Damm, credits to you. Kepping this PR up to date for about 6 months. I hope this gets merged soon, it's one of the better PRs we have out here.

bryanforbes:
> Damm, credits to you.

Thanks. I use it for my own bot, so it's not a big deal keeping it up to date. If Danny doesn't end up merging it, I can take the work here and create a `discord-stubs` package after 1.0 is released.

discord/embeds.pyi
```diff
@@ -0,0 +1,65 @@
...
+    def __getattr__(self, attr: str) -> _EmptyEmbed: ...
+
+_E = TypeVar('_E', bound=Embed)
+
+class Embed:
+    title: Union[str, _EmptyEmbed]
+    type: str
+    description: Union[str, _EmptyEmbed]
+    url: Union[str, _EmptyEmbed]
+    colour: Union[int, Colour, _EmptyEmbed]
```
NCPlayz: `Embed.colo(u)r` and `Embed.timestamp` are properties rather than attributes, but this doesn't seem to cause an issue, so I guess this can pass.

discord/emoji.pyi
```diff
@@ -0,0 +1,51 @@
...
+    @property
+    def created_at(self) -> datetime.datetime: ...
+    @property
+    def url(self) -> Asset: ...
+    @property
+    def roles(self) -> List[Role]: ...
+    @property
+    def guild(self) -> Guild: ...
+    async def delete(self, *, reason: Optional[str] = ...) -> None: ...
+    async def edit(self, *, name: str, reason: Optional[str] = ...) -> None: ...
```
NCPlayz: `roles` parameter was not implemented.

discord/enums.pyi
```diff
@@ -0,0 +1,191 @@
...
+    default: int
+    recipient_add: int
+    recipient_remove: int
+    call: int
+    channel_name_change: int
+    channel_icon_change: int
+    pins_add: int
+    new_member: int
+
+class VoiceRegion(Enum):
```
NCPlayz: japan, southafrica, and india are missing from this enum.

discord/ext/commands/bot.py
```diff
@@ -94,7 +96,9 @@ def __repr__(self):
 
 _default = _DefaultRepr()
 
-class BotBase(GroupMixin):
+_CT = TypeVar('_CT', bound=Context)
```
NCPlayz: was this supposed to go in discord/ext/commands/bot.py? I see the same code in the pyi file.

discord/ext/commands/bot.pyi
```diff
@@ -0,0 +1,89 @@
...
+    command_prefix: _CommandPrefix
+    extra_events: Dict[str, List[Callable[..., Coroutine[Any, Any, None]]]]
+    case_insensitive: bool
+    description: str
+    owner_id: Optional[int]
+    help_command: Optional[HelpCommand[_CT]]
+
+    def __init__(self, command_prefix: _CommandPrefix, help_command: Optional[HelpCommand[_CT]] = ...,
+                 description: Optional[str] = ..., **options: Any) -> None: ...
+    def dispatch(self, event: str, *args: Any, **kwargs: Any) -> None: ...
```
NCPlayz:
```suggestion
    def dispatch(self, event_name: str, *args: Any, **kwargs: Any) -> None: ...
```

discord/widget.pyi
```diff
@@ -0,0 +1,51 @@
...
+from .invite import Invite
+from .user import BaseUser
+from .activity import Activity, Game, Streaming, Spotify
+from .channel import VoiceChannel
+from .invite import Invite
+
+from datetime import datetime
+from typing import Any, Optional, Union, List, Set, NamedTuple
+
+class WidgetChannel(NamedTuple):
```
NCPlayz: `WidgetChannel.__str__`?

discord/backoff.pyi
```diff
@@ -0,0 +1,5 @@
+from typing import Union
+
+class ExponentialBackoff:
```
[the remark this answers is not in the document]
bryanforbes: I'm not sure what you mean here. I've typed the `base` and `integral` parameters for `__init__`, but they are stored in "private" attributes.

discord/client.pyi
```diff
@@ -0,0 +1,264 @@
...
+_FuncType = Callable[..., Coroutine[Any, Any, Any]]
+_F = TypeVar('_F', bound=_FuncType)
+_GuildChannels = Union[TextChannel, VoiceChannel, CategoryChannel, StoreChannel]
+
+class Client:
+    ws: Optional[DiscordWebSocket]
+    loop: asyncio.AbstractEventLoop
+    shard_id: Optional[int]
+    shard_count: Optional[int]
+    activity: Union[Activity, Game, Streaming, Spotify]
```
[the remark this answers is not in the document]
bryanforbes: If the property has a getter and setter, it's easier to type it as an attribute since it's effectively the same thing from an interface perspective

discord/embeds.pyi
bryanforbes: See comment about `Client.activity`.

0x5c:
Is there information on whether or not this will be merged in the future? It would be great to be able to use `mypy` when working on bots.

discord/client.pyi
```diff
@@ -210,8 +210,8 @@ class Client:
                  timeout: Optional[float] = ...) -> asyncio.Future[Tuple[Role, Role]]: ...
     @overload
     def wait_for(self, event: Literal['guild_emojis_update'], *,
-                 check: Optional[Callable[[Guild, List[Emoji], List[Emoji]], bool]] = ...,
-                 timeout: Optional[float] = ...) -> asyncio.Future[Tuple[Guild, List[Emoji], List[Emoji]]]: ...
+                 check: Optional[Callable[[Guild, Tuple[Emoji], Tuple[Emoji]], bool]] = ...,
```
henribru: If this is a variable-length Tuple it should be `Tuple[Emoji, ...]`. But how come the type hints use `Tuple` when the documentation says `Sequence`?

bryanforbes: Good catch. I'm not sure why the docs are `Sequence` because what is produced are `Tuple[Emoji, ...]` (the event emits the before and after state of `Guild.emojis`) so I made the typings reflect the code.

SnowyLuma: It was changed to sequence in a [recent PR](https://github.com/Rapptz/discord.py/pull/2370) which corrected the docstring, as it was wrong before.

bryanforbes: `Sequence` is generally used (in typings) for an argument where you don't care about the exact type, but only that it implements that protocol. It's not generally used in return statements/callback arguments where the type is known. If the sequence of emojis could at some point be a list and another point a tuple, I could see using `Sequence`, however the sequnce of emojis will [only ever be a tuple](https://github.com/Rapptz/discord.py/blob/master/discord/state.py#L665).

jw2476:
this should be merged

DrBluefall:
Sorry to be a small bother, but are there any projections as to when this will be finished?

ShadowJonathan:
```
lint run-test: commands[0] | mypy -p '.../snek/snek' --config-file '.../snek/mypy.ini'
snek\cogs\identity.py:20: error: Untyped decorator makes function "on_message" untyped
```
`identity.py:20`;
```py
20:    @Cog.listener()
21:    async def on_message(self, message: Message) -> None:
```
Just a heads-up and a bump, since I found myself not able to figure out this problem for an hour or so, till i finally realized this lays with the library itself.

discord/member.pyi
```diff
@@ -0,0 +1,110 @@
...
+
+class Member(discord.abc.Messageable, discord.abc.User):
+    joined_at: Optional[datetime.datetime]
+    activities: Tuple[Union[Activity, Game, Streaming, Spotify], ...]
+    nick: Optional[str]
+    premium_since: Optional[datetime.datetime]
+    guild: Guild
+
+    # From discord.user.BaseUser
+    name: str
```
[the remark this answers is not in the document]
bryanforbes: These are needed because mypy says they're abstract properties without them

discord/member.pyi
```diff
@@ -0,0 +1,110 @@
...
+    @property
+    def desktop_status(self) -> Status: ...
+    def is_on_mobile(self) -> bool: ...
+    @property
+    def colour(self) -> Colour: ...
+    @property
+    def color(self) -> Colour: ...
+    @property
+    def roles(self) -> List[Role]: ...
+    @property
```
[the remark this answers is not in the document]
bryanforbes: These are needed because mypy says they're abstract properties without them

discord/user.pyi
```diff
@@ -0,0 +1,119 @@
...
+    def early_supporter(self) -> bool: ...
+    @property
+    def hypesquad(self) -> bool: ...
+    @property
+    def hypesquad_houses(self) -> List[HypeSquadHouse]: ...
+
+_BaseUser = discord.abc.User
+
+class BaseUser(_BaseUser):
+    name: str
```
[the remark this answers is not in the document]
bryanforbes: These are needed because mypy says they're abstract attributes without them

discord/user.pyi
```diff
@@ -0,0 +1,119 @@
...
+    def default_avatar_url(self) -> Asset: ...
+    @property
+    def colour(self) -> Colour: ...
+    @property
+    def color(self) -> Colour: ...
+    @property
+    def mention(self) -> str: ...
+    def permissions_in(self, channel: discord.abc.GuildChannel) -> Permissions: ...
+    @property
+    def created_at(self) -> datetime.datetime: ...
```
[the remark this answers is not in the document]
bryanforbes: These are needed because mypy says they're abstract properties without them

bryanforbes:
@ShadowJonathan can you share more of your code? I can't reproduce with the latest code from my repo. Or were you simply pointing out that the official library doesn't have typings?

ShadowJonathan:
> @ShadowJonathan can you share more of your code? I can't reproduce with the latest code from my repo. Or were you simply pointing out that the official library doesn't have typings?
both, but here's the file which fails to go through `mypy` linting without adding the `# type: ignore` comment:
https://git.jboi.dev/ShadowJonathan/SNEK/src/commit/58272b4f5d001d8e3a6ea40176b533becb52b19f/snek/cogs/identity.py

jack1142:
I think `User`, `PrivateChannel` and `GuildChannel` abc classes should also have `id` and `created_at` attribute as docs say that their subclasses must also implement `Snowflake`.

0x5c:
> > Damm, credits to you.
> 
> Thanks. I use it for my own bot, so it's not a big deal keeping it up to date. If Danny doesn't end up merging it, I can take the work here and create a `discord-stubs` package after 1.0 is released.

Is this still a possibility?

iomintz:
> > Thanks. I use it for my own bot, so it's not a big deal keeping it up to date. If Danny doesn't end up merging it, I can take the work here and create a `discord-stubs` package after 1.0 is released.
> 
> Is this still a possibility?

If it's not, maybe the PR could be released under the MIT license, same as discord.py itself, so that others can pick it up?

slice:
I noticed that `Client.http` isn't exposed, yet `HTTPClient` is typed. Is this intentional?

discord/ext/commands/core.pyi
```diff
@@ -0,0 +1,161 @@
...
+    params: Mapping[str, Parameter]
+    _buckets: CooldownMapping
+    cog: Optional[Cog[_CT]]
+
+    def __init__(self, func: _CoroType, *, name: str = ..., enabled: bool = ...,
+                 help: Optional[str] = ..., brief: Optional[str] = ..., usage: Optional[str] = ...,
+                 aliases: List[str] = ..., description: str = ..., hidden: bool = ...,
+                 rest_is_raw: bool = ..., ignore_extra: bool = ..., cooldown_after_parsing: bool = ...,
+                 checks: List[_CheckPredicate[_CT]] = ..., cooldown: Cooldown = ..., parent: _BaseCommand = ...,
+                 cog: Optional[Cog[_CT]] = ...) -> None: ...
```
slice, line 52: I think all of these kwargs need to be in `Optional`—it's possible to create a `Command` with just `func`.

henribru: `Optional` is for things that can be None. The reason you can create a `Command` with just `func` is that these arguments all have defaults, which is reflected in these type annotations.

slice: Oh, you're right. I confused defaults with optionals.

bryanforbes:
> I noticed that `Client.http` isn't exposed, yet `HTTPClient` is typed. Is this intentional?

I don't remember why I typed `HTTPClient` since it's not part of the public API. It's probably a hold-over from when I first started doing the typings and I typed everything.

iomintz:
@bryanforbes would you mind commenting on the proposal to license the pull request so that others can legally pick it up?

Gelbpunkt:
Why did you decide to use typing stub files and not to type the actual codebase? It would be a great way to find potential caveats and retype (https://pypi.org/project/retype/) seems to be a great starting point to work on that. I, however, love this PR and still wonder why it hasn't been merged yet. Untyped imports are one of the main issues with static typing in python, and anyone who's ever had that problem will understand me and is surely waiting for a merge.

bryanforbes:
> @bryanforbes would you mind commenting on the proposal to license the pull request so that others can legally pick it up?

I'm not sure what you mean. In order for me to contribute this to the project, I'm licensing it under the project's license. However, I've been thinking over the last couple of days that publishing these stubs under a `discord-stubs` package is probably the best bet.

bryanforbes:
> Why did you decide to use typing stub files and not to type the actual codebase?

I started out doing that, but after talking with Danny, he didn't want the annotations in the library itself. I did find a few issues in the codebase while I was typing the actual `.py` files that I submitted as fixes, but ultimately it boiled down to the wants of the maintainer of the project.

> It would be a great way to find potential caveats and retype (https://pypi.org/project/retype/) seems to be a great starting point to work on that.

Thanks for showing me this. This could help in other projects I've got going :).

> I, however, love this PR and still wonder why it hasn't been merged yet. Untyped imports are one of the main issues with static typing in python, and anyone who's ever had that problem will understand me and is surely waiting for a merge.

Thanks for the encouragement. I'll talk with Danny this week and see if we can't get this merged sooner, rather than later.

Gelbpunkt:
Having a discord-stubs package is the worst solution, but definitely better than no solution at all. The problem with that is that when changing the discord.py codebase, it would be easy to update the stubs on the go if it's in the same project. Having them separate means continuous work on updating them as people PR and merge into discord.py, which is a lot of work over time.

ItWasAllIntended:
> I, however, love this PR and still wonder why it hasn't been merged yet.
The original PR comment states Python 3.6 is required whereas discord.py still supports 3.5.3 and up

Gelbpunkt:
> The original PR comment states Python 3.6 is required whereas discord.py still supports 3.5.3 and up.
Python 3.5 goes EOL in September, might be about time to ditch support soon in favour of 3.6+. :smile:

bryanforbes:
> Having a discord-stubs package is the worst solution

I completely agree. I maintain two stubs packages in repos because I don't want to publish them and have the maintenance nightmare later on.

> , but definitely better than no solution at all.

Also agree.

> The problem with that is that when changing the discord.py codebase, it would be easy to update the stubs on the go if it's in the same project. Having them separate means continuous work on updating them as people PR and merge into discord.py, which is a lot of work over time.

There's also the fact that you wouldn't be able to properly use generics and would either have to use string type annotations or use `if TYPE_CHECKING:` statements. So to be clear: I don't want to do a stubs package.

bryanforbes:
> The original PR comment states Python 3.6 is required whereas discord.py still supports 3.5.3 and up.

I was mistaken when I first did this PR. I just tested and the `.pyi` files don't have to follow the same syntax as the Python version being checked. @Gelbpunkt, this is one advantage of using stub files instead of having the annotations in the code: we get to use variable type annotation syntax instead of type comments (until 3.5 is deprecated for discord.py).

iomintz:
> I'm not sure what you mean. In order for me to contribute this to the project, I'm licensing it under the project's license.

That's not required. PRs aren't released under any license until they're merged (and arguably not even then, without a CLA).

discord/member.pyi
```diff
@@ -0,0 +1,114 @@
...
+    # From discord.user.BaseUser
+    name: str
+    id: int
+    discriminator: str
+    avatar: Optional[str]
+    bot: bool
+    system: bool
+
+    @property
+    def avatar_url(self) -> str: ...
```
jack1142, line 44: This returns `Asset` object, like in its superclass `discord.user.BaseUser`
```suggestion
    def avatar_url(self) -> Asset: ...
```

discord/member.pyi
```diff
@@ -0,0 +1,114 @@
...
+    def is_avatar_animated(self) -> bool: ...
+    def avatar_url_as(self, *, format: Optional[str] = ..., static_format: str = ...,
+                      size: int = ...) -> str: ...
```
jack1142, line 47: This returns `Asset` object, like in its superclass `discord.user.BaseUser`
```suggestion
                      size: int = ...) -> Asset: ...
```

discord/member.pyi
```diff
@@ -0,0 +1,114 @@
...
+    @property
+    def default_avatar(self) -> DefaultAvatar: ...
+    @property
+    def default_avatar_url(self) -> str: ...
```
jack1142, line 51: This returns `Asset` object, like in its superclass `discord.user.BaseUser`
```suggestion
    def default_avatar_url(self) -> Asset: ...
```

discord/member.pyi
bryanforbes: Fixed

discord/member.pyi
bryanforbes: Fixed

discord/member.pyi
bryanforbes: Fixed

Spazzlo:
Any news on when this will be added to discord.py? Typing is very useful.

jack1142:
I don't have a news on that, but now there's a [discord.py-stubs package](https://github.com/bryanforbes/discord.py-stubs) from the author of this PR, which you could use if you need d.py to be typed.

bryanforbes:
After a few months of inactivity here, I've decided that it will be easier for me to maintain the stubs in their own package on pypi named [discord.py-stubs](https://pypi.org/project/discord.py-stubs/). This does add some complexity in using discord.py with mypy, namely using [generics from the stubs when they aren't generic at runtime](https://mypy.readthedocs.io/en/stable/common_issues.html#using-classes-that-are-generic-in-stubs-but-not-at-runtime). For instance, to create a `Cog` that could be type checked you would have to do the following:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    Cog = commands.Cog[commands.Context]
else:
    Cog = commands.Cog

class MyCog(Cog):
    ...
```

To alleviate this issue, I have also created [discord-ext-typed-commands](https://pypi.org/project/discord-ext-typed-commands/) which mirrors the `commands` module, but adds the ability to subscript for generics:

```python
from discord.ext import typed_commands

class MyCog(typed_commands.Cog[commands.Context]):
    ...
```

Thanks to @Rapptz for the help here as well as everyone who reviewed this PR (especially @NCPlayz and @Werseter )!

PR #1498, opened 2018-08-10
Add support for converters working with PEP-563

bryanforbes:
PEP-563 allows postponing evaluation of annotations with `from __future__ import annotations`, but this causes all annotations to be stored at runtime as strings. This causes `Parameter.annotation` to be a string and converters to fail. This PR resolves the type hints for the `Command` callback and updates the annotation with the resolved type hint, if needed.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -173,8 +173,20 @@ class Command:
 
         self.description = inspect.cleandoc(kwargs.get('description', ''))
         self.hidden = kwargs.get('hidden', False)
+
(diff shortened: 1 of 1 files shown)


bmintz:
Do converters even work with static type checkers, since they may return any type?

Modelmat:
Some do, such as `discord.Member`, and basic most certainly would (if the function itself is type hinted)
The converters would still work with static type checkers however it would be expected of the type of the converter, not the return type; hence it would be an incorrect type reported.

PR #1500, opened 2018-08-11
Delete  requirements.txt

GrayHatter:
In after upgrading to python 3.7 and istalling with `pip install https://github.com/Rapptz/discord.py/archive/async.zip`, discord.py wouldn't play nicely with EITHER lib at these versions. When I told pip to ignore dependencies, discord.py started working again. 

I'd be happy to fix up this commit with the actual required dependencies if you need them. But no matter what. these values are invalid for the async branch.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-aiohttp>=1.0.0,<1.1.0
-websockets>=3.1,<4.0
+aiohttp>=1.0.0
+websockets>=3.1


ItWasAllIntended:
If 3.7 is a requirement you can always switch over to the [rewrite](https://github.com/Rapptz/discord.py/tree/rewrite) branch

GrayHatter:
I was told not to use the rewrite branch yet before I was banned from the discord server. It the rewrite branch now safe to use?

Is there a test suite, or a standard test system that I could do to verify the async branch works with up to date dependencies? 

Python 3.7 is now the primary version of python, (ignoring that python 2 still exists for some reason). But discord.py doesn't work with it at all (because pinned libs don't). So while I agree that this isn't the best solution, (given I already agreed to fix up the commit as needed). It seems to me, the options are

1. do nothing: and discord.py is broken for new or recently updated installs.
2. merge this as is: and it might work, it's working for me
3. tell me how you'd like the dependencies to be updated/tested?

Unless I'm missing something?

ItWasAllIntended:
Rewrite has been safe to use for a long while. It implements everything async does and is fine for production.

mental32:
Rewrite is perfectly fine to use it’s the most up to date branch since async is no longer getting new features and, barring voice, works fine on 3.7

Your only issue migrating is porting the code because rewrite is not backwards compatible with async at all. It’s documentation even includes a migration guide

ClementJ18:
Testing wise: Because creating mocks of the discord models without emulating the api itself is complex.

GrayHatter:
I did fix the pull to just remove the upper limit of the dependencies. However...

If async is gonna be replaced soon, there's no need to find out what versions async is gonna need. 
I think this pull is safe close this then. I still think merging this is better than nothing,  as it wont hurt existing installs, but new installs on python 3.7 will start to work. 

I'm going to retract my offer to verify the exact versions of libs needed, but I'm happy to replace it with adding audio support to the rewrite branch. Only the work/specific tasks still needing work on rewrite isn't listed anywhere obvious.

Tobotimus:
@GrayHatter If you want to run async, you need to run it with the dependency constraints listed. The author put them there for a reason, and they're not going to change for async. In case you are not aware, *it's entirely possible to have multiple versions of python installed on the same system* - virtualenvs will help you out here. Not trying to be rude but if you want to save Rapptz some time, close the PR

Werseter:
Have you tried doing so on a fresh machine? :)

Seriously, drop it. We have pointed out your mistakes already. You have not fixed anything, you are basing your findings on a false positive.

GrayHatter:
Yes, just setup a fresh system to test it.... It worked

PR #1503, opened 2018-08-11
guild: remove chunked attribute

bmintz:
All the guilds in my bot are chunked now, even the smallest ones.
Removes Guild.chunked from the class and the repr.
This is a breaking change.


khazhyk:
This seems like a bad change

```py
plsbeta eval return sum(not g.chunked for g in bot.guilds)
✅
1
spoo.py🍅 betaBOTToday at 12:56 PM
74
```

PR #1504, opened 2018-08-11
guild: remove chunked from the repr

bmintz:
it's not useful enough to show in the repr, bloats the repr of Member, and can still be retrieved from the repr using Client.get_guild


Rapptz:
Useful debug info, which is pretty much what `chunked`  is for.

PR #1505, opened 2018-08-12
Supports awaitable setup function for extension

Modelmat:
By returning the possible generator the load_extension
can be awaited and thus await the possible generator;
this allows for the setup function to be asynchronous.


Modelmat:
Closing this as if the extension setup fails it will be in the extension list. No way to do this while keeping backwards compatability.

PR #1520, opened 2018-08-17
Fix Member.roles order on GUILD_ROLE_UPDATE

### LINKED ISSUE #1518: Author.top_role not updating on role hierarchy changes
Hello. I've moved one role that I have to the top of the roles in my server, and when I print ``author.top_role``, it shows the old role.

Screenshot 1: https://cdn.discordapp.com/attachments/305571047535673345/479841129341255681/unknown.png

Screenshot 2:
https://cdn.discordapp.com/attachments/305571047535673345/479841165114605568/unknown.png

I added an ``on_guild_role_update`` event to track if the role positional update gets fired, and it does, so the members' roles aren't updated when ``on_guild_role_update`` fires.

Harmon758:
Fixes #1518

### DIFF
--- a/discord/state.py
+++ b/discord/state.py
@@ -763,6 +763,9 @@ class ConnectionState:
             if role is not None:
                 old_role = copy.copy(role)
                 role._update(role_data)
+                if role.position != old_role.position:
+                    for member in role.members:
+                        member.roles.sort()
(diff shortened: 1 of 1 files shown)


Rapptz:
I think there's a better fix for this which probably involves changing the underlying storage.

PR #1521, opened 2018-08-17
Added change/leave hypesquad house for non bot clients

mental32:
Please review.

This PR is due to the new Hypesquad 2.0, where it's open to everybody and delegated into hypesquad houses (bravery, brilliance, balance).

What this PR does is add two methods to both the HTTPClient and the Client allowing the action of changing a hypesquad house and leaving a hypesquad house. Also adding an Enum for hypesquad houses (see enums.py)

Hopefully with further review and scrutiny this PR can allow differentiating between the type of current hypesquad house and joining one remotely and or decide if this functionality is even needed in the library.

### DIFF
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -220,12 +220,10 @@ class ActivityType(IntEnum):
     listening = 2
     watching = 3
 
-
 class HypesquadHouse(Enum):
     bravery = 1
     brilliance = 2
     balance = 3
-
 
 def try_enum(cls, val):
     """A function that tries to turn the value into enum ``cls``.
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -212,6 +212,9 @@ class UserFlags(Enum):
     staff = 1
     partner = 2
     hypesquad = 4
+    hypesquad_bravery = 64
+    hypesquad_brilliance = 128
+    hypesquad_balance = 256
 
 class ActivityType(IntEnum):
     unknown = -1
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -29,7 +29,7 @@ from enum import Enum, IntEnum
 __all__ = ['ChannelType', 'MessageType', 'VoiceRegion', 'VerificationLevel',
            'ContentFilter', 'Status', 'DefaultAvatar', 'RelationshipType',
            'AuditLogAction', 'AuditLogActionCategory', 'UserFlags',
-           'ActivityType', ]
+           'ActivityType', 'HypesquadHouse']
 
 class ChannelType(Enum):
     text     = 0
@@ -219,6 +219,12 @@ class ActivityType(IntEnum):
     streaming = 1
     listening = 2
     watching = 3
+
+
+class HypesquadHouse(Enum):
+    bravery = 1
+    brilliance = 2
+    balance = 3
 
 
 def try_enum(cls, val):
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -29,7 +29,7 @@ from enum import Enum, IntEnum
 __all__ = ['ChannelType', 'MessageType', 'VoiceRegion', 'VerificationLevel',
            'ContentFilter', 'Status', 'DefaultAvatar', 'RelationshipType',
            'AuditLogAction', 'AuditLogActionCategory', 'UserFlags',
-           'ActivityType', 'HypesquadHouse']
+           'ActivityType', 'HypeSquadHouse']
 
 class ChannelType(Enum):
     text     = 0
@@ -223,7 +223,7 @@ class ActivityType(IntEnum):
     listening = 2
     watching = 3
 
-class HypesquadHouse(Enum):
+class HypeSquadHouse(Enum):
     bravery = 1
     brilliance = 2
     balance = 3
--- a/discord/http.py
+++ b/discord/http.py
@@ -769,13 +769,8 @@ class HTTPClient:
     def get_user_profile(self, user_id):
         return self.request(Route('GET', '/users/{user_id}/profile', user_id=user_id))
 
-    def change_hypesquad_house(self, house_id, first_name, last_name):
-        payload = {
-            'house_id': house_id,
-            'first_name': first_name,
-            'last_name': last_name
-        }
-
+    def change_hypesquad_house(self, house_id):
+        payload = {'house_id': house_id}
         return self.request(Route('POST', '/hypesquad/online'), json=payload)
 
     def leave_hypesquad_house(self):
--- a/discord/http.py
+++ b/discord/http.py
@@ -768,3 +768,15 @@ class HTTPClient:
 
     def get_user_profile(self, user_id):
         return self.request(Route('GET', '/users/{user_id}/profile', user_id=user_id))
+
+    def change_hypesquad_house(self, house_id, first_name, last_name):
+        payload = {
+            'house_id': house_id,
+            'first_name': first_name,
+            'last_name': last_name
+        }
+
+        return self.request(Route('POST', '/hypesquad/online'), json=payload)
+
+    def leave_hypesquad_house(self):
+        return self.route(Route('DELETE', '/hypesquad/online'))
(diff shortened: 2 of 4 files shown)


Vexs:
Seems to me this would better be added as part of [clientuser.edit](https://discordpy.readthedocs.io/en/rewrite/api.html#discord.ClientUser.edit) which would match with everything else related to editing user profiles. 
Further, this PR does not include support for telling which house someone is in, and should be included. See: [profile.hypesquad](https://discordpy.readthedocs.io/en/rewrite/api.html#discord.Profile.hypesquad)

mental32:
Yeah I like the idea of it being part of clientuser.edit I’ll move it over to there.

As for detecting what house you’re already in I talked with some folks over in the dapi server and it seems that there are inconsistencies in the profile flags when there are other badges present (staff, bug hunter) but I will further look into this.

Also btw this commit wasn’t to be accepted as is I’m merely looking for the best way to implement this functionality if at all

mental32:
TLDR we can recognise which house users belong to

I’ve addressed the issue of figuring out what house you are in in the first place and I can’t seem to recreate the issues that were apparent before.

So I narrowed it down to 7, 8 and 9 bits in the user flag to represent the bravery, brilliance and balance houses also I have performed several tests confirming my assumptions

discord/user.py
```diff
@@ -61,6 +61,9 @@ def hypesquad(self):
     def partner(self):
         return self._has_flag(UserFlags.partner)
 
+    @property
+    def hypesquad_houses(self):
+        return [house for house, flag in zip(HypesquadHouse, (64, 128, 256)) if (self.flags & flag) == flag]
```
bmintz: don't hardcode this

mental32: user flags are already hard coded in the enums

discord/user.py
```diff
@@ -383,6 +391,24 @@ def blocked(self):
...
+                raise ClientException('When changing hypesquad houses first_name and last_name fields are required.')
+
+            else:
+                if isinstance(house, HypesquadHouse):
+                    value = house.value
+                else:
+                    value = int(house)
+
+                await http.change_hypesquad_house(value, fields['first_name'], fields['last_name'])
+        except KeyError:
```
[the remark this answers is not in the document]
mental32: Can't return early, we still have to send out a request to edit the ClientUser. can't bring back the except farther too, since we'd have a NameError which that entire block inside the try except relies on `house` being present and a valid house type.

This block is supposed to simply fail silently if we get no house in the kwargs, it's not supposed to affect the success of the regular edit call

discord/enums.py
```diff
@@ -221,6 +221,12 @@ class ActivityType(IntEnum):
     watching = 3
 
 
+class HypesquadHouse(Enum):
+    bravery = 1
+    brilliance = 2
+    balance = 3
+
+
```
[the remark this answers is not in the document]
mental32: okay the extra line break i will remove

discord/http.py
```diff
@@ -768,3 +768,15 @@ def get_user_info(self, user_id):
...
+        payload = {
+            'house_id': house_id,
+            'first_name': first_name,
+            'last_name': last_name
+        }
+
+        return self.request(Route('POST', '/hypesquad/online'), json=payload)
+
+    def leave_hypesquad_house(self):
+        return self.route(Route('DELETE', '/hypesquad/online'))
```
bmintz: should be `self.request`

mental32: ouch good catch

discord/http.py
```diff
@@ -779,4 +779,4 @@ def change_hypesquad_house(self, house_id, first_name, last_name):
         return self.request(Route('POST', '/hypesquad/online'), json=payload)
 
     def leave_hypesquad_house(self):
-        return self.route(Route('DELETE', '/hypesquad/online'))
+        return self.request(self.route(Route('DELETE', '/hypesquad/online')))
```
bmintz: now remove self.route :rofl:

discord/user.py
bmintz: i think maybe they should be their own enum, or maybe the powers of two could replace the values in HypesquadHouse

Rapptz: Seconding, this needs to be done better. These are, after all, user flags that have their own enum.

discord/user.py
```diff
@@ -383,6 +391,24 @@ def blocked(self):
 
         http = self._state.http
 
+        try:
+            house = fields['house']
+            if house is None:
+                await http.leave_hypesquad_house()
+
```
Rapptz: Superfluous newline.

discord/user.py
```diff
@@ -383,6 +391,24 @@ def blocked(self):
...
+            elif 'first_name' not in fields or 'last_name' not in fields:
+                raise ClientException('When changing hypesquad houses first_name and last_name fields are required.')
+
```
Rapptz: Superfluous newline.

discord/user.py
```diff
@@ -383,6 +391,24 @@ def blocked(self):
 
         http = self._state.http
 
+        try:
+            house = fields['house']
+            if house is None:
+                await http.leave_hypesquad_house()
+
+            elif 'first_name' not in fields or 'last_name' not in fields:
```
Rapptz: Generally speaking I prefer accessing and returning than checking if they key exists. I think this probably applies here too since it's two fields.

discord/user.py
```diff
@@ -383,6 +391,24 @@ def blocked(self):
...
+                raise ClientException('When changing hypesquad houses first_name and last_name fields are required.')
```
Rapptz: This exception change is undocumented.

discord/enums.py
```diff
@@ -220,6 +220,10 @@ class ActivityType(IntEnum):
     listening = 2
     watching = 3
 
+class HypesquadHouse(Enum):
```
Rapptz: Proper capitalisation is `HypeSquad`.

discord/user.py
```diff
@@ -338,6 +341,11 @@ def blocked(self):
         email: str
             The new email you wish to change to.
             Only applicable to user accounts.
+        house: Union[int, :class:HypesquadHouse]
```
Rapptz:
```
Optional[:class:`HypeSquadHouse`]
```

Not sure about having the ability to pass an integer.

discord/user.py
mental32: I'm actually not too sure what you mean by this @Rapptz

discord/user.py
```diff
@@ -338,6 +342,11 @@ def blocked(self):
         email: str
             The new email you wish to change to.
             Only applicable to user accounts.
+        Optional[:class:`HypeSquadHouse`]
+            If this is supplied also supply ``first_name`` and ``last_name`` fields of str type.
```
bmintz: ditto

discord/user.py
```diff
@@ -383,6 +393,21 @@ def blocked(self):
 
         http = self._state.http
 
+        try:
+            house = fields['house']
+            if house is None:
+                await http.leave_hypesquad_house()
+            elif 'first_name' not in fields or 'last_name' not in fields:
+                raise ClientException('When changing hypesquad houses first_name and last_name fields are required.')
```
bmintz: In my experience the first_name and last_name fields are neither required or sent by the official client.

discord/http.py
```diff
@@ -768,3 +768,15 @@ def get_user_info(self, user_id):
 
     def get_user_profile(self, user_id):
         return self.request(Route('GET', '/users/{user_id}/profile', user_id=user_id))
+
+    def change_hypesquad_house(self, house_id, first_name, last_name):
+        payload = {
+            'house_id': house_id,
+            'first_name': first_name,
+            'last_name': last_name
```
bmintz: ditto

discord/user.py
mental32: When i was told about the non user endpoint they told me to include these fields. I am not sure if changing houses would fail houses without them I'm going to go check on this rn

mental32: Can confirm a POST to `/hypesquad/online` does not require first_name and last_name fields updating PR now

PR #1522, opened 2018-08-18
Space Insensitivity

fourjr:
Allows for commands to be called with space insensitivity

**Examples**
`!ping` works
`!          ping` works too

This PR adds a kwarg for `commands.Bot`, `space_insensitive`. It is set to `False` on default.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -161,6 +161,7 @@ class BotBase(GroupMixin):
         self.owner_id = options.get('owner_id')
         self.command_not_found = options.pop('command_not_found', 'No command called "{}" found.')
         self.command_has_no_subcommands = options.pop('command_has_no_subcommands', 'Command {0.name} has no subcommands.')
+        self.space_insensitive = options.get('space_insensitive', False)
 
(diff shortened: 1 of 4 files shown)


Hornwitser:
Is there a good reason to have this?  Also seems like it would be much simpler implemented with `view.skip_ws()`.

Vexs:
I can't say I've ever seen someone request this before, and the general policy on conditionals getting added to the command processor is "not unless it's absolutely necessary"- and this strikes me as an extreme edge case.
Further, as a simple alternative, you can just add `"prefix "` to your prefixes.
As a more advanced alternative, you could implement a callable prefix like this:
```py
async def get_prefix(bot, message):
     match = re.match(r'prefix\s+?')
     if match:
         return [match]
     return ['!','/']
```
While neither options are exactly optimal, they're both simple options that would cover the same thing-ish this does. 

But really, it comes down to the fact I've yet to see someone ask for this. (Outside of mention-prefixes, which is solved elegantly by one of the above)

Rapptz:
Sorry.

I think this is part of the functionality that dynamic prefixes should cover (either with a regex or single-space forgiveness).

Issue #1529, opened 2018-08-22
Signal handling in run() prevents threading


Theriaca:
[Signal handling](https://github.com/Rapptz/discord.py/blob/rewrite/discord/client.py#L550-L554) (for non-win32 platforms) should atleast be made optional, in case the users don't want to run their bot instance in the main thread.

Perhaps might have missed an option myself and there is another proper workaround (and I apologize if so), but as it stands the easiest way I've found to work around this (when updating a bot to use the rewrite branch) was to comment it out.

```
Traceback (most recent call last):
  File "/usr/lib/python3.6/asyncio/unix_events.py", line 100, in add_signal_handler
    signal.set_wakeup_fd(self._csock.fileno())
ValueError: set_wakeup_fd only works in main thread
```

Ikusaba-san:
What are you doing that requires `client.run()` to be run on a separate thread? Usually it never needs to be on a separate thread unless you're doing something dubious.

Theriaca:
Basically, running bots for different platforms together and leaving the main thread open for other operations. Also, for example, some actions/initializations require waiting on the different clients to be ready and I'm using threading events for that purpose. One thing I'm actually wondering though, is why a library is calling signals in the first place.

bmeh:
I concur. Firstly, signal handlers such as these have no place in a library, and libraries shouldn't even exit on their own, they should return an error (to the user of the library), throw an exception, or anything that would make the user responsible for handling those errors, and secondly, one of the top priorities of a library should be eliminating those error handling paths to begin with, not adding them without giving control to the user. It's not the job of a library to handle errors by exiting or handling anything at the expense of reducing the control of any user of a library, at least in the case of error handling.

bmeh:
Except you are adding unnecessary inconsistency between platforms for no reason whatsoever other than "I like it that way". If you are not willing to make it optional - which is quite easy to do so -, then expect your library to be put aside by many people (no, I'm not going to pull percentages out of my a$$). I'm sorry you could not be convinced to implement something so simple, but judging from your latest commit history, and the decent amount of unanswered pull requests, it seems to me that you have no interest in maintaining it anymore.

Your claim of what I have said being irrelevant is just outright false. On what grounds did you conclude that it adds "very little harm", and what option are you talking about if you are not even convinced to add an optional bool variable? It's apparently an issue, otherwise this wouldn't have been reported, but I guess not many people have reported it. Is that why you think it's irrelevant?

[...]

PR #1530, opened 2018-08-22
Ignore exceptions related to registering signal handlers.

Gorialis:
This deals with #1529 by making successful signal registry no longer a requirement for `Client.run` to operate.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -547,11 +547,12 @@ class Client:
         is blocking. That means that registration of events or anything being
         called after this function call will not execute until it returns.
         """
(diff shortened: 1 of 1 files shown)


Hornwitser:
This seems like a bad idea.  If you start the bot in a thread on Linux nothing in `self._do_cleanup` will be done on exit.

Gorialis:
It doesn't get performed anyway, as signal handler registration fails. This not only doesn't run `self._do_cleanup`, it doesn't run *any* bot code at all (because it raises). This holds true for the current alternative as well, which is to use `client.start` manually.

PR #1552, opened 2018-08-30
Fix PartialEmoji._as_reaction() for emojis the bot can't see

CarlGroth:
Sorry about the last PR.

Using `Message.add_reaction(PartialEmoji)` on an already existing reaction fails if the bot does not have access to the emoji. This PR fixes that.


Rapptz:
I fixed your commit message for you.

Thanks.

PR #1564, opened 2018-09-08
Added support for discordapp.com/invite invites

PumPum7:


PumPum7:
It didnt work for me when I didnt change it so I changed the numbers

Gorialis:
That's because you added a capturing group. The proposition is to make it a noncapturing group.

PR #1566, opened 2018-09-08
Make implicit raw backslashes explicit

BeatButton:
In Python 3.8 implicit raw backslashes in strings will be deprecated. So, this.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -533,7 +533,7 @@ class GuildChannel:
             The :class:`Member` or :class:`Role` to overwrite permissions for.
         overwrite: :class:`PermissionOverwrite`
             The permissions to allow and deny to the target.
-        \*\*permissions
+        \\*\\*permissions
             A keyword argument list of permissions to set for ease of use.
             Cannot be mixed with ``overwrite``.
         reason: Optional[str]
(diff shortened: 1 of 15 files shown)


Hornwitser:
This was deprecated in Python 3.6.

BeatButton:
Sorry, I misspoke. They'll be a straight-up error in 3.8.

Hornwitser:
I'm unable to find any place saying this will be an error in 3.8.  [Issue32912](https://bugs.python.org/issue32912) has suggested changing it to a SyntaxWarning instead of a non-silent DeprecationWarning in 3.8 but there doesn't appear to be any movement on it.

Hornwitser:
That does not say _when_ it'll become an error, only that it _will eventually_ become an error.

BeatButton:
Ah, you're right, I had conflated this deprecation warning with a different one. But, regardless, it's been deprecated. There's no reason not to fix it.

Rapptz:
This PR is wrong.

1. You shouldn't change `r` prefixed strings.
2. It's better to instead turn the docstrings into `r"""..."""` instead of the current `"""..."""` and changing every backslash into a double-backslash, per [PEP-257](https://www.python.org/dev/peps/pep-0257/).

BeatButton:
Haha, oops, I forgot about the raw-strings thing. And that's a good point, too. Sorry for having the wrong idea.

PR #1570, opened 2018-09-09
Change docstrings to raw-strings

BeatButton:
Second try at #1566. I deleted the branch for that one though so I just figured I'd do a new PR.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -489,7 +489,7 @@ class GuildChannel:
         await self._state.http.delete_channel(self.id, reason=reason)
 
     async def set_permissions(self, target, *, overwrite=_undefined, reason=None, **permissions):
-        """|coro|
+        r"""|coro|
 
         Sets the channel specific permission overwrites for a target in the
         channel.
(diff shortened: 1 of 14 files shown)


Hornwitser:
You forgot three docstrings with backslashes.  One in calls.py (line 135) and two in client.py (line 66, 291).  Furthermore the `\N{name}` escapes in client.py line 709-719 looks like they should be escaped as well.

BeatButton:
Thanks! For some reason they didn't show up as ```DeprecationWarning```s, which is what I was using to test them. As for ```\N{name}``` eacapes, those are valid Python escapes and don't need to be turned into raw strings as far as I can figure.

Hornwitser:
It's a code example.  When not escaped they turn into the emoji in the built docs, which makes me wonder if that was unintentional, as the emoji chars might as well have been used directly when the docs shows them as the emoji.

BeatButton:
Oh, I see. I wouldn't know whether that's intentional or not.

bmintz:
Leaving them unescaped would preserve their current appearance in the docs. Adding the r prefix for those docstrings would require changing the emojis to their actual Unicode form.

BeatButton:
Unless, of course, the intent is to show the escapes as raw text to demonstrate how it can be done.

PR #1573, opened 2018-09-10
Make setup.py executable

bmintz:


Hornwitser:
Why do you want this change?

bmintz:
Cause I had to run `setup.py sdist` to test something, and it's easier to run `./setup.py sdist` than to run `python3 setup.py sdist`. Also, a file which is intended to be run as a script should generally be marked as such.

bmintz:
The argument for marking this as a script is the same as for any other script: it's an executable file. You're not meant to import it. Same goes for main.py or bot.py or launcher.py or whatever.

c-x-berger:
No, you;re not meant to import it.  
You're meant to execute it with `python setup.py dist`.

bmintz:
Right. Which is why a shebang is a good idea, since executing ./setup.py sdist inside  a venv would run it with the current python (and is shorter)

bmintz:
This pull request is merely for convenience.

A possible issue is trying to execute setup.py but you meant a different python than the one on your PATH. Example: venv/bin/python setup.py would not be the same as ./setup.py if the venv is not activated. However, this is a rare use case so I don't anticipate that this drawback would outweigh the convenience benefit.

Gorialis:
I don't mind this change. It's out of typical convention, but it's not like doing this removes or changes any existing functionality or workflow. My only argument would be that if we're going down this route, we might as well add the [encoding header](https://github.com/Rapptz/discord.py/blob/rewrite/discord/__init__.py#L1) as well:

https://github.com/Rapptz/discord.py/blob/rewrite/discord/__init__.py#L1

bmintz:
Good fight everyone 🙇

PR #1576, opened 2018-09-11
Provide note about current vs develepment version of discord.py

vanous:
Attempt to educate people about rewrite.
Thank you
Petr

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,12 +11,14 @@ This was written to allow easier writing of bots or chat logs. Make sure to fami
 
 ### Breaking Changes
 
-The discord API is constantly changing and the wrapper API is as well. There will be no effort to keep backwards compatibility in versions before `v1.0.0`.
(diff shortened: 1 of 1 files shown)


vanous:
Thank you for all the work done on discord.py , this is great and i appreciate all the hours that have been put into it. Unfortunately, decision like in #946 do dis-value time of all people who have been willing to put their time into using discord.py . No developer should develop piece of software (bot) only to learn that if the bot scales above 2500 servers, the code will have to be updated to rewrite, thus put many extra hours into rewriting code, rather then starting on correct bases.

Please reconsider this approach.

scragly:
It is not necessary for a bot to use in-development code just because of the reasoning that it requires sharding. Sharding is possible in the async library, it is just not involving so much brain-dead handholding as the rewrite library due to the existence of AutoShardingBot.
Due to this, your reasoning is void.

vanous:
Thank you, i appreciate your time.

Vexs:
It's also worth noting that the rewrite-only feature of `autoshardedclient` that you are presumably referring to only really works for around 20k servers- at which point it runs into other limitations. 

Further, to detail what @scragly said, a cursory look at the docs for client shows kwargs for shard_count and shard_id- for [both](https://discordpy.readthedocs.io/en/latest/api.html#discord.Client) [branches.](https://discordpy.readthedocs.io/en/rewrite/api.html#discord.Client)

PR #1584, opened 2018-09-15
New Colour

BluePhoenixGame:
A new colour to discord.colour named darker_greyple. Its the same colour as the discord dark-theme background. Which makes a pretty cool effect:
screenshot_20180915_183138
And in discord light-theme it also looks pretty okay:
screenshot_20180915_183323

### DIFF
--- a/discord/colour.py
+++ b/discord/colour.py
@@ -227,4 +227,13 @@ class Colour:
         """A factory method that returns a :class:`Colour` with a value of ``0x99aab5``."""
         return cls(0x99aab5)
 
+    @classmethod
+    def dark_theme(cls):
+        """A factory method that returns a :class:'Colour' with a value of ``0x36393F``.
+        Will appear transparent on Discord's dark theme and be the text colour on Discord's light theme.
+        
+        .. versionadded:: 1.5
+        """
+        return cls(0x36393F)
+
 Color = Colour


BluePhoenixGame:
Yeah, i named it darker_greyple as it looks like a darker tone of the already existing greyple.

Aimarekin:
Looks neat!

discord/colour.py
```diff
@@ -226,5 +226,13 @@ def blurple(cls):
     def greyple(cls):
         """A factory method that returns a :class:`Colour` with a value of ``0x99aab5``."""
         return cls(0x99aab5)
+    
+    @classmethod
+    def darker_greyple(cls):
+        """A factory method that returns a :class:`Colour` with a value of ``0x36393F``.
+        Which is the value of discords dark-theme background."""
```
Gorialis: - Documentation should avoid using poorly formed run-on sentences.
- Reference to Discord as a company should be capitalized, and an apostrophe is required here (`Discord's`).
- Explicit reference should be made to the Discord web/desktop client, as the background color varies between the web and mobile native clients.
- "Dark theme" is not one word and thus doesn't need to be hyphenated.

discord/colour.py
```diff
@@ -226,5 +226,13 @@ def blurple(cls):
...
+        return cls(0x36393F)
+
+Colour.transparent = Colour.darker_greyple
```
Gorialis: Aliasing this to `transparent` is misleading. It is not *actually* transparent, having shading differences on mobile and not working at all on light theme. We should not make the assumption that users are using dark theme, especially considering that light theme is still the default theme for Discord, but even so it is inaccurate.

BluePhoenixGame:
Let me try to come up with a more fitting name.

MusicOnline:
Not `default_dark` or `transparent` please.
I think `dark_greyple` is better to stay consistent with `dark_green`, `dark_teal` and etc which are already present in the library,

The 'transparent-ness' of the colour can be added as a note in its docstring instead. Refer to other parts of the documentation to get a hold on the writing style,

BluePhoenixGame:
Thanks! I I'll try to fix it when I get home.

BluePhoenixGame:
I added a note saying that its the text colour wich you can see here in the D:
image

BluePhoenixGame:
I also have made tests on my phone and there the background also is ``0x36393F``

BluePhoenixGame:
Hope i have completed your requests/opinions. Is it better now?

discord/colour.py
```diff
@@ -226,5 +226,15 @@ def blurple(cls):
     def greyple(cls):
         """A factory method that returns a :class:`Colour` with a value of ``0x99aab5``."""
         return cls(0x99aab5)
+    
+    @classmethod
+    def darker_greyple(cls):
+        """
+        A factory method that returns a :class:'Colour' with a value of ``0x36393F``.
+        Will appear transparent on Discord's dark theme and be the text colour on Discord's light theme.
```
TurnrDev, line 234: To avoid confusion in the future, maybe you should clarify the version of Discord that it's been tested against. This classmethod could easily become ignored by the devs at d.py and if Discord changed the background colour on their client, this statement would no longer be true.

TurnrDev (changes requested):

discord/colour.py
BluePhoenixGame: I see the concern, but I believe that adding such a note would be unnecessary. There are other stuff in this library that isn't based on official "data", which doesn't do this (writing a date/version). Doing this would rather make it more of a hassle because it would preferably be updated each update, as seeing an older version would make it seem outdated. When the case would just be that Discord hasn't changed it.

At this point I am extremely confident that changing the color wouldn't really happen and if it did then we would know, because of the fact that this pull request is 2 years old in a week. And during these years I have yet to update the colour ever.

It's also made clear in the documentation that it will "appear transparent on Discord's dark theme", it is in fact not transparent.

discord/colour.py
```diff
@@ -226,5 +226,15 @@ def blurple(cls):
     def greyple(cls):
         """A factory method that returns a :class:`Colour` with a value of ``0x99aab5``."""
         return cls(0x99aab5)
+    
+    @classmethod
+    def darker_greyple(cls):
```
Rapptz, line 231: This was brought up a few days ago again in the discord.py server and the idea generally doesn't seem too bad. Incidentally, a better name was mentioned there so I'll propose it here.

```suggestion
    def dark_theme(cls):
```

The documentation can then be updated to refer to the fact that this is the dark theme's background colour and the appropriate versionadded string should you want it.

Rapptz (changes requested):

BluePhoenixGame:
Renamed and squashed all previous commits.

TurnrDev (approved):

Rapptz (approved):

PR #1586, opened 2018-09-15
Updating the alias descriptions for discord.Member.color

itsVale:
Like the title says, the docs for `discord.Member.colour` denote that there is an alias for this under `color`. So far so good, but `discord.Member.color` is denoting the exact same thing. So I changed that for discord.Member.colour to denote that there is the alias `color` and for discord.Member.color to denote that there is the alias `colour`.

### DIFF
--- a/docs/locale/ja/LC_MESSAGES/api.po
+++ b/docs/locale/ja/LC_MESSAGES/api.po
@@ -7350,8 +7350,12 @@ msgid ""
 "instance of :meth:`Colour.default` is returned."
 msgstr ""
 
-#: discord.Member.color:5 discord.Member.colour:5 of
+#: discord.Member.colour:5 of
 msgid "There is an alias for this under ``color``."
+msgstr ""
+
+#: discord.Member.color:5 of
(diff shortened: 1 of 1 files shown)


Crypti-x:
In my opinion it is unnecessary to list aliases separately, because the reason for the existence of this alias is a non-standard language. This non-standardization should not be encouraged.

itsVale:
Well, here are the changes. I think it's necessary.

http://prntscr.com/kuufhr

Crypti-x:
I don't think so.
The way to go would be the removal of the alias notice

eliasfrank:
Why is ``colour`` an alias in the first place? Every standard spellchecker denotes it as an error because it's a UK regional spelling of the word ``color``.

ItWasAllIntended:
No, colour is the international standard. Please don't spread your agenda and 'patriotism' in a PR

Rapptz:
This is not how you do this for starters.

Second, it seems like a bug that `color` is even output.

PR #1588, opened 2018-09-16
Add colour attribute for BaseUser

MusicOnline:
This PR makes `ctx.me.colour` or `ctx.author.colour` possible in `DMChannel`. For instance, using them for an embed colour (without try/except). Also adds a bit of consistency between `User` and `Member` (for lack of a better word). It always returns `Colour.default()` because there's no role colours.

Also, color alias.


slice:
Any motivation/reasoning behind this? Just the consistency?

MusicOnline:
```
try:
    embed = discord.Embed(colour=ctx.author.colour)  # or ctx.me.colour
except AttributeError:
    embed = discord.Embed()
```
As stated in the OP, this PR reduces the above codeblock to one line. I don't think there's any other use cases than for embeds, but I also doubt this 'devalues' the library in performance or design (again for lack of a better word).

PR #1600, opened 2018-09-20
Add add_reaction method to Context

romangraef:
This adds an `add_reaction` method to `discord.ext.commands.Context` which just calls `Message.add_reaction`. I think this could be useful as many bots react to a message with an :x: or an :white_check_mark: to indicate success/failure of a command.

### DIFF
--- a/discord/ext/commands/context.py
+++ b/discord/ext/commands/context.py
@@ -167,7 +167,7 @@ class Context(discord.abc.Messageable):
             to_call = cmd.root_parent or cmd
             view.index = len(self.prefix)
             view.previous = 0
-            view.get_word() # advance to get the root command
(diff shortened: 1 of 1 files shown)


romangraef:
@crrapi Not really, i just typed it way to often, and thought this was an easier way for my future me and potentially others.

TomsUsername:
I think most of us already do this with a custom ctx

bmintz:
If you're going to add this I think remove_reaction should be added also. But I don't think this is worth merging in the first place.

Rapptz:
If you want to do this, it's better to just use your own subclass. I don't want to add a bunch of shortcuts unless I believe it benefits the majority of people and in this case I don't think it does much except make `Context` seem like a super class of sorts.

PR #1613, opened 2018-09-24
Remove loop calls

bmintz:
Removes loop from the following places:
1. discord.utils.sane_wait_for (commit 1, safest to merge)
2. Every call to a built in asyncio callable. This includes Lock, Event, wait_for, wait, sleep, and ensure_future.

This is possible because these functions just call asyncio.get_event_loop(), which, post 3.5.3, always return the current running event loop in the current thread, as discussed in #general.

Loop arguments to websockets, aiohttp, and discord.py classes such as ConnectionState have been kept. Some of these may be removable, except for aiohttp.ClientSession (warns about creation outside of a coroutine otherwise).

Note: this code is **untested**. I wanted to file a pull request now, so I can gather feedback. I will post the results of my tests later.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -768,12 +768,12 @@ class Messageable(metaclass=abc.ABCMeta):
         ret = state.create_message(channel=channel, data=data)
         if delete_after is not None:
             async def delete():
-                await asyncio.sleep(delete_after, loop=state.loop)
+                await asyncio.sleep(delete_after)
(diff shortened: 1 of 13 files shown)


bmintz:
Danny: this is pretty much just always making the library call asyncio.get_event_loop() for no reason when it's already right fucking there as a variable
Danny: all for ""readability""

PR #1623, opened 2018-09-30
(Help Formatter) Check cog for 'cog_name' attribute and use that instead

uniflare:
Simple change to allow the cog to specify a name to use as the category in the default help formatter.

Could be cleaner to specify this in the add_cog method.


uniflare:
Duplicate of https://github.com/Rapptz/discord.py/pull/1422

PR #1624, opened 2018-09-30
Allow quoted arguments

Nenrikido:
This allows discord.py to allow commands with quoted arguments like
`!foo "bar" bar2 "bar3"`


Rapptz:
Hi.

Thanks for the PR.

What, pray tell, is this PR doing? It doesn't seem right.

Rapptz:
I see.

This is by design.

PR #1628, opened 2018-09-30
Allow custom cog names and multiple cog instances

uniflare:
Cogs can now specify a custom cog name when added
Additionally, this allows cog instances of the same type to be added with different names

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -517,7 +517,7 @@ class BotBase(GroupMixin):
 
     # cogs
 
-    def add_cog(self, cog):
+    def add_cog(self, cog, name=None):
         """Adds a "cog" to the bot.
 
         A cog is a class that has its own event listeners and commands.
@@ -536,9 +536,16 @@ class BotBase(GroupMixin):
         -----------
         cog
             The cog to register to the bot.
+        name
+            The name under which to register the cog. If not present, defaults to the name of the cog class.
         """
 
-        self.cogs[type(cog).__name__] = cog
+        cog_name = str(name) if name else type(cog).__name__
+        type(cog).cog_name = cog_name
+
+        assert cog_name not in self.cogs
+
+        self.cogs[cog_name] = cog
 
         try:
             check = getattr(cog, '_{.__class__.__name__}__global_check'.format(cog))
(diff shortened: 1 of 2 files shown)


Hornwitser:
Asserts are usually only used to guard against conditions that should not be possible, mostly as a sanity check.  Failing an assert generally means the _code doing the assertion is broken_, which is not the case here.

Secondly, Python will not automatically optimize out asserts.  They are either all checked at runtime, or all ignored entirely (which happens when the `-O` option is passed on the command line.)

---

PS: Did you even try the code you're proposing?
```
  File "test.py", line 95, in <module>
    bot.add_cog(cog)
  File "discord/ext/commands/bot.py", line 541, in add_cog
    cog_name = str(name) if name else type(cog).__name__
UnboundLocalError: local variable 'name' referenced before assignment
```

uniflare:
My apologies for the missing changes in this commit. The PR should be complete now.

Regarding the assert;
1. I do not understand why you wouldn't run your script without -O/-OO optimization in a live environment.
2. If you provide the same name twice you overwrite an original cog, this is broken code in my opinion since the original cog will not be accessible.

Hornwitser:
For no. 1 this isn't common because there's pretty much no measurable performance difference between running with `-OO` and running without it.  The libs discord.py uses doesn't have a lot of expensive asserts to optimize out.  Another reason is that the command lib's default help handler doesn't work with `-OO`, help texts are based on docstrings and `-OO` strips them out.

For no. 2 yes that is broken.  But using an assert to guard it isn't really appropriate as this is a user interface.   Having the traceback read
```
  File "my_bot.py", line 95, in <module>
    bot.add_cog(my_cog)
  File "discord/ext/commands/bot.py", line 542, in add_cog
    assert cog_name not in self.cogs
AssertionError
```
is not acceptable when something far more easily understood like `ValueError: Cog with name 'Example' is already added` can be raised instead.  Secondly, if you run python with `-O` the assertion is not done and you'll be back to the `add_cog` code being just as broken as before.

-----

Keep in mind that because of #426 this doesn't really allow multiple cog instances to be used as that is currently completely broken.

ir-3:
I think this has now been done in [caf3d17](https://github.com/Rapptz/discord.py/commit/caf3d17d4aa3ce45435bb96ea6d99317bfef8618)

PR #1629, opened 2018-09-30
Add GuildConverter

Tobotimus:
Hi!

I know I'm not the first to want `discord.Guild` to magically work as a converter. I can see #574 was closed, I'm sure there's a reason for it, I'm not really sure what it is though.

Thanks :smiley:


Tobotimus:
Thanks for the detailed explanation. I would argue none of those problems are limited to the guild converter and already exist in other converters, however I'm happy to withdraw this PR. I understand it's a simple converter - from my perspective this is all the more reason for it to be accepted upstream for the convenience of library users

PR #1634, opened 2018-10-02
Correct Guild.create_custom_emoji documentation

MusicOnline:
Changed limit of 50 local emotes to 50 static and animated emojis respectively.
Also added GIF to formats accepted.


SnowyLuma:
There should probably also be a note that guilds with the `MORE_EMOJI` feature can have up to 200 emoji of each type.

MusicOnline:
Added.

PR #1649, opened 2018-10-06
Allow flagging messages to keep them in the cache

DXsmiley:
Nobody else asked for this but here's a new feature. I'm very willing to accept comments on how this should be implemented if there's a more efficient / more idiomatic method.

**Rationale** (why I want this feature)
Bots are only able to respond edit events of messages that are in the cache. If a message is not in the cache, the bot is unable to respond to it being edited. My bot is able to tell in advance if it needs to watch for changes on particular messages. In this situation, it would be useful for these messages to be kept in the cache longer. With a small cache size, "important" messages are often pushed out by the constant stream of unimportant messages coming in. The current solution to this problem is to make the message cache larger, however this is memory intensive and only scales so far before it becomes impractical. This PR allows messages to be marked, so that they are not pushed out of the cache by unimportant.

**Notes**
I'm a little uneasy adding a mutable value to the Message class. Maybe maintaining a set of important IDs somewhere would be a better move, but this is open to discussion.

In the current implementation, there's a potential race-like condition where a message is pushed out of the "mixed messages" section of the cache before it can be marked as "important". This is likely to happen if the cache is too small, or there are a lot of messages coming in, or if the bot is only able to know if a message is important or not after a long-running computation, or a combination of the above.


Werseter:
What I have been doing for now, when needing a handler for possibly uncached messages, was having a check in the raw handler, for whether or not the message was in the cache, pushing it back in, and dispatching an event manually, just like the bot's backend would do it. But having such a behavior as a feature, and not a copy-pasteable solution wouldn't actually be bad. _(in my case, it mostly had to do with reaction events)_

But I do not expect this would have a usability on a more widespread level, rendering it quite a niche that may not warrant being useful enough for a push to the lib.

DXsmiley:
@EvieePy Thankyou for pointing that out. It solves my particular use case, and if this PR is knocked back I'll probably end up using it. However it raises the question of how much wheel-reinventing one would have to do if they wanted to use a custom cache for __all__ incoming events.

@Werseter Nicheness is one of my concerns here too. I had the idea of being able to [pass in a custom message cache](https://en.wikipedia.org/wiki/Dependency_injection), for complete customization, but that seemed like overkill.

Rapptz:
I don't exactly like this implementation.

Main reason being that now everyone suffers a penalty for what is essentially double lookup and double filtering which will just spike CPU usage up when for most people using the library this is a very niche use case that can be accomplished using the corresponding raw events.

There is something to be said about message edits being entirely non-trivial -- afterall, the field is entirely full of gotchas and partial information which would make it hard for the user to know what to do with a partial event such as `MESSAGE_EDIT`. 

The concept of message cache persistence has been brought up in the [official server](https://discord.gg/r3sSKJJ) a few times, but the main problem is that there hasn't really been a good way to do it without sacrificing something. I was thinking of changing the message cache into a LRU cache which would mitigate some issues but not all of them. 

The idea of having your own custom cache being injected through DI is pretty novel and I suppose a better idea but it kind of makes me wonder how I'd go about providing a "standardised" interface for people to inject with, since there could be a chance that the underlying data structure could change (such as a list to a dictionary).

PR #1651, opened 2018-10-08
Sizumita

sizumita:
**Embed add function**
add insert_field_in


Gorialis:
You have managed to merge async over rewrite in your own branch. While I question the motive of this PR in the first place, it has no chance of being merged like this.

PR #1656, opened 2018-10-10
Embed Limits

BluePhoenixGame:
Changed the documentation to add all the embed limits from [here](https://discordapp.com/developers/docs/resources/channel#embed-limits)

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -100,9 +100,11 @@ class Embed:
 
         self.colour = colour
         self.title = kwargs.get('title', EmptyEmbed)
+        if len(self.title) > 256: raise EmbedError('Embed title is over 256 limit.')
         self.type = kwargs.get('type', 'rich')
         self.url = kwargs.get('url', EmptyEmbed)
         self.description = kwargs.get('description', EmptyEmbed)
+        if len(self.descripẗion) > 2048: raise EmbedError('Embed description is over 2048 limit.')
 
         try:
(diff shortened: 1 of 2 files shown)


Rapptz:
Hi. Thanks for the PR.

I don't really like the wording used here. It doesn't really flow well with the rest of the prose. Chances are I'll just add it myself (since the commit message also doesn't really meet guidelines).

BluePhoenixGame:
Okay, that seems to be my problem here, that I can't find the documentation flow.

BluePhoenixGame:
Did this fix your problems with the PR

BluePhoenixGame:
Using error EmbedError will be easier for newstarters as getting a HTTPException will not make that much sense. This will also make it easy for other automatic stuff.

MusicOnline:
I don't think the library should be the one policing embed character limits. I mean, they may change at any time.

BluePhoenixGame:
Ah, yes that is true. Although we have many things documented that could change any time, like we have here.

BluePhoenixGame:
I now have documented EmbedError and how it is raised if the embed is over the limit

BluePhoenixGame:
Closing this PR because discord updated their error log. New error includes the info i wanted to give out with this PR: `discord.errors.HTTPException: BAD REQUEST (status code: 400): Invalid Form Body
In embed: Embed size exceeds maximum size of 6000`

PR #1659, opened 2018-10-10
[commands] Ensure removal of extensions if there is an error upon setting up

Phxntxm:


Rapptz:
This is not really atomic enough to solve the issue. A real atomic operation would put the bot state as it was before the error happened.

Phxntxm:
Good call, didn't think of that...simplest way to do that I think is the last change I just made then, add the library to the extensions dict then run `unload_extensions`

Harmon758:
Resolved by https://github.com/Rapptz/discord.py/commit/26e9b5bfac840c95744bbdd0ab356842e9efa72f.

PR #1671, opened 2018-10-15
clean up parenthesized await expressions

bmintz:
Removes superfluous parens surrounding await expressions.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -340,7 +340,7 @@ class BotBase(GroupMixin):
         if len(data) == 0:
             return True
 
-        return (await discord.utils.async_all(f(ctx) for f in data))
+        return await discord.utils.async_all(f(ctx) for f in data)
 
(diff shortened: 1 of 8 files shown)


BluePhoenixGame:
There is a reason for the parentece. If you know the priority rule in math then you would know that you have parentece as the number one priority. And it's kinda the same for python. The parentece are there since it's await that the function will get ran first before it gets returned, not risking any bug/error.

DavidArchibald:
No, they are not needed. Using the same math analogy you used: return is the lowest priority. The only case it would be needed is something such as `return (await my_coroutine())["prop"]` but that isn't a good usage either: it's doing too much on one line.
This entire PR likely boils down to a style opinion.

BluePhoenixGame:
Ah, yes. That is true. Well then I do approve of this.

Hornwitser:
If you look a bit into the history of that code you'll find it's left overs from when yield from was used (which did require parenthesis).

PR #1672, opened 2018-10-16
Change "command.can_run" to return False if any of the checks fail, rather than allowing the check to raise an exception.

XuaTheGrate:
Basically, running `command.can_run(ctx)` would currently raise an error if any of the checks fail.
Here's an example list comp I'm using to determine a list of every function the invoker can run.
```python
[c.name for c in self.bot.commands if await c.can_run(ctx)]
```
Normally, this will raise an error due to some of the commands not being accessible to certain people, i.e. an owner only command.
This request changes this behaviour to instead of raising an error, just returning False to show that the command can not be invoked.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -770,18 +770,13 @@ class Command:
         """|coro|
 
         Checks if the command can be executed by checking all the predicates
-        inside the :attr:`.checks` attribute.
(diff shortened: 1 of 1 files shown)


Rapptz:
No.

Raising is **by design** so the library propagates it to the error handlers.

PR #1677, opened 2018-10-17
add support for Bug Hunter and Early Supporter flags

bmintz:


bmintz:
## Test cases

245559087524216832
Bug Hunter: False
Early Supporter: True

110813477156720640
Bug Hunter: True
Early Supporter: True

493848183932321804
Bug Hunter: False
Early Supporter: False

PR #1682, opened 2018-10-22
Document `from_data`

DavidArchibald:
The docs includes the `to_dict`, but doesn't include `from_data`. This PR fixes that.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -110,6 +110,13 @@ class Embed:
 
     @classmethod
     def from_data(cls, data):
+        """Returns an embed.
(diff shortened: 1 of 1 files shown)


DavidArchibald:
This shouldn't be on the async branch, however switching the base branch doesn't give the expected result. So I will close the PR, and create a new one on the correct branch.

Hornwitser:
`Embed.from_data` is an internal function.  It is intentional that it's not documented.

DavidArchibald:
I will not open the PR on the correct branch then.

PR #1689, opened 2018-10-25
Add support for multiple activities

LostLuma:
Discord recently added an additional `activities` field to presences and allows members to have more than one activity at a time: https://github.com/discordapp/discord-api-docs/pull/707.

This causes some problems with the current single-activity based model as an activity that a user wants to interact with may not be the primary one and be inaccessible.

### DIFF
--- a/discord/member.py
+++ b/discord/member.py
@@ -142,7 +142,7 @@ class Member(discord.abc.Messageable, _BaseUser):
     status : :class:`Status`
         The member's status. There is a chance that the status will be a :class:`str`
         if it is a value that is not recognised by the enumerator.
-    activities: Tuple[Union[:class:`Game`, :class:`Streaming`, :class:`Activity`]]
+    activities: Tuple[Union[:class:`Game`, :class:`Streaming`, :class:`Spotify`, :class:`Activity`]]
         The activities that the user is currently doing.
     guild: :class:`Guild`
         The guild that the member belongs to.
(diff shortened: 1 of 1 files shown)


bmintz:
I haven't looked at the code yet, but these files should not be marked executable.

discord/member.py
```diff
@@ -142,15 +142,15 @@ class Member(discord.abc.Messageable, _BaseUser):
     status : :class:`Status`
         The member's status. There is a chance that the status will be a :class:`str`
         if it is a value that is not recognised by the enumerator.
-    activity: Union[:class:`Game`, :class:`Streaming`, :class:`Activity`]
-        The activity that the user is currently doing. Could be None if no activity is being done.
+    activities: Tuple[Union[:class:`Game`, :class:`Streaming`, :class:`Activity`]]
+        The activities that the user is currently doing.
```
bmintz: “are currently doing”

bmintz: I mean that it should say "user are currently doing." rather than "… is currently doing.", since "activities" is plural.

bmintz: wait
i think i'm wrong here. i rescind my comment cause english is hard

Rapptz:
I believe most users will have one activity. To account for that, I feel like minimising the breaking change is worth it. I suggest adding a `Member.activity` shortcut that fetches the top activity.

SnowyLuma:
I've gone ahead and added the changes from #1693 as it modifies a docstring that was removed in this PR.

PR #1695, opened 2018-10-29
SnowflakeList: rename has to __contains__

bmintz:

PR #1696, opened 2018-10-29
deduplicate SnowflakeList.get

bmintz:
Note: although cleaner, this way is slightly slower.

Conflicts with #1695.


Rapptz:
The way it is written is fine.

PR #1710, opened 2018-11-03
Update user.py

sKirua:
Added tag support to allow user to get the whole username + discriminator.


Gorialis:
`discriminator` has a chance of being an `int`. Nevertheless, this functionality is already implemented better in `__str__`, so I don't see the point in this.

https://github.com/Rapptz/discord.py/blob/async/discord/user.py#L71

BluePhoenixGame:
You probably wanna make it a `@property` too.

Rapptz:
Just cast the user instance to a string.

PR #1715, opened 2018-11-09
move Bot.is_owner to Client

bmintz:
I'm not sure why this was in BotBase to begin with.
It only depends on application_info, which is in Client.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -966,6 +966,25 @@ class Client:
                        bot_require_code_grant=data['bot_require_code_grant'],
                        owner=User(state=self._connection, data=data['owner']))
 
+    async def is_owner(self, user):
+        """Checks if a :class:`.User` or :class:`.Member` is the owner of
+        this bot.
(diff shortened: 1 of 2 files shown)


bmintz:
Why is that?

Rapptz:
Because it's not part of the `Client`? A client represents just that -- a client. The concept of an owner is something extra that only bots have. You can argue that `application_info` is similar but it's an API call and does not require extraneous storage on my side.

bmintz:
I don't understand. Is it the storage as an attribute that makes this part of Bot? Discord doesn't care whether you use Bot or Client, and so the concept of an owner applies to both, right?

Rapptz:
Yes it's the storage that changes this. The whole point of the `Bot` and `Client` separation is that if people don't want to use any of the command framework stuff then they shouldn't incur the cost of its existence and that attribute and extra non-Discord-specific function is part of the cost.

PR #1717, opened 2018-11-10
Made the wait_for_reaction coro doc less confusing

DevilXD:
The current "returns the first reaction that meets the requirements" information can be really confusing to someone reading the docs, since it isn't what it actually returns - only later down the line it's mentioned that it returns a "namedtuple with reaction and user attributes". Because of that, I wanted to propose a small change that would make it less confusing for other people using the docs.

This is for the current `discord.py`, not rewrite. I wanted to contribute so that people won't spend 2 hours like me figuring out why I'm getting something else than written in the docs.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -758,7 +758,7 @@ class Client:
         then the first reaction emoji that is in the list is returned. If ``None`` is
         passed then the first reaction emoji used is returned.
 
-        This function returns the **first reaction that meets the requirements**.
+        This function returns the **first result that meets the requirements**.
 
         Examples
         ---------


ClementJ18:
Async is feature frozen and will be deprecated once rewrite comes out. It is unlikely that this PR will be accepted or merged. If you wish to contribute to the docs you should do so for the rewrite branch.

DevilXD:
I understand, but this is just a doc change, so doesn't affect anything in the code - it's not related to any feature other than the method description used by/for developers only.

I am not using the rewrite, I didn't saw the docs there and I'm not interested in getting into it just yet, so I can't really contribute there. I wanted to make this less confusing for people still using this version of `discord.py` like me, so that they won't waste their time because of slightly inconsistent docs.

LyricLy:
It does return the first reaction, in the form of a namedtuple. I don't know what you were expecting it to return... just a string?

OrangutanGaming:
The last commit to async was practically a year ago. If the edit was applicable to rewrite then it might get merged, but wait_for was changed

DevilXD:
@LyricLy The docs say that it returns the first reaction that matches - there is nothing in that sentence about it being a tuple that also includes the user, so I expected a `discord.Reaction` object. I spent around 2 hours making a system fetching all reactions and checking for the user who reacted, only to realize that the user already gets returned by that coro as well, and the only thing I needed to do is to reference it properly.

@OrangutanGaming Is it bad? You sound like making a helpful update to a year old docs is something no one should ever do and it should never happen, but I disagree - this change does literally nothing to the current code functionality, other than making it less confusing for other people, so that they won't waste time looking for something that is there already up for a grab. That sentence is really misleading and discussing it with other people who code as well (mostly TwentySix), lead them to the same state I was in before reading the latter part of the docs, where "what it really returns" is explained.

ItWasAllIntended:
It's not about how helpful or insignificant the change is, the async branch does not receive changes. It's as simple as that

OrangutanGaming:
And also rewrite will overwrite async if it's not archived soon anyway

PR #1718, opened 2018-11-11
Translated some parts of api.po, migrating.po, commands api.po and commands.po.

mistio100:
As mentioned title.


mistio100:
Thank you for fixing my mistakes.
and I am so sorry to trouble you.

Gorialis:
Don't worry about it 👍

PR #1721, opened 2018-11-12
Add support for IDs in commands.has_role and commands.has_any role.

XuaTheGrate:
Basically, this will add support for role IDs in the following: 
```
commands.has_role
commands.has_any_role
commands.bot_has_role
commands.bot_has_any_role
```
As an example:
```python
@bot.command()
@commands.has_role(492212595072434186)
async def etc(ctx):
    pass

@bot.command()
@commands.bot_has_any_role("Moderator", 492212595072434186)
async def abc(ctx):
    pass
```

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1207,42 +1207,47 @@ def check(predicate):
         return func
     return decorator
 
-def has_role(name):
+def has_role(item):
     """A :func:`.check` that is added that checks if the member invoking the
-    command has the role specified via the name specified.
+    command has the role specified via the name or ID specified.
 
-    The name is case sensitive and must be exact. No normalisation is done in
-    the input.
+    If a string is specified, you must give an exact name of the role, including
+    caps and spelling.
+
+    If an integer is specified, you must give the exact snowflake ID of the role.
 
     If the message is invoked in a private message context then the check will
     return ``False``.
 
     Parameters
     -----------
-    name: str
-        The name of the role to check.
+    item: Union[int, str]
+        The name or ID of the role to check.
     """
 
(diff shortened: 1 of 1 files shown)


discord/ext/commands/core.py
```diff
@@ -1207,50 +1207,55 @@ def decorator(func):
...
 
-def has_role(name):
+def has_role(item):
     """A :func:`.check` that is added that checks if the member invoking the
-    command has the role specified via the name specified.
+    command has the role specified via the name or ID specified.
 
-    The name is case sensitive and must be exact. No normalisation is done in
-    the input.
+    If a string is specified, you must give an exact name of the role, including
```
crrapi: >If a string is specified, you must give an exact name of the role

This should probably be changed to `the exact name of the role`.

Gelbpunkt:
I like this change, it allows role checks for specific roles in a server instead of only simply a role name that works in any server and it makes some custom checks unnecessary.

discord/ext/commands/core.py
```diff
@@ -1207,50 +1207,55 @@ def decorator(func):
...
+    item: Union[int, str]
+        The name or ID of the role to check.
     """
 
     def predicate(ctx):
         if not isinstance(ctx.channel, discord.abc.GuildChannel):
             return False
 
-        role = discord.utils.get(ctx.author.roles, name=name)
+        if isinstance(item, int):
```
bmintz: I would do it like this:
```py
        if isinstance(item, int):
            role = discord.utils.get(ctx.author.roles, id=item)
        else:
            role = discord.utils.get(ctx.author.roles, name=item)
    
        return role is not None
```

PR #1725, opened 2018-11-14
Add pass_context=True to command decorator

Tadaboody:
Spoilers - it's all of them.

It was very frustrating finding this out from outside sources after a lot of trail and error

Not sure if the more advanced types of commands need editing


Vexs:
Eh? The rewrite branch (1.0.0a) removed pass_context completely as a kwarg. It doesn't do anything. 

You're likely confusing this with the async branch. (0.16.x)

PR #1739, opened 2018-11-25
Invites....

BluePhoenixGame:
Added support for the `discordapp.com/invite/` invites.
_(btw, in the preview the indents look weird for me. But here's how it looks when i edit the file. Dont know if it looks like that for you too.)_
image

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -172,10 +172,11 @@ class Client:
         if isinstance(invite, Invite) or isinstance(invite, Object):
             return invite.id
         else:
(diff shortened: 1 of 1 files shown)


Modelmat:
Why do you have 8 and six space indents?
Use four spaces.

Vexs:
Not only are the indents still buggered, but you could do this with a single regex....

Rapptz:
Thanks for the attempt but I can probably just do this one myself.

PR #1744, opened 2018-11-28
Proposed fix to STDIN audio

JellyWX:
Proposed fix to piping audio not working: feeding BytesIO objects in would raise an `io.UnsupportedOperation: fileno`, now raw binary can be fed in alongside a filename

### DIFF
--- a/discord/player.py
+++ b/discord/player.py
@@ -30,6 +30,7 @@ import audioop
 import logging
 import shlex
 import time
+import io
 
 from .errors import ClientException
 from .opus import Encoder as OpusEncoder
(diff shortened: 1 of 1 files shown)


JellyWX:
Ah yes, I presumed this wouldn't be the case and it'd share the attributes of Popen but clearly not.

I'm unaware of another way to implement this using just Subprocess

JellyWX:
Not  currently at my laptop, but this looks promising?
https://asyncio.readthedocs.io/en/latest/subprocess.html

JellyWX:
The only other ways I can think is to use some bash scripts which would probably also be OS dependant or something like a threadpool executor, however using this within the library is almost definitely not preferable since the rest of the library doesn't threadpool (?) so it'd make other efforts to remain async futile potentially (and I believe threadpool executors are CPU consuming).

I know how much data I'm sending in and out of the players via pipes, so I'm okay using my approach but I understand that it isn't a belonging approach in the rest of the library. There's this thing https://pypi.org/project/async_subprocess/ but that's only adding more dependancies and it was last updated in 2011.

I've hit a dead end since I don't know much about Async. I can transfer this to an issue and close the PR?

PR #1762, opened 2018-12-09
[Rewrite] member.has_role method

FoggyIO:
Returns whether the member has a role object or not.

This is useful for commands that result in different outputs based on role


Gorialis:
Calling this function takes up more characters, and is less pythonic, than just doing `role in member.roles` in the first place.

TheRockettek:
but this seems kinda useless, you could just do it in your bot code we dont need an extra function just for this

Rapptz:
Sorry, I don't see a point in having this as a built-in function. It goes against how things are done in Python (e.g. it's unpythonic).

PR #1772, opened 2018-12-13
Update Message.mention_everyone docs to include @here

gamescom15:
The documentation for Message.mention_everyone does not include @here, even though it is True if @here is mentioned.
The disocord UI also mentions both @here and @everyone.
14a9cd2b0b33cfe0d49bdb07ea433970

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -142,8 +142,8 @@ class Message:
 
         .. note::
 
-            This does not check if the ``@everyone`` text is in the message itself.
-            Rather this boolean indicates if the ``@everyone`` text is in the message
+            This does not check if the ``@everyone`` or the ``@here`` text is in the message itself.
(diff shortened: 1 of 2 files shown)


crrapi:
Although I don't oppose this PR, simply documenting `@everyone` would also imply that `@here` is included, as they essentially have the same purpose

discord/message.py
```diff
@@ -142,8 +142,8 @@ class Message:
 
         .. note::
 
-            This does not check if the ``@everyone`` text is in the message itself.
-            Rather this boolean indicates if the ``@everyone`` text is in the message
+            This does not check if the ``@everyone`` or the ``@here`` text is in the message itself.
+            Rather this boolean indicates if either the ``@everyone`` or the ``@here`` text is in the message
             **and** it did end up mentioning everyone.
```
```suggestion
            **and** it did end up mentioning.
```

Should probably update this part as well.

PR #1778, opened 2018-12-16
commands.clean_content: escape {{ spoilers }}

bmintz:


Rapptz:
I'm not sure why all braces have to escaped. Surely it should be double braces?

bmintz:
fixed. What should be done about ~, in that case?

bmintz:
Closing until the syntax is finalized.

fixator10:
Spoilers also now uses `|| tag ||`

PR #1781, opened 2018-12-16
Fix webhook file sending

diceroll123:
Currently, the webhook file-sending code will close the file(s) before they're sent, causing an I/O error.

To fix this, I had a look at discord.http and noticed the files it sends never get closed, so I've just had this do the same. 🤷‍♂️

---

I didn't want to go and change too much, but I'd like to add that Webhook's `request` differs from discord.http's for files.

Webhook file tuple: `(file.filename, file.open_file(), 'application/octet-stream')`
discord.http file tuple: `(file.open_file(), file.filename)` (and the content type`application/octet-stream` is added to the FormData directly)


diceroll123:
After looking into it more (and being told in the server 😂) this isn't a proper solution, and I see why. Files should be explicitly closed, and it slipped by me that they're closed outside of the method `discord.http.send_files`.

---

For future context since my explanation was bad:

The send code for Webhook is busted with AsyncWebhookAdapter with at least one file attached.
```python
try:
    to_pass = (file.filename, file.open_file(), 'application/octet-stream')
    return self._adapter.execute_webhook(wait=wait, file=to_pass, payload=payload)
finally:
    file.close()
```
`return self._adapter.execute_webhook` returns an unawaited coroutine, which immediately fires the `finally` statement, closing the file before it's sent, throwing an I/O error.

---

The `Webhook.send` method itself (and/or adapters) must be reworked to be more like `discord.http.send_files`, which is a big change that I'm sure Danny would prefer to do.

I'll see what I can do in the meantime, in any case.

PR #1783, opened 2018-12-16
Added User.can_dm()

CapnS:
I added support for users to check whether or not they can DM a user without actually sending them a message.

**Testing**:
I have tested this and it works as expected.

**Weaknesses**:
My doc-making skills are sub-par so these docs are probably wrong. 

_As always, all recommendations are welcome._

### DIFF
--- a/discord/user.py
+++ b/discord/user.py
@@ -555,7 +555,7 @@ class User(BaseUser, discord.abc.Messageable):
 
         Returns
         -------
-        :class:`Bool`
+        :class:`bool`
            Whether or not the bot can DM this :class:`User`
         """
         try:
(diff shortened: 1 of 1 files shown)


discord/user.py
```diff
@@ -550,6 +550,22 @@ def dm_channel(self):
         data = await state.http.start_private_message(self.id)
         return state.add_dm_channel(data)
 
+    async def can_dm(self):
+        """Determines if you can DM this :class:`User`
+
+        Returns
+        -------
+        :class:`Bool`
```
slice: this should be:

```suggestion
        :class:`bool`
```

discord/user.py
```diff
@@ -550,6 +550,22 @@ def dm_channel(self):
...
+           Whether or not the bot can DM this :class:`User`
+        """
+        try:
+            await self.send("")
+        except Exception as e:
```
slice: you should be catching `discord.HTTPException` here

Rapptz:
1. This implementation is hacky and subpar at best.
2. This goes as a violation of Python best practices which is to use EAFP (Easier to Ask For Permission than forgiveness) over LBYL (Look before you leap) style.

PR #1790, opened 2018-12-18
Add additional parameters to channel creation

TheTrain2000:
This PR adds parameters to both channel creation methods, `create_text_channel` and `create_voice_channel`, that were previously missing.

`create_text_channel` gained:
- `topic`: str, the channel's topic
- `slowmode_delay`: int, the channels' slowmode_delay
- `nsfw`: bool, the channel's nsfw status
- `position`: int, the channel's position in the channel list

`create_voice_channel` gained:
- `bitrate`: int, the channel's bitrate in bits per second
- `user_limit`: int, the user limit of the channel
- `position`: int, the channel's position in the channel list

Note:
Functionality surrounding the `position` parameters may not be intuitive to users. [See the following conversation that took place in DAPI/#api](https://canary.discordapp.com/channels/81384788765712384/381887113391505410/524675554176270336). I am looking for feedback on this parameter.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -555,7 +555,7 @@ class Guild(Hashable):
 
         return utils.find(pred, members)
 
-    def _create_channel(self, name, overwrites, channel_type, category=None, reason=None):
+    def _create_channel(self, name, overwrites, channel_type, category=None, **options):
         if overwrites is None:
             overwrites = {}
         elif not isinstance(overwrites, dict):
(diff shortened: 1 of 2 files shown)


discord/guild.py
```diff
@@ -580,11 +580,16 @@ def _create_channel(self, name, overwrites, channel_type, category=None, reason=
...
+        try:
+            options['rate_limit_per_user'] = options.pop('slowmode_delay')
+        except KeyError:
+            pass
+
         parent_id = category.id if category else None
-        return self._state.http.create_channel(self.id, name, channel_type.value, parent_id=parent_id,
-                                               permission_overwrites=perms, reason=reason)
+        return self._state.http.create_channel(self.id, channel_type.value, name=name, parent_id=parent_id,
+                                               permission_overwrites=perms, **options)
```
diceroll123: Deleted the reason?

TheTrain2000: `reason` is in `options`.

discord/http.py
```diff
@@ -508,17 +508,16 @@ def bulk_channel_update(self, guild_id, data, *, reason=None):
...
             'type': channel_type
         }
 
-        if permission_overwrites is not None:
-            payload['permission_overwrites'] = permission_overwrites
-
-        if parent_id is not None:
-            payload['parent_id'] = parent_id
+        valid_keys = ('name', 'parent_id', 'topic', 'bitrate', 'nsfw',
+                      'user_limit', 'position', 'permission_overwrites', 'rate_limit_per_user')
```
diceroll123: `reason` should be in here then, no?

TheTrain2000: It's in that function's definition.

PR #1805, opened 2018-12-28
Adds after_options to FFmpegPCMAudio

AngellusMortis:
Simple change to add new option to `FFmpegPCMAudio`: `after_options`.

`after_options` will allow you to add extra `ffmpeg` CLI args after all of the other ones that get made by `FFmpegPCMAudio`. A good use case (what I need it for) is save the `ffmpeg` source off to a file in addition to passing via `pipe` into to a `VoiceClient`.


AngellusMortis:
Closing as I figured out another way to do this know that I have figured out `ffmpeg` a lot.

PR #1808, opened 2018-12-30
Allow embeds from default help

Quantomistro3178:
When trying to override the format() function by extending the HelpFormatter, I found out that it wasn't possible to send Embed messages due to the inner functionality of the `_default_help_command()` function in `discord/ext/commands/bot.py`.

In this PR I tried to solve that by checking if the returned value from `<HelpFormatter>.format_help_for()` was an instance of discord.Embed.
I also implemented the `len()` function for the discord.Embed class, since it was also being used on the returned values from `<HelpFormatter>.format_help_for()`. This `len()` function returns the total length of strings from the author.name, footer.text, title, description, and all field names + field values of the Embed.

I also think it might be worth mentioning in the docstring that if a user decides to override `<HelpFormatter>.format()`, they should make sure that they return a list.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -473,3 +473,19 @@ class Embed:
             result['title'] = self.title
 
         return result
+
+    def __len__(self):
+        """Returns total length of strings from the Embed's author.name,
+        footer.text, description, title, all field names and all field
+        values.        
+        """
+        values = [f.name for f in self.fields]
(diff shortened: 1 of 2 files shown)


MusicOnline:
This won't be pulled because the async branch (v0.16.x) is completely frozen. Only the rewrite branch (v1.0.0a) will receive updates.

PR #1811, opened 2019-01-02
Add member.is_on_web() and member.is_on_desktop()

ir-3:


Rapptz:
Unlike the mobile methods, this one is wrong. I discussed the implementation of the method on the d.py server and the reason why that exists but not the other ones (there is a mobile indicator but there isn't an indicator for anything else). The implementation for these are too broad. If someone is away on desktop or web but active on mobile then these will return True. Whereas on mobile, there is no concept of being "away" since the connection gets terminated in 2 minutes or so, which is why the implementation takes advantage of this fact. 

So I don't see the point of adding these.

ir-3:
Okay, I understand what you mean. Thanks!

PR #1815, opened 2019-01-03
Implement PartialEmoji == Emoji (alternative to #1627)

bmintz:
I'm not sure I like that `__eq__` implementation, with its hardcoded `(Emoji, PartialEmoji)` tuple. I guess `tuple(ComparableEmoji.__subclasses__())` could be used instead, however, that's a CPython impl detail.

### DIFF
--- a/discord/emoji.py
+++ b/discord/emoji.py
@@ -29,7 +29,15 @@ from collections import namedtuple
 from . import utils
 from .mixins import Hashable
 
-class PartialEmoji(namedtuple('PartialEmoji', 'animated name id')):
+class ComparableEmoji:
+    def __eq__(self, other):
+        return isinstance(other, (Emoji, PartialEmoji)) and self.id == other.id
+
+    def __ne__(self, other):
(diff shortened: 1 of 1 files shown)


Ikusaba-san:
This would break `==` on unicode PartialEmojis (yes that's possible) -- those have `.name` defined but the `.id` is None.

bmintz:
OK, the new commit fixes that issue. Feedback is appreciated.
The following cog now works as expected:
```py
class EmojiEqTest:
    def __init__(self, bot):
        self.bot = bot

    async def on_ready(self):
        e = self.bot.emojis[0]
        pe = discord.PartialEmoji(animated=e.animated, name=e.name, id=e.id)
        assert pe == e
        joy = discord.PartialEmoji(animated=None, name='😂', id=None)
        assert joy != e
        joy2 = discord.PartialEmoji(animated=None, name='😂', id=None)
        assert joy2 is not joy and joy2 == joy
        print('tests pass')
```

PR #1823, opened 2019-01-06
Added Member.owner

OrangutanGaming:


MusicOnline:
Eh, you can compare using `Guild.owner` already.
Also, `Member.owner` is misleading. At least make it `Member.is_owner`.

Still, I don't think this is needed.

laggycomputer:
Agreed. I cannot think of many uses for this.

discord/member.py
```diff
@@ -381,7 +386,7 @@ def guild_permissions(self):
         administrator implication.
         """
 
-        if self.guild.owner == self:
+        if self.owner:
```
Helehelehele: You forgot to update this.

Also, ideally, you'd rebase your commits.

laggycomputer:
This makes an already very short comparison even shorter, and even potentially confusing.

Issue #1761, opened 2018-12-09
No way to edit webhook channel


Vexs:
There presently is no provided way to edit a webhook channel post creation as far as I can see- primarily because its the one thing you cannot edit when using tokens to edit webhooks. [Relevant documentation](https://discordapp.com/developers/docs/resources/webhook#modify-webhook-with-token)

As a workaround you can do
`client.http.request(Route('PATCH', '/webhooks/{webhook_id}', webhook_id=webhook.id), json={'channel_id':'123456789'})`

ExDomino:
The workaround is not working because name 'Route' is not defined.

apple502j:
@ExDomino `discord.http.Route`

ExDomino:
Thanks.

PR #1824, opened 2019-01-07
Add support for a channel in Webhook.edit

crrapi:
This adds an optional channel kwarg in Webhook.edit. Proposed solution for #1761


crrapi:
Closing this as I was informed (Thanks @CapnS) that this wouldn't work with token authentication.

PR #1832, opened 2019-01-12
discord.Activity application_id should be INT

BluePhoenixGame:
For some reason application_id is documented to being string. But in rewrite all other IDs are integers meaning that application_id should follow that too. I didn't even see any reason why it was string and not integer. So here we are with a pull request.

### DIFF
--- a/discord/activity.py
+++ b/discord/activity.py
@@ -94,7 +94,7 @@ class Activity(_ActivityTag):
 
     Attributes
     ------------
-    application_id: :class:`str`
+    application_id: :class:`int`
         The application ID of the game.
     name: :class:`str`
         The name of the activity.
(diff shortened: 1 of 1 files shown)


MusicOnline:
Right now, `discord.Activity.application_id` returns a string. I won't comment whether it should or not, however this PR is invalid if you incorrectly change the documentation.

discord/activity.py
```diff
@@ -138,7 +138,7 @@ def __init__(self, **kwargs):
         self.timestamps = kwargs.pop('timestamps', {})
         self.assets = kwargs.pop('assets', {})
         self.party = kwargs.pop('party', {})
-        self.application_id = kwargs.pop('application_id', None)
+        self.application_id = int(kwargs.pop('application_id', None))
```
Rapptz: If there is no `application_id` and it returns `None` then you do `int(None)` which will raise an exception.

Rapptz: This is why there is a helper for this code -- `utils._get_as_snowflake`

BluePhoenixGame: I did not know about this function until now and have now updated my code.

PR #1849, opened 2019-01-26
[RFC] [commands] custom default arguments

khazhyk:
Modeled slightly after Converters, allow specifying Converter-like class
for context-based default parameters. e.g.

```
class Author(ParamDefault):
  async def default(self, ctx):
    return ctx.author

async def my_command(ctx, user: discord.Member=Author):
  ...
```

### DIFF
--- a/discord/ext/commands/default.py
+++ b/discord/ext/commands/default.py
@@ -34,7 +34,19 @@ __all__ = (
     'Call',
 )
 
-class CustomDefault:
+class CustomDefaultMeta(type):
+    def __new__(cls, *args, **kwargs):
+        name, bases, attrs = args
+        attrs['display'] = kwargs.pop('display', name)
+        return super().__new__(cls, name, bases, attrs, **kwargs)
+
+    def __repr__(cls):
+        return str(cls)
+
+    def __str__(cls):
+        return cls.display
+
+class CustomDefault(metaclass=CustomDefaultMeta):
     """The base class of custom defaults that require the :class:`.Context`.
 
     Classes that derive from this should override the :meth:`~.CustomDefault.default`
--- a/discord/ext/commands/default.py
+++ b/discord/ext/commands/default.py
@@ -1,0 +1,88 @@
+# -*- coding: utf-8 -*-
+
+"""
+The MIT License (MIT)
+
+Copyright (c) 2015-2019 Rapptz
+
+Permission is hereby granted, free of charge, to any person obtaining a
+copy of this software and associated documentation files (the "Software"),
+to deal in the Software without restriction, including without limitation
+the rights to use, copy, modify, merge, publish, distribute, sublicense,
+and/or sell copies of the Software, and to permit persons to whom the
+Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+"""
+
+from .errors import MissingRequiredArgument
+
+__all__ = (
+    'CustomDefault',
+    'Author',
+    'CurrentChannel',
+    'CurrentGuild',
+    'Call',
+)
+
+class CustomDefault:
+    """The base class of custom defaults that require the :class:`.Context`.
+
+    Classes that derive from this should override the :meth:`~.CustomDefault.default`
+    method to do its conversion logic. This method must be a coroutine.
+    """
+
+    async def default(self, ctx, param):
+        """|coro|
+
+        The method to override to do conversion logic.
+
+        If an error is found while converting, it is recommended to
+        raise a :exc:`.CommandError` derived exception as it will
+        properly propagate to the error handlers.
+
+        Parameters
+        -----------
+        ctx: :class:`.Context`
+            The invocation context that the argument is being used in.
+        """
+        raise NotImplementedError('Derived classes need to implement this.')
+
+
+class Author(CustomDefault):
+    """Default parameter which returns the author for this context."""
+
+    async def default(self, ctx, param):
+        return ctx.author
+
+class CurrentChannel(CustomDefault):
+    """Default parameter which returns the channel for this context."""
+
+    async def default(self, ctx, param):
+        return ctx.channel
+
+class CurrentGuild(CustomDefault):
+    """Default parameter which returns the guild for this context."""
+
+    async def default(self, ctx, param):
+        if ctx.guild:
+            return ctx.guild
+        raise MissingRequiredArgument(param)
+
+class Call(CustomDefault):
+    """Easy wrapper for lambdas/inline defaults."""
+
+    def __init__(self, callback):
+        self._callback = callback
+
+    async def default(self, ctx, param):
+        return self._callback(ctx, param)
--- a/discord/ext/commands/default.py
+++ b/discord/ext/commands/default.py
@@ -24,6 +24,8 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 DEALINGS IN THE SOFTWARE.
 """
 
+import discord
+
 from .errors import MissingRequiredArgument
 
 __all__ = (
@@ -49,9 +51,11 @@ class CustomDefaultMeta(type):
 class CustomDefault(metaclass=CustomDefaultMeta):
     """The base class of custom defaults that require the :class:`.Context`.
 
-    Classes that derive from this should override the :meth:`~.CustomDefault.default`
-    method to do its conversion logic. This method must be a coroutine.
+    Classes that derive from this should override the :attr:`~.CustomDefault.converters` attribute to specify
+    converters to use and the :meth:`~.CustomDefault.default` method to do its conversion logic.
+    This method must be a coroutine.
     """
+    converters = (str,)
 
     async def default(self, ctx, param):
         """|coro|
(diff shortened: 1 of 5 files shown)


Gelbpunkt:
@Rapptz any update/statement on this? Would make some discord.py code far prettier and seems a not-breaking change

khazhyk:
I might have time to finish this up this week/end

khazhyk:
- Added example and docs
- Fixed issue where non-specified converter type would default to `type` instead of `str`

Would appreciate others testing and comments.

Gelbpunkt:
I may have found a bug?
```py
from discord.ext.commands.default import Author

@bot.command()
async def testme(ctx, thing: discord.Member = Author):
    await ctx.send(thing.joined_at)
```
This raises `AttributeError: type object 'Author' has no attribute 'joined_at'`.
When I print `dir(thing)`, it's all the proper attributes there, even `joined_at`

khazhyk:
I can't reproduce ( see https://canary.discordapp.com/channels/336642139381301249/381963689470984203/574499828172587038 and following) - the behavior your experiencing seems like the command framework isn't evaluating the customdefaults, does this behavior persist past restart?

khazhyk:
[Ah - ctx.invoke() is what isn't working](https://canary.discordapp.com/channels/336642139381301249/381963689470984203/574501979678572544)

Gelbpunkt:
Was an issue on my end - the code proposed in the PR works fine. Thanks!

khazhyk:
Pushed an update to fix a missing `isclass` check which could cause some commands with defaults to break... whoops

Gelbpunkt:
I also found this to look weird on `command.signature`. Here's what it looks like:
`[argument=<class 'discord.ext.commands.default.Author'>]`

Is there a way to prettify that or do I have to write my own signature generator?

NCPlayz:
The classes should probably define their own `__repr__` to *prettify* that

Gelbpunkt:
Yea, I know that. Not sure if they want to write a new repr for the "prepacked" defaults and what it would be, though. `[argument=Author]` isn't very user-friendly either...

MinerChAI:
@khazhyk I created a PR to your fork to fix the problem with signature https://github.com/khazhyk/discord.py/pull/2

khazhyk:
Does the default HelpCommand already have special casing for Converter? It would make sense to treat it similarly to Converter (which does not override __repr__ just for HelpCommand)

MinerChAI:
Type hints (including converters) are not displayed in help lol

MinerChAI:
I tested your example from docs and there are some name errors

MinerChAI:
When I ran it after fixing them I got ```
!echo_image [image=LastImage]

```
Image converter isn't there

**Comment by khazhyk** [comment]
hmm, still not sure if this should be handled by overwriting __repr__ rather than being handled in the help formatter

**Comment by Gelbpunkt** [comment]
Handling it in the help formatter won't fix `command.signature` yet. I personally would prefer a `CustomDefault.name` attribute/property that every class could set. Therefore, signatures would need to be changed.

**Comment by Rapptz** [comment]
Every type implements this via `__str__`.

If you want to play nice with the current system then implement `__str__`. As a default you can have ```py
def __str__(self):
    return self.__class__.__name__
```

MinerChAI:
Oh I fordot `__str__`... But is it necessary?

Rapptz:
It's literally how every type does it. Why should this one be any special?

MinerChAI:
Ok

MinerChAI:
Added. 
@khazhyk please merge my PR

docs/ext/commands/commands.rst
```diff
@@ -585,6 +585,60 @@ handlers that allow us to do just that. First we decorate an error handler funct
...
 :exc:`~ext.commands.CommandError`. A list of errors is found in the :ref:`ext_commands_api_errors` page of the documentation.
 
+
+Custom Defaults
+---------------
+
+Custom defaults allow us to specify :class:`.Context`-based defaults. Custom defaults are always classes which inherit from
+:class:`.CustomDefault`.
+
+The library provides some simple default implementations in the :module:`default` - :class:`default.Author`, :class:`default.Channel`,
```
NCPlayz:
```suggestion
The library provides some simple default implementations in ``ext.commands.default`` - :class:`.default.Author`, :class:`.default.Channel`,
```

`:module:` is not a valid tag... Is this necessary? (won't make a difference if `:mod:` is used).

docs/ext/commands/commands.rst
```diff
@@ -585,6 +585,60 @@ handlers that allow us to do just that. First we decorate an error handler funct
...
+and :class:`default.Guild` returning the corresponding properties from the Context. These can be used along with Converters to
```
NCPlayz:
```suggestion
and :class:`.default.Guild` returning the corresponding properties from the Context. These can be used along with Converters to
```

docs/ext/commands/commands.rst
```diff
@@ -585,6 +585,60 @@ handlers that allow us to do just that. First we decorate an error handler funct
...
+simplify your individual commands. You can also use :class:`default.Call` to quickly wrap existing functions.
```
NCPlayz:
```suggestion
simplify your individual commands. You can also use :class:`.default.Call` to quickly wrap existing functions.
```

docs/ext/commands/commands.rst
```diff
@@ -585,6 +585,60 @@ handlers that allow us to do just that. First we decorate an error handler funct
...
+
+A DefaultParam returning `None` is valid - if this should be an error, raise :class:`.MissingRequiredArgument`.
```
NCPlayz:
```suggestion
A DefaultParam returning ``None`` is valid - if this should be an error, raise :class:`.MissingRequiredArgument`.
```

docs/ext/commands/commands.rst
```diff
@@ -585,6 +585,60 @@ handlers that allow us to do just that. First we decorate an error handler funct
...
+
+.. code-block:: python3
```
NCPlayz: I feel like specific lines should be highlighted for this as it is a long codeblock.

MinerChAI: Btw how do you highlight them

NCPlayz:
```rst
.. code-block:: python3
    :emphasize-lines: 1,2
        
        print('hello world!')
        print('test.')
```

khazhyk:
Thanks for the review

discord/ext/commands/default.py
```diff
@@ -0,0 +1,100 @@
...
+        raise MissingRequiredArgument(param)
+
+class Call(CustomDefault):
+    """Easy wrapper for lambdas/inline defaults."""
+
+    def __init__(self, callback):
+        self._callback = callback
+
+    async def default(self, ctx, param):
+        return self._callback(ctx, param)
```
NCPlayz: I feel like it would be good if this supported async functions as well, although it isn't a strong enough opinion.

avayert:
I don't like the duplicate name for `Guild` (and to a certain extent `Channel`) as it already exists as `discord.Guild`. `discord.Guild` also has converter support built into it in the commands API which would make this inconsistent and a little confusing, even when using the namespaced `default.Guild`.

I don't currently know how or if it would even made sense to attach the defaults to said class, but a simple fix could be to rename them to `CurrentGuild` and `CurrentChannel`, which I find makes the signature more comprehensible to read.

bmintz:
Attaching the defaults to the classes would be far worse IMO, as you'd end up with signatures like "x: discord.Guild = discord.Guild" which to me.is very confusing. I think it would be better to rename them to Current\* yes.

discord/ext/commands/default.py
bmintz: I disagree. In order to use this with a new async function you'd have to define it outside of the command signature anyway. At that point just make a Default class. If you want this to support async functions, I suggest you file a PEP for async lambdas :^)

khazhyk:
> I don't currently know how or if it would even make sense to attach the defaults to said class, but a simple fix could be to rename them to `CurrentGuild` and `CurrentChannel`, which I find makes the signature more comprehensible to read.

Doing `CurrentGuild` and `CurrentChannel` makes sense to me, done.

MinerChAI:
Why isn't it merged

TheRockettek:
Because it needs to be reviewed. The only that there is my dude

sairam4123:
Is this dead?

MinerChAI:
It’s still waiting for review

mouncg:
Hi chai

sairam4123:
Because it has been 1 year

khazhyk:
Rebased this. I recall some people claimed issues between customdefaults and some of the newer argument types (greedy, etc.). Is that the case/still the case? Is there an example that behaves weirdly?

sairam4123:
Let me test the code.

sairam4123:
Just a suggestion, instead of Author, you could do, `CurrentAuthor` would be nice to see.

sairam4123:
I got an error when trying to use Greedy with this
```py
Ignoring exception in command my_command:
  File "F:\PyCharm Python Works\OpenCityBot\discord.py\discord\ext\commands\core.py", line 86, in wrapped
  File "F:\PyCharm Python Works\OpenCityBot\discord.py\test.py", line 14, in my_command
    await ctx.send(member.display_name)
AttributeError: type object 'Author' has no attribute 'display_name'

, line 903, in invoke
    await ctx.command.invoke(ctx)
  File "F:\PyCharm Python Works\OpenCityBot\discord.py\discord\ext\commands\core.py", line 872, in invoke                                                             , line 903, in invoke
    await injected(*ctx.args, **ctx.kwargs)
  File "F:\PyCharm Python Works\OpenCityBot\discord.py\discord\ext\commands\core.py", line 872, in invoke", line 95, in wrapped
    raise CommandInvokeError(exc) from exc                                         ", line 95, in wrapped
discord.ext.commands.errors.CommandInvokeError: Command raised an exception: AttributeError: type object 'Author' has no attribute 'display_name'    
```
Here's the code I used.
```py
@bot.command()
async def my_command(ctx, member: commands.Greedy[discord.Member] = Author, channel: discord.TextChannel = CurrentChannel, guild = CurrentGuild):
    await ctx.send(member.display_name)
    await ctx.send(channel.name)
    await ctx.send(guild.name)
```

khazhyk:
> 
> 
> Just a suggestion, instead of `Author`, you could do, `CurrentAuthor` would be nice to see.

The reason I went for "Author" instead of "CurrentAuthor" is that Author I think should be unambiguous, vs. Channel/Guild, which would be ambiguous. And stuff like CurrentUser also could be misunderstood. So CurrentAuthor seems too wordy, imo, but does look slightly more consistent. I'm indifferent, do others have thoughts on this?

khazhyk:
@sairam4123 figured out and fixed the issue with Greedy, thanks! That should work now.

khazhyk:
I also added a commit from @sgtlaggy which I thought was useful, although there could be some bikeshedding to do here. In particular, I'm not so sure that having a member variable to override is the best api. 

one option, which I don't think works in this current PR, is if the defaultargument was both a converter and a default, then detect if the default is (also) a converter, and use that, e.g.

```py

class MyDefault(DefaultArgument, MyConverter):
  ...
```

Tari-dev:
If we are going to add this couldn't we support attachment with ext.commands?
Also we could support for slash-ext hybrid?

khazhyk:
Sorry, as evidenced by my lack of touching this for a while, I'm not really motivated to push this to completion. I still think something like this is a good idea, and if anyone else wants to implement something like this (or polish this code up for the lib), I would be happy to see it.

Issue #1770, opened 2018-12-12
Cannot send local files via webhooks.


XuaTheGrate:
As the name suggests, you cannot send files on your disk via webhooks.
Running the following code: (file.txt is a local file)
```python
await webhook.send(file=discord.File("file.txt"))
```
Will raise the following traceback:
```python
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/discord/ext/commands/core.py", line 62, in wrapped
    ret = await coro(*args, **kwargs)
  File "/home/fedora/Thanatos/modules/test.py", line 9, in test
    await webhook.send(file=discord.File("file.txt"))
  File "/usr/local/lib/python3.7/site-packages/discord/webhook.py", line 194, in handle_execution_response
    data = await response
  File "/usr/local/lib/python3.7/site-packages/discord/webhook.py", line 162, in request
    async with self.session.request(verb, url, headers=headers, data=data) as r:
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/client.py", line 855, in __aenter__
    self._resp = await self._coro
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/client.py", line 361, in _request
    ssl=ssl, proxy_headers=proxy_headers, traces=traces)
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/client_reqrep.py", line 233, in __init__
    self.update_body_from_data(data)
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/client_reqrep.py", line 405, in update_body_from_data
    body = body()
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/formdata.py", line 141, in __call__
    return self._gen_form_data()
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/formdata.py", line 135, in _gen_form_data
    self._writer.append_payload(part)
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/multipart.py", line 745, in append_payload
    size = payload.size
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/payload.py", line 284, in size
    return os.fstat(self._value.fileno()).st_size - self._value.tell()
ValueError: I/O operation on closed file
```
This issue does also occur when using the `files` kwarg as well.
Side note that using io objects (BytesIO, StringIO, etc) do work normally.

PR #1861, opened 2019-01-30
Fixed webhook local file send

CapnS:
This fixes the problem from #1770. All I did was change the way files are closed and instead of closing them in a finally (which runs before the return and therefore before the file is actually used), I close them right after the file is used. I have tested this a few times and it works with both the file and files kwargs.

It isn't the cleanest because I had to add another kwarg to get the file/files to handle_execution_response() so I could close them right after the files are used. If anyone has any style suggestions, or any suggestions in general, I am open to criticism.

### DIFF
--- a/discord/webhook.py
+++ b/discord/webhook.py
@@ -123,7 +123,10 @@ class WebhookAdapter:
             multipart = None
         url = '%s?wait=%d' % (self._request_url, wait)
         maybe_coro = self.request('POST', url, multipart=multipart, payload=data)
-        return self.handle_execution_response(maybe_coro, wait=wait, f=f)
+        if isinstance(self.webhook._adapter, RequestsWebhookAdapter):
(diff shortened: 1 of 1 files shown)


CapnS:
I changed up how the webhook sends, which fixes the problem in a different way.

discord/webhook.py
```diff
@@ -104,7 +104,7 @@ def store_user(self, data):
         # mocks a ConnectionState for appropriate use for Message
         return BaseUser(state=self.webhook._state, data=data)
 
-    def execute_webhook(self, *, payload, wait=False, file=None, files=None, f=None):
+    async def execute_webhook(self, *, payload, wait=False, file=None, files=None):
```
Hornwitser: Making this a coroutine breaks the RequestWebhookAdapter.

PR #1866, opened 2019-02-01
commands.clean_content: escape || spoilers ||

bmintz:


khazhyk:
Won't this break code blocks with || in them

shikhir-arora:
Yeah, you'd probably want something along these lines if you want to check specifically:

`input = 'some test text'`
`match = re.match(r"(\|{2}(.*)\|{2})",  input)` 
`sanitize = re.sub(r"(\|{2}(.*)\|{2})", '"\u200B"', input)`

https://regexr.com/47hme

bmintz:
Unmatched \~, \*, and \` are already escaped. I see no need to start only escaping matched spoilers.

Rapptz:
I wonder if it would be better to escape things only if they're matched.

PR #1868, opened 2019-02-01
Fix rich presence asset image resolution

kairusds:


ir-3:
Correct me if I am wrong but, aren't the assets already the necessary sizes?

kairusds:
@ir-3 not its not, it only returns low res images

ir-3:
I think those are just for viewing the specific assets for both the large and small rich presence images, not for changing the size of them.

kairusds:
@ir-3 oh okay

PR #1869, opened 2019-02-03
Remove notes about Client.messages

XuaTheGrate:
Since `Client.messages` no longer exists, I think we should remove the note about the cache being named that.


gamescom15:
https://github.com/Rapptz/discord.py/projects/1#card-3199584
As a note, the decision to keep Client.messages removed is not final

PR #1876, opened 2019-02-04
Use non-deprecated method of acquiring lock

Tobotimus:
Sometimes the current use of `await lock` here prints this deprecation warning in the console on python 3.7:
```
discord/http.py:155: DeprecationWarning: 'with await lock' is deprecated use 'async with lock' instead
```
What `await lock` does is simply call `await lock.acquire()` and return a context manager. In this case the return value of `await lock` is being discarded anyway, so I see no reason not to make this change.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -148,7 +148,7 @@ class HTTPClient:
             # wait until the global lock is complete
             await self._global_over.wait()
 
-        await lock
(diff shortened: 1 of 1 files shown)


Rapptz:
This is a left-over from the `yield from` days. Good catch. Thanks.

PR #1884, opened 2019-02-06
Fix ext.commands help page full-width indentation

### LINKED ISSUE #1883: [ext.commands]help page incorrect indentation(in case of full-width named command)
In rewrite branch, ext.commands Bot help page may show some incorrect indentation when bot command name has a full-width characters.

I need: :laughing:

Now: :sob:

or I shouldn't give command name that contains full-width characters?

cod-sushi:
This fixed #1883.

### DIFF
--- a/discord/ext/commands/formatter.py
+++ b/discord/ext/commands/formatter.py
@@ -26,6 +26,7 @@ DEALINGS IN THE SOFTWARE.
 
 import itertools
 import inspect
+import discord.utils
 
 from .core import GroupMixin, Command
 from .errors import CommandError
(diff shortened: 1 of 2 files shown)


Vexs:
It seems wise to me to either make this an opt-in flag, as this won't affect the vast majority of discord.py users, or just providing a note in the helpformatter class.

As an optimization, checking if the char is ascii before checking width results in an overall speedup and replacing `UNICODE_WIDE_CHAR_TYPE` with a set() also helps. You can further improve this by using a sum() wrapping a list comp. Additionally, the `u` prefix on `UNICODE_WIDE_CHAR_TYPE` is vestigial.

Rapptz:
I'm going to merge this but I'm going to do some optimisations myself. Thanks for the patch!

Issue #610, opened 2017-06-23
Add Reaction.remove() on rewrite


Nekorooni:
Is this something to consider so you don't have to call `await message.remove_reaction(reaction.emoji, user)`?
Or would this not be convinient since there's no user linked to the Reaction?

Mainly for this:
```py
r, u = await self.bot.wait_for('reaction_add', check=lambda r, u: u.id == ctx.author.id)
await m.remove_reaction(r.emoji, u)
# or
await r.remove()
```
I can imagine this functionality not being worth the hassle of recoding though.

Harmon758:
This was added with https://github.com/Rapptz/discord.py/commit/806295a1be2cfb0ebc8659aa9abb605bd59ff3b6 (#1891).

BatteTarte:
this dont work

Ender2K89:
you do realize this issue is from 2019.. right?

PR #1888, opened 2019-02-07
Adding Reaction.remove()

CapnS:
This addresses #610 about adding a .remove() to Reactions. This will make it cleaner and easier to remove reactions from their messages. I have tested this and it works as intended. As always, suggestions and criticisms are always welcome!

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -679,7 +679,7 @@ class Message:
             The emoji parameter is invalid.
         """
 
-        emoji = self._emoji_reaction(emoji)
+        emoji = utils.emoji_reaction(emoji)
         await self._state.http.add_reaction(self.id, self.channel.id, emoji)
 
     async def remove_reaction(self, emoji, member):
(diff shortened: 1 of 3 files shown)


Nekorooni:
I think it's better to use .format() instead of % style formatting.

CapnS:
> I think it's better to use .format() instead of % style formatting.
> 
> `'{0.name}:{0.id}'.format(emoji)`

I copied that static method straight from [message.py](https://github.com/Rapptz/discord.py/blob/rewrite/discord/message.py#L724-L736) where it does the same thing.

laggycomputer:
If we want to establish a clear library standard (examples use `.format`, some lib code uses `%s`, that's another issue. `%s` is fine for now.)
.

Vexs:
I'm reasonably certain this won't work. 

Your addition of `user_id` ties the reaction object to a specific user- which, in this case, I suspect is just going to be the first person who added the reaction originally. (Or, if removed, the second person, etc. Whatever discord dispatches.) Reaction objects as it stands are not tied to users- instead just having a count, and a method to get reaction users. This is why [message.remove_reaction](https://discordpy.readthedocs.io/en/rewrite/api.html#discord.Message.remove_reaction) takes an abc.Snowflake.
In light of that, you need to consider the NotFound exception in your documentation as well. 

This also fails to cover conditions where you are removing your own reaction. 

Additionally, I find the example fairly pointless. It's a rather self-explanatory method.

CapnS:
It has been brought to my attention that this will not work. I only tested it with myself and did not realize that this will not work if there are more than one of the same reaction on the message.

PR #1889, opened 2019-02-07
Removed fixed string from HelpFormatter

### LINKED ISSUE #1886: Add property to HelpFormatter for i18n
I'm developing japanese language discord bot.

`ext.commands.Bot` class has `command_not_found` and `command_has_no_subcommands` property.
Document says, They are useful for i18n. I use them usefully.

But, in `HelpFormatter` class's `get_ending_note` method, message `Type {0}{1} command for more info on a command. You can also type {0}{1} category for more info on a category.` is fixed text.
It's not useful for i18n.
I want `help_more_info` property to `HelpFormat` class, like `command_not_found` of `Bot`.

cod-sushi:
Removed fixed strings `"Commands:"` and help page ending note(`"Type ?help command for more info on a command ..."`)
and added properties modify these strings.
default behavior is not changed. fix #1886

I wrote these codes referencing code about `Bot.command_prefix` .

I added comment about added properties too.(for document)
I apologize if there are any mistakes in my English and document writing.

### DIFF
--- a/discord/ext/commands/formatter.py
+++ b/discord/ext/commands/formatter.py
@@ -145,13 +145,17 @@ class HelpFormatter:
     commands_heading: :class:`str`
         The command list's heading string used when the help command is invoked with a categoly name.
         Useful for i18n. Defaults to ``"Commands:"``
+    no_categoly: :class:`str`
+        The string used when there is a command which does not belong to any categoly(cog).
+        Useful for i18n. Defaults to ``"No Categoly"``
     """
     def __init__(self, show_hidden=False, show_check_failure=False, width=80,
-                 commands_heading="Commands:"):
+                 commands_heading="Commands:", no_categoly="No Categoly"):
         self.width = width
         self.show_hidden = show_hidden
         self.show_check_failure = show_check_failure
         self.commands_heading = commands_heading
+        self.no_categoly = no_categoly
 
     def has_subcommands(self):
         """:class:`bool`: Specifies if the command has subcommands."""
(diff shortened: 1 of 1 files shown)


discord/ext/commands/formatter.py
```diff
@@ -142,11 +142,31 @@ class HelpFormatter:
...
+        The command list's heading string used when the help command is invoked with a categoly name.
+        Useful for i18n. Defaults to ``"Commands:"``
     """
-    def __init__(self, show_hidden=False, show_check_failure=False, width=80):
+    def __init__(self, show_hidden=False, show_check_failure=False, width=80,
+                 ending_note=None, commands_heading="Commands:"):
         self.width = width
         self.show_hidden = show_hidden
         self.show_check_failure = show_check_failure
+        self.ending_note = ending_note
```
Rapptz: I feel like this shouldn't be an attribute but something that should be changed through inheritance and subclassing. This is why it's in a separate method, so you can do something like the following:

```py
class MyHelpFormatter(commands.HelpFormatter):
    def get_ending_node(self):
      return '...'
```

I do agree with the `commands_heading` though so I think that's good.

cod-sushi:
Thanks for reviewing.
I removed `ending_note` property.
And, I notice there is "No Categoly", so I added property `no_categoly` to HelpFormatter to remove this fixed text.

Vexs:
Typo: spelled `category` not `categoly`

cod-sushi:
so embarrassed... I'll fix it soon. thank you vex :sob:

Rapptz:
Thanks for the pull request!

Issue #1738, opened 2018-11-25
If a Spotify song being listened to is longer than 128 characters, member.activity returns None instead


numbermaniac:
Another user was initially having this issue in the discord server, which I then was able to reproduce.

https://open.spotify.com/track/5YnS9oWrGcwf1tgode1F7z?si=bp0Zkk3qSF68yxSZjOuVzA

This song's name is exactly 128 characters long, and if you have it playing while connected to Discord, `member.activity` will correctly return a `discord.activity.Spotify` instance with the song details.

https://open.spotify.com/track/47dPsSo7cEDNkvIOSB4O2k?si=id9VCJHiSiWpedthzYQcUA

This song title, however, is 129 characters long, and if the user is listening to this song while connected to Discord, `member.activity` will always return `None`.

[activity.py](https://github.com/Rapptz/discord.py/blob/rewrite/discord/activity.py#L41) indicates that this may be an intentional limit. Is this a Discord API limitation or an issue with discord.py? If it's an API limitation, perhaps it should be noted in the documentation as this could be an unexpected surprise.

Rapptz:
Right. This is an API limitation, not one that has to do with the library. The library just presents the information as-is.

numbermaniac:
Thanks for the answer! Since it's an API limitation, maybe a note should be added to the documentation somewhere warning of this behaviour?

SilicalNZ:
Music said this well and I feel like it applies to this situation as well.
https://github.com/Rapptz/discord.py/pull/1656#issuecomment-428910927

PR #1890, opened 2019-02-07
Adding docs about Member.activity returning None with long song names

CapnS:
These docs address #1738 about Member.activity returning None if a Spotify song name is longer than 128 characters. As always, suggestions and criticisms are welcome.


Rapptz:
There are a lot more limits than just the song name. I figure it would probably be best to have a more generic sounding note that it can be `None` if these limits are found, but when I see the wording like this it makes me wonder if this is a bug within the library after all.

PR #1891, opened 2019-02-08
Add `remove()` to Reaction

EvieePy:
Added a coro, `remove()` which takes in a sole parameter, `user`.
This new coro will remove the reaction by the provided user/member from the reactions message.

`message.remove_reaction(reaction, member)` was not removed as to not introduce breaking changes.

### DIFF
--- a/discord/reaction.py
+++ b/discord/reaction.py
@@ -93,6 +93,29 @@ class Reaction:
     def __repr__(self):
         return '<Reaction emoji={0.emoji!r} me={0.me} count={0.count}>'.format(self)
 
+    async def remove(self, member):
+        """|coro|
+
+        Remove the reaction by the provided :class:`discord.member.Member` from the message.
+
+        If the reaction is not your own (i.e. ``member`` parameter is not you) then
+        the :attr:`discord.permissions.Permissions.manage_messages` permission is needed.
+
+        The ``member`` parameter must represent a member and meet
+        the :class:`abc.Snowflake` abc.
+
+        Raises
+        -------
+        HTTPException
+            Removing the reaction failed.
+        Forbidden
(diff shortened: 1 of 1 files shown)


discord/reaction.py
```diff
@@ -93,28 +93,33 @@ def __str__(self):
     def __repr__(self):
         return '<Reaction emoji={0.emoji!r} me={0.me} count={0.count}>'.format(self)
 
-    async def remove(self, member):
+    async def remove(self, user):
         """|coro|
 
-        Remove the reaction by the provided :class:`discord.member.Member` from the message.
+        Remove the reaction by the provided user from the message.
```
gamescom15:
```suggestion
        Remove the reaction by the provided :class:`User` from the message.
```

gamescom15: not sure if the reference needs to be absolute like :class:`discord.user.User`

discord/reaction.py
```diff
@@ -93,28 +93,33 @@ def __str__(self):
...
+
         Raises
         -------
         HTTPException
             Removing the reaction failed.
         Forbidden
             You do not have the proper permissions to remove the reaction.
         NotFound
-            The member you specificed, or the reactions message was not found.
+            The user you specificed, or the reactions' message was not found.
```
crrapi: `specified`, and `reaction's`.

BluePhoenixGame:
Honestly, i think we should remove `message.remove_reaction(reaction, member)` as it does not follow the way that rewrite is built like how all the other functions are made. We do `guild.get_member(id)`, we do `emoji.delete(reason)` - so why should we do `message.remove_reaction(reaction, member)` and not `reaction.remove(member)`. But a poll on that one might be needed.

Rapptz:
Thanks.

PR #1894, opened 2019-02-11
Bump aiohttp requirement and update AsyncWebhookAdapter to use non-deprecated property

bryanforbes:
3.5 brings a bunch of bug fixes and only one small API change that affects this library

Relevant changes: https://github.com/aio-libs/aiohttp/blob/f6f647eb828fa738610d61481f11fa51e42599e9/CHANGES.rst#350-2018-12-22

### DIFF
--- a/discord/webhook.py
+++ b/discord/webhook.py
@@ -141,7 +141,7 @@ class AsyncWebhookAdapter(WebhookAdapter):
 
     def __init__(self, session):
         self.session = session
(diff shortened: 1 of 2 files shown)


Rapptz:
Thanks for the PR.

Since I'm using 3.5.2+ now, it's probably better to just use `asyncio.get_event_loop()` now.

As an aside, can commit messages be kept at a 72 column line limit?

Issue #953, opened 2017-12-16
login() should validate kwargs


khazhyk:
User was confused why this wasn't setting their status
```py
bot.run(token, status=discord.Status.idle)
```

The answer being, status is not a valid kwarg for bot.run() (which just passes args/kwargs down to bot.login())

NCPlayz:
This feature request has been added at 38eb2b4.

PR #1896, opened 2019-02-13
Validate keyword arguments in `client.run`

crrapi:
See #953

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -476,6 +476,12 @@ class Client:
 
         bot = kwargs.pop('bot', True)
         reconnect = kwargs.pop('reconnect', True)
+        try:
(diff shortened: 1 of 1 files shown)


Rapptz:
From the way CPython does this error it seems to be done one of two ways:

```python
raise TypeError('got an unexpected keyword argument {arg!r}'.format(arg=next(iter(kwargs))))
# or 
raise ValueError("unexpected keyword argument(s) %s" % kwargs.keys())
```

Either way, there is no `try...except` in the implementation. I'm not sure which one is better, but I'm leaning slightly towards the latter though I have no strong opinion. 

On the topic of exception being raised, I feel like `TypeError` is more appropriate.

PR #1897, opened 2019-02-13
 Bot check optional

tjubutsi:
Provides the ability to not have to overwrite process_commands, reducing clutter. Use when overriding on_message (or don't, see if I care).


Rapptz:
I would prefer people to have their own `on_message` when doing this type of overriding. Especially since for complicated bots most provide their own context and the like already, rendering this function invalid. 

Considering how rare it is to not ignore bots, this doesn't seem to have many gains.

Issue #1927, opened 2019-02-25
Update __main__.py to match new cog changes. 


Rapptz:
During the cog update I forgot to update the entry point script that generates cogs and the like to have the new syntax and design.

PR #1929, opened 2019-02-25
Update __main__.py to use new Cog syntax

Gorialis:
Fix for #1927 and an example.
This doesn't demonstrate the new `commands.Cog`-specific functionality, only updating names and conventions that have changed.

### DIFF
--- a/discord/__main__.py
+++ b/discord/__main__.py
@@ -95,7 +95,7 @@ cog_template = '''# -*- coding: utf-8 -*-
 from discord.ext import commands
 import discord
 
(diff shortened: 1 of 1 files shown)


Gorialis:
Update: was pointed out that `playlist.py` is actually not updated to rewrite at all. Since this change would introduce a weird inconsistency, I'm going to rebase out the example change, but really this example should either be properly ported or removed.

Gorialis:
This PR is obsoleted by the rebasing of #1931.

PR #1946, opened 2019-03-01
Fix Signature for Greedy/Optional converters annotations.

### LINKED ISSUE #1653: Command.signature reports Greedy and Optional params as required
```py
@commands.command()
async def foo(context, a: typing.Optional[discord.Member], b: commands.Greedy[discord.TextChannel], c):
    pass
```

Expected `foo.signature`:
`foo [a] [b...] <c>`

Actual `foo.signature`:
`foo <a> <b> <c>`

EvieePy:
Should fix #1653.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -795,11 +795,23 @@ class Command(_BaseCommand):
             return self.help.split('\n', 1)[0]
         return ''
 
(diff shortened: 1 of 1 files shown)


Rapptz:
I feel like to separate from `*args`, `Greedy` should use `[a]...` and `[a=b]...` notation instead of `[a...]` and `[a=b...]`.

PR #1949, opened 2019-03-02
Add disable_mentions to Messagable.send

IngeniousCoder:
This allows a easier way for users to make sure their bots do not trigger @everyone / @here or any other tags accidentally.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -685,7 +685,7 @@ class Messageable(metaclass=abc.ABCMeta):
     async def _get_channel(self):
         raise NotImplementedError
 
-    async def send(self, content=None, *, tts=False, embed=None, file=None, files=None, delete_after=None, nonce=None):
+    async def send(self, content=None, *, tts=False, embed=None, file=None, files=None, delete_after=None, nonce=None, disable_mentions=False):
         """|coro|
 
         Sends a message to the destination with the content given.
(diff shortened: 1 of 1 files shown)


BluePhoenixGame:
Isn't that what `Message.clean_content` do?

Vexs:
This largely seems to be an idiot tax of overhead for... pretty much everyone. I see no benefit to this. 

Also I think this was PR'd before and declined for that reason but it might have been something similar and not this.

laggycomputer:
Even if this were to be merged, I disagree with the implementation. Adding a zero-width space after EVERY @ symbol is too broad of a search. Look at how [`Message.clean_content`](https://discordpy.readthedocs.io/en/rewrite/api.html#discord.Message.clean_content) is [implemented](https://github.com/Rapptz/discord.py/blob/rewrite/discord/message.py#L393-L444), for example. It uses a regex.

mikeshardmind:
Already done with better handling in another PR: https://github.com/Rapptz/discord.py/pull/1429 , by all means try and advocate for the change if you want it, but it can also be reasonably accomplished by just overriding send in a custom context class (and ensuring your bot uses that context class).

You can also add a function to your bot object for named something like `send_filtered` which would do this for you and take a destination, as to not incur this overhead on every message, even those which use hardcoded responses.

IngeniousCoder:
That’s why it’s made so you need to specify if disable_mentions is True or false.   
  
  
  
>   
>   
>

PR #1959, opened 2019-03-06
Add cached saving

makusu2:
Currently it's difficult to save an attachment from a deleted message; you'll get a `Forbidden` exception if you try after the server's deleted the attachment. 

In order to save an attachment after it's been deleted, you have to manually save it. Instead, I propose we add an option to use the proxy URL of the attachment in the attachment.save function. 

This solution is not perfect by any means; there is still a chance the proxy_url will fail. However, it is far less likely than the base url failing.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -91,6 +91,12 @@ class Attachment:
         seek_begin: bool
             Whether to seek to the beginning of the file after saving is
             successfully done.
+        use_cached: bool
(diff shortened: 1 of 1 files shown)


Rapptz:
I've rebased and changed the documentation a bit. Also made `use_cached=True` the default. Thanks for the PR.

PR #1964, opened 2019-03-08
Added support for guild descriptions

Tarek1337:
As of yesterday/today verified servers can add descriptions (found in server settings -> server overview).

This PR adds support for it for discord.py, it has been tested and works as it should, though this is my first PR, let me know if I missed something. :x

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -759,7 +759,7 @@ class Guild(Hashable):
         name: str
             The new name of the guild.
         description: str
-            The new description of the guild.
+            The new description of the guild. This is only available to guilds that contain `VERIFIED` in :attr:`Guild.features`.
         icon: bytes
             A :term:`py:bytes-like object` representing the icon. Only PNG/JPEG supported.
             Could be ``None`` to denote removal of the icon.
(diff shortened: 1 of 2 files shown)


discord/guild.py
```diff
@@ -119,7 +119,7 @@ class Guild(Hashable):
     """
 
     __slots__ = ('afk_timeout', 'afk_channel', '_members', '_channels', 'icon',
-                 'name', 'id', 'unavailable', 'name', 'region', '_state',
+                 'name', 'description', 'id', 'unavailable', 'name', 'region', '_state',
```
Rapptz: Since you're not actually setting `description` anywhere, there is no need for it to be in the `__slots__`. Is this received on `GUILD_CREATE` by any chance? If so you can add an attribute and document it.

discord/guild.py
```diff
@@ -758,6 +758,8 @@ def _create_channel(self, name, overwrites, channel_type, category=None, **optio
         ----------
         name: str
             The new name of the guild.
+        description: str
+            The new description of the guild.
```
Rapptz: I feel like this needs slightly more detail. What is a description? Where is it in the GUI?

If this is an attribute we can receive, then the documentation could be on the attribute.

discord/guild.py
Tarek1337: From what I can tell, there is nothing sent over `GUILD_CREATE`. I'll remove it, thank you.

PR #1965, opened 2019-03-08
Documentation change: "any" to "all" permissions

tjubutsi:
has_permissions doesn't check if any of the permission:bool combinations is satisfied, it checks all of them.

Edit: :)

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1337,7 +1337,7 @@ def has_any_role(*items):
     return check(predicate)
 
 def has_permissions(**perms):
-    """A :func:`.check` that is added that checks if the member has any of
+    """A :func:`.check` that is added that checks if the member has all of
     the permissions necessary.
 
     The permissions passed in must be exactly like the properties shown under


SilicalNZ:
Where is the evidence for said inconsistancy. A smiley face has no justifications.

CapnS:
I mean, he's right isn't he? That check requires you to have all of the permissions, not any.

laggycomputer:
Notice that in the code for `has_permissions`, it is possible to *disallow* permissions as well as require them. If anything, the docs should remain the same or mention that disallowing perms is possible.

CapnS:
But why would they stay the same, all of the permissions (whether True or False) need to be correct for this to return True, not any of them. I do agree that there should be a note about being able to disallow certain permissions, but I definitely think it needs to be switched to all.

Gorialis:
Issues (and particularly pull requests) should be descriptive, including a summary of the changes made and a rationale for doing so. Short, inspecific titles and little or no description make it more difficult to find and catalogue issues and pull requests later.

Please see the GitHub guidelines and [merged pull requests](https://github.com/Rapptz/discord.py/pulls?q=is%3Apr+is%3Amerged) to see how a good pull request should be created.

PR #1966, opened 2019-03-09
Add support for guild banners

SnowyLuma:
### Summary

This PR adds support for guild banners which show up behind the guild name in the client.

### Checklist

- [ ] This PR fixes an issue.

In addition to adding support for banners I've also increased the maximum valid icon size to 4096, as it seems to have been bumped and banners support larger sizes.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -294,8 +294,8 @@ async def sane_wait_for(futures, *, timeout, loop):
         raise asyncio.TimeoutError()
 
 def valid_icon_size(size):
-    """Icons must be power of 2 within [16, 2048]."""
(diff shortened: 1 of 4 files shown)


discord/utils.py
```diff
@@ -294,8 +294,8 @@ def _parse_ratelimit_header(request):
         raise asyncio.TimeoutError()
 
 def valid_icon_size(size):
-    """Icons must be power of 2 within [16, 2048]."""
-    return not size & (size - 1) and size in range(16, 2049)
+    """Icons must be power of 2 within [16, 4086]."""
```
Rapptz:
```suggestion
    """Icons must be power of 2 within [16, 4096]."""
```

PR #1968, opened 2019-03-09
Change superclass to subclass in some documentation

kaeptmblaubaer1000:
### Summary

It makes no sense that methods accept "Command or superclass objects", because this for example includes a plain `object()`.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/context.py
+++ b/discord/ext/commands/context.py
@@ -53,13 +53,13 @@ class Context(discord.abc.Messageable):
     prefix: :class:`str`
         The prefix that was used to invoke the command.
     command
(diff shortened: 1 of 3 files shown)


Rapptz:
Lgtm.

kaeptmblaubaer1000:
I changed the source of the translations, thus I needed to change the translations too... wdym with "ruin the diffing"?

Rapptz:
These are built by a program occasionally. The program does a comparison and then applies it.

kaeptmblaubaer1000:
This should fully continue to work AFAIK.

PR #1972, opened 2019-03-10
Allow some decorators to be used without calling them

bryanforbes:
### Summary

Currently, to use `Bot.listen()`, `Cog.listener()`, `commands.command()`, `commands.group()`, `Group.command()`, `Group.group()`, `commands.guild_only()`, `commands.is_owner()`, and `commands.is_nsfw()`, you must call them as part of the decorator:

```python
@bot.command()
@commands.guild_only()
async def foo(ctx):
    ...
```

This PR changes all of these decorators so they can be used without being called:

```python
@bot.command
@commands.guild_only
async def foo(ctx):
    ...
```

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -517,6 +517,11 @@ class BotBase(GroupMixin):
             self.add_listener(func, name)
             return func
 
+        if callable(name):
+            func = name
+            name = None
+            return decorator(func)
+
         return decorator
 
     # cogs
--- a/discord/ext/commands/cog.py
+++ b/discord/ext/commands/cog.py
@@ -207,7 +207,7 @@ class Cog(metaclass=CogMeta):
             the name.
         """
 
-        if name is not None and not isinstance(name, str):
+        if name is not None and not callable(name) and not isinstance(name, str):
             raise TypeError('Cog.listener expected str but received {0.__class__.__name__!r} instead.'.format(name))
 
         def decorator(func):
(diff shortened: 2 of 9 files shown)


Gorialis:
I personally disagree with this change as I think the inconsistency will make the confusion not worth the convenience. If a decorator works such that it **can** accept arguments, making it optionally not have brackets may mislead people into not knowing or not understanding that such a thing is possible.

In this sense, I think that this actually works *against* QOL, as it merely scrapes up mistakes instead of actually teaching people how to use the library properly.

Rapptz:
This discussion came about because the standard library already implements this behaviour in some of its decorators, mainly `@dataclass` in 3.7, which can be called `@dataclass()`, `@dataclass(...)` or `@dataclass`.

Since there is pre-existing standard for this behaviour to exist in the Python standard library itself, along with removing a common noob trap that aims to just have cognitive overhead. A lot of decorators can be called with arguments but a lot of them can't be called with arguments, e.g. `@property`, `@functools.lru_cache()` but do you have to care to memorize all of them?

Overall I feel like the pros outweighs the cons here -- which from what I'm gathering seems to be worrying over a hypothetical scenario of preventing people from learning something, when in reality it does no such thing as usage and documentation will prove that it *can* be called with arguments but it is not *required* to.

bryanforbes:
I added back the `()` in the docs and added notes in a few places to note that the decorators can be used without `()`.

GetRektByMe:
I approve this also

NCPlayz:
Reopen this perhaps? Was closed as @bryanforbes did not change the merging branch from `rewrite` to `master` before `rewrite` was deleted.

PR #1994, opened 2019-03-17
Implement Widget

NCPlayz:
### Summary

This implements the widget API via `/guilds/{guild_id}/widget.json`

```py
>>> g = bot.get_guild(81384788765712384) 
... w = await g.widget() # you can also do Client.get_widget()
... print(w.invite)
http://discord.gg/discord-api
```

This also implements simple members and channels that are provided by the Widget API.

```py
>>> for i in w.members:
...     print(i.name)
Danny
Voltana
Sinister Rectus
[...]
>>> discord.utils.get(w.channels, name="API Gaming")
WidgetChannel(id=183361433059328000, name='API Gaming', position=6)
```

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -974,7 +974,6 @@ class Client:
             The widget.
         """
         data = await self.http.get_widget(guild_id)
-        print(data)
         invite = None
         if data['instant_invite']:
             invite = await self.get_invite(data['instant_invite'])
--- a/discord/client.py
+++ b/discord/client.py
@@ -27,7 +27,6 @@ DEALINGS IN THE SOFTWARE.
 import asyncio
 from collections import namedtuple
 import logging
-import re
 import signal
 import sys
 import traceback
@@ -38,7 +37,6 @@ import websockets
 from .user import User, Profile
 from .invite import Invite
 from .widget import Widget
-from .object import Object
 from .guild import Guild
 from .errors import *
 from .enums import Status, VoiceRegion
@@ -168,16 +166,6 @@ class Client:
 
     def _handle_ready(self):
         self._ready.set()
-
-    def _resolve_invite(self, invite):
-        if isinstance(invite, Invite) or isinstance(invite, Object):
-            return invite.id
-        else:
-            rx = r'(?:https?\:\/\/)?discord(?:\.gg|app\.com\/invite)\/(.+)'
-            m = re.match(rx, invite)
-            if m:
-                return m.group(1)
-        return invite
 
     @property
     def latency(self):
@@ -915,7 +903,7 @@ class Client:
             The invite from the URL/ID.
         """
 
-        invite_id = self._resolve_invite(url)
+        invite_id = utils.resolve_invite(url)
         data = await self.http.get_invite(invite_id, with_counts=with_counts)
         return Invite.from_incomplete(state=self._connection, data=data)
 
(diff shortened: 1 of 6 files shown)


discord/widget.py
```diff
@@ -0,0 +1,217 @@
...
+            _id = int(member.get('id'))
+            bot = bool(member.get('bot'))
+            status = try_enum(Status, member.get('status'))
+            deafened = bool(member.get('deaf')) or bool(member.get('self_deaf'))
+            muted = bool(member.get('mute')) or bool(member.get('self_mute'))
+            suppress = bool(member.get('mute'))
+
+            game = None
+            if member.get('game'):
+                game = Game(**member.get('game'))
```
Rapptz: Who is to say these only support `Game` objects? Likewise, it is inefficient to keep fetching the same value.

discord/widget.py
```diff
@@ -0,0 +1,217 @@
...
+
+            connected_channel = None
+            if member.get('channel_id'):
+                connected_channel = get(self.channels, id=int(member.get('channel_id')))
```
Rapptz: This is accidentally O(N^2).

You probably want to build a temporary `channel_id: Channel` mapping and use that.

discord/widget.py
```diff
@@ -0,0 +1,217 @@
...
+            status = try_enum(Status, member.get('status'))
+            deafened = bool(member.get('deaf')) or bool(member.get('self_deaf'))
+            muted = bool(member.get('mute')) or bool(member.get('self_mute'))
+            suppress = bool(member.get('mute'))
+
+            game = None
+            if member.get('game'):
+                game = Game(**member.get('game'))
+
+            connected_channel = None
```
Rapptz:
```suggestion
            connected_channel = utils._get_as_snowflake(member, 'channel_id')
```

Followed by `if connected_channel:`

discord/widget.py
```diff
@@ -0,0 +1,217 @@
...
+            if member.get('channel_id'):
+                connected_channel = get(self.channels, id=int(member.get('channel_id')))
+
+            self.members.append(WidgetMember(id=_id,
```
Rapptz: This large spaghetti can probably be replaced by a proper `__init__` from the `__slot__` class I mentioned earlier so you wouldn't have to do all this.

discord/widget.py
```diff
@@ -0,0 +1,217 @@
...
+                                             avatar=member.get('avatar'),
+                                             nick=member.get('nick'),
+                                             game=game,
+                                             deafened=deafened,
+                                             muted=muted,
+                                             suppress=suppress,
+                                             connected_channel=connected_channel))
+
+    def __str__(self):
+        return self.json_url
```
Rapptz: This needs to be documented. Usually there is an `.. container:: operations` table that documents the python ops like `str`, `hash`, `==`, etc you can do on a type.

discord/client.py
```diff
@@ -946,6 +947,39 @@ def event(self, coro):
...
+            Retrieving the widget failed.
+
+        Returns
+        --------
+        :class:`Widget`
+            The widget.
+        """
+        data = await self.http.get_widget(guild_id)
+        invite = None
+        if data['instant_invite']:
```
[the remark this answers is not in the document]
NCPlayz: It is always available from what I have tested, if it has no invite the value given back is `null`.

discord/client.py
```diff
@@ -946,6 +947,39 @@ def event(self, coro):
...
+            invite = await self.get_invite(data['instant_invite'])
```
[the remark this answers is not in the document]
NCPlayz: I originally planned to move it to `Widget` itself, but `_resolve_invite` was attached to `Client`, which I felt like would add extra effort just to access it. I feel like if I were to move it to `Widget`, it would be better if `_resolve_invite` was moved to `utils.py`.

PR #1995, opened 2019-03-18
Cause ext.commands.ColourConverter to fail if colour value out of range

bijij:
### Summary

changes make the `ext.commands.ColourConverter` Converter fail when user input is outside the acceptable value range 0x000000 - 0xFFFFFF

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -294,6 +294,8 @@ class ColourConverter(Converter):
             arg = arg[1:]
         try:
             value = int(arg, base=16)
+            if not 0 <= value <= 16777215:
+                raise ValueError()
             return discord.Colour(value=value)
         except ValueError:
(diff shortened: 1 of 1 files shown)


discord/ext/commands/converter.py
```diff
@@ -294,6 +294,8 @@ class ColourConverter(Converter):
             arg = arg[1:]
         try:
             value = int(arg, base=16)
+            if not 0 <= value <= 16777215:
```
```suggestion
            if not (0 <= value <= 0xFFFFFF):
```

Please use hex.

discord/ext/commands/converter.py
```diff
@@ -294,6 +294,8 @@ class ColourConverter(Converter):
...
+                raise ValueError()
```
Rapptz: Instead of double raising (`ValueError` -> `BadArgument`), raise `BadArgument` here.

bijij:
Requested changes resolved.

PR #2012, opened 2019-03-21
Refactor saving assets

### LINKED ISSUE #1997: Emoji.save, to mimic Attachment.save
### The Problem

I believe Emoji.save should be a thing.

### The Ideal Solution

Emoji.save, a la Attachment.save. Like, exactly the same, just abstracted away from being HTTPClient.get_attachment. I wouldn't make a PR which changes/removes public methods, so I'm making this issue.

### The Current Solution

```python
fp = BytesIO()
async with session.get(emoji.url) as resp:
    fp.write(await resp.read())
# now do stuff with the emoji data
```

### Summary

I want to save emojis to a file, and/or bytesIO etc in a clean, lib-provided way.

Use cases include backing up emojis, fun emoji commands like deepfrying them or whatever... 

---

In the d.py server, it was also brought up by others to be able to do this for other things such as avatars, https://discordapp.com/channels/336642139381301249/336642776609456130/557351275247829012

NCPlayz:
The London Summer Olympics PR

### Summary

This resolves #1997 feature request.

It does this in a manner of returning an `Asset` for URLs, and subclassing
things such as `Emoji`, and `PartialEmoji`.

### New
- all image types for `Guild`, `PartialInviteGuild`, and `BaseUser`'s available images return `abc.Asset`, allowing for:
```py
>>> await g.icon_url.save('icon.png')
13225
>>> await g.banner_url.save('banner.png')
32145
>>> await g.splash_url_as(format='png', size=1024)
<Asset url='https://cdn.discordapp.com/splashes/12344310703324/1cvx48e12gsg4.png?size=1024'>
>>> await u.default_avatar_url.save('my_default_avy.webp')
12456
>>> return f'<{u.avatar_url}>'
'<https://cdn.discordapp.com/avatars/301449773200834561/75dbedbad336ae40cc25cea88fbc2e4f.webp?size=1024>'
```
- `Emoji` and `PartialEmoji` can directly access `Asset.save`
```py
>>> await my_emoji.save('emoji.png')
12456
```

### Checklist

- [ ] This PR fixes an issue.


discord/asset.py
```diff
@@ -0,0 +1,128 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+"""
+
+import io
+from .errors import InvalidArgument
+from . import utils
+
+VALID_STATIC_FORMATS = {"jpeg", "jpg", "webp", "png"}
```
Rapptz:
```suggestion
VALID_STATIC_FORMATS = frozenset({"jpeg", "jpg", "webp", "png"})
```

discord/asset.py
```diff
@@ -0,0 +1,128 @@
...
+        """Creates a :class:`.Asset` from the requested guild image."""
+        if not utils.valid_icon_size(size):
+            raise InvalidArgument("size must be a power of 2 between 16 and 4096")
+        if format not in VALID_STATIC_FORMATS:
+            raise InvalidArgument("format must be one of {}".format(VALID_STATIC_FORMATS))
+
+        if getattr(guild, image_type) is None:
+            return ''
+
+        if image_type == 'splash':
```
Rapptz: You should probably just specify the route through the parameter with the suffix instead of doing this.

discord/asset.py
```diff
@@ -0,0 +1,128 @@
...
+
+    @classmethod
+    def from_guild_image(cls, guild, state, image_type, *, format='webp', size=1024):
+        """Creates a :class:`.Asset` from the requested guild image."""
+        if not utils.valid_icon_size(size):
+            raise InvalidArgument("size must be a power of 2 between 16 and 4096")
+        if format not in VALID_STATIC_FORMATS:
+            raise InvalidArgument("format must be one of {}".format(VALID_STATIC_FORMATS))
+
+        if getattr(guild, image_type) is None:
```
Rapptz: This `if` block can probably be contained inside the property itself and then the property's value can be passed as a parameter. Much cleaner and readable code than this strange `getattr` call.

discord/asset.py
```diff
@@ -0,0 +1,128 @@
...
+            return ''
+
+        if image_type == 'splash':
+            suffix = "es"
+        else:
+            suffix = "s"
+
+        return cls(state, 'https://cdn.discordapp.com/{0}/{1.id}/{1.icon}.{2}?size={3}'.format(image_type+suffix, guild, format, size))
```
Rapptz: This line is too long (136/125).

discord/asset.py
```diff
@@ -0,0 +1,128 @@
...
+            raise InvalidArgument("size must be a power of 2 between 16 and 1024")
+        if format is not None and format not in VALID_AVATAR_FORMATS:
+            raise InvalidArgument("format must be None or one of {}".format(VALID_AVATAR_FORMATS))
+        if format == "gif" and not user.is_avatar_animated():
+            raise InvalidArgument("non animated avatars do not support gif format")
+        if static_format not in VALID_STATIC_FORMATS:
+            raise InvalidArgument("static_format must be one of {}".format(VALID_STATIC_FORMATS))
+
+        if user.avatar is None:
+            return user.default_avatar_url
```
Rapptz:
```suggestion
            return Asset(state, user.default_avatar_url)
```

Consistent return types and allows people to save the default URL for whatever reason rather than a weird special case.

discord/asset.py
```diff
@@ -0,0 +1,128 @@
...
+
+        if format is None:
+            if user.is_avatar_animated():
```
Rapptz: Can probably combine this into one line
```suggestion
            format = 'gif' if user.is_avatar_animated() else static_format
```

discord/asset.py
```diff
@@ -0,0 +1,128 @@
...
+    @classmethod
+    def from_guild_image(cls, guild, state, image_type, *, format='webp', size=1024):
+        """Creates a :class:`.Asset` from the requested guild image."""
+        if not utils.valid_icon_size(size):
+            raise InvalidArgument("size must be a power of 2 between 16 and 4096")
+        if format not in VALID_STATIC_FORMATS:
+            raise InvalidArgument("format must be one of {}".format(VALID_STATIC_FORMATS))
+
+        if getattr(guild, image_type) is None:
+            return ''
```
Rapptz:
```suggestion
            return Asset(state)
```

Again, consistent return types.

Following the addition of this change though, the following should be added:

- `__len__`, to query the length of the string or 0 if it's `None` like in this snippet.
- `__bool__` to check if the asset is `None`, since it saves a `__len__` call and every instantiation of this type seems to have a non-empty string as its invariant.

This allows for code to not break when checking for it in an if statement or what have you.

Be sure to update the operations container when doing this.

discord/emoji.py
```diff
@@ -103,11 +111,11 @@ def url(self):
         _format = 'gif' if self.animated else 'png'
         return "https://cdn.discordapp.com/emojis/{0.id}.{1}".format(self, _format)
 
-class Emoji:
+class Emoji(Asset):
     """Represents a custom emoji.
 
     Depending on the way this object was created, some of the attributes can
-    have a value of ``None``.
+    have a value of ``None``. This inherits from :class:`Asset`.
```
Rapptz: Put the `inherits from...` line on its own like everything else.

discord/http.py
```diff
@@ -220,16 +220,16 @@ def recreate(self):
...
             raise HTTPException(r, data)
 
-    async def get_attachment(self, url):
+    async def get_from_cdn(self, url):
         async with self._session.get(url) as resp:
             if resp.status == 200:
                 return await resp.read()
             elif resp.status == 404:
-                raise NotFound(resp, 'attachment not found')
+                raise NotFound(resp, 'media not found')
```
Rapptz:
```suggestion
                raise NotFound(resp, 'asset not found')
```

The class name is `Asset` not `Media`. Same for the rest of the strings.

discord/emoji.py
```diff
@@ -65,7 +66,14 @@ class PartialEmoji(namedtuple('PartialEmoji', 'animated name id')):
         The ID of the custom emoji, if applicable.
     """
 
-    __slots__ = ()
+    __slots__ = ('animated', 'name', 'id')
+
+    def __init__(self, state, *, animated, name, id=None):
```
Rapptz: By changing this constructor you have to ensure that all instantiations are also updated to meet the changes.

You missed one in `discord/ext/commands/converter.py`.

NCPlayz:
Update:
- I have completed all requests by @Rapptz to the best of my abilities.
- `AppInfo` is no longer a namedtuple; it is a fully documented dataclass, as well as having the state attached to it.
- `GroupChannel.icon_url` now returns `Asset` via `Asset.from_icon`.
    - Same for `AppInfo.icon_url`.
- `Webhook.avatar_url_as` returns an `Asset`.
- `Webhook.edit` now accepts `Asset`.

bmintz:
I'm against PartialEmoji requiring `state` for construction. For instance, a configuration file containing custom emojis might look like this:

```py
confirm_button = '✅'
deny_button = PartialEmoji(animated=False, name='x', id=2345678)
```
which is more self-documenting than having to do `'<:x:2345678>'`.

NCPlayz:
Ah. I didn't seem to think of that use case.

[As Danny has mentioned](https://discordapp.com/channels/336642139381301249/336642776609456130/560944182030630915), it is a Discord object, and should not be instantiated yourself, but I will look into this, perhaps making a classmethod providing state, but it feels like it wouldn't be clean if you simply `PartialEmoji(animated=False, name='x', id=2345678)` and attempt to save it, since I would have to ask the user to provide state in `PartialEmoji.save`. 

Alternatively, I could just remove the Asset functionality of `PartialEmoji`.

discord/emoji.py
Rapptz: It seems to me that people do not want this to be changed since they use it for their own purposes. I don't consider the use case invalid but it does put a bit of a dent in how you could implement this so it works fine for both groups of people.

After some thinking I've come up with the following solution:

1. Don't inherit from `Asset`.
2. Make `PartialEmoji.url` return an `Asset` instead.
3. Make a classmethod that constructs the `PartialEmoji` for the library's internal uses with the `ConnectionState` being passed in.
4. Default to `None` for the `PartialEmoji._state`
5. When calling `PartialEmoji.url`, the `None` connection state should propagate over to it.
6. Inside `Asset.save` check if the state parameter is `None` and raise if so.

This should satisfy both parties and make the behaviour consistent (it's an asset) but also raises during invalid usage.

As another note, since `namedtuple` was removed from the inheritance chain, this class no longer provides a `__ne__` as per the documentation. (Also it seems to me that `Emoji.__ne__` does not exist either despite being documented so maybe get that one as well)

discord/ext/commands/converter.py
```diff
@@ -400,7 +400,7 @@ class PartialEmojiConverter(Converter):
             emoji_name = match.group(2)
             emoji_id = int(match.group(3))
 
-            return discord.PartialEmoji(animated=emoji_animated, name=emoji_name, id=emoji_id)
+            return discord.PartialEmoji(ctx.bot._connection, animated=emoji_animated, name=emoji_name, id=emoji_id)
```
Rapptz: Make sure not to miss this one when you do the changes above.

discord/asset.py
```diff
@@ -0,0 +1,144 @@
...
+    @classmethod
+    def _from_icon(cls, state, object, path):
+        if object.icon is None:
+            return cls(state)
+
+        url = 'https://cdn.discordapp.com/{0}-icons/{1.id}/{1.icon}.jpg'.format(path, object)
+        return cls(state, url)
+
+    @classmethod
+    def _from_guild_image(cls, state, id, hash, endpoint, *, format='webp', size=1024):
```
Rapptz:
```suggestion
    def _from_guild_image(cls, state, id, hash, key, *, format='webp', size=1024):
```

Since all the endpoints passed to this seem to follow a format of:

```py
'https://cdn.discordapp.com/{key}/{0}/{1}.{2}?size={3}' 
```

It seems wise to just create it inside and pass a key instead, ergo inside the body:

```py
url = 'https://cdn.discordapp.com/{key}/{0}/{1}.{2}?size={3}' 

# ...
url.format(a, b, c, d, key=key)
```

That way you don't have to keep typing the complete format everywhere.

discord/asset.py
```diff
@@ -0,0 +1,144 @@
...
+        NotFound
+            The asset was deleted.
+
+        Returns
+        --------
+        :class:`int`
+            The number of bytes written.
+        """
+        if not self._url:
+            raise DiscordException('a URL was not provided')
```
Rapptz: This exception is undocumented.

```suggestion
            raise DiscordException('Invalid asset (no URL provided)')
```

Rapptz:
Another note:

If `AppInfo` is now its own data class then it should be documented in `api.rst`, exported in `__init__.py`, and linked to in `Client.application_info`.

discord/emoji.py
bmintz: If not provided, `__ne__` defaults to the negation of `__eq__` so that's not necessary.

NCPlayz: I was nearly done with it and was about to test everything, but then I realised it would cause an inconsistency between `Emoji` being an `Asset` itself, and `PartialEmoji.url` returning an `Asset`. This could affect people who do `typing.Union[discord.Emoji, discord.PartialEmoji]` for example.

Should I also revert `Emoji` to match with `PartialEmoji`?

Rapptz: Yeah probably. If truly necessary there can be `save` helper functions that just redirect to the appropriate `Asset.save` functions.

NCPlayz: Gotcha, pre-emptively reverted it anyway. Should be available in the next commit!

discord/__init__.py
```diff
@@ -20,7 +20,7 @@
 from collections import namedtuple
 import logging
 
-from .client import Client, AppInfo
+from .client import Client
```
Rapptz: You're missing `from .appinfo import AppInfo`.

discord/asset.py
```diff
@@ -92,7 +92,9 @@ def _from_guild_image(cls, state, id, hash, endpoint, *, format='webp', size=102
         if hash is None:
             return Asset(state)
 
-        return cls(state, endpoint.format(id, hash, format, size))
+        url = 'https://cdn.discordapp.com/{key}/{0}/{1}.{2}?size={3}'
+
```
Rapptz: Unnecessary newline here.

discord/asset.py
```diff
@@ -120,6 +122,15 @@ def __repr__(self):
 
         Raises
         --------
+        DiscordException
+            The URL or :class:`ConnectionState` was not provided.
+
+            .. note::
+
+                :class:`PartialEmoji` will not have a state if you make
+                your own instance via ``PartialEmoji(animated=False, name='x', id=2345678)``.
```
Rapptz: This note seems misplaced. The example is also superfluous since it's implied by stating "create your own instance", everyone should know how a class is created.

I'm not sure if this note is a good idea in general really.

discord/asset.py
```diff
@@ -120,6 +122,15 @@ def __repr__(self):
 
         Raises
         --------
+        DiscordException
+            The URL or :class:`ConnectionState` was not provided.
```
Rapptz: "There was no valid URL or internal connection state"

discord/client.py
```diff
@@ -1044,7 +1044,7 @@ def fetch_guilds(self, *, limit=100, before=None, after=None):
 
         Returns
         --------
-        :class:`AppInfo`
+        :class:`appinfo.AppInfo`
```
Rapptz: No need for this `appinfo.` when you fix the export in `__init__.py`

discord/emoji.py
```diff
@@ -27,26 +27,20 @@
...
     - "Raw" data events such as :func:`on_raw_reaction_add`
     - Custom emoji that the bot cannot see from e.g. :attr:`Message.reactions`
 
-    This inherits from :class:`Asset`.
-
     .. container:: operations
 
         .. describe:: x == y
 
             Checks if two emoji are the same.
```
Rapptz: Add back the `x != y` tidbit, @bmintz was right in that it's implicitly defined so there was no reason to remove it, albeit being explicit is better. When I offered the suggestion last time I meant that you should define `__ne__` not that you should remove the documentation.

discord/emoji.py
```diff
@@ -66,14 +60,20 @@ class PartialEmoji(Asset):
...
         self.name = name
         self.id = id
+        self._state = None
 
-        super().__init__(state, self.url)
+    @classmethod
+    def with_state(cls, state, *, animated, name, id=None):
+        self = cls(animated=animated, name=name, id=id)
+        self._state = state
+
```
Rapptz: Unnecessary newline.

discord/emoji.py
```diff
@@ -107,31 +107,27 @@ def _as_reaction(self):
...
-        """Returns a URL version of the emoji, if it is custom."""
+        """:class:`Asset`:Returns an asset of the emoji, if it is custom."""
         if self.is_unicode_emoji():
-            return None
+            return Asset(self._state)
 
         _format = 'gif' if self.animated else 'png'
-        return "https://cdn.discordapp.com/emojis/{0.id}.{1}".format(self, _format)
+        url = "https://cdn.discordapp.com/emojis/{0.id}.{1}".format(self, _format)
+
```
Rapptz: Unnecessary newline.

discord/emoji.py
```diff
@@ -210,7 +204,9 @@ def created_at(self):
     def url(self):
         """Returns a URL version of the emoji."""
         _format = 'gif' if self.animated else 'png'
-        return "https://cdn.discordapp.com/emojis/{0.id}.{1}".format(self, _format)
+        url = "https://cdn.discordapp.com/emojis/{0.id}.{1}".format(self, _format)
+
```
Rapptz: Unnecessary newline.

discord/emoji.py
```diff
@@ -107,31 +107,27 @@ def _as_reaction(self):
...
 
-    This inherits from :class:`Asset`.
-
     .. container:: operations
 
         .. describe:: x == y
 
             Checks if two emoji are the same.
 
-        .. describe:: x != y
```
Rapptz: Same tidbit I mentioned above.

docs/api.rst
```diff
@@ -40,6 +40,9 @@ Client
 .. autoclass:: AutoShardedClient
     :members:
 
+.. autoclass:: discord.appinfo.AppInfo
```
Rapptz: No need for explicit namespacing here when you export via `__init__.py`

discord/emoji.py
NCPlayz: Ahh! I completely misinterpreted it. That should be easy enough to do.

PR #2020, opened 2019-03-24
Bulk delete change

### LINKED ISSUE #1807: No good way to separate bulk deletes from single message deletes for logging purposes
Presently I'm not aware of any clean way to separate bulk message deletes from single message deletes for logging purposes. For example, if someone has a basic deleted message logger that just dumps messages in `on_message_delete` a purge will ratelimit it quite quickly. 

Rewrite provides `on_raw_message_delete` and `raw_bulk_message_delete` which are dispatched separately, however, deleted message logging doesn't work well here, as the message is immediately purged from cache after the event is dispatched- which will lead attempts to fetch said message from cache in any `raw_bulk_message_delete` listener failing. 

An easy fix is just replacing the `message_delete` dispatch inside `ConnectionState.parse_message_delete` with `bulk_message_delete`, which would be dispatched with a list of message objects bulk deleted. 

Alternatively, as a non-breaking change, `RawBulkMessageDeleteEvent` could have an optional list of cached message objects deleted attached to it.

Vexs:
### Summary

This PR changes the behavior of the `message_delete`, event, adds the `message_bulk_delete` event and adds an attribute to `raw_message_delete`, `raw_bulk_message_delete`. It also fixes a typo in the on_message_delete documentation.

`message_delete` will no longer receive individual messages bulk deleted. Instead, those will be dispatched to `message_bulk_delete`. This is technically(?) a breaking change.

`raw_message_delete` and `raw_bulk_message_delete`'s payload objects now have a `cached_message(s)` attribute, containing the relevant cached message. This is explicitly done as there is no way to "manually" fetch a message object in raw_(bulk_)message_delete events, as the message is purged from cache immediately when the event is called. This enables people to do different logging events depending on if it is cached.

With the changes to include cached messages in the raw_ events, it's questionable if the non-raw methods are required at all. I also feel like adding the cached objects to the raw payloads is questionable in and of itself, however, I find it necessary as there is no other way to work around it, short of users managing a second cache. The cached messages were added as an attribute of the raw payload objects instead of passing through the event to avoid a larger breaking change.

Resolves #1807

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/raw_models.py
+++ b/discord/raw_models.py
@@ -35,14 +35,16 @@ class RawMessageDeleteEvent:
         The guild ID where the deletion took place, if applicable.
     message_id: :class:`int`
         The message ID that got deleted.
+    cached_message: Optional[:class:`Message`]
+        The cached message, if found in the internal message cache.
     """
 
-    __slots__ = ('message_id', 'channel_id', 'guild_id')
+    __slots__ = ('message_id', 'channel_id', 'guild_id', 'cached_message')
 
     def __init__(self, data):
         self.message_id = int(data['id'])
         self.channel_id = int(data['channel_id'])
-
+        self.cached_message = None
         try:
             self.guild_id = int(data['guild_id'])
         except KeyError:
@@ -59,13 +61,16 @@ class RawBulkMessageDeleteEvent:
         The channel ID where the message got deleted.
     guild_id: Optional[:class:`int`]
         The guild ID where the message got deleted, if applicable.
+    cached_messages: :class:`list`
+        The cached messages, if found in the internal message cache.
     """
 
-    __slots__ = ('message_ids', 'channel_id', 'guild_id')
+    __slots__ = ('message_ids', 'channel_id', 'guild_id', 'cached_messages')
 
     def __init__(self, data):
         self.message_ids = {int(x) for x in data.get('ids', [])}
         self.channel_id = int(data['channel_id'])
+        self.cached_messages = []
 
         try:
             self.guild_id = int(data['guild_id'])
--- a/discord/raw_models.py
+++ b/discord/raw_models.py
@@ -61,7 +61,7 @@ class RawBulkMessageDeleteEvent:
         The channel ID where the message got deleted.
     guild_id: Optional[:class:`int`]
         The guild ID where the message got deleted, if applicable.
-    cached_messages: :class:`list`
+    cached_messages: List[:class:`Message`]
         The cached messages, if found in the internal message cache.
     """
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -235,13 +235,13 @@ to handle it, which defaults to print a traceback and ignoring the exception.
 
     Called when messages are bulk deleted. If none of the messages deleted
     are found in the internal message cache, then this event will not be called.
-    If some individual messages are not found in the internal message cache,
-    the event will still be called, but the messages will not be included in
-    the messages list. Messages will not be in cache if the message is too old
+    If individual messages were not found in the internal message cache,
+    this event will still be called, but the messages not found will not be included in
+    the messages list. Messages might not be in cache if the message is too old
     or the client is participating in high traffic guilds.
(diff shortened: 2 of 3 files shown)


discord/raw_models.py
```diff
@@ -59,13 +61,16 @@ class RawBulkMessageDeleteEvent:
         The channel ID where the message got deleted.
     guild_id: Optional[:class:`int`]
         The guild ID where the message got deleted, if applicable.
+    cached_messages: :class:`list`
```
Rapptz:
```suggestion
    cached_messages: List[:class:`Message`]
```

docs/api.rst
```diff
@@ -224,13 +224,25 @@ to handle it, which defaults to print a traceback and ignoring the exception.
...
+
+    Called when messages are bulk deleted. If none of the messages deleted
+    are found in the internal message cache, then this event will not be called.
+    If some individual messages are not found in the internal message cache,
+    the event will still be called, but the messages will not be included in
+    the messages list. Messages will not be in cache if the message is too old
+    or the client is participating in high traffic guilds.
+    To fix this, increase the ``max_messages`` option of :class:`Client`.
+
+    :param messages: A List[:class:`Message`] of the deleted messages.
```
Rapptz:
```suggestion
    :param messages: A :class:`list` of :class:`Message` that have been deleted.
```

docs/api.rst
```diff
@@ -224,13 +224,25 @@ to handle it, which defaults to print a traceback and ignoring the exception.
...
 
+.. function:: on_bulk_message_delete(messages)
+
+    Called when messages are bulk deleted. If none of the messages deleted
+    are found in the internal message cache, then this event will not be called.
+    If some individual messages are not found in the internal message cache,
+    the event will still be called, but the messages will not be included in
+    the messages list. Messages will not be in cache if the message is too old
+    or the client is participating in high traffic guilds.
+    To fix this, increase the ``max_messages`` option of :class:`Client`.
```
Rapptz:
```suggestion

    If this occurs increase the :attr:`Client.max_messages` attribute.
```

Needs a newline as well.

docs/api.rst
```diff
@@ -224,13 +224,25 @@ to handle it, which defaults to print a traceback and ignoring the exception.
...
 
     :param message: A :class:`Message` of the deleted message.
 
+.. function:: on_bulk_message_delete(messages)
+
+    Called when messages are bulk deleted. If none of the messages deleted
+    are found in the internal message cache, then this event will not be called.
+    If some individual messages are not found in the internal message cache,
+    the event will still be called, but the messages will not be included in
+    the messages list. Messages will not be in cache if the message is too old
```
Rapptz:
```suggestion
    the messages list. Messages might not be in cache if the message is too old
```

Wording is too strong.

docs/api.rst
```diff
@@ -224,13 +224,25 @@ to handle it, which defaults to print a traceback and ignoring the exception.
...
     traffic guilds. To fix this, increase the ``max_messages`` option of
     :class:`Client`.
 
     :param message: A :class:`Message` of the deleted message.
 
+.. function:: on_bulk_message_delete(messages)
+
+    Called when messages are bulk deleted. If none of the messages deleted
+    are found in the internal message cache, then this event will not be called.
+    If some individual messages are not found in the internal message cache,
```
Rapptz:
```suggestion
    If individual messages were not found in the internal message cache,
```

docs/api.rst
```diff
@@ -224,13 +224,25 @@ to handle it, which defaults to print a traceback and ignoring the exception.
...
+    the event will still be called, but the messages will not be included in
```
Rapptz:
```suggestion
    this event will still be called, but the messages not found will not be included in
```

docs/api.rst
Vexs: Should I update existing documentation to follow this? Also, why a newline here?

Rapptz: Obviously.

Newline because it stands out.

PR #2047, opened 2019-04-07
Add coodlown_bypass decorators and cog local.

EvieePy:
### Summary

Allow for cooldowns to be bypassed by several methods:

`@bot.cooldown_bypass` (Applies to all commands)
`@command.cooldown_bypass` (Applies to a single command)
`cog_cooldown_bypass` (Applies to all cog commands)

These checks allow the user to determine whether or not a cooldown(s) should be bypassed.

If the check returns True, the cooldown will be bypassed and no ratelimits will apply, else the cooldown will run as normal.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -105,6 +105,7 @@ class BotBase(GroupMixin):
         self._check_once = []
         self._before_invoke = None
         self._after_invoke = None
+        self._cooldown_bypass = None
         self._help_command = None
         self.description = inspect.cleandoc(description) if description else ''
         self.owner_id = options.get('owner_id')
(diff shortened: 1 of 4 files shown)


CapnS:
My personal opinion is that the check should go the other way e.g. the cooldown gets run if the check returns True. This would keep in line with the other checks in this library where the action happens if the check returns True.

discord/ext/commands/bot.py
```diff
@@ -362,6 +363,48 @@ def after_invoke(self, coro):
...
+            async def my_command(ctx):
+                await ctx.send('This message can be sent unlimitedly by the guild owner, due to my cooldown bypass.)
+
+        Raises
+        -------
+        discord.ClientException
+            The coroutine is not actually a coroutine.
+        """
+        if not asyncio.iscoroutinefunction(coro):
+            raise discord.ClientException('The cooldown bypass must be a coroutine.')
```
Rapptz:
```suggestion
            raise TypeError('The cooldown bypass must be a coroutine.')
```

Same everywhere else.

discord/ext/commands/bot.py
```diff
@@ -362,6 +363,48 @@ def after_invoke(self, coro):
...
+                return False
+
+            @bot.command()
+            @commands.cooldown(1, 30, type=commands.BucketType.guild)
+            async def my_command(ctx):
+                await ctx.send('This message can be sent unlimitedly by the guild owner, due to my cooldown bypass.)
+
+        Raises
+        -------
+        discord.ClientException
```
Rapptz:
```suggestion
        TypeError
```

Same everywhere else.

discord/ext/commands/bot.py
```diff
@@ -362,6 +363,48 @@ def after_invoke(self, coro):
...
+            The coroutine is not actually a coroutine.
```
Rapptz:
```suggestion
            The coroutine passed is not actually a coroutine.
```

Same everywhere else.

scragly:
Since we don't want the cooldown check to be primarily identified with bypassing in it's functionality and naming, we can just no call it a bypass and treat it the same as every other predicated check in the library where returning True applies the thing it's checking, in this case the cooldown.

Effectively, this flips the current True/False behaviour, and allows us to adjust the name to something primarily about it being the predicate check of the cooldown.

PR #2049, opened 2019-04-08
Add function insert_field

sizumita:
### Summary

User can insert field in fields on embed

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -483,6 +483,40 @@ class Embed:
         field['inline'] = inline
         return self
 
+    def insert_field(self, index, *, name, value, inline=True):
+        """Insert a field to the embed object before index.
+
+        The index must point to a valid pre-existing field.
+
+        This function returns the class instance to allow for fluent-style
+        chaining.
+        If not fields, add a field to fields.
+
+        Parameters
+        -----------
+        index: :class:`int`
+            The position of the field to insert.
+        name: :class:`str`
+            The name of the field.
+        value: :class:`str`
+            The value of the field.
+        inline: :class:`bool`
+            Whether the field should be displayed inline.
+        """
+
+        field = {
+            'inline': inline,
+            'name': str(name),
+            'value': str(value)
+        }
+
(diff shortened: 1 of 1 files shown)


SnowyLuma:
This method already exists as [insert_field_at](https://discordpy.readthedocs.io/en/rewrite/api.html#discord.Embed.set_field_at).

discord/embeds.py
```diff
@@ -483,6 +483,44 @@ def set_field_at(self, index, *, name, value, inline=True):
         field['inline'] = inline
         return self
 
+    def insert_field(self, index, *, name, value, inline=True):
```
Rapptz:
```suggestion
    def insert_field_at(self, index, *, name, value, inline=True):
```

discord/embeds.py
```diff
@@ -483,6 +483,44 @@ def set_field_at(self, index, *, name, value, inline=True):
...
+        """Insert a field to the embed object before index.
```
Rapptz:
```suggestion
        """Insert a field to the embed object at the index given.
```

discord/embeds.py
```diff
@@ -483,6 +483,44 @@ def set_field_at(self, index, *, name, value, inline=True):
...
+
+        The index must point to a valid pre-existing field.
```
Rapptz:
```suggestion
```

It doesn't, but going out-of-bounds in `list.insert` will just insert at the end so this should be documented.

discord/embeds.py
```diff
@@ -483,6 +483,44 @@ def set_field_at(self, index, *, name, value, inline=True):
...
+            The name of the field.
+        value: :class:`str`
+            The value of the field.
+        inline: :class:`bool`
+            Whether the field should be displayed inline.
+
+        Raises
+        -------
+        IndexError
+            An invalid index was provided.
```
Rapptz: I'm not exactly aware of a situation where this is raised.

PR #2064, opened 2019-04-10
Added fetch_url_message

makusu2:
### Summary

This pull request adds a fetch_url_message function, which allows a message link (https://discordapp.com/channels/336642139381301249/381974649019432981/381976391303823360) to be used to fetch a message.

Returns None on non-found channel to reflect get_channel's behavior.
Raises NotFound on non-found message to reflect fetch_message's behavior.
This difference will also allow the user to differentiate issues.

The benefit of this function, in my opinion, is to allow users to specify a certain message when using a bot command without having to enable Discord developer mode. It also allows a single argument for channel and message IDs.

Possible usage of this addition:
```
@commands.command()
async def reject_message(self, ctx, message_url):
    try:
        target_message = await ctx.bot.fetch_url_message(message_url)
    except NotFound:
        await ctx.send("That message does not appear to exist anymore! :o")
    else:
        if target_message is None:
            await ctx.send("I don't have access to that channel!")
        else:
            target_message.add_reaction(❌)
```

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -982,7 +982,7 @@ class Client:
     async def fetch_url_message(self, url: str):
         """|coro|
 
-        Retrieves a :class:`Message` from a URL.
+        Retrieves a :class:`.Message` from a jump URL.
         If channel or message is not found, returns ``None``.
 
         Parameters
--- a/discord/client.py
+++ b/discord/client.py
@@ -983,7 +983,7 @@ class Client:
         """|coro|
 
         Retrieves a :class:`Message` from a URL.
-        If channel is not found, returns None.
+        If channel or message is not found, returns None.
 
         Parameters
         -----------
--- a/discord/client.py
+++ b/discord/client.py
@@ -994,8 +994,6 @@ class Client:
         -------
         ValueError
             The URL does not match any recognized format.
-        :exc:`.NotFound`
-            The specified message was not found.
         :exc:`.Forbidden`
             You do not have the permissions required to get a message.
         :exc:`.HTTPException`
@@ -1010,7 +1008,10 @@ class Client:
         message_channel = self.get_channel(int(channel_id))
         if message_channel is None:
             return None
-        return await message_channel.fetch_message(int(message_id))
+        try:
+            return await message_channel.fetch_message(int(message_id))
+        except NotFound:
+            return None
 
     # Invite management
 
--- a/discord/client.py
+++ b/discord/client.py
@@ -979,7 +979,7 @@ class Client:
         data = await self.http.create_guild(name, region, icon)
         return Guild(data=data, state=self._connection)
 
-    async def fetch_url_message(self, url: str):
+    async def fetch_message_from_url(self, url: str):
         """|coro|
 
         Retrieves a :class:`.Message` from a jump URL.
--- a/discord/client.py
+++ b/discord/client.py
@@ -979,6 +979,39 @@ class Client:
         data = await self.http.create_guild(name, region, icon)
         return Guild(data=data, state=self._connection)
 
+    async def fetch_url_message(self, url: str):
+        """|coro|
+
+        Retrieves a :class:`Message` from a URL.
+        If channel is not found, returns None.
+
+        Parameters
+        -----------
+        url: :class:`str`
+            The message URL (message link).
+
+        Raises
+        -------
+        ValueError
+            The URL does not match any recognized format.
+        :exc:`.NotFound`
+            The specified message was not found.
+        :exc:`.Forbidden`
+            You do not have the permissions required to get a message.
+        :exc:`.HTTPException`
+            Retrieving the message failed.
+
+        Returns
+        -------
+        :class:`.Message`
+            The message referenced by the URL.
+        """
+        channel_id, message_id = url.split(r"/")[-2:]
+        message_channel = self.get_channel(int(channel_id))
+        if message_channel is None:
+            return None
+        return await message_channel.fetch_message(int(message_id))
+
     # Invite management
 
     async def fetch_invite(self, url, *, with_counts=True):
--- a/discord/client.py
+++ b/discord/client.py
@@ -1002,7 +1002,7 @@ class Client:
         Returns
         -------
         :class:`.Message`
-            The message referenced by the URL.
+            The message referenced by the jump URL.
         """
         channel_id, message_id = url.split(r"/")[-2:]
         message_channel = self.get_channel(int(channel_id))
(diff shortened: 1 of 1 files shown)


makusu2:
Updated to reflect the request ^_^

discord/client.py
```diff
@@ -979,6 +979,40 @@ def fetch_guilds(self, *, limit=100, before=None, after=None):
         data = await self.http.create_guild(name, region, icon)
         return Guild(data=data, state=self._connection)
 
+    async def fetch_url_message(self, url: str):
```
NCPlayz:
```suggestion
    async def fetch_message_from_jump_url(self, url: str):
```

I feel like this should be called `fetch_message_from_jump_url` (or something shorter), as though it is [`Message.jump_url`](https://github.com/Rapptz/discord.py/blob/d37839f38acd79d49bc19ae06d3067a455695e9b/discord/message.py#L471)

discord/client.py
```diff
@@ -979,6 +979,40 @@ def fetch_guilds(self, *, limit=100, before=None, after=None):
...
+        """|coro|
+
+        Retrieves a :class:`Message` from a URL.
```
NCPlayz:
```suggestion
        Retrieves a :class:`.Message` from a jump URL.
```

discord/client.py
```diff
@@ -979,6 +979,40 @@ def fetch_guilds(self, *, limit=100, before=None, after=None):
...
+        If channel or message is not found, returns None.
```
NCPlayz:
```suggestion
        If channel or message is not found, returns ``None``.
```

discord/client.py
```diff
@@ -979,6 +979,40 @@ def fetch_guilds(self, *, limit=100, before=None, after=None):
...
+            The URL does not match any recognized format.
+        :exc:`.Forbidden`
+            You do not have the permissions required to get a message.
+        :exc:`.HTTPException`
+            Retrieving the message failed.
+
+        Returns
+        -------
+        :class:`.Message`
+            The message referenced by the URL.
```
NCPlayz:
```suggestion
            The message referenced by the jump URL.
```

discord/client.py
makusu2: I feel like `fetch_url_message` is decently well-fitting, and I can't think of a way to make a short enough name that isn't difficult to read. Adding "jump url" to the docstring should still allow anyone looking for the method to find it easily, and it shouldn't be too difficult for someone reading code to tell what "fetch_url_message" is doing.
Do you have any suggestions on how to improve the name without making it too long or sacrificing readability?

MusicOnline: `fetch_url_message` is weird (what's a url message?)
`fetch_message_url` is misleading (fetching the url?)
`fetch_message_from_url` should be fine without the `jump` imo.

makusu2: That works for me! @NCPlays how about you?

NCPlayz: I'm fine with that! :grinning:

Vexs:
I don't like this PR. Not only does it have slightly difficult and nonstandard error handling as pointed out by @Gorialis, but this really sticks too deep into obscure helper functions to my mind, as stuff people should handle themselves. If something like this is to be implemented, a commands ext converter would be a good addition as well.

Further, it only handles a single possible URL syntax, and even then, questionably. 
`channel_id, message_id = url.split(r"/")[-2:]` is just... lazy. 
For starters, it matches invalid not-even-urls, like `381963689470984203/565671434899161099`
It also fails to match valid URLs, like `https://discordapp.com/channels/336642139381301249/381963689470984203/565671434899161099/`
`r"/"` doesn't need to be a raw string either.
Bare minimum a basic regex should be used here.

Discord's jump links don't have to be valid message IDs either. They can jump to the message nearest(?) to some ID or, as a special case,  `https://discordapp.com/channels/336642139381301249/381963689470984203/0` Will jump to the beginning of the channel. (And break the client in the process)

Rapptz:
I don't think this helper should be in the library. It feels like the wrong thing to have. Maybe there's a helper out there that could be done for these URLs but it isn't this one.

makusu2:
> I don't like this PR. Not only does it have slightly difficult and nonstandard error handling as pointed out by @Gorialis, but this really sticks too deep into obscure helper functions to my mind, as stuff people should handle themselves. If something like this is to be implemented, a commands ext converter would be a good addition as well.
> 
> Further, it only handles a single possible URL syntax, and even then, questionably.
> `channel_id, message_id = url.split(r"/")[-2:]` is just... lazy.
> For starters, it matches invalid not-even-urls, like `381963689470984203/565671434899161099`
> It also fails to match valid URLs, like `https://discordapp.com/channels/336642139381301249/381963689470984203/565671434899161099/`
> `r"/"` doesn't need to be a raw string either.
> Bare minimum a basic regex should be used here.
> 
> Discord's jump links don't have to be valid message IDs either. They can jump to the message nearest(?) to some ID or, as a special case, `https://discordapp.com/channels/336642139381301249/381963689470984203/0` Will jump to the beginning of the channel. (And break the client in the process)

A lot of those changes I can definitely make, and will. 
I disagree that this is something people should do themselves; this removes the need for every user wanting to do this to do research into how jump urls work. Given that it's something Discord-specific and not just logic, I don't think there's anything wrong with an API handling it.
A regex would work too, but it's extremely easy to read either way, and it feels simpler this way, not to mention faster (by an insignificant amount). Nothing wrong with being "lazy" if the lazy solution works well.
We could have it force an exact match to the required urls, but I don't see any need to require it if the beginning of the URL doesn't really matter. That part I don't really feel strongly about, though.
Thank you for pointing out the issues! I'll update it soon

makusu2:
@Rapptz can you share why you feel like this specific helper method shouldn't be in the library, but one that fetches messages from a jump URL could be? Is it the bad code style? "It isn't this one" just feels vague...
I'd very much like to work on this. What can do to fix the issue you have with it, aside from the issues already mentioned? I'd really like to improve it.

Rapptz:
Your pull request assumes that URLs have one purpose, which is to fetch a message to it.

In reality, a URL is just a URL with a specific format that can be parsed into different components. A proper helper can just be something that extracts those components similar to [`urllib.parse.urlparse`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse).

makusu2:
@Vexs may I ask where you learned about the special cases with jump links, such as 0 replacing the message ID? A search of Discord's API knowledge base has revealed nothing, although I may simply be using the wrong search terms.
Or anyone else, for that matter. My apologies for being unable to find a detailed explanation of how Discord's client handles them.

PR #2065, opened 2019-04-10
[commands] Add __str__ magic method to Cog

PendragonLore:
### Summary

It feels kind of inconsistent to have many objects in the lib, including `Command`s, return their name except for cogs when the `str()` method is used on them, this PR adds the `__str__` magic method to the Cog class, which returns qualified_name.
This also makes easier to avoid errors when using `Cog.qualified_name` in a custom `HelpCommand` class since, even if one's of `mapping` keys is `None`, it will not give out any errors when using the `str()` method unlike when trying to access `qualified_name`.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/cog.py
+++ b/discord/ext/commands/cog.py
@@ -183,6 +183,9 @@ class Cog(metaclass=CogMeta):
                 parent.add_command(command)
 
         return self
+    
+    def __str__(self):
+        return self.qualified_name
 
     def get_commands(self):
         r"""Returns a :class:`list` of :class:`.Command`\s that are


Rapptz:
I don't really agree with this change.

Since I consider Cogs to be classes that users create themselves then my philosophy is to avoid adding unnecessary things. Considering there is already one way of doing this and that there could be personal use cases to overriding `__str__` then I don't know what this buys outside of laziness in doing a None check.

PendragonLore:
I see, well then I'm closing it, thanks for your opinion though.

Issue #2069, opened 2019-04-11
(Re-)add support for aiohttp.ClientResponse in discord.File


Terrance:
### The Problem

During the rewrite alpha, one could pass the content from an `aiohttp` response object directly to a new `discord.File` instance:

```python
with aiohttp.ClientSession() as sess:
    with sess.get(...) as resp:
        f = discord.File(resp.content, filename)
```

It looks like the file code has been refactored shortly before the stable release (5e65ec9) and this is no longer possible.  Judging from the docs, this was probably an unintended feature.

### The Ideal Solution

It would be nice to re-add proper support of this, such that a response body can be streamed into the upload to Discord without fetching the entire file into memory.

### The Current Solution

The content needs to be loaded into a `BytesIO` instance:

```python
with aiohttp.ClientSession() as sess:
    with sess.get(...) as resp:
        content = io.BytesIO(await resp.content.read())
        f = discord.File(content, filename)
```

Terrance:
I'm assuming this hasn't been implemented, and is being closed as part of tidying up old issues?

(I can't see any related commits; this would still be useful for what it's worth, unless there's a better pattern now.)

Rapptz:
Correct, I was just doing some cleaning and this one was caught. I'm still not really opposed to this.

LostLuma:
I would like to reimplement this feature, however I have a question due to the ideal solution you've outlined.

The `StreamReader` provided by aiohttp is single-use, as it just passes the data from one end to the other, meaning if the request to send the message fails for a reason we usually retry on we can't do so.

Would storing the data while streaming it to allow retrying at the cost of temporarily using more memory be okay, or are you looking for this in order to completely avoid that?

PR #2073, opened 2019-04-12
Add `File.from_response` constructor

mental32:
### Summary

This PR addresses #2069 directly by adding a constructor to `discord.File` that takes an `aiohttp.ClientResponse` response object and any subsequent arguments and reads the response into an `io.BytesIO` object.

Additionally I could not confirm that the code blocks presented in #2069 would work pre-https://github.com/Rapptz/discord.py/commit/5e65ec978cf278fefcc5586e2df732bc7a8bed4e
as I kept encountering an ClientConnectionError exception being thrown with a traceback similar to [this](https://hastebin.com/qepebeneva.sql) (HEAD: https://github.com/Rapptz/discord.py/commit/8ba48c14a7c59c2fc3062f3e8aea09cbbc910c16 code: [here](https://hastebin.com/nokahinuye.py)) while trying to design a more elegant implementation without the need for a new method.

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/file.py
+++ b/discord/file.py
@@ -87,6 +87,26 @@ class File:
         if spoiler and self.filename is not None and not self.filename.startswith('SPOILER_'):
             self.filename = 'SPOILER_' + self.filename
 
+    @classmethod
+    async def from_response(cls, resp, *args, **kwargs):
+        """|coro|
+
+        Alternate constructor for a :class:`discord.File` object.
+
+        Basically equivalent to:
+
+        .. code-block:: python3
+
+            discord.File(io.BytesIO(await resp.content.read()), filename, spoiler=spoiler)
+
+        Parameters
+        -----------
+        resp: :class:`aiohttp.ClientResponse`
+            The response to read the content from.
+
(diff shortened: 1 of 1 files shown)


Vexs:
This seems far too narrow in scope- this more or less only works with aiohttp responses, and I can't think of any user-exposed function in the lib that so specifically relies on a class being a specific class from a third-party library.
It seems more sensible to me to make a generic from_bytes method, or just change the behavior of File to allow bytes by default. 

Implementation wise, I'm fairly sure `:class:`aiohttp.ClientResponse`` won't link correctly, and `filename` should be made a required arg as this method doesn't work without it being passed. Or, it should attempt to read from the response what the filename is and pass that by default.

MusicOnline:
I support Vex's generic `from_bytes` constructor idea instead of a specific one for aiohttp responses.

Rapptz:
This isn't how I want to support #2069 

Notice how the issue mentions not loading everything into memory, which this PR does.

PR #2075, opened 2019-04-12
[tasks] Add Loop.restart()

diceroll123:
### Summary

It's a convenience method to cancel and then restart a Loop. Doubles as a `Loop.start()` alias.

Heavily relies on #2074 being merged. 😜 

Documentation phrasing criticism is welcome. _I may not have gotten the Sphinx linking correct._

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/tasks/__init__.py
+++ b/discord/ext/tasks/__init__.py
@@ -133,6 +133,21 @@ class Loop:
         if self._task:
             self._task.cancel()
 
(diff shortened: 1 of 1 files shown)


Rapptz:
This one seems a bit weird. A task isn't necessary done at the cancel() point so maybe this needs a different implementation.

PR #2083, opened 2019-04-13
Add MessageConverter to the command converters

retke:
### Summary

This PR adds a message converter to easily get a message object from 3 possible ways:
- Use the combo {channel ID}-{message ID} (you can get this by shift-clicking on "Copy ID" on a message)
- Use the message link
- Use the single message ID (in this case, the context channel will be used)

If one of these matches, a `discord.Message` object is returned.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -31,7 +31,7 @@ import discord
 
 from .errors import BadArgument, NoPrivateMessage
 
-__all__ = ['Converter', 'MemberConverter', 'UserConverter',
+__all__ = ['Converter', 'MemberConverter', 'UserConverter', 'MessageConverter'
            'TextChannelConverter', 'InviteConverter', 'RoleConverter',
            'GameConverter', 'ColourConverter', 'VoiceChannelConverter',
            'EmojiConverter', 'PartialEmojiConverter', 'CategoryChannelConverter',
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -170,52 +170,52 @@ class MessageConverter(IDConverter):
     """
 
     async def convert(self, ctx, argument):
-        if len(argument.split()) > 1:
-            raise BadArgument("The message must be an ID or an URI.")
-        exception = BadArgument('Message "{msg}" not found.'.format(msg=argument))
-        if "-" in argument:
-            channel_id, message_id = [self._get_id_match(x) for x in argument.split("-")]
-            if not channel_id or not message_id:
-                raise exception
-            channel_id = int(channel_id.group(1))
-            message_id = int(message_id.group(1))
-            channel = ctx.bot.get_channel(channel_id)
+        prog = re.compile(r'^(?:(?P<channel_id>[0-9]{15,21})-)?(?P<message_id>[0-9]{15,21})$')
+        match = prog.match(argument)
+        if match:
+            message_id = int(match.group("message_id"))
+            message = ctx.bot._connection._get_message(message_id)
+            if message:
+                return message
+            channel_id = match.group("channel_id")
+            if channel_id:
+                channel = ctx.bot.get_channel(int(channel_id))
+                if not channel:
+                    raise BadArgument('Channel "{channel}" not found.'.format(channel=channel_id))
+                try:
+                    message = await channel.fetch_message(message_id)
+                except discord.errors.NotFound:
+                    raise BadArgument('Message "{msg}" not found.'.format(msg=argument))
+                except discord.errors.Forbidden:
+                    raise BadArgument("Can't read messages in {channel}".format(channel=channel.mention))
+            else:
+                channel = ctx.channel
+                try:
+                    message = await channel.fetch_message(message_id)
+                except discord.errors.NotFound:
+                    raise BadArgument('Message "{msg}" not found.'.format(msg=argument))
+        else:
+            prog = re.compile(
+                r'^https?://(?:(ptb|canary)\.)?discordapp\.com/channels/'
+                r'(?P<guild_id>[0-9]{15,21})/(?P<channel_id>[0-9]{15,21})/(?P<message_id>[0-9]{15,21})$'
+            )
+            match = prog.match(argument)
+            if not match:
+                raise BadArgument('Message "{msg}" not found'.format(msg=argument))
+            channel_id = int(match.group("channel_id"))
+            message_id = int(match.group("message_id"))
+            message = ctx.bot._connection._get_message(message_id)
+            if message:
+                return message
+            channel = ctx.bot.get_channel(int(match.group("channel_id")))
             if not channel:
-                raise exception
+                BadArgument('Message "{msg}" not found'.format(msg=argument))
             try:
                 message = await channel.fetch_message(message_id)
             except discord.errors.NotFound:
-                raise exception
+                BadArgument('Message "{msg}" not found'.format(msg=argument))
             except discord.errors.Forbidden:
                 raise BadArgument("Can't read messages in {channel}".format(channel=channel.mention))
-        elif "://discordapp.com/channels/" in argument:
-            try:
-                guild_id, channel_id, message_id = [
-                    self._get_id_match(x)
-                    for x in argument.split("://discordapp.com/channels/")[1].split("/")
-                ]
-            except (KeyError):
-                raise exception
-            guild = ctx.bot.get_guild(int(guild_id.group(1)))
-            if not guild:
-                raise exception
-            channel = guild.get_channel(int(channel_id.group(1)))
-            if not channel:
-                raise exception
-            try:
-                message = await channel.fetch_message(int(message_id.group(1)))
-            except discord.errors.NotFound:
-                raise exception
-            except discord.errors.Forbidden:
-                raise BadArgument("Can't read messages in {channel}".format(channel=channel.mention))
-        else:
-            channel = ctx.channel
-            try:
-                message = await channel.fetch_message(int(argument))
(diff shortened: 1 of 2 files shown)


discord/ext/commands/converter.py
```diff
@@ -159,6 +159,80 @@ class UserConverter(IDConverter):
 
         return result
 
+class MessageConverter(IDConverter):
+    """Converts to a :class:`discord.Message`.
+
+    The lookup strategu is as follows (in order):
```
Rapptz: I feel like this lookup strategy is not optimal. It should be reordered to 1, 3, 2. 

Also URI is actually URL.

discord/ext/commands/converter.py
```diff
@@ -159,6 +159,80 @@ class UserConverter(IDConverter):
...
+
+    1. Lookup by "{channel ID}-{message ID}" (got by shift-clicking on "Copy ID")
+    2. Lookup by message URI
+    3. Lookup by message ID (the message **must** be in the context channel)
+    """
+
+    async def convert(self, ctx, argument: str) -> discord.Message:
```
Rapptz: Type hinting is done by the stubs somewhere else.
```suggestion
    async def convert(self, ctx, argument):
```

discord/ext/commands/converter.py
```diff
@@ -159,6 +159,80 @@ class UserConverter(IDConverter):
...
+        """|coro|
```
Rapptz: Converter.convert doesn't get documentation.

discord/ext/commands/converter.py
```diff
@@ -159,6 +159,80 @@ class UserConverter(IDConverter):
...
+        properly propagate to the error handlers.
+
+        Parameters
+        -----------
+        ctx: :class:`.Context`
+            The invocation context that the argument is being used in.
+        argument: :class:`str`
+            The argument that is being converted.
+        """
+        if len(argument.split()) > 1:
```
Rapptz: It makes no sense to call `split` here. You're calling a semi-expensive operation and then just throwing it away.

discord/ext/commands/converter.py
```diff
@@ -159,6 +159,80 @@ class UserConverter(IDConverter):
...
+            raise BadArgument("The message must be an ID or an URI.")
+        exception = BadArgument('Message "{msg}" not found.'.format(msg=argument))
```
Rapptz: I don't believe I use this idiom anywhere in the library, ergo nowhere do I make an exception object ahead of time and then raise it.

discord/ext/commands/converter.py
```diff
@@ -159,6 +159,80 @@ class UserConverter(IDConverter):
...
+        if "-" in argument:
```
Rapptz: This substring match (and the entire block following this) is better suited with a custom regex.

```py
x = re.compile(r'^(?:(?P<channel_id>[0-9]{15,21})-)?(?P<message_id>[0-9]{15,21})$')`
```

This regex will catch both `ChannelID-MessageID` and bare `MessageID`.

discord/ext/commands/converter.py
```diff
@@ -159,6 +159,80 @@ class UserConverter(IDConverter):
...
+            channel = ctx.bot.get_channel(channel_id)
+            if not channel:
+                raise exception
+            try:
+                message = await channel.fetch_message(message_id)
+            except discord.errors.NotFound:
+                raise exception
+            except discord.errors.Forbidden:
+                raise BadArgument("Can't read messages in {channel}".format(channel=channel.mention))
+        elif "://discordapp.com/channels/" in argument:
```
Rapptz: Once again this is better suited with a regex rather than this approach. This also does not work for `ptb` and `canary` links.

discord/ext/commands/converter.py
numbermaniac: (also, "strategu" should be "strategy")

retke:
Alright, I modified almost everything following what you said here and on the server, it's completly working with regex now, and the #testing gang of Red tested a lot of situations, it works quite well!

Vexs:
This still seems pretty undercooked. You have quite a lot of code replication in here, for starters. 
Your error messages also need work, it doesn't make much sense for it to say a message can't be found when it's the channel that couldn't be found, for instance.
Also, you could do this with a single regex search if you're so inclined.

discord/ext/commands/converter.py
```diff
@@ -159,6 +159,65 @@ class UserConverter(IDConverter):
 
         return result
 
+class MessageConverter(IDConverter):
+    """Converts to a :class:`discord.Message`.
+
+    The lookup strategy is as follows (in order):
+
+    1. Lookup by "{channel ID}-{message ID}" (got by shift-clicking on "Copy ID")
```
laggycomputer: `(got by shift-clicking on "Copy ID")` seems unnecessary here.

discord/ext/commands/converter.py
```diff
@@ -159,6 +159,65 @@ class UserConverter(IDConverter):
...
+
+    async def convert(self, ctx, argument):
+        prog = re.compile(r'^(?:(?P<channel_id>[0-9]{15,21})-)?(?P<message_id>[0-9]{15,21})$')
+        match = prog.match(argument)
+        if match:
+            message_id = int(match.group("message_id"))
+            message = ctx.bot._connection._get_message(message_id)
+            if message:
+                return message
+            channel_id = match.group("channel_id")
```
```suggestion
            channel_id = match.group("channel_id")
            channel = ctx.bot.get_channel(int(channel_id)) or ctx.channel
```

Do this instead of duplicating all that code down there.

discord/ext/commands/converter.py
```diff
@@ -159,6 +159,65 @@ class UserConverter(IDConverter):
...
+            else:
+                channel = ctx.channel
+                try:
+                    message = await channel.fetch_message(message_id)
+                except discord.errors.NotFound:
+                    raise BadArgument('Message "{msg}" not found.'.format(msg=argument))
+        else:
+            prog = re.compile(
+                r'^https?://(?:(ptb|canary)\.)?discordapp\.com/channels/'
+                r'(?P<guild_id>[0-9]{15,21})/(?P<channel_id>[0-9]{15,21})/(?P<message_id>[0-9]{15,21})$'
```
Rapptz: This regex does not support `@me` in the URLs, which is for private message contexts.

discord/ext/commands/converter.py
retke: Using `int()` would raise `TypeError` if there's no channel ID.
```py
>>> import re
>>> id_regex = re.compile(r'^(?:(?P<channel_id>[0-9]{15,21})-)?(?P<message_id>[0-9]{15,21})$')
>>> match = id_regex.match("987654321012345")
>>> int(match.group("channel_id")) or 42
  File "<stdin>", line 1, in <module>
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

retke:
Didn't know `regex1 or regex2` was a thing. I rewrote the function with this, and now it's way shorter. I tested it and it works well on DMs (and I'm not grabbing the guild ID since it's useless now).

Rapptz:
Thanks for the PR and good job.

discord/ext/commands/converter.py
```diff
@@ -159,6 +159,41 @@ class UserConverter(IDConverter):
...
+    2. Lookup by message ID (the message **must** be in the context channel)
+    3. Lookup by message URL
+    """
+
+    async def convert(self, ctx, argument):
+        id_regex = re.compile(r'^(?:(?P<channel_id>[0-9]{15,21})-)?(?P<message_id>[0-9]{15,21})$')
+        link_regex = re.compile(
+            r'^https?://(?:(ptb|canary)\.)?discordapp\.com/channels/'
+            r'(?:([0-9]{15,21})|(@me))'
+            r'/(?P<channel_id>[0-9]{15,21})/(?P<message_id>[0-9]{15,21})$'
```
makusu2: `message_id` can be 0, as pointed out in https://github.com/Rapptz/discord.py/pull/2064#issuecomment-481901818. If this is the case, it should return the first message in the given channel.
It takes a long time to do so, and I don't think it should be added, but it is technically a valid URL. /shrug

makusu2:
```suggestion
            r'/(?P<channel_id>[0-9]{15,21})/(?P<message_id>[0-9]{15,21})/?$'
```
URL can end in /, such as https://discordapp.com/channels/336642139381301249/336642139381301249/564977228014682112/

discord/ext/commands/converter.py
```diff
@@ -159,6 +159,41 @@ class UserConverter(IDConverter):
...
+    1. Lookup by "{channel ID}-{message ID}" (retrieved by shift-clicking on "Copy ID")
+    2. Lookup by message ID (the message **must** be in the context channel)
+    3. Lookup by message URL
+    """
+
+    async def convert(self, ctx, argument):
+        id_regex = re.compile(r'^(?:(?P<channel_id>[0-9]{15,21})-)?(?P<message_id>[0-9]{15,21})$')
+        link_regex = re.compile(
+            r'^https?://(?:(ptb|canary)\.)?discordapp\.com/channels/'
+            r'(?:([0-9]{15,21})|(@me))'
```
makusu2: There could be a decent speed increase if we use the guild ID (the first of the 3 IDs), since the bot wouldn't have to search through the guilds. Might be nice in the future.
(Sometimes the guild ID will just be the channel ID for dumb reasons, so it's probably fine if this isn't added)
See https://github.com/Rapptz/discord.py/pull/2091/files

PR #2086, opened 2019-04-15
add read-only cached_messages property to Client

khazhyk:
### Summary

For those of us who want access to this sweet trove of zero hop messages

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -228,6 +228,14 @@ class Client:
         return self._connection.emojis
 
     @property
+    def cached_messages(self):
+        """Sequence[:class:`~discord.Message`]: Read-only list of messages the connected client has cached.
+
+        .. versionadded:: 1.1.0
+        """
+        return utils.SequenceProxy(self._connection._messages)
+
+    @property
     def private_channels(self):
         """List[:class:`abc.PrivateChannel`]: The private channels that the connected client is participating on.
 
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -26,6 +26,7 @@ DEALINGS IN THE SOFTWARE.
 
 import array
 import asyncio
+import collections.abc
 import unicodedata
 from base64 import b64encode
 from bisect import bisect_left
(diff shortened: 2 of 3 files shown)


discord/utils.py
```diff
@@ -78,6 +79,17 @@ def decorator(func):
         return CachedSlotProperty(name, func)
     return decorator
 
+class SequenceProxy(collections.abc.Sequence):
```
Rapptz: This is missing the following to satisfy the interface:

1. `__iter__`
2. `__reversed__`
3. `__contains__`

discord/client.py
```diff
@@ -230,6 +230,11 @@ def emojis(self):
         """List[:class:`Emoji`]: The emojis that the connected client has."""
         return self._connection.emojis
 
+    @property
+    def cached_messages(self):
+        """SequenceProxy[:class:`Message`]: Read-only list of messages the connected client has cached."""
```
Rapptz: Needs a `..versionadded::` string.

discord/utils.py
khazhyk: Those are implemented by the abc

NCPlayz:
In [migrating.rst](https://discordpy.readthedocs.io/en/latest/migrating.html#models-are-stateful), it is mentioned that `Client.messages` was removed, but would it be feasible to write that it has returned in some form?

khazhyk:
Updated migrating.rst. I don't have an rst renderer but it looks right...

discord/utils.py
```diff
@@ -78,6 +79,32 @@ def decorator(func):
...
+    def __contains__(self, item):
+        return item in self.__proxied
+
+    def __iter__(self):
+        return iter(self.__proxied)
+
+    def __reversed__(self):
+        return reversed(self.__proxied)
+
+    def index(self, value, *args, **kwargs):
```
bmintz: Index takes no kwargs.

Rapptz: This is a forwarding function so it's fine.

docs/migrating.rst
```diff
@@ -206,8 +206,6 @@ A list of these changes is enumerated below.
 +---------------------------------------+------------------------------------------------------------------------------+
 | ``Client.wait_until_login``           | Removed                                                                      |
 +---------------------------------------+------------------------------------------------------------------------------+
-| ``Client.messages``                   | Removed                                                                      |
```
Rapptz: Why not just point to the new attribute in the table instead of adding it somewhere else?

khazhyk: This changed didn't feel like it fit under the "models are stateful" heading so I moved it to the misc section

PR #2088, opened 2019-04-15
Document remaining public enums.

NCPlayz:
### Summary

I have documented the remaining public enums, which are:

- UserContentFilter
- FriendFlags
- PremiumType
- Theme

UserFlags is not documented as though it is only used internally.

### Checklist

- [ ] This PR fixes an issue.


Rapptz:
These probably need a note that this only applies to users and not bots.

PR #2091, opened 2019-04-16
Parse message links

makusu2:
### Summary

This is a rewrite of https://github.com/Rapptz/discord.py/pull/2064. It aims to fix the issues presented there.

"Jump URL" refers to a message link (https://discordapp.com/channels/336642139381301249/381974649019432981/381976391303823360).

PR additions:
parse_jump_url method: Parses a jump URL and returns a tuple of (guild_id, channel_id, message_id)
fetch_message_from_url method: Returns a message object from a jump URL
MessageConverter: Converter for messages, which returns a message from a jump URL argument.

Updates:

As pointed out by @Vexs, there is a special case where replacing message_id with 0 will be treated as the first message in a channel. Discord does not have any public documentation on special cases with message links, but more special cases may come up in the future. For the special case with 0, as well as in order to support future special cases, the client method to retrieve the Message object will be included in addition to the parser, since it will take care of special cases.

Sometimes the guildID will match the channelID in a jump URL. However, in cases like this, get_guild will still accept the channel ID, so no additional steps must be taken.

Retrieving the first message (using the special 0 case) will take quite a long time if the channel has many messages. It may be desired to ignore this special case to avoid long requests; thoughts?

As requested by some reviewers, the netloc of the URL must match discordapp.com else a ValueError will be raised. I am personally of the opinion that this should not be checked, as it is unlikely to be confused if the rest of the format matches, and will invalidate future alias URLs. 

Casing for "channel" (/channel/...) is ignored, since incorrect casing will work on the client. Incorrect casing WILL fail off-client (on browsers) most of the time, but I believe "Channel" is a typo that should be allowed to pass.

As requested by @Gorialis, the method will return None regardless of what ID it failed to retrieve (guild, channel, message)

As requested by @NCPlayz, the method has been renamed to fetch_message_from_url for clarity, and the URL is referred to as both "message link" and "jump URL" to make it easy to find.

discord.com is not included as a valid network location for a jump link. This is because it will redirect to discordapp.com, but will not do so in-client. If anyone disagrees with this, it can certainly be changed.

Please let me know if you have any issues with it! 

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -29,7 +29,6 @@ from collections import namedtuple
 import logging
 import signal
 import sys
-import re
 import traceback
 
 import aiohttp
--- a/discord/client.py
+++ b/discord/client.py
@@ -982,7 +982,7 @@ class Client:
     async def fetch_url_message(self, url: str):
         """|coro|
 
-        Retrieves a :class:`Message` from a URL.
+        Retrieves a :class:`.Message` from a jump URL.
         If channel or message is not found, returns ``None``.
 
         Parameters
--- a/discord/client.py
+++ b/discord/client.py
@@ -983,7 +983,7 @@ class Client:
         """|coro|
 
         Retrieves a :class:`Message` from a URL.
-        If channel is not found, returns None.
+        If channel or message is not found, returns None.
 
         Parameters
         -----------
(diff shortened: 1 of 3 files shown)


Rapptz:
This PR is doing way way way way too many things. 

Please only do only one thing per PR.

PR #2092, opened 2019-04-16
Add URL to Asset

XuaTheGrate:
### Summary

Adds `Asset.url` to obtain the URL, as an alternative to `str(Asset)`.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/asset.py
+++ b/discord/asset.py
@@ -106,6 +106,11 @@ class Asset:
 
         url = 'https://cdn.discordapp.com/{key}/{0}/{1}.{2}?size={3}'
         return cls(state, url.format(id, hash, format, size, key=key))
+    
+    @property
+    def url(self):
+        """Returns the URL of the CDN asset."""
+        return self.__str__()
 
     def __str__(self):
         return self._url if self._url is not None else ''


apple502j:
Just use `str(Asset)` ~~or `Asset._url`~~

XuaTheGrate:
Using `Asset.__str__()` was on average `.001` seconds faster according to this, but if its too small to matter then I'll change it.
test
(I ran it a few times, got roughly the same result.)

Rapptz:
Don't really see the point in this one.

BluePhoenixGame:
This would improve readability in code I would say. It's easier to understand what `Asset().url` does VS `str(Asset)`, both is as easy to read for someone who knows all this. But it's easier for new people to read someones code and understand it.

NCPlayz:
The behaviour is documented. You could say the same about [``str(User)``](https://discordpy.readthedocs.io/en/latest/api.html#discord.User), which gives back ``name#discrim``, but this behaviour is also documented so I feel like you can't *really* be confused here.

This also follows the style of [yarl.URL](https://yarl.readthedocs.io/en/latest/api.html), where ``str(URL)`` returns the full URL in string format.


On Apr 16, 2019 4:32 PM, BluePhoenixGame <notifications@github.com> wrote:

This would improve readability in code I would say. It's easier to understand what Asset().url does VS str(Asset), both is as easy to read for someone who knows all this. But it's easier for new people to read someones code and understand it.

Rapptz:
The readability argument has no grounds in the actual reality of the library. You receive these through properties that end in `_url`. What you're proposing is practically a readability boost when you do `m.avatar_url.url` which is asinine.

XuaTheGrate:
My main reason I requested this was because in the Assets repr, it would show `<Asset url=https://.../>`, and I don't know about anyone else but this makes me think that `url` is an attribute of Asset and I should be able to access it that way.

I do understand your point that it makes readability kinda wack and thinking back, it would end up being kinda weird.

PR #2094, opened 2019-04-17
Add delay kwarg to message.delete()

Vexs:
### Summary

This PR adds a `delay` kwarg to message.delete that delays the deletion of a message while yielding back to the event loop. This is the same behavior used in `delete_after` kwargs, but for more generic usage. It also refactors existing `delete_after` behaviors to use the new kwarg. 

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -563,7 +563,7 @@ class Message:
             else:
                 return '{0.author.name} started a call \N{EM DASH} Join the call.'.format(self)
 
-    async def delete(self, delay=None):
+    async def delete(self, *, delay=None):
         """|coro|
 
         Deletes the message.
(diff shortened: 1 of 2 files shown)


discord/message.py
```diff
@@ -563,7 +563,7 @@ def system_content(self):
             else:
                 return '{0.author.name} started a call \N{EM DASH} Join the call.'.format(self)
 
-    async def delete(self):
+    async def delete(self, delay=None):
```
```suggestion
    async def delete(self, *, delay=None):
```

This needs to be a keyword only argument. `await m.delete(10.0)` is bad readability and could be ambiguous in the future.

Vexs: Oh bugger I completely forgot. Yes, this should be that.

PR #2096, opened 2019-04-18
Make Guild.features a frozenset

bmintz:
### Summary

The most frequent use-case for Guild.features is containment checks, which are slow on a list.
The API does not guarantee any order for guild features anyway, so it's unlikely that anyone would use indexing on this list, making it unlikely to be a breaking change.
Additionally, using a set eases use cases such as the following, which checks if a given guild is partnered:

```py
def is_partnered(guild):
	partnered_features = {'VIP_REGIONS', 'VANITY_URL', 'INVITE_SPLASH'}
	return guild.features.issuperset(partnered_features)
```

Finally, a frozenset is used instead of a set because it should not be modified. I am open to feedback on whether this should be a set instead. `GroupMixin.commands` already uses a regular set, however, that set is copied from an existing dict values view, so mutating it would have no lasting effect.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -113,8 +113,8 @@ class Guild(Hashable):
         The guild's explicit content filter.
     default_notifications: :class:`NotificationLevel`
         The guild's notification settings.
-    features: List[:class:`str`]
-        A list of features that the guild has. They are currently as follows:
+    features: FrozenSet[:class:`str`]
+        A unique set of features that the guild has. They are currently as follows:
 
         - ``VIP_REGIONS``: Guild has VIP voice regions
         - ``VANITY_URL``: Guild has a vanity invite URL (e.g. discord.gg/discord-api)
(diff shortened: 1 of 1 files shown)


SnowyLuma:
PartialInviteGuild also has the `features` attribute, the type of it should probably also change in order to be consistent.

Rapptz:
Changing the return type of something is indeed a breaking change. Some trivial code that used to work and now no longer does:

```py
x = guild.features
if len(x) == 0:
    print(x[0])
```

I agree that this change is probably for the better but I am unsure whether it's worth doing the breaking change which would break the versioning scheme and require a major version bump.

makusu2:
It wouldn't be too difficult to add an OrderedSet class to utils that has all the attributes of list, but I feel like that'd be kinda dumb, and it still wouldn't pass every test (`isinstance(guild.features, list)`) unless it did really crazy stuff like extend both list and frozenset.

Could we add a private _features attribute and have it be the frozenset, and then change features to be a property that returns `list(_features)`? That'd require quite a few extra methods which might get messy :?

bmintz:
@makusu2 The point of this change is to benefit user-facing code, especially since the internal code doesn't use Guild.features at all. Changing the internal behavior so that it only benefits those who depend on implementation details is ultimately not worth it.

Closing this since the API is finalized and this is a breaking change.

PR #2097, opened 2019-04-18
Implement "Partial Messages"

Vexs:
### Summary

This PR implements a "Partial message". A partial message is a message object with no meaningful content attributes, but are instead designed to be user-created and have all available message-related API calls. 
It also refactors `Message` to be a subclass of the new `PartialMessage` class.

With the current design, message objects are not to be user created, but instead generated by the library- either from API calls returning messages, (`.send`, `.history`, .`fetch_message`). However, this creates a non-ideal situation when attempting to do anything to message objects outside of cache, without having to do needless, and in the case of `fetch_message`, heavily ratelimited API calls. 

The `PartialMessage` class has a number of benefits over the other method for the same concept, `channel.raw_do_thing`. It avoids cluttering the existing channel objects with more methods, and maintains the stateful concepts already present in the library. Further, in the event the user wants to do multiple API calls for some "raw" message, it is, to my mind, far cleaner to use a partial- IE
```py
partial = channel.create_partial_message(12345)
await partial.edit(content='hello')
await partial.pin()
```
Versus 
```py
channel.raw_edit_message(message_id=12345, content='hello')
channel.raw_pin_message(message_id=12345)
```
I also believe that `PartialMessage`s will be far less likely to result in the following antipattern:
`channel.raw_edit_message(message.id, content='hello')` Where `message` is a fully-established message object. 

I'm slightly unsure of the implementation- I believe the exposed API to be everything required and more or less clean, but due to DM channels, `_get_channel()` has to be a coro. To avoid making creating the class a coro, (the original reason for `channel.create_partial_message` existing) destinations are now resolved when an API call is attempted to be made by the `@resolve_destination` decorator.
Without this change, single-liners are `await (await channel.create_partial_message(id))` which is more than a little odd. It might be a good idea to split this class entirely from `Message`, instead of making it a parent class. 

The documentation also needs work. 
~~also the commit history for this is a mess yikes~~
### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/__init__.py
+++ b/discord/__init__.py
@@ -29,7 +29,7 @@ from .channel import *
 from .guild import Guild
 from .relationship import Relationship
 from .member import Member, VoiceState
-from .message import Message, Attachment
+from .message import Message, Attachment, PartialMessage
 from .asset import Asset
 from .errors import *
 from .calls import CallMessage, GroupCall
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -854,7 +854,7 @@ class Messageable(metaclass=abc.ABCMeta):
         data = await self._state.http.get_message(channel.id, id)
         return self._state.create_message(channel=channel, data=data)
 
-    async def create_partial_message(self, id):
+    def create_partial_message(self, id):
         """|coro|
         Creates a single :class:`PartialMessage` from the destination.
 
@@ -870,8 +870,7 @@ class Messageable(metaclass=abc.ABCMeta):
         """
         # Avoid circular import
         from .message import PartialMessage
-        channel = await self._get_channel()
-        return PartialMessage(state=self._state, channel_id=channel.id, message_id=id)
+        return PartialMessage(messageable=self, message_id=id)
 
     async def pins(self):
         """|coro|
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -32,6 +32,7 @@ from collections import namedtuple
 from .iterators import HistoryIterator
 from .context_managers import Typing
 from .errors import InvalidArgument, ClientException, HTTPException
+from .message import PartialMessage
 from .permissions import PermissionOverwrite, Permissions
 from .role import Role
 from .invite import Invite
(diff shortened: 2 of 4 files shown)


makusu2:
I feel like if you have the channel ID and you're worried about performance, it'd be cleaner to just pass the ID to the `raw` methods. 
Basically, I prefer the "before" in the before/after you provided. I'd just store 12345 in a variable, which eliminates the hardcodedness of the example you provided.

Rapptz:
I like the idea behind this PR and we've discussed it before in the server. At the time I implemented it a bit differently. However, this implementation is not one I want to merge as it stands. There are many things that do not make sense and I feel could be implemented differently.

1. The naming "PartialMessage" is disingenuous. The name "Partial" is an actual Discord API term for a data model that is returned without full information. This naming scheme is also present in the library. There is nothing "partial" about this message. It is bad naming.
2. The inheritance scheme does not make sense with this class, frankly speaking. They are not that similar except that this is used as a mixin for the HTTP methods. But by inheriting you've caused a few issues:
    1. You no longer redefine `__slots__` in `PartialMessage` which causes the `Message` class to have some weird behaviours, including having a `__dict__` -- something that is undesirable.
    2. Since this inheritance scheme is actually based on a mixin principle you don't actually call `super()`
    3. However, in certain places (despite it being a mixin) you do end up calling `super()` anyway.
    4. A lot of the methods that were moved to the `PartialMessage` class do not have anything to do with the fundamental functionality that the class was meant to have.
3. There is a lot of juggling with the `Messageable` for some reason, something that flat out makes no sense to me. Why does `Messageable` matter? The fact that there's this juggling has tainted almost every method around a strange decorator.
4. There is a call, `abc.Messageable.create_partial_message` which creates this class but once again -- it makes no sense to me why the design was made this way. Every `create_` method in the library as far as I'm aware is a coroutine. This one however is not, quite frankly I'm not even sure *why* it exists. If a class is meant to be user constructible then why shouldn't the user create it using... well the familiar syntax they're all used to? `msg = T(a, b)`.

Since I consider the fundamental design of this PR to be incorrect I think it's better suited if it's implemented differently entirely.

Vexs:
I'll write out thoughts here to add to the discord convo-
1. Send help names suck

2. The inheritance scheme was used to reduce code dupe, and to ensure that it behaves exactly the same as existing message methods- subpoints stem from my misunderstandings of py, and can be resolved, but I think the parent class is still mostly the right way to go.

3/4. Originally this class was not user-createable, but instead only made through the `messageable.create_partial_message` coroutine- which must be a coro to resolve DMs.
 However, this results in unsightly single-liners such as:
`await (await channel.create_partial_message(id).do_thing()`
To avoid this, I moved to `PartialMessage(channel, message_ID)`, but this puts the onus of handling DMs on the user, which the library avoids everywhere else. Thus: the decorator, and Messageable-juggling . 
And yeah, messageable.create_partial_message is vestigial.

PR #2098, opened 2019-04-18
[commands] Update default on_command_error

sirtezza451:
### Summary

Since the cogs update, the default `on_command_error` has never been updated.
This means that the check for a cog error handler is essentially useless.

This pull request makes use of the `_get_overridden_method` method within
the default `on_command_error` to check if `cog_command_error` has been
overridden.

I have ticked `I have updated the documentation to reflect the changes` since
none were needed.

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -161,10 +161,7 @@ class BotBase(GroupMixin):
 
         cog = context.cog
         if cog:
-            handler = getattr(cog, 'cog_command_error')
-            overridden = Cog._get_overridden_method(handler)
-
-            if overridden:
+            if Cog._get_overridden_method(cog.cog_command_error) is not None:
                 return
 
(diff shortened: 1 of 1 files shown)


discord/ext/commands/bot.py
```diff
@@ -161,8 +161,10 @@ def dispatch(self, event_name, *args, **kwargs):
 
         cog = context.cog
         if cog:
-            attr = '_{0.__class__.__name__}__error'.format(cog)
-            if hasattr(cog, attr):
+            handler = getattr(cog, 'cog_command_error')
+            overridden = Cog._get_overridden_method(handler)
```
xxcodianxx:
```suggestion
            if Cog._get_overridden_method(handler):
                return
```

sirtezza451:
Made requested changes.

PR #2103, opened 2019-04-19
PartialEmoji: make constructor args positional again

bmintz:
### Summary

As part of the Asset change, `PartialEmoji.__init__`'s args were made keyword-only.
This reverts a change made as part of be227ebcf0c8bad6b56798339b5414b8da414dc0.

I think that three-argument unnamed PartialEmoji construction is OK because the order matches the format of an emoji string (animated, name, id).

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/emoji.py
+++ b/discord/emoji.py
@@ -67,7 +67,7 @@ class PartialEmoji:
 
     __slots__ = ('animated', 'name', 'id', '_state')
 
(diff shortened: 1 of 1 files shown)


Rapptz:
The only fix that I can imagine doing for this is making `animated` optional and default to `False`.

PR #2104, opened 2019-04-19
[commands] Add command.parents

Vexs:
### Summary

This PR adds the command.parents property, which returns list of all the parents a command has.
Also adjusts command.root_parent to use the new property.

### Checklist

- [ ] This PR fixes an issue.


BluePhoenixGame:
Is it a list of command objects? And in what order is the list, like Is the root parent first or last in the list if you iterate through it. Could be good to document.

PR #2118, opened 2019-04-26
Add read method to attachment objects

Vexs:
### Summary

This adds a `read` method to `Attachment` objects. 
Useful for when you want to read the attachment object and not save it to some file-like object.

Also noticed that the exceptions `http.get_from_cdn` raise are all hardcoded to say `asset`, despite usage in areas besides assets. Seems out of scope for this PR though. 

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -123,6 +123,42 @@ class Attachment:
             with open(fp, 'wb') as f:
                 return f.write(data)
 
+    async def read(self, *, use_cached=False):
+        """|coro|
+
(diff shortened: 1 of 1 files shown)


discord/message.py
```diff
@@ -123,6 +123,42 @@ def is_spoiler(self):
             with open(fp, 'wb') as f:
```
apple502j: `save()` should use `read()`.

PR #2119, opened 2019-04-27
overwrites_for now returns None if no overwrites are set

DevilXD:
### Summary

`overwrites_for()` now returns None if the overwrite for the tested object isn't set. This is to distinguish between a case where an overwrite is set for a member or role but is empty, and a case where no overwrites are set for them at all.
Currently, the only way to distinguish between those is to iterate over all overwrites manually (using the `overwrites()` method), checking if one exists. I feel like this change would make the behavior more logical, because when testing this method, I never expected it to return an actual overwrite if none has been set for the role I was testing.

I updated the documentation to reflect the change, however I'm not that good with doc-strings so I'm not sure if the change is and looks clear enough. It's a subject to change for sure.

This change might be breaking for code that doesn't expect this method to return None.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -314,17 +314,19 @@ class GuildChannel:
 
     def overwrites_for(self, obj):
         """Returns the channel-specific overwrites for a member or a role.
-
+        
+        If no overwrites are set, returns None.
+        
         Parameters
         -----------
         obj
             The :class:`Role` or :class:`abc.User` denoting
             whose overwrite to get.
-
+            
         Returns
         ---------
         :class:`PermissionOverwrite`
-            The permission overwrites for this object.
+            The permission overwrites for this object, if set. Returns None otherwise.
         """
 
         if isinstance(obj, User):
@@ -340,7 +342,7 @@ class GuildChannel:
                 deny = Permissions(overwrite.deny)
                 return PermissionOverwrite.from_pair(allow, deny)
 
-        return PermissionOverwrite()
+        return None
 
     @property
     def overwrites(self):


discord/abc.py
```diff
@@ -324,7 +326,7 @@ def overwrites_for(self, obj):
         Returns
         ---------
         :class:`PermissionOverwrite`
```
Gorialis: Functions/properties that can return None are usually marked `Optional`.
```suggestion
         Optional[:class:`PermissionOverwrite`]
```

discord/abc.py
```diff
@@ -340,7 +342,7 @@ def overwrites_for(self, obj):
                 deny = Permissions(overwrite.deny)
                 return PermissionOverwrite.from_pair(allow, deny)
 
-        return PermissionOverwrite()
+        return None
```
Gorialis: `return None` is redundant, a function will implicitly return None if it returns nothing at all.

discord/abc.py
```diff
@@ -315,6 +315,8 @@ def created_at(self):
     def overwrites_for(self, obj):
         """Returns the channel-specific overwrites for a member or a role.
 
+        If no overwrites are set, returns None.
```
Gorialis: I feel this should probably be merged with the above sentence, e.g.
```
Returns the channel-specific overwrites for a member or role, or None if no overwrites are set.
```
It might not be necessary at all, however, since it's mentioned in the return signature.

discord/abc.py
DevilXD: Interesting, didn't know that. Shouldn't it be left for clarity tho? I kinda want to leave it there to explicitly state that if no overwrites match, I want to return None instead of the usual overwrite that'd be returned two lines higher.

discord/abc.py
DevilXD: Changed per suggestion

PR #2124, opened 2019-04-29
TextChannel edit method now specifies slowmode unit

DevilXD:
### Summary

The doc string for the edit method of a TextChannel only specified the maximum amount, but was missing the unit, so could be confusing. A `seconds` unit is now clearly specified, just like in the TextChannel doc string itself.

### Checklist

- [ ] This PR fixes an issue.


Tarek1337:
While you're at it, could you do the same for https://github.com/Rapptz/discord.py/blob/746e20a826d1b0168f7e0566d502dfabc5f75184/discord/guild.py#L678-L680 as well? The max value is also incorrect.

DevilXD:
Of course, it should be good now. Not sure in what other places can slowmode setting be specified.

PR #2126, opened 2019-04-29
Fix MessageConverter

davidecotten:
### Summary

MessageConverter was not working for me. I dug through this module and this change made it work. Thanks!

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -172,7 +172,7 @@ class UserConverter(IDConverter):
 
         return result
 
-class MessageConverter:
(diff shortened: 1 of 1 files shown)


khazhyk:
What was the error you saw, that is fixed by this? I see no reason we should inherit from IDConverter here

khazhyk:
(that said, it probably should inherit from Converter at the least)

Rapptz:
This is actually the wrong way to fix this. There's no reason to inherit from `IDConverter`, instead it should inherit `commands.Converter`.

PR #2147, opened 2019-05-11
Remove Member.unban()

sirtezza451:
### Summary

This PR removes `Member.unban()` since it is useless due to the fact that the `Member` object doesn't exist anymore after it gets banned.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/member.py
+++ b/discord/member.py
@@ -432,13 +432,6 @@ class Member(discord.abc.Messageable, _BaseUser):
         """
         await self.guild.ban(self, **kwargs)
 
-    async def unban(self, *, reason=None):
-        """|coro|
-
-        Unbans this member. Equivalent to :meth:`Guild.unban`
-        """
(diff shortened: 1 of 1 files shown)


scragly:
This is definitely not useless :thinking: 
It's use in a lot of bots. Just because you're of the opinion that it doesn't suit your personal usage doesn't mean it doesn't suit others.

Gorialis:
This PR is misguided. Consider, for instance, the following example:
```py
@bot.command()
@commands.has_permissions(ban_members=True)
async def softban(ctx: commands.Context, *, member: discord.Member):
    await member.ban(delete_message_days=7)
    await member.unban()
```
The `Member` object does not become invalid once it leaves the guild. It is ejected from the cache, but continues to exist until its reference count reaches zero (such as when it leaves the scope in which it was assigned).

PR #2148, opened 2019-05-11
Use tasks extension for examples

apple502j:
### Summary
Background Task example now uses Tasks.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/background_task.py
+++ b/examples/background_task.py
@@ -1,12 +1,13 @@
 import discord
+from discord.ext import tasks
 import asyncio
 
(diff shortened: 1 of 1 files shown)


Rapptz:
I'm not sure I want to completely replace the original pure `asyncio` example. If it were to be replaced I think it'd be good to have a second example showing the pure `asyncio` version without the tasks extension.

TAG-Epic:
Maby make another example for this rather than rewriting it?

PR #2150, opened 2019-05-11
Update colour.py (Improved one)

KowlinMC:
Add the colour black


avayert:
The value `0x000000` doesn't represent black, it represents no colour at all (default)

A-Trash-Coder:
just rename to black

PR #2152, opened 2019-05-12
Fixed Middle Dot for MinimalHelpCommand

Soheab:
from U+2002 to U+2022 so it works and shows now and added a space before and after it so it looks better.

### Summary

Fixes the dot between commands of the MinimalHelpCommand

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/help.py
+++ b/discord/ext/commands/help.py
@@ -1105,8 +1105,8 @@ class MinimalHelpCommand(HelpCommand):
             The heading to add to the line.
         """
         if commands:
-            # U+2002 Middle Dot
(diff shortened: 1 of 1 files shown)


Rapptz:
The comment is just out of date. 

The first version of this used the middle dot as you proposed but I changed it because people in the server loudly complained that it was ugly and they didn't like it, even if I personally didn't think so.

bmintz:
Why is it joined with an EN SPACE currently, rather than an ascii space?

Rapptz:
Because it's more spaced out for readability purposes. 

Should be noted that if someone wants to override it they're free to do so. This function is documented and pretty easy to override.

PR #2154, opened 2019-05-13
[tasks] Return coro from before and after decorators

SnowyLuma:
### Summary

This PR allows registering the before or after coroutine to multiple tasks, or calling it outside of task context. Previously this was not possible as the decorators did not return the registered coroutine.

### Checklist

- [x] This PR fixes an issue.


Rapptz:
Not sure how I missed this. Thanks.

MinerChAI:
Also would be good to make command be callable, so you can have multiple `.command` decorators for one callable

Issue #2135, opened 2019-05-03
Issue with token verification


VoxelPrismatic:
### Summary

Able to log in and do whatever I want, but upon trying to send a message a `411 Length Required` error is thrown

### Reproduction Steps

Just add `\n` to the end of your token
**EDIT** Apperently you need 2 newlines, so put `\n\n` at the end of your token

### Expected Results

I expect that the invalid token was caught earlier, preferably before logging in

### Actual Results

I could log in to my bot with an invalid key, and only sending a message threw an error
Traceback is [here](https://imgur.com/gallery/kAe87A9) (imgur cuz I don't have the copy-paste)

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [x] I have removed my token from display, if visible.

### System Information

- discord.py version: `1.0.1a RW`
- Python version: `3.7.3`
- Operating system: `Kubuntu 18.10 Cosmic Cuttlefish`

CapnS:
I have tested this with two bots, and I cannot reproduce this. It throws an error on login because the authentication is incorrect.

VoxelPrismatic:
I'm not too sure, but given that you cannot reproduce this, it may be a Linux specific issue

CapnS:
On linux, I get a similar, yet slightly different error when having a newline at the end of my token.

Traceback:
```
Traceback (most recent call last):
  File "testbot.py", line 10, in <module>
    bot.run(TOKEN_REDACTED)
  File "/usr/local/lib/python3.6/dist-packages/discord/client.py", line 544, in run
    self.loop.run_until_complete(runner())
  File "/usr/lib/python3.6/asyncio/base_events.py", line 473, in run_until_complete
    return future.result()
  File "/usr/local/lib/python3.6/dist-packages/discord/client.py", line 539, in runner
    await self.start(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/discord/client.py", line 510, in start
    await self.login(*args, bot=bot)
  File "/usr/local/lib/python3.6/dist-packages/discord/client.py", line 376, in login
    await self.http.static_login(token, bot=bot)
  File "/usr/local/lib/python3.6/dist-packages/discord/http.py", line 258, in static_login
    data = await self.request(Route('GET', '/users/@me'))
  File "/usr/local/lib/python3.6/dist-packages/discord/http.py", line 222, in request
    raise HTTPException(r, data)
[...]

PR #2161, opened 2019-05-16
A possible fix for #2135

davfsa:
Makes sure that the token has no new lines when it is passed so no errors are encountered

This pull request will fix the issue stated in #2135

### Checklist

- [x] This PR fixes an issue.


Rapptz:
I don't quite get the point of this. Likewise this isn't really the proper solution even if I wanted to change it.

PR #2174, opened 2019-05-20
Implement `Webhook.move_to`.

### LINKED ISSUE #1761: No way to edit webhook channel

NCPlayz:
### Summary

This allows you to move a webhook to another channel, which could resolve #1761. Unlike the previous Pull Requests, this seperates it from `Webhook.edit`. 
Two reasons: 

- This technically uses a different endpoint.
- This also requires ConnectionState. 
    - Since ConnectionState relies on being fetched from `Client`, `Guild`, or `TextChannel` (in this case), this would mean it could raise an exception, interrupting the rest of `Webhook.edit`. 
    - By seperating it, it would only raise an exception when you specifically want to move the webhook. 

However, **I am willing to incorporate it as part of `Webhook.edit`** if preferred.

Another benefit of my PR, I set `Webhook.channel_id`. :)

### Sample code

```py
import discord

client = discord.Client()

@client.event
async def on_ready():
    global client # dont do this :P

    channel = client.get_channel(id)  # another channel
    webhook = await client.fetch_webhook(id)
    await webhook.move_to(channel=channel)

client.run("token")
```

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/webhook.py
+++ b/discord/webhook.py
@@ -653,16 +653,12 @@ class Webhook:
         return self._adapter.edit_webhook(**payload)
 
     async def _async_move(self, channel_id, token, payload):
-        try:
-            await self._adapter.move_webhook(token, **payload)
-        finally:
-            self.channel_id = channel_id
+        await self._adapter.move_webhook(token, **payload)
+        self.channel_id = channel_id
 
     def _sync_move(self, channel_id, token, payload):
-        try:
-            self._adapter.move_webhook(token, **payload)
-        finally:
-            self.channel_id = channel_id
+        self._adapter.move_webhook(token, **payload)
+        self.channel_id = channel_id
 
     def move_to(self, channel):
         """|maybecoro|
@@ -680,7 +676,7 @@ class Webhook:
 
         Parameters
         -----------
-        channel: Union[:class:`TextChannel`]
+        channel: :class:`TextChannel`
             Where to move the webhook. Must be a text channel.
 
         Raises
--- a/discord/webhook.py
+++ b/discord/webhook.py
@@ -58,6 +58,7 @@ class WebhookAdapter:
         self._webhook_id = webhook.id
         self._webhook_token = webhook.token
         self._request_url = '{0.BASE}/webhooks/{1}/{2}'.format(self, webhook.id, webhook.token)
+        self._move_url = '{0.BASE}/webhooks/{1}'.format(self, webhook.id)
         self.webhook = webhook
 
     def request(self, verb, url, payload=None, multipart=None):
@@ -87,6 +88,9 @@ class WebhookAdapter:
 
     def edit_webhook(self, **payload):
         return self.request('PATCH', self._request_url, payload=payload)
+
+    def move_webhook(self, token, **payload):
+        return self.request('PATCH', self._move_url, payload=payload, token=token)
 
     def handle_execution_response(self, data, *, wait):
         """Transforms the webhook execution response into something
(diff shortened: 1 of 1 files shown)


discord/webhook.py
```diff
@@ -642,6 +652,66 @@ def edit(self, **kwargs):
...
+
+        .. warning::
+
+            This will only work if you fetch the webhook from a client, as there is a state attached.
+
+        .. versionadded:: 1.2.0
+
+        Parameters
+        -----------
+        channel: Union[:class:`TextChannel`]
```
MinerChAI: Don't need Union i think

NCPlayz: Good catch 😄

Vexs:
As a thought, making this a method of channels instead of webhook removes a fair bit of nastiness. You need a logged in bot to get the channels anyway.

NCPlayz:
Well, a reason I made it as part of Webhook is because the same style is used for [`VoiceClient.move_to`](https://github.com/Rapptz/discord.py/blob/master/discord/voice_client.py#L290). It didn't really make sense to make it part of Channel since you would usually create/fetch the object there. Though, I do agree that it would make it much more tidy.

discord/webhook.py
```diff
@@ -642,6 +652,62 @@ def edit(self, **kwargs):
 
         return self._adapter.edit_webhook(**payload)
 
+    async def _async_move(self, channel_id, token, payload):
+        await self._adapter.move_webhook(token, **payload)
+        self.channel_id = channel_id
+
+    def _sync_move(self, channel_id, token, payload):
```
apple502j: No need, see below.

discord/webhook.py
```diff
@@ -642,6 +652,62 @@ def edit(self, **kwargs):
...
+        self._adapter.move_webhook(token, **payload)
+        self.channel_id = channel_id
+
+    def move_to(self, channel):
+        """|maybecoro|
+
+        Moves this Webhook.
+
+        If the webhook is constructed with a :class:`RequestsWebhookAdapter` then this is
```
apple502j: If you get Webhook from Client/TextChannel/Guild, it will always use Async adapter, so this should be omitted.

discord/webhook.py
```diff
@@ -642,6 +652,62 @@ def edit(self, **kwargs):
...
+        self.channel_id = channel_id
+
+    def _sync_move(self, channel_id, token, payload):
+        self._adapter.move_webhook(token, **payload)
+        self.channel_id = channel_id
+
+    def move_to(self, channel):
+        """|maybecoro|
+
+        Moves this Webhook.
```
apple502j: Moving webhook is not a discord term (Moving members is) , so `Moves this Webhook to other channel.` may be better.

discord/webhook.py
```diff
@@ -642,6 +652,62 @@ def edit(self, **kwargs):
...
+        .. warning::
+
+            This will only work if you fetch the webhook from a client, as there is a state attached.
+
+        .. versionadded:: 1.2.0
+
+        Parameters
+        -----------
+        channel: :class:`TextChannel`
+            Where to move the webhook. Must be a text channel.
```
apple502j: No need to say "Must be a text channel." because the line just above says TextChannel.

discord/webhook.py
```diff
@@ -642,6 +652,62 @@ def edit(self, **kwargs):
...
+        payload = {'channel_id': channel.id}
+
+        token = channel._state.http.token
+
+        if self._state.is_bot:
+            token = 'Bot ' + token
+
+        if isinstance(self._adapter, AsyncWebhookAdapter):
+            return self._async_move(channel.id, token, payload)
+        elif isinstance(self._adapter, RequestsWebhookAdapter):
```
apple502j: No need to include sync version as it will never work.

PR #2181, opened 2019-05-22
Change system information for bug reports.

NCPlayz:
### Summary

This changes the system information section for bug reports. Mainly due to the built-in CLI now returning formatted version info via `python -m discord -v`.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -30,6 +30,5 @@ about: Report broken or incorrect behaviour
 
 ### System Information
 
-- discord.py version:
-- Python version:
-- Operating system:
+<!-- Run `python -m discord -v` and paste this information below. -->
+<!-- This command is available in v1.1.0 or higher. If you are unable to run this command, paste basic info (ie. Python version, library version, and your operating system -->


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -30,6 +30,10 @@ about: Report broken or incorrect behaviour
 
 ### System Information
 
-- discord.py version:
-- Python version:
-- Operating system:
+<!-- Do `python -m discord -v` and paste this information.-->
```
laggycomputer: I'm not sure if "do" is the most appropriate here. It might not be obvious to beginners that this is a shell command.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -30,6 +30,10 @@ about: Report broken or incorrect behaviour
...
+
+- Python
```
laggycomputer: Should there be colons after these lines? I also think that the word "version" should still be included as well.

.github/ISSUE_TEMPLATE/bug_report.md
RDIL:
```suggestion
<!-- Run `python -m discord -v` and paste this information.-->
```

laggycomputer: Sounds good to me. Might need `at the command line` after the command though.

Modelmat: I don't think `at` is what should be used for that; in or on would be better.

Rapptz:
```suggestion
<!-- Run `python -m discord -v` and paste this information.-->
```

PR #2188, opened 2019-05-26
[commands] Make Command.brief default to the first line of the help doc

sirtezza451:
### Summary

According to the documentation for `Command.brief`, it defaults to the first line of `Command.help`:
```
brief
str – The short help text for the command. If this is not specified then the first line of the long help text is used instead.
```

In reality, that is not the case. This PR makes it so.  
No documentation changes were needed.

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -215,7 +215,7 @@ class Command(_BaseCommand):
 
         self.help = help_doc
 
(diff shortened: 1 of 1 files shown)


Rapptz:
The documentation is slightly incorrect here.

This behaviour is what powers `short_doc`.

PR #2190, opened 2019-05-27
Update Permissions to accept kwargs

Gorialis:
### Summary

This adds `**kwargs` to the Permissions signature, aligning it with the similar `PermissionOverwrite` signature that functions in a similar manner.

This allows shortening of some more verbose snippets:
```diff
- p = discord.Permissions()
- p.update(manage_messages=True)
- await guild.create_role(name='Janitor', permissions=p)
+ await guild.create_role(name='Janitor', permissions=discord.Permissions(manage_messages=True))
```

At the moment this functions using a direct call to `Permissions.update`, which does not raise on invalid permissions. It may be a good idea to adjust this behavior to align with `PermissionOverwrite` (which does raise), either in `__init__` or by changing `Permission.update` itself.

It may also be a good idea to implement this functionality as a `classmethod` if this signature change is undesirable.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/permissions.py
+++ b/discord/permissions.py
@@ -25,7 +25,7 @@ DEALINGS IN THE SOFTWARE.
 """
 
 class Permissions:
-    """Wraps up the Discord permission value.
+    r"""Wraps up the Discord permission value.
 
     The properties provided are two way. You can set and retrieve individual
     bits using the properties as if they were regular bools. This allows
@@ -59,6 +59,17 @@ class Permissions:
                Returns an iterator of ``(perm, value)`` pairs. This allows it
                to be, for example, constructed as a dict or a list of pairs.
 
+    Parameters
+    -----------
+    permissions: :class:`int`
+        The raw permissions value to instantiate this class with.
+        See :attr:`value` for information.
+    \*\*kwargs
+        A list of key/value pairs to set permission values with.
+        The values of the permissions passed will take priority over their
+        respective bit flag in ``permissions``, if it is provided.
+        Permissions not passed default to ``False``.
+
     Attributes
     -----------
     value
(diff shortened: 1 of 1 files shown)


Rapptz:
This PR is a duplicate of a denied feature request seen here: https://github.com/Rapptz/discord.py/pull/237

Unfortunately the discussion of its rejection took place in DAPI and due to the date the history regarding its removal has been removed due to the November 19th incident.

At the time the original contentions were:

1. Raising `AttributeError` inside `__init__` is very peculiar for invalid inputs.
2. Silently discarding invalid input (especially in `__init__`) can be a bit confusing considering how error prone a lot of the keyword arguments can be (missing plural? server vs guild?, etc.).

Point 2 was dismissed for `update` due to a combination of bad design and wanting to mimic the non-throwing `update` API of `dict`.

So I'm not sure how I feel about this PR.

bmintz:
I don't think this is doing more than one thing per se. It's just initializing the permissions with a value, as in any `__init__`. Same goes for dict or PartialEmoji. (2) is still tricky though. Maybe an internal helper could be added, which validates each kwarg before passing it to .update().

Rapptz:
Consider: `Permissions(value=0xFF, manage_messages=True)`

This is what I meant by "more than one thing".

bmintz:
Oh I see what you mean. Perhaps update() could return self then, as an alternative. Or,

```
🐍 dict({'x': 1}, x=2, y=3)
{'x': 2, 'y': 3}
```

allow both? I dunno.

Rapptz:
This has been superseded by https://github.com/Rapptz/discord.py/commit/2770137bd6d53dbfbeb325f74f302cf1ffd26f99

PR #2193, opened 2019-05-30
[commands] Allow multiple checks to be added at once.

Bluenix2:
This PR allows multiple checks to be added at once, making code a little bit more compact and pretty: 
```py
def check_if_it_is_me(ctx):
    return ctx.message.author.id == 85309593344815104

def check_if_in_discord_py(ctx):
    return ctx.message.guild.id == 336642139381301249

@bot.command()
@commands.check(check_if_it_is_me, check_if_in_discord_py)
async def only_for_me(ctx):
    await ctx.send('I know you are in the discord.py guild!')
```
It does this by turning the predicate argument into an argument list and iterating through it when adding the checks.

If have tested the code i have changed and adding mutliple checks works, adding just one works, and turning one into it's own decorator works. I have also updated the documentation and added an example that i have tested.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1447,8 +1447,8 @@ def group(name=None, **attrs):
     attrs.setdefault('cls', Group)
     return command(name=name, **attrs)
 
-def check(predicate):
-    r"""A decorator that adds a check to the :class:`.Command` or its
+def check(*predicates):
+    r"""A decorator that adds a check or more to the :class:`.Command` or its
     subclasses. These checks could be accessed via :attr:`.Command.checks`.
 
     These checks should be predicates that take in a single parameter taking
@@ -1483,6 +1483,9 @@ def check(predicate):
     .. versionchanged:: 1.3
         The ``predicate`` attribute was added.
 
+    .. versionchanged:: 1.7
+        Allow several checks to be added at once.
+
     Examples
     ---------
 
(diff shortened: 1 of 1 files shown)


EEKIM10:
why not just have multiple decors?

BluePhoenixGame:
Well, why not just "compress" them into one line?

BluePhoenixGame:
Resolved conflicts with 77a0931

BluePhoenixGame:
Updated the versionchanged to 1.3.0 since pull request was never merged before the 1.2.0 release.

discord/ext/commands/core.py
```diff
@@ -1304,18 +1323,19 @@ async def only_me(ctx):
...
-                func.__commands_checks__ = []
+        for predicate in predicates:
+            if isinstance(func, Command):
+                func.checks.append(predicate)
+            else:
+                if not hasattr(func, '__commands_checks__'):
+                    func.__commands_checks__ = []
 
-            func.__commands_checks__.append(predicate)
+                func.__commands_checks__.append(predicate)
```
avayert, line 1338:
```py
if isinstance(func, Command):
    func.checks.extend(predicates)
else:
    if not hasattr(func, '__commands_checks__'):
        func.__commands_checks__ = []
    func.__commands_checks__.extend(predicates)
```
This removes a level of nesting and avoids repeated conditional calls

BluePhoenixGame: I see, great!

BluePhoenixGame:
Updated the versionchanged to the most recent upcoming update 1.3.4

Gobot1234:
Shouldn't it be 1.4 as 1.3.4 was just a bug hot fix?

Harmon758:
v1.3.4 was already released:
https://github.com/Rapptz/discord.py/releases/tag/v1.3.4
https://pypi.org/project/discord.py/1.3.4/
and yeah, this would almost certainly be part of a minor rather than patch version bump.

BluePhoenixGame:
My bad guys, sorry.

PR #2194, opened 2019-05-31
Rename invoke_without_command

BluePhoenixGame:
Renames `invoke_without_command` to `only_invoke_without_command`. This is as `invoke_without_command` makes it sound like parents of subcommands normally don't run without a subcommand which is not the case. This PR adds a `only` to the name of the parameter to ensure that anyone don't get the wrong idea and misunderstand while learning discord.py. The extra character length i get might be a problem. 

I have tested the code, and updated the documentation to reflect the changes. This PR is a breaking change though which i know is not preferable.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1107,7 +1107,7 @@ class Group(GroupMixin, Command):
 
     Attributes
     -----------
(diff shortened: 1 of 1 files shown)


Rapptz:
This isn't gonna happen.

PR #2201, opened 2019-06-06
Add role cooldown bucket

BluePhoenixGame:
### Summary

This pull request adds another variable to the BucketType class. And modifies _bucket_key to  support that. Meaning that this pull request adds another cooldown bucket.

I chose to use top_tole as the role to use for the bucket. It wasn't possible to use all roles as the only way would be to return a tuple of all role IDs which would mean it would be a cooldown for every combination of roles so instead I chose the role that I feel mostly represents a members roles, the top role.

I have updated the documentation and tested my code. Also this pull request is not a breaking change but more of a new feature.


TAG-Epic:
Would be really useful

discord/ext/commands/cooldowns.py
```diff
@@ -127,6 +128,8 @@ def _bucket_key(self, msg):
             return ((msg.guild and msg.guild.id), msg.author.id)
         elif bucket_type is BucketType.category:
             return (msg.channel.category or msg.channel).id
+        elif bucket_type is BucketType.role:
+            return msg.author.top_role.id
```
Rapptz: This breaks when it's in a DM context.

For some reason so does the `BucketType.category` one. I must have missed that one during review.

BluePhoenixGame: Just want to know your thoughts on this, would checking if you're in a DM context and then returning the user id be a good fix for that bug?

Rapptz: Sure

BluePhoenixGame: Fixed with the latest commit 6c5fee0

PR #2205, opened 2019-06-08
Document command_prefix type

BluePhoenixGame:
### Summary
This pr documents what type the bots command_prefix should be. I didn't really know how to document a callable correctly but i looked at other documentation and i think i got it right. Let me know if i didn't.
### Checklist

- [ ] This PR fixes an issue.


Rapptz:
This type is incorrect. It's left out because the signature would be too complicated and would just be noise.

Issue #133, opened 2016-03-14
Expose metadata endpoints


khazhyk:
api/guilds/:id has metadata for servers (this seems to require you're joined to the server) (this also has full role info)
api/guilds/:id/members/:id can be used to get member metadata, this might be useful for non-ws usage of the lib. (this requires you are joined to the server)
api/guilds/:id/channels - useful for non-ws (requires you're joined to server)

/api/v6/users/@me/guilds - some others too

NCPlayz:
Although how old it is, it does affect #1822, so I thought of two ways this could be implemented (in rewrite).

## Implementation No. 1

```py
...
    async def get_guild(self, id, *, metadata=False):
        """Returns a :class:`Guild` with the given ID. If not found, returns None."""
        if metadata:
            data = await self.http.get_guild(id)
            return Guild(data=data, state=self._connection)
        return self._connection._get_guild(id)
```

This would allow the user to change between if they want to use the WS content, or the http request. The only problem with this is that it could induce a breaking change, since you would have to `await` them, which currently is not required.

## Implementation No. 2

```py
...
    async def get_guild_metadata(self, id):
        """Returns a :class:`Guild` with the given ID via the API. If not found, returns None."""
        data = await self.http.get_guild(id)
        return Guild(data=data, state=self._connection)
```

This would not be a breaking change, but it does mean that it would be longer to type, if that's a valid issue.

Both implementations use this in `discord.http.HTTPClient`:
```py
...
    def get_guild(self, guild_id):
        return self.request(Route('GET', '/guilds/{guild_id}', guild_id=guild_id))
```


Would any of these implementations be viable in your eyes, @Rapptz?

khazhyk:
Can we have a more obvious difference between get_ methods which touch
local cache vs get_ methods which cause an http request? (I guess that
would be a breaking change but 1.0 isn't out yet...) E.g. people get
confused that get_user_info is an http request, vs. get_user is cache, and
this would introduce a second case. Maybe fetch_ prefix or something...

NCPlayz:
You raise an interesting point, since they should be to avoid confusion. Using `get_` for a http request would definitely make sense since you are *literally* sending a `GET` request.
As you mentioned, discord.py 1.0.0 has not been released yet, which could mean that there's an excuse to change it now rather than after release. 
Although I am with you on this, I feel like changing how it is named would require *a lot* of changes (around 38 function name changes, and a possible more than 50 call changes based off what IntelliJ is telling me).

[...]

PR #2208, opened 2019-06-08
Implement `Guild.fetch_roles`

NCPlayz:
### Summary

This implements yet another endpoint listed in #133, `/guilds/:id/roles`, in the form of `Guild.fetch_roles`. There are two reasons for why this is not an iterator:

- There can only be a finite amount of roles per guild.
- There are no parameters available for the endpoint.

However, I am willing to add any parameters to the function, since I planned to be able to reverse it but I wasn't too sure about it.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -1469,6 +1469,30 @@ class Guild(Hashable):
         data = await self._state.http.create_custom_emoji(self.id, name, img, roles=roles, reason=reason)
         return self._state.store_emoji(self, data)
 
+    async def fetch_roles(self):
+        """|coro|
+
+        Retrieves all :class:`Role` that the guild has.
+
+        .. note::
+
+            This method is an API call. For general usage, consider :attr:`roles` instead.
+
+        .. versionadded:: 1.3.0
+
+        Raises
+        -------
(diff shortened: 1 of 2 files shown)


discord/guild.py
```diff
@@ -1469,6 +1469,31 @@ def convert(d):
...
+    async def fetch_roles(self):
+        """|coro|
+
+        Retrieves all :class:`Role` that the guild has.
+
+        .. note::
+
+            This method is an API call. For general usage, consider :attr:`roles` instead.
+
+        .. versionadded:: 1.2.0
```
Rapptz:
```suggestion
        .. versionadded:: 1.3.0
```

discord/guild.py
```diff
@@ -1469,6 +1469,31 @@ def convert(d):
...
+        HTTPException
+            Retrieving the roles failed.
+
+        Returns
+        -------
+        List[:class:`Role`]
+            All roles in the guild.
+        """
+        data = await self._state.http.get_roles(self.id)
+
```
Rapptz:
```suggestion
```

PR #2209, opened 2019-06-09
Update docs to note Enum derivation change

Tobotimus:
### Summary

The docs previously stated that all enumerations derived from `enums.Enum`. I've updated the docs to state this isn't technically true.

### Checklist

- [x] I have updated documentation.


BluePhoenixGame:
Yeah, that is needed.

PR #2218, opened 2019-06-13
catch asyncio.CancelledError in 3.8 when calling fut.exception() on a Task object

bmintz:
In python 3.8, asyncio.CancelledError is a subclass of BaseException rather than Exception,
so `except Exception:` will not swallow CancelledError. This change prevents an error in 3.8 from being printed
to the console when the following is run:

```
async with ctx.typing():
	pass
```

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/context_managers.py
+++ b/discord/context_managers.py
@@ -30,7 +30,7 @@ def _typing_done_callback(fut):
     # just retrieve any exception and call it a day
     try:
         fut.exception()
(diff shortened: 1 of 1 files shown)


xoxfaby:
https://bugs.python.org/issue32528

Makes sense to me, this change is definitely needed for 3.8 compatibility.

PR #2219, opened 2019-06-13
Option to disable message cache

JellyWX:
### Summary

Message caching is not a feature required by all bots. This PR adds an option to the client that disables message caching within the connection state.

Extended: increase the options for 'pickiness' regarding caching, including options for which attributes should be stored.

### Checklist

- [ ] This PR fixes an issue.

I'm unsure if forced message caching is something enforced by Discord, but [Serenity](https://github.com/serenity-rs/serenity) has all caching as optional and for a bot of mine the cache goes completely unused :woman_shrugging:


Vexs:
There isn't really much reason to disable it completely- it already has comparatively little ram usage, the member/roles/user caches are generally many times larger. It can also be set to 100, which will make it's usage almost null by comparison.

Implementation wise, makes more sense to me to just reduce the minimum to zero, or replace the cache deque with a null iterator.

JellyWX:
Okay- just ran some calculations and I predict the message cache in the most ideal situation is ~1% of the total memory usage.

Even if disabling it isn't an option, I still think having the minimum bound is a little unnecessary- people should be able to set the cache to whatever length they want (minimum bound is currently 100).

And your points on implementation are far superior to my actual implementation. I'll consider this in my fork

Rapptz:
Hello.

I don't really like the implementation of this PR in terms of performance and I also don't like how it's crept in scope with other things that I definitely do not agree with on principle.

PR #2222, opened 2019-06-15
Add class `Setup` and function `setup`

sizumita:
You can add cog without function `setup(bot)` like this:
```py
@commands.setup
class TestCog(commands.Cog):
    def __init__(self, bot):
        self.bot = bot
```

### Summary

In this pull request, you don't have to write function `setup`.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/__init__.py
+++ b/discord/ext/commands/__init__.py
@@ -10,7 +10,7 @@ An extension module to facilitate creation of bot commands.
 :license: MIT, see LICENSE for more details.
 """
 
-from .bot import Bot, AutoShardedBot, when_mentioned, when_mentioned_or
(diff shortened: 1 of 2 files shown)


ShineyDev:
i have a strong feeling this won't work correctly if there are multiple cogs in the extension. on top of this i don't know if it'll be taken in at all since it's another breaking change.

for these reasons i disapprove.

MusicOnline:
Extensions aren't only for cogs. I have a `help.py` extension which loads a custom `HelpCommand` in setup. In teardown, it will reload the original `DefaultHelpCommand`. This PR does not consider when extensions do not want to load a cog or when extensions want to other setup work before and after adding the cog.

sizumita:
I did't think multiple cogs. Thanks.

PR #2223, opened 2019-06-15
Added default_setup parameter to load_extension

OrangutanGaming:
### Summary

Adds `default_setup` parameter to `load_extension` and the other relavent methods.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -568,13 +568,21 @@ class BotBase(GroupMixin):
                 if _is_submodule(name, module):
                     del sys.modules[module]
 
-    def _load_from_module_spec(self, lib, key):
+    def _load_from_module_spec(self, lib, key, default):
         # precondition: key not in self.__extensions
         try:
             setup = getattr(lib, 'setup')
         except AttributeError:
-            del sys.modules[key]
-            raise errors.NoEntryPointError(key)
+            if not default:
+                del sys.modules[key]
+                raise errors.NoEntryPointError(key)
+            else:
+                def setup(bot):
+                    for var_name in dir(lib):
+                        var = getattr(lib, var_name)
+                        if (inspect.isclass(var) and issubclass(var, Cog) and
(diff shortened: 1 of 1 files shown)


OrangutanGaming:
Could also pass if the variable starts with `__`, wasn't 100% sure if that's needed.

Hornwitser:
What is the rationale behind this?  How would you use this?  Have you considered any alternatives to this?

OrangutanGaming:
It's based off #2222 and I realised that since practically everyone uses the same code anyway, why even have people do it when it can be done by the library.
If you have lots of cogs in one file or just want a cleaner file, this would provide a way to have the library deal with it instead, so you would just not define a `setup` and set the `kwarg` to True.
Comparing it to some of the other ideas suggested in the discord, I think this is one of the best ways this idea could be implemented.

Hornwitser:
I still do not understand how you're supposed to use this.

OrangutanGaming:
You use `bot.load_extension('cog', default_setup=True)`, and if the extension does not have a `setup` defined, it will make one for you

Hornwitser:
You didn't actually test this did you?  Your setup function looks for an instance of `Cog` (or any subclasses of it), not a class itself, and then treats it like it was a cog class.

Hornwitser:
My main objection to this is that it's unnecessary troublesome magic.  What you save is approximately 2 lines of code per extension.  But what you get is poorly behaving magic, where the right behavior is difficult to even imagine.  Consider a cog implementation that uses a base class, will the base class be included?  What if you have an extension with multiple cogs imported from separate files?  What if you don't find any cogs, do you ignore that or error out?

Rapptz:
This type of functionality is not one I ever want to propagate.

The idea that a cog is equivalent of a lazy extension is one that I go out of my way to dispel not one that I want to propagate. For this reason any PR offering shortcuts that blur the line between them even more is not one I want to accept.

PR #2239, opened 2019-06-21
Support team members data in application info

fourjr:
### Summary
New data has been added to `/oauth2/applications/@me` with team data (See https://github.com/discordapp/discord-api-docs/issues/787#issuecomment-503660497)

This PR allows for access within the team and adds some shop-related attributes into AppInfo.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/appinfo.py
+++ b/discord/appinfo.py
@@ -26,6 +26,7 @@ DEALINGS IN THE SOFTWARE.
 
 from .user import User
 from .asset import Asset
+from .team import Team
 
 
 class AppInfo:
@@ -40,6 +41,8 @@ class AppInfo:
         The application name.
     owner: :class:`User`
         The application owner.
+    team: Optional[:class:`Team`]
+        The application's team.
     icon: Optional[:class:`str`]
         The icon hash, if it exists.
     description: Optional[:class:`str`]
@@ -52,9 +55,28 @@ class AppInfo:
         grant flow to join.
     rpc_origins: Optional[List[:class:`str`]]
         A list of RPC origin URLs, if RPC is enabled.
+    summary: Optional[:class:`str`]
+        If this application is a game sold on Discord,
+        this field will be the summary field for the store page of its primary sku
+    verify_key: Optional[:class:`str`]
+        The base64 encoded key for the GameSDK's GetTicket
+    guild_id: Optional[:class:`int`]
+        If this application is a game sold on Discord,
+        this field will be the guild to which it has been linked
+    primary_sku_id: Optional[:class:`int`]
+        If this application is a game sold on Discord,
+        his field will be the id of the "Game SKU" that is created, if exists
+    slug: Optional[:class:`str`]
+        If this application is a game sold on Discord,
+        this field will be the URL slug that links to the store page  
+    cover_image: Optional[:class:`str`]
+        If this application is a game sold on Discord,
+        this field will be the hash of the image on store embeds    
     """
     __slots__ = ('_state', 'description', 'id', 'name', 'rpc_origins',
-                 'bot_public', 'bot_require_code_grant', 'owner', 'icon')
+                 'bot_public', 'bot_require_code_grant', 'owner', 'icon',
+                 'summary', 'verify_key', 'team', 'guild_id', 'primary_sku_id',
+                 'slug', 'cover_image')
 
     def __init__(self, state, data):
         self._state = state
@@ -67,6 +89,16 @@ class AppInfo:
         self.bot_public = data['bot_public']
         self.bot_require_code_grant = data['bot_require_code_grant']
         self.owner = User(state=self._state, data=data['owner'])
+        team = data.get('team')
+        self.team = Team(state, team) if team else None
+        self.summary = data['summary']
+        self.verify_key = data['verify_key']
+        guild_id = data.get('guild_id')
+        self.guild_id = int(guild_id) if guild_id else None
+        primary_sku_id = data.get('primary_sku_id')
+        self.primary_sku_id = int(primary_sku_id) if primary_sku_id else None
+        self.slug = data.get('slug')
+        self.cover_image = data.get('cover_image')
 
     def __repr__(self):
         return '<{0.__class__.__name__} id={0.id} name={0.name!r} description={0.description!r} public={0.bot_public} ' \
@@ -76,3 +108,8 @@ class AppInfo:
     def icon_url(self):
         """:class:`.Asset`: Retrieves the application's icon asset."""
         return Asset._from_icon(self._state, self, 'app')
+
+    @property
+    def cover_image_url(self):
+        """:class:`.Asset`: Retrieves the cover image on a store embed."""
+        return Asset._from_cover_image(self._state, self)
--- a/discord/appinfo.py
+++ b/discord/appinfo.py
@@ -57,15 +57,15 @@ class AppInfo:
         A list of RPC origin URLs, if RPC is enabled.
     summary: Optional[:class:`str`]
         If this application is a game sold on Discord,
-        this field will be the summary field for the store page of its primary sku
+        this field will be the summary field for the store page of its primary SKU
     verify_key: Optional[:class:`str`]
         The base64 encoded key for the GameSDK's GetTicket
-    guild_id: Optional[:class:`int`]
+    guild: Optional[:class:`Guild`]
         If this application is a game sold on Discord,
         this field will be the guild to which it has been linked
     primary_sku_id: Optional[:class:`int`]
         If this application is a game sold on Discord,
-        his field will be the id of the "Game SKU" that is created, if exists
+        this field will be the id of the "Game SKU" that is created, if exists
     slug: Optional[:class:`str`]
         If this application is a game sold on Discord,
         this field will be the URL slug that links to the store page
--- a/discord/appinfo.py
+++ b/discord/appinfo.py
@@ -78,42 +78,6 @@ class AppInfo:
         team = data.get('team')
         self.team = Team(state, team) if team else None
 
-        self.game_info = GameInfo(state, data) if data['summary'] else None
-
-    def __repr__(self):
-        return '<{0.__class__.__name__} id={0.id} name={0.name!r} description={0.description!r} public={0.bot_public} ' \
-               'owner={0.owner!r}>'.format(self)
-
-    @property
-    def icon_url(self):
-        """:class:`.Asset`: Retrieves the application's icon asset."""
-        return Asset._from_icon(self._state, self, 'app')
-
-
-class GameInfo:
-
-    """Represents the game information if the
-    application is a game sold on Discord.
-
-    Attributes
-    -------------
-    summary: :class:`str`
-        The summary field for the store page of its primary SKU.
-    verify_key: :class:`str`
-        The base64 encoded key for the GameSDK's GetTicket.
-    guild: Optional[:class:`Guild`]
-        The guild to which it has been linked.
-    primary_sku_id: Optional[:class:`int`]
-        The id of the "Game SKU" that is created, if exists.
-    slug: Optional[:class:`str`]
-        The URL slug that links to the store page.
-    cover_image: Optional[:class:`str`]
-        The hash of the image on store embeds.
-    """
-
-    def __init__(self, state, data):
-        self._state = state
-
         self.summary = data['summary']
         self.verify_key = data['verify_key']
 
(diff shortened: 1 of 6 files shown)


discord/appinfo.py
```diff
@@ -52,9 +55,28 @@ class AppInfo:
         grant flow to join.
     rpc_origins: Optional[List[:class:`str`]]
         A list of RPC origin URLs, if RPC is enabled.
+    summary: Optional[:class:`str`]
+        If this application is a game sold on Discord,
+        this field will be the summary field for the store page of its primary sku
+    verify_key: Optional[:class:`str`]
+        The base64 encoded key for the GameSDK's GetTicket
+    guild_id: Optional[:class:`int`]
```
NCPlayz: You don't ever set `AppInfo.guild_id`.
```suggestion
    guild: Optional[:class:`Guild`]
```

discord/team.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Attributes
+    -------------
+    id: :class:`int`
+        The team ID.
+    name: :class:`str`
+        The team name
+    icon: Optional[:class:`str`]
+        The icon hash, if it exists.
+    owner_id: :class:`int`
```
NCPlayz: You don't ever set `Team.owner_id`.
```suggestion
    owner: :class:`User`
```

discord/team.py
```diff
@@ -0,0 +1,99 @@
...
+        return Asset._from_icon(self._state, self, 'team')
+
+
+class TeamMember:
+    """Represents a team member in a team.
+
+
+    Attributes
+    -------------
+    id: :class:`int`
```
NCPlayz: You don't ever set `TeamMember.id`.

discord/appinfo.py
```diff
@@ -52,9 +55,28 @@ class AppInfo:
...
+        If this application is a game sold on Discord,
+        this field will be the guild to which it has been linked
+    primary_sku_id: Optional[:class:`int`]
+        If this application is a game sold on Discord,
+        his field will be the id of the "Game SKU" that is created, if exists
```
BluePhoenixGame: I assume you meant "this"?

discord/appinfo.py
```diff
@@ -52,9 +55,28 @@ class AppInfo:
         grant flow to join.
     rpc_origins: Optional[List[:class:`str`]]
         A list of RPC origin URLs, if RPC is enabled.
+    summary: Optional[:class:`str`]
+        If this application is a game sold on Discord,
+        this field will be the summary field for the store page of its primary sku
```
BluePhoenixGame: SKU is an acronym.
```suggestion
        this field will be the summary field for the store page of its primary SKU
```

BluePhoenixGame:
This looks good. But isn't it easier to make a new class and subclass AppInfo instead of having a lot of "If this applications is a game sold on Discord" in several attributes?

discord/asset.py
```diff
@@ -94,6 +94,14 @@ def _from_icon(cls, state, object, path):
         url = 'https://cdn.discordapp.com/{0}-icons/{1.id}/{1.icon}.jpg'.format(path, object)
         return cls(state, url)
 
+    @classmethod
+    def _from_cover_image(cls, state, object):
```
Rapptz:
```suggestion
    def _from_cover_image(cls, state, obj):
```

discord/asset.py
```diff
@@ -94,6 +94,14 @@ def _from_icon(cls, state, object, path):
...
+        if object.cover_image is None:
```
Rapptz:
```suggestion
        if obj.cover_image is None:
```

discord/asset.py
```diff
@@ -94,6 +94,14 @@ def _from_icon(cls, state, object, path):
...
+            return cls(state)
+
+        url = 'https://cdn.discordapp.com/app-assets/{0.id}/store/{0.cover_image}.jpg'.format(object)
```
Rapptz:
```suggestion
        url = 'https://cdn.discordapp.com/app-assets/{0.id}/store/{0.cover_image}.jpg'.format(obj)
```

discord/appinfo.py
```diff
@@ -52,9 +55,13 @@ class AppInfo:
         grant flow to join.
     rpc_origins: Optional[List[:class:`str`]]
         A list of RPC origin URLs, if RPC is enabled.
+    game_info: Optional[:class:GameInfo]
```
Rapptz:
```suggestion
    game_info: Optional[:class:`GameInfo`]
```

discord/appinfo.py
```diff
@@ -68,6 +75,11 @@ def __init__(self, state, data):
         self.bot_require_code_grant = data['bot_require_code_grant']
         self.owner = User(state=self._state, data=data['owner'])
 
+        team = data.get('team')
+        self.team = Team(state, team) if team else None
+
+        self.game_info = GameInfo(state, data) if data['summary'] else None
```
Rapptz: I don't really get the rationale for splitting this into two separate classes. I suppose it saves some extraneous key checks but this class name seems peculiar because it's not an actual API object. Not to mention that this key check seems suspicious and I feel like it tosses information aside as well.

Looking at the API documentation (https://discordapp.com/developers/docs/topics/oauth2#get-current-application-information) the following keys are always present regardless of the key being there or not:

- `summary`
- `verify_key`

I don't really want to lose information on the `verify_key` just because the `summary` is empty. Even if you could argue that it's unnecessary for those not making a game, it's still some form of intentionally designed data loss which I don't want.

I suggest reverting the second class entirely and bringing the PR back to its mono-class state.

discord/ext/commands/bot.py
```diff
@@ -108,6 +108,7 @@ def __init__(self, command_prefix, help_command=_default, description=None, **op
         self._help_command = None
         self.description = inspect.cleandoc(description) if description else ''
         self.owner_id = options.get('owner_id')
+        self.team_member_ids = {}
```
Rapptz: Personally I feel as if using `owner_ids` as a key would have been better, along with making it possible to pass it at `__init__`. That way those who are not using teams for whatever reason can also just pass in some iterable or list and the library can cast it to a `set` without hitting any endpoint much like the current `owner_id` today.

This will probably require a check to make sure the following post-conditions are true:

- `owner_id` and `owner_ids` are not both set
- `owner_ids` can successfully be cast to an `set` or `int`.

Also requires changes to the stuff below.

discord/team.py
```diff
@@ -0,0 +1,97 @@
...
+        The team ID.
+    name: :class:`str`
+        The team name
+    icon: Optional[:class:`str`]
+        The icon hash, if it exists.
+    owner: :class:`User`
+        The team's owner ID
```
Rapptz:
```suggestion
        The team's owner ID.
```

discord/team.py
```diff
@@ -0,0 +1,97 @@
...
+
+    Attributes
+    -------------
+    id: :class:`int`
+        The team ID.
+    name: :class:`str`
+        The team name
+    icon: Optional[:class:`str`]
+        The icon hash, if it exists.
+    owner: :class:`User`
```
Rapptz:
```suggestion
    owner_id: :class:`int`
```

discord/team.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    __slots__ = ('_state', 'id', 'name', 'icon', 'owner_id', 'members')
+
+    def __init__(self, state, data):
+        self._state = state
+
+        self.id = int(data['id'])
+        self.name = data['name']
+        self.icon = data['icon']
+        self.owner = self._state.get_user(int(data['owner_user_id']))
```
Rapptz: `owner` is not in the `__slots__` so this will fail.

```suggestion
        self.owner_id = int(data['owner_user_id'])
```

To bring back the `Team.owner` functionality you'd have to do something like this:

```py
@property
def owner(self):
    """Optional[:class:`User`]: The team's owner, if available from the cache."""
    return self._state.get_user(self.owner_id)
```

discord/team.py
```diff
@@ -0,0 +1,97 @@
...
+    """Represents a team member in a team.
+
+
+    Attributes
+    -------------
+    team: :class:`team`
+        The team that the member is from.
+    membership_state: :class:`TeamMembershipState`
+        The membership state of the member (e.g. invited or accepted)
+    permissions: List[:class:`str`]
```
Rapptz: Since this is hard coded and can change in the future I suggest undocumenting it completely so its data type and the like can change in a future version without breaking changes.

discord/appinfo.py
fourjr: Since we are removing GameInfo, I'm resolving this comment.

discord/ext/commands/bot.py
fourjr: Why would `owner_ids` be able to be casted to an `int`?

Rapptz: It's a typo -- I meant `set` of `int` not "or".

discord/appinfo.py
```diff
@@ -52,9 +55,28 @@ class AppInfo:
         grant flow to join.
     rpc_origins: Optional[List[:class:`str`]]
         A list of RPC origin URLs, if RPC is enabled.
+    summary: Optional[:class:`str`]
```
Rapptz:
```suggestion
    summary: :class:`str`
```

This seems to always be there as a string.

discord/appinfo.py
```diff
@@ -52,9 +55,28 @@ class AppInfo:
...
+        If this application is a game sold on Discord,
+        this field will be the summary field for the store page of its primary SKU
+    verify_key: Optional[:class:`str`]
```
Rapptz:
```suggestion
    verify_key: :class:`str`
```

This field is also always there.

discord/appinfo.py
```diff
@@ -68,6 +90,20 @@ def __init__(self, state, data):
         self.bot_require_code_grant = data['bot_require_code_grant']
         self.owner = User(state=self._state, data=data['owner'])
 
+        team = data.get('team')
+        self.team = Team(state, team) if team else None
+
+        self.summary = data['summary']
+        self.verify_key = data['verify_key']
+
+        guild_id = data.get('guild_id')
```
Rapptz: There is a utility for this that casts to an integer and what not.
```suggestion
        self.guild_id = utils._get_as_snowflake(data, 'guild_id')
```

discord/appinfo.py
```diff
@@ -68,6 +90,20 @@ def __init__(self, state, data):
...
+        self.guild_id = int(guild_id) if guild_id else None
+
+        primary_sku_id = data.get('primary_sku_id')
```
Rapptz: Same as the above:

```suggestion
        self.primary_sku_id = utils._get_as_snowflake(data, 'primary_sku_id')
```

discord/appinfo.py
```diff
@@ -68,6 +90,20 @@ def __init__(self, state, data):
...
+        self.primary_sku_id = int(primary_sku_id) if primary_sku_id else None
```
Rapptz:
```suggestion
```

discord/appinfo.py
```diff
@@ -68,6 +90,20 @@ def __init__(self, state, data):
...
         self.owner = User(state=self._state, data=data['owner'])
 
+        team = data.get('team')
+        self.team = Team(state, team) if team else None
+
+        self.summary = data['summary']
+        self.verify_key = data['verify_key']
+
+        guild_id = data.get('guild_id')
+        self.guild_id = int(guild_id) if guild_id else None
```
Rapptz:
```suggestion
```

discord/appinfo.py
```diff
@@ -76,3 +112,14 @@ def __repr__(self):
...
+    @property
+    def cover_image_url(self):
+        """:class:`.Asset`: Retrieves the cover image on a store embed."""
+        return Asset._from_cover_image(self._state, self)
+
+    @property
+    def guild(self):
+        """Optional[:class:`Guild`]: If this application is a game sold on Discord,
+        this field will be the guild to which it has been linked"""
+        return self._state._get_guild(int(guild_id))
```
Rapptz:
```suggestion
        return self._state._get_guild(int(self.uild_id))
```

NameError there.

discord/ext/commands/bot.py
```diff
@@ -295,11 +299,18 @@ def whitelist(ctx):
             Whether the user is the owner.
         """
 
-        if self.owner_id is None:
+        if self.owner_id:
+            return user.id == self.owner_id
+        elif isinstance(self.owner_ids, (list, tuple, set)):
```
Rapptz: You seem to have misunderstood my previous review.

I said to do this *at* `__init__` time so the performance of *this* function is not impacted. It's an invariant when a *user passes a parameter* so it should be checked for correctness.

discord/ext/commands/bot.py
```diff
@@ -959,6 +970,9 @@ class Bot(BotBase, discord.Client):
         The ID that owns the bot. If this is not set and is then queried via
         :meth:`.is_owner` then it is fetched automatically using
         :meth:`~.Bot.application_info`.
+    owner_ids: Optional[:class:`set`]
+        The IDs that owns the bot. This is similar to `owner_id`.
+        If both `owner_id` and `owner_ids` are set, `owner_id` is preferred.
```
Rapptz: An exception should be raised instead.

PR #2243, opened 2019-06-23
Add O(log n) Member.has_role(id) method

Tobotimus:
### Summary

Adds a method for quickly checking if a user has a role with the given ID.

I did find #1762 and I agree that this is less pythonic than `role in member.roles`, but also note that it's a somewhat costly operation of creating a new sorted list of roles, then doing a linear search on it, as opposed to just an O(log n) binary search on an already sorted array of ints.

I've updated the `@has_role`, `@bot_has_role` et al. check decorators to use this new method, when a snowflake is provided. My (Red's) use case is somewhat similar to the `@has_any_roles` decorator, where a lot of commands will require checking whether a user has "mod" or "admin" role - this would be a welcome performance boost for those sorts of commands.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1352,7 +1352,7 @@ def has_role(item):
             raise NoPrivateMessage()
 
         if isinstance(item, int):
-            role = discord.utils.get(ctx.author.roles, id=item)
+            role = ctx.author.has_role(item) or None
         else:
             role = discord.utils.get(ctx.author.roles, name=item)
         if role is None:
@@ -1396,8 +1396,9 @@ def has_any_role(*items):
         if not isinstance(ctx.channel, discord.abc.GuildChannel):
             raise NoPrivateMessage()
 
-        getter = functools.partial(discord.utils.get, ctx.author.roles)
-        if any(getter(id=item) is not None if isinstance(item, int) else getter(name=item) is not None for item in items):
+        author = ctx.author
+        getter = functools.partial(discord.utils.get, author.roles)
+        if any(author.has_role(item) is True if isinstance(item, int) else getter(name=item) is not None for item in items):
             return True
         raise MissingAnyRole(items)
 
@@ -1424,7 +1425,7 @@ def bot_has_role(item):
 
         me = ch.guild.me
         if isinstance(item, int):
-            role = discord.utils.get(me.roles, id=item)
+            role = me.has_role(item) or None
         else:
             role = discord.utils.get(me.roles, name=item)
         if role is None:
@@ -1452,7 +1453,7 @@ def bot_has_any_role(*items):
 
         me = ch.guild.me
         getter = functools.partial(discord.utils.get, me.roles)
-        if any(getter(id=item) is not None if isinstance(item, int) else getter(name=item) is not None for item in items):
+        if any(me.has_role(item) is True if isinstance(item, int) else getter(name=item) is not None for item in items):
             return True
         raise BotMissingAnyRole(items)
     return check(predicate)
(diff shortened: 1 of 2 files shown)


Tobotimus:
Thanks. I forgot to push :grimacing:

discord/ext/commands/core.py
```diff
@@ -1396,8 +1396,9 @@ def predicate(ctx):
         if not isinstance(ctx.channel, discord.abc.GuildChannel):
             raise NoPrivateMessage()
 
-        getter = functools.partial(discord.utils.get, ctx.author.roles)
-        if any(getter(id=item) is not None if isinstance(item, int) else getter(name=item) is not None for item in items):
+        author = ctx.author
```
BluePhoenixGame: Basically useless.
```suggestion
```

discord/ext/commands/core.py
```diff
@@ -1396,8 +1396,9 @@ def predicate(ctx):
...
+        getter = functools.partial(discord.utils.get, author.roles)
```
BluePhoenixGame: And if you chose to remove that line just change this too.
```suggestion
        getter = functools.partial(discord.utils.get, ctx.author.roles)
```

discord/member.py
```diff
@@ -402,6 +402,22 @@ def permissions_in(self, channel):
         """
         return channel.permissions_for(self)
 
+    def has_role(self, role_id):
```
BluePhoenixGame: Maybe allow you to pass the name of the role or the role object itself too? That would allow you to use the function in all decorators as in some you still have to use `discord.utils.get()`

discord/member.py
```diff
@@ -402,6 +402,22 @@ def permissions_in(self, channel):
...
+        """Check if a user has a role with the given ID.
```
NCPlayz:
```suggestion
        """Check if the user has a role with the given ID.
```

discord/member.py
```diff
@@ -402,6 +402,22 @@ def permissions_in(self, channel):
...
+
+        Parameters
+        ----------
+        role_id: :class:`int`
+            The ID for the role.
+
+        Returns
+        -------
+        bool
```
NCPlayz:
```suggestion
        :class:`bool`
```

discord/member.py
```diff
@@ -402,6 +402,22 @@ def permissions_in(self, channel):
...
+            :code:`True` if the user has a role with the
```
NCPlayz:
```suggestion
            ``True`` if the user has a role with the given ID.
```

discord/member.py
```diff
@@ -402,6 +402,22 @@ def permissions_in(self, channel):
...
+            given ID.
```
NCPlayz:
```suggestion
```

discord/member.py
BluePhoenixGame: Seems like we all missed this. It's a member.
```suggestion
        """Check if a member has a role with the given ID.
```

discord/member.py
BluePhoenixGame: And again, it's a member.
```suggestion
            :code:`True` if the member has a role with the
```

Tobotimus:
I'm gonna close this PR, I've had a discussion with Danny and I think his reasons are well and truly justified :smiley:

In summary, he said if he was considering anything it would be `Member.get_role(id)`, which would serve the same purpose in the same time-complexity on average, but also be more useful. But he's not sure if he wants to keep the underlying data structure as a sorted `SnowflakeList`, and that makes it difficult for him to settle on an interface at this time.

Issue #2211, opened 2019-06-10
Broaden scope of ExtensionFailed


Travus:
### The Problem

Currently, when you load an extension with bot.load_extension, the ExtensionFailed exception is given if an error occurs inside the setup entrypoint. When something outside of the setup entrypoint encounters an error, like some code in the global scope, the error that is encountered is raised instead.

Additionally, if an ImportError is raised during the loading process of an extension, the ExtensionNotFound error is given by bot.load_extension, despite the fact the extension was in fact found and only ran into an error. I'm not sure if this can actually be fixed since I believe the absence of a module is also signaled by an ImportError so both would look the same to the library, but I can't be sure.

Edit: I did some more research into this after waking up, and as suspected is the current way the library detects ExtensionNotFound just going by importlib ImportErrors so it currently cannot tell the difference.

### The Ideal Solution

I would like for the ExtensionFailed exception to be called even if the extension encounters an error outside of the setup entrypoint during loading, and if at all possible even if the error encountered is an ImportError. This way you can just listen for the ExtensionFailed exception in order to report back to users that the extension they tried to load failed to execute properly. (And you don't wrongly tell them it doesn't exist because the extension has incorrect imports.)

### The Current Solution

Add a catchall after loading exceptions that reports back to the user when any undefined exception is raised by bot.load_extension. And check if the file is present manually instead of using ExceptionNotFound.

bmintz:
Actually it is possible to detect if a module is importable without executing it, according to the [importlib docs](https://docs.python.org/3.7/library/importlib.html#checking-if-a-module-can-be-imported)

bmintz:
I made a PR which does the opposite of this issue: #2244

PR #2244, opened 2019-06-24
bot.load_extension: don't raise ExtensionNotFound for ImportErrors inside the module

bmintz:
Now loading an extension that _contains_ a failed import will fail with ExtensionFailed, rather than ExtensionNotFound.

Open question: I'm not sure if that `unload()` local function inside Bot#_load_from_module_spec should be there, or if its code should be duplicated.

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -570,31 +570,25 @@ class BotBase(GroupMixin):
 
     def _load_from_module_spec(self, spec, key):
         # precondition: key not in self.__extensions
-        def unload(e):
-            self._remove_module_references(lib.__name__)
-            self._call_module_finalizers(lib, key)
-            raise errors.ExtensionFailed(key, e) from e
-
         lib = importlib.util.module_from_spec(spec)
         try:
             spec.loader.exec_module(lib)
         except Exception as e:
-            unload(e)
+            raise errors.ExtensionFailed(key, e) from e
 
         try:
             setup = getattr(lib, 'setup')
         except AttributeError:
             raise errors.NoEntryPointError(key)
 
-        sys.modules[key] = lib
-
         try:
             setup(self)
         except Exception as e:
-            del sys.modules[key]
-            unload(e)
+            self._remove_module_references(lib.__name__)
+            self._call_module_finalizers(lib, key)
+            raised errors.ExtensionFailed(key, e) from e
         else:
-            self.__extensions[key] = lib
+            sys.modules[key] = self.__extensions[key] = lib
 
     def load_extension(self, name):
         """Loads an extension.
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -27,7 +27,7 @@ DEALINGS IN THE SOFTWARE.
 import asyncio
 import collections
 import inspect
-import importlib
+import importlib.util
 import sys
 import traceback
 import re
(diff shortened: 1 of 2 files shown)


Travus:
I'm in support of this. ExtensionFailed more closely describes what the error is.
This also addresses part of #2211.

bmintz:
Regarding ExtensionNotFound.original, should
 I set it to None, or self, or something like that, for backwards compatibility?

discord/ext/commands/bot.py
```diff
@@ -568,20 +568,31 @@ def _call_module_finalizers(self, lib, key):
...
+        def unload(e):
+            self._remove_module_references(lib.__name__)
+            self._call_module_finalizers(lib, key)
+            raise errors.ExtensionFailed(key, e) from e
+
+        lib = importlib.util.module_from_spec(spec)
+        try:
+            spec.loader.exec_module(lib)
+        except Exception as e:
+            unload(e)
```
Rapptz: What is this even doing here?

If the import failed then the `lib` reference is probably broken.

discord/ext/commands/bot.py
```diff
@@ -568,20 +568,31 @@ def _call_module_finalizers(self, lib, key):
...
+
         try:
             setup = getattr(lib, 'setup')
         except AttributeError:
-            del sys.modules[key]
             raise errors.NoEntryPointError(key)
 
+        sys.modules[key] = lib
```
Rapptz: Why is this being assigned here if you're going to delete it later anyway? Just add it when everything is all done.

discord/ext/commands/bot.py
```diff
@@ -568,20 +568,31 @@ def _call_module_finalizers(self, lib, key):
...
+
         try:
             setup(self)
         except Exception as e:
-            self._remove_module_references(lib.__name__)
-            self._call_module_finalizers(lib, key)
-            raise errors.ExtensionFailed(key, e) from e
+            del sys.modules[key]
+            unload(e)
```
Rapptz: Unroll this function since it's unnecessary up there.

discord/ext/commands/errors.py
```diff
@@ -521,19 +521,15 @@ def __init__(self, name, original):
...
     Attributes
     -----------
     name: :class:`str`
         The extension that had the error.
-    original: :exc:`ImportError`
-        The original exception that was raised. You can also get this via
-        the ``__cause__`` attribute.
     """
-    def __init__(self, name, original):
-        self.original = original
```
Rapptz: Since this is technically a breaking change the only way to get away with it is to explicitly set it to `None` and updating the documentation to say it's hard-coded to `None` for backwards compatibility (along with a version changed in the top level class docstring I suppose).

The documented return type was always wrong anyway, given that `Exception.__cause__` is `Optional[Exception]` it doesn't seem entirely out of line that this could also be set to explicitly `None`.

discord/ext/commands/errors.py
```diff
@@ -521,19 +521,18 @@ def __init__(self, name, original):
...
         The extension that had the error.
-    original: :exc:`ImportError`
-        The original exception that was raised. You can also get this via
-        the ``__cause__`` attribute.
+    original: :class:`NoneType`
+        Always ``None`` for backwards compatibility.
     """
-    def __init__(self, name, original):
-        self.original = original
+    def __init__(self, name):
```
Rapptz:
```suggestion
    def __init__(self, name, original=None):
```

I've given this some thought and I think to absolutely minimize breaking changes even further the parameter count should stay the same with a default and then subsequently ignored.

bmintz:
This commit caused a regression, discovered by @Gorialis:
[importsfucked.zip](https://github.com/Rapptz/discord.py/files/3366264/importsfucked.zip).
The issue has to do with subpackages causing an AttributeError despite existing in sys.modules. Cause unknown as of yet.

bmintz:
Regression fixed in 6f71552c508c61d9b6bf024fc259063ad056b7c4.

PR #2246, opened 2019-06-26
Removed Returns from all decorator docstrings

OrangutanGaming:
### Summary

This removes all `Returns` in any docstring for a decorator. This is because they specify what the function returns, not the decorator, as the decorator returns a function. This also means that linters e.g. PyCharm's can work out the correct type of the return instead of receiving the wrong one, leading to errors when it thinks the decorator returns e.g. a `Loop` instance instead of a function.
It also makes it more consistent as `tasks.loop` is the only decorator with a `Returns` in the docstring. For example, `commands.command`:
https://github.com/Rapptz/discord.py/blob/0cf38241d00df3c89c74889a2a73a4c33bb649f9/discord/ext/commands/core.py#L1205-L1234

For example, if you use `tasks.loop()`, the linter will warn saying:
```
'Loop' object is not callable
Inspection info: This inspection highlights attempts to call objects which are not callable, like, for example, tuples.
```
as this is the return type described in the docstring instead of a callable that returns `Loop`.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/tasks/__init__.py
+++ b/discord/ext/tasks/__init__.py
@@ -348,7 +348,7 @@ class Loop:
 
 def loop(*, seconds=0, minutes=0, hours=0, count=None, reconnect=True, loop=None):
     """A decorator that schedules a task in the background for you with
-    optional reconnect logic.
+    optional reconnect logic. The decorator returns a :class:`Loop`.
 
     Parameters
     ------------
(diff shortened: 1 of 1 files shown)


OrangutanGaming:
Ah, I forgot to include it in the main description.

PR #2247, opened 2019-06-26
Added ignore_others argument

MinerChAI:
### Summary

It adds `ignore_others` parameter, that insicates if bot must ignore bots and in case of selfbot other people

### Checklist

- [ ] This PR fixes an issue.


Rapptz:
I'm not sure what the point or motivation of this PR is. `skip_check` does two things:

1. It ignores yourself if you're a bot
2. It ignores others if you're a selfbot

This PR for some reason makes this optional (something that I do not want) with a rather confusing name. This seems rather niche. You can just override `process_commands` yourself. I don't even see the use case of listening to yourself.

PR #2254, opened 2019-06-30
Add User.set_note()

apple502j:
### Summary
This adds `await user.set_note("foo bar")`, a function which sets a "note" (shown in profile).

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/user.py
+++ b/discord/user.py
@@ -861,7 +861,7 @@ class User(BaseUser, discord.abc.Messageable):
 
         Parameters
         -----------
-        note: :class:`str`
+        note: Optional[:class:`str`]
             The note to set. Pass ``None`` to remove the note.
 
         Raises
--- a/discord/user.py
+++ b/discord/user.py
@@ -872,5 +872,5 @@ class User(BaseUser, discord.abc.Messageable):
             Setting the note failed.
 
         """
-        note = str(note) or ""
+        note = "" if note is None else str(note)
         await self._state.http.set_note(self.id, note)
--- a/discord/user.py
+++ b/discord/user.py
@@ -849,3 +849,28 @@ class User(BaseUser, discord.abc.Messageable):
                        mutual_guilds=mutual_guilds,
                        user=self,
                        connected_accounts=data['connected_accounts'])
+
+    async def set_note(self, note):
+        """|coro|
+
+        Sets the note.
+
+        .. note::
+
+            This only applies to non-bot accounts.
+
+        Parameters
+        -----------
+        note: :class:`str`
+            The note to set. Pass ``None`` to remove the note.
+
+        Raises
+        -------
+        Forbidden
+            Not allowed to set note.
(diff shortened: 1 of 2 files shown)


CapnS:
I don't know if it's worth the effort because this is only for userbots, but could you also  add User.fetch_note()?

Soheab:
Ew selfbots.. but fetch note would cool? https://github.com/Rapptz/discord.py/pull/2254#issuecomment-507030695

EEKIM10:
i dont really see a point in this. who even uses notes anyway?

BluePhoenixGame:
I feel like discord.py should support the whole discord api. And that includes the notes!

Phxntxm:
Personally...I actually *don't* want this included, I have already seen times where people think they can try to use notes as some kind of internal-to-discord database, which is dumb. I understand the reasoning behind leaving the already-included selfbot only, but I don't like adding another that I see people trying to abuse.

Maybe I have the abnormal opinion here, but just putting it out there.

discord/user.py
```diff
@@ -849,3 +849,28 @@ def transform(d):
...
+
+        Sets the note.
+
+        .. note::
+
+            This only applies to non-bot accounts.
+
+        Parameters
+        -----------
+        note: :class:`str`
```
Rapptz:
```suggestion
        note: Optional[:class:`str`]
```

discord/user.py
```diff
@@ -849,3 +849,28 @@ def transform(d):
...
+
+        Raises
+        -------
+        Forbidden
+            Not allowed to set note.
+        HTTPException
+            Setting the note failed.
+
+        """
+        note = str(note) or ""
```
Rapptz:
```suggestion
        note = "" if note is None else str(note)
```

discord/user.py
```diff
@@ -849,3 +849,28 @@ def transform(d):
                        mutual_guilds=mutual_guilds,
                        user=self,
                        connected_accounts=data['connected_accounts'])
+
+    async def set_note(self, note):
```
Rapptz: Honestly not sure I like this name.

Let me disregard the fact that this is a user-bot only feature for a minute.

In this library whenever you want to edit something about the model you don't use any `set_` methods, you have an `edit` method with a kwarg-only interface. Now, `User.edit` doesn't make much sense to add for only this thing but it's what would be the most consistent. The problem is that the concept of editing a `User` instance and adding an `edit` to it is that it breaks expectations when using `Member.edit`. Why couldn't you edit the note on a Member with the same interface?

These design decisions are why personally I'm not sure this is a good fit. Maybe if the name was something else. Maybe it's unimportant.

bmintz:
I would rather have User.note which is None on bot accounts than User.fetch_note, unless this information is not dispatched along with the other fields of User.

AtieP:
Why is this still not approved?

PR #2258, opened 2019-07-03
Add FFmpegOpusAudio and other improvements

### LINKED ISSUE #1846: Add a specific AudioSource for opus
Real-time encoding is a very heavy operation, and encoding before into opus is a good idea, but there isn't a specific AudioSource for opus, only through ffmpeg at the moment so it's useless.

imayhaveborkedit:
### Summary

This PR adds several voice related improvements, the main one being the addition of the `FFmpegOpusAudio` AudioSource.  This source skips the opus encoding step in the library by having ffmpeg encode to opus and parsing the output.  This way libopus is not a hard library requirement when using voice.  Other improvements include:
- New `FFmpegAudio` base class to assist with making custom ffmpeg based audio sources.
- Using more accurate and monotonic `time.perf_counter()` in AudioPlayer instead of `time.time()`.
- Lazy opus loading.  The library no longer loads libopus on import, now only on `opus.Encoder` creation or manually.  Additionally, `VoiceClient` now creates its encoder object in `play()` instead of in `__init__`.
  - This also fixes the issue (on Windows at least) where importing the library in a repl or having a bot running loads the libopus binary, placing a file lock it it even when not actively using voice, preventing updates to the library.

### Checklist

- [X] This PR fixes an issue. (Fixes #1846, more or less.)

### DIFF
--- a/discord/player.py
+++ b/discord/player.py
@@ -148,8 +148,8 @@ class FFmpegAudio(AudioSource):
 
         try:
             proc.kill()
-        except:
-            pass
+        except Exception:
+            log.exception("Ignoring error attempting to kill ffmpeg process %s", proc.pid)
 
         if proc.poll() is None:
             log.info('ffmpeg process %s has not terminated. Waiting to terminate...', proc.pid)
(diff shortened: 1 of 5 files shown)


thegergo02:
On the discord server I already heard it improves a lot on audio, and I would love to use it so I'm waiting for the merge!

discord/player.py
```diff
@@ -159,36 +211,246 @@ def __init__(self, source, *, executable='ffmpeg', pipe=False, stderr=None, befo
...
+        codec, bitrate = await cls.probe(source, method=method, executable=executable)
+        return cls(source, bitrate=bitrate, codec=codec, **kwargs)
+
+    @classmethod
+    async def probe(cls, source, *, method=None, executable=None):
+        """Probes the input source for bitrate and codec information.
+
+        Parameters
+        ------------
+        source:
```
Rapptz:
```suggestion
        source
```

If no type hint is given then remove the colons.

discord/player.py
```diff
@@ -159,36 +211,246 @@ def __init__(self, source, *, executable='ffmpeg', pipe=False, stderr=None, befo
...
+            Identical to the ``source`` parameter for :class:`FFmpegOpusAudio`.
+        method:
```
Rapptz:
```suggestion
        method
```

discord/player.py
```diff
@@ -159,36 +211,246 @@ def __init__(self, source, *, executable='ffmpeg', pipe=False, stderr=None, befo
...
+            Identical to the ``method`` parameter for :meth:`FFmpegOpusAudio.from_probe`.
+        executable:
```
Rapptz:
```suggestion
        executable
```

discord/player.py
```diff
@@ -159,36 +211,246 @@ def __init__(self, source, *, executable='ffmpeg', pipe=False, stderr=None, befo
...
-            log.info('ffmpeg process %s successfully terminated with return code of %s.', proc.pid, proc.returncode)
+            raise TypeError("Expected str or callable for parameter 'probe', " \
+                            "not '{0.__class__.__name__}'" .format(method))
 
-        self._process = None
+        codec = bitrate = None
+        loop = asyncio.get_event_loop()
+        try:
+            codec, bitrate = await loop.run_in_executor(None, lambda: probefunc(source, executable))
+        except:
```
Rapptz: Probably want to avoid using bare except here, though I know it's done in other parts of the library it shouldn't be done in general.

imayhaveborkedit: The reason here is that it doesn't matter why it failed, just that it failed and we didn't get the data we were looking for.  The error was logged and that's as much as we can do.  Even if there were specific exceptions to catch I don't know what they would be and why they would need specific handling.

PR #2266, opened 2019-07-11
Fix fetch_members missing async

Geroyuni:
### Summary
fetch_members coroutine is missing async.

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -1157,7 +1157,7 @@ class Guild(Hashable):
 
         return [convert(d) for d in data]
 
-    def fetch_members(self, *, limit=1, after=None):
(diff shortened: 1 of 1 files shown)


Rapptz:
This function returns an async iterator. Marking it with `async` does not do what you expect.

bmintz:
No it's not. Functions that return async iterators should not be async; the asynchronous code is in MemberIterator#\_\_anext\_\_. With this PR you would have to do `async for member in await guild.fetch_members()`. If that's what you wanted then I suggest you update the docs too.

Geroyuni:
Understood, sorry.

PR #2267, opened 2019-07-11
Fix BaseUser.permissions_in

tedle:
### Summary

User objects don't have roles, so they need to be converted to Member objects first.

Example:
```
import discord

client = discord.Client()

@client.event
async def on_ready():
    for guild in client.guilds:
        print(f"{guild.name}")
        for channel in guild.channels:
            permissions = client.user.permissions_in(channel)
            print(f"\t{channel.name}:{permissions}")
    await client.logout()

client.run("token")
```
Results in:
```
Ignoring exception in on_ready
  File "/discord.py/discord/client.py", line 296, in _run_event
    await coro(*args, **kwargs)
  File "test.py", line 10, in on_ready
    permissions = client.user.permissions_in(channel)
  File "/discord.py/discord/user.py", line 223, in permissions_in
    return channel.permissions_for(self)
  File "/discord.py/discord/abc.py", line 426, in permissions_for
    roles = member.roles
AttributeError: 'ClientUser' object has no attribute 'roles'
```

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/user.py
+++ b/discord/user.py
@@ -220,7 +220,7 @@ class BaseUser(_BaseUser):
         channel: :class:`abc.GuildChannel`
             The channel to check your permissions for.
         """
-        return channel.permissions_for(self)
+        return channel.permissions_for(channel.guild.get_member(self.id))
 
     @property
     def created_at(self):


BluePhoenixGame:
This doesn't make sense, a User object represents a discord user. Not a guild member, and so it shouldn't be linked to any channel. If you want the bot as a member you can use attribute `.me` of guild. Or it might be of channel.

IngeniousCoder:
To refer to client.user in the current guild, use guild.me

tedle:
> This doesn't make sense, a User object represents a discord user.

This is just a fix for broken code that already exists in the User class. It's not adding anything.

> To refer to client.user in the current guild, use guild.me

I thought about this but wasn't sure if all instances of User were guaranteed to be the client, since ClientUser and User are 2 separate classes and this fix applies to both of them.

IngeniousCoder:
> all instances of User were guaranteed to be the client

No. A user object can represent any discord user.

Rapptz:
This is working by design. It's a forwarding function. Your code is just a bandaid towards a misunderstanding on the function. If in a DM context for example the code equally breaks.

If you want to fix *your* code then you should probably use `guild.me.permissions_in(channel)`.

tedle:
@Rapptz In that case shouldn't `permissions_in` be removed from BaseUser and defined exclusively in Member? Right now ClientUser and User are both have a documented function that cannot work in any circumstance since it tries to access a `.role` attribute that understandably does not exist.

Rapptz:
https://discordpy.readthedocs.io/en/latest/api.html#discord.GroupChannel.permissions_for

tedle:
@Rapptz Ah, okay thank you for clearing up my misunderstanding. I didn't know the API well enough to realize that permissions existed outside of guild channels. The docs for the ClientUser and User implementations of `permissions_in` both claim it's an alias of `abc.GuildChannel` which only takes Members, but I guess theres no `abc.PrivateChannel` implementation to point to since `DMChannel.permissions_for` and `GroupChannel.permissions_for` have different signatures.

PR #2270, opened 2019-07-16
Add CustomStatus class and document it

apple502j:
### Summary
Add CustomStatus class that represents custom status feature. This is a slimmed down version of Activity.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/activity.py
+++ b/discord/activity.py
@@ -35,6 +35,7 @@ __all__ = (
     'Streaming',
     'Game',
     'Spotify',
+    'CustomStatus'
 )
 
 """If curious, this is the current schema for an activity.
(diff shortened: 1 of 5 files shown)


Rapptz:
It's too early for this feature since it's subject to change.

Also, there is the fact that there's the guild/channel specific thing that is bound to change.

khazhyk:
could we merge the enum so that stuff like Activity.to_dict() isn't broken

```
  File "/var/lib/spoobeta/dango.py/venv/lib/python3.5/site-packages/discord/activity.py", line 166, in to_dict
    ret['type'] = self.type.value
AttributeError: 'int' object has no attribute 'value'
```

Rapptz:
This was a silent change in the 1.2.3 release: https://github.com/Rapptz/discord.py/commit/1e982e0042ae6c50e59b9b77fb77aa6f361ef0e9

Update your library.

OrangutanGaming:
Custom status is now released
Also, will this be added to self bots or are they feature locked?

discord/activity.py
```diff
@@ -582,6 +583,77 @@ def party_id(self):
...
+
+    @property
+    def status(self):
+        """:class:`str` : The custom status.
+
+        There is an alias for this named :meth:`state`."""
+        return self.state
+
+    def __eq__(self, other):
+        return (isinstance(other, CustomStatus) and other._state === self._state)
```
algmyr: === is not valid python

apple502j:
Closing this because Wumpus asked me to do

PR #2281, opened 2019-07-19
[commands] Update channel converter regexes.

mikeshardmind:
### Summary

Updates channel converter regex.

See: https://github.com/discordapp/discord-api-docs/issues/1029

### Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -226,7 +226,7 @@ class TextChannelConverter(IDConverter):
     async def convert(self, ctx, argument):
         bot = ctx.bot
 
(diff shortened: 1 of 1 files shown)


TheRockettek:
We arent really sure if this type of format is fully confirmed yet :/

mikeshardmind:
Night's comment in that thread suggests that even if it isn't, that there's no guarantee they won't do similar things unannounced in the future as well. While I can understand not wanting to support it, converters are currently broken, and this does not break prior existing converter use.

Rapptz:
This change is being worked on locally because the regex change goes further than just converters. I planned on committing it before I went to work but time ran out.

PR #2286, opened 2019-07-23
[docs] add new FAQ entries

iomintz:
## Summary

- How to DM
- Lack of search and its implications
  - This one raises a concern, namely that most people asking about the channel search API are asking bout overall message counts. Instead of not suggesting `Messageable.history()`, I opted to answer the question at face value and include a warning for those looking for this particular use case. In the warning, should I mention that EUD should be encrypted?
- Where to find examples
- Getting the ID of a sent message
  - Should I include a specific example of something to do with the ID? I couldn't think of anything…

(There is also an Easter egg in this PR, can you find it?)

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -85,12 +85,12 @@ in the repository.
 Can I search messages in channels?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Unfortunately, Discord does not provide a channel search API for bots. There are some workarounds:
+Unfortunately, Discord does not expose a channel search API for bots. There are some workarounds:
 
 - Use :meth:`abc.Messageable.history` and check the :attr:`~Message.content` of each message.
-  This may be slow depending on the number of messages checked.
-- Listen for new messages that match your condition in :func:`on_message`.
+  This may be slow depending on the number of messages checked, as more messages require more HTTP calls.
 - Iterate through :attr:`Client.cached_messages`.
+  This has a limited number of messages and it is likely the message you want will not be cached here.
 
 .. warning::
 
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -76,6 +76,28 @@ General
 
 General questions regarding library usage belong here.
 
+Are there any usage examples?
```
Rapptz:
```suggestion
Where can I find examples?
```

docs/faq.rst
```diff
@@ -76,6 +76,28 @@ General
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Example code can be found in the `examples/ <https://github.com/Rapptz/discord.py/tree/master/examples>`_ folder
+in the repository.
+
+Can I search messages in channels?
```
Rapptz: I don't like this question's title nor its content. It needs a rewrite.

docs/faq.rst
```diff
@@ -104,8 +126,27 @@ You must fetch the channel directly and then call the appropriate method. Exampl
     channel = client.get_channel(12324234183172)
     await channel.send('hello')
 
+How do I send a DM?
+~~~~~~~~~~~~~~~~~~~
+
+Get the :class:`User` object and call :meth:`User.send`. For example: ::
```
Rapptz:
```suggestion
Get the :class:`User` or :class:`Member` object and call :meth:`abc.Messageable.send` on it. For example: ::
```

docs/faq.rst
```diff
@@ -104,8 +126,27 @@ You must fetch the channel directly and then call the appropriate method. Exampl
...
+
+	user = client.get_user(381870129706958858)
+	await user.send('👀')
```
Rapptz: Also needs an example showing `message.author.send` since that's 99.99999% the more common use case.

docs/faq.rst
```diff
@@ -104,8 +126,27 @@ You must fetch the channel directly and then call the appropriate method. Exampl
...
+
+If the user is not cached (:meth:`Client.get_user` returns ``None``), you can use :meth:`Client.fetch_user` instead.
```
Rapptz: This entire sentence needs to be removed. If it's not in the cache then the DM will just fail.
```suggestion
```

docs/faq.rst
```diff
@@ -104,8 +126,27 @@ You must fetch the channel directly and then call the appropriate method. Exampl
...
+
+If you are responding to an event, such as :func:`on_message`, you already have the :class:`User` object via :attr:`Message.author`..
```
Rapptz:
```suggestion
```

This sentence is better off being shown in the code sample rather than written text.

docs/faq.rst
```diff
@@ -104,8 +126,27 @@ You must fetch the channel directly and then call the appropriate method. Exampl
...
+Use :meth:`User.send`.
+
+How do I get the ID of a sent message?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:meth:`abc.Messageable.send` returns the :class:`Message` that was sent.
+The ID of a message can be accessed via :attr:`Message.id`.
```
Rapptz: As silly as this is, this needs a code example again. Remember the target audience.

docs/faq.rst
iomintz: I ended up deleting that entry because I'm not sure how best to go about it.

PR #2300, opened 2019-08-07
Add permissions_synced to abc.GuildChannel

CapnS:
### Summary

This allows users to check whether or not the permissions for a Guild Channel are synced with the permissions for its category. Discord automatically syncs the permissions when the overwrites are equal so just checking if the two overwrites are equal will determine if they are synced. 

I feel as if the name isn't perfect, but I couldn't think of something that worked better. If anyone has any suggestions you can just leave them in the comments.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -391,7 +391,10 @@ class GuildChannel:
         If there is no category then this is ``False``.
         """
         category = self.guild.get_channel(self.category_id)
-        return self._overwrites == category._overwrites
+        if category is not None:
+            return self._overwrites == category._overwrites
+        else:
+            return False
 
     def permissions_for(self, member):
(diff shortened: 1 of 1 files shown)


discord/abc.py
```diff
@@ -383,6 +383,19 @@ def category(self):
...
 
+    @property
+    def permissions_synced(self):
+        """:class:`bool`: Whether or not the permissions for this channel are synced with the
+        category it belongs to.
+
+        If there is no category then this is ``False``.
+        """
+        category = self.guild.get_channel(self.category_id)
+        if category is not None:
```
Rapptz: This entire function can be summed up as 
```py
return category and category._overwrites == self._overwrites
```

PR #2314, opened 2019-08-17
Add new message type, CHANNEL_FOLLOW_ADD

diceroll123:
### Summary

Adds new MessageType enum for following an announcement channel.

### Checklist

- [ ] This PR fixes an issue.


docs/api.rst
```diff
@@ -803,6 +803,9 @@ of :class:`enum.Enum`.
 
         The system message denoting that a member has "nitro boosted" a guild
         and it achieved level 3.
+    .. attribute:: channel_follow_add
+
+        The system message denoting that an announcement channel has been followed.
```
apple502j: versionadded!

diceroll123: Looks like maybe `versionchanged` should be used instead? Under the MessageType class, similar to https://discordpy.readthedocs.io/en/latest/api.html#discord.RawMessageUpdateEvent

BluePhoenixGame: No, that's used when something was changed. So if this gets merged and discord adds another flag. And you make another PR then you should use versionchanged saying that you added the new flag. This time you added something new.

diceroll123: I see reasons for both, but this is just following suit, RawReactionActionEvent existed before 1.3.0 yet it has `Changed in version 1.3.0: The event_type attribute was added.` under it.

BluePhoenixGame: Oh, yes. This should be a versionchanged. Since you didn't add a new object/function, you changed one. My message above i seem to have mixed up this PR with your other.

apple502j:
And Message.system_content missing for that.

PR #2326, opened 2019-08-27
Fix _get_message returning the message cache.

PendragonLore:
### Summary

This PR fixes ``ConnectionState._get_message`` returning the message cache, instead of a message or ``None``.
This happens when the cache is empty.

### Checklist

- [x] This PR fixes an issue.


PendragonLore:
Actually, this seems to happen whenever ``max_messages`` is not ``None`` and a message, which is not in the cache, is edited.

PR #2355, opened 2019-09-22
Add TextChannel.jump_url, DMChannel.jump_url and GroupChannel.jump_url

PgBiel:
### Summary

Adds TextChannel.jump_url, DMChannel.jump_url and GroupChannel.jump_url, links that let the user jump to those channels when clicked. (With help from @ShineyDev)

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -141,7 +141,7 @@ class TextChannel(discord.abc.Messageable, discord.abc.GuildChannel, Hashable):
     @property
     def jump_url(self):
         """:class:`str`: Returns a URL that allows the client to jump to this channel."""
-        return 'https://discordapp.com/channels/{0.guild.id}/{0.id}'.format(self)
+        return "https://discordapp.com/channels/{0.guild.id}/{0.id}".format(self)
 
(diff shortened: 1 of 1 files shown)


Rapptz:
I don't get the point of this one. You know you can mention a channel right? It does the same thing in less characters and isn't an unsupported hack.

PgBiel:
> I don't get the point of this one. You know you can mention a channel right? It does the same thing in less characters and isn't an unsupported hack.

You can mention a channel, but you cannot link to it without the jump url. It's useful for:

1. Adding links to embeds such as on embed author url or by masking a link in the embed body, such that, when clicked, the user can navigate to the channel quickly;
2. Linking to channels in websites and other locations;
3. Supporting more of the API.

That's why I think this should be implemented.

Rapptz:
1. Jump URLs don't even consistently work when masked in the desktop client
2. If it's in a website then it doesn't have much to do with this wrapper.
3. This isn't part of the API. 

I'm not really convinced.

PR #2356, opened 2019-09-25
Support Attachment in send()

apple502j:
### Summary
Supports `Attachment` in `send()` 's file and files arguments.

### Checklist
- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -25,6 +25,7 @@ DEALINGS IN THE SOFTWARE.
 """
 
 import abc
+import io
 import copy
 import asyncio
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


Yuhanun:
There's an empty elif:
https://github.com/Rapptz/discord.py/pull/2356/commits/b1307fddc6a3a2992ce8f3b1e74cd5ca8e55ea3e#diff-8359e9b5c3ab56e3edee2a02209a7041R845
```py

        elif files is not None:
            if len(files) > 10:
                raise InvalidArgument('files parameter must be a list of up to 10 elements')
            elif:
                processed_file```

Should update it to an `else:`

Rapptz:
Personally not a fan of supporting this, especially like this. It's too niche and this code is too complicated when it can be done user-end a bit more simpler.

PR #2362, opened 2019-10-04
Add discord.Emoji.use()

Denayder:
### Summary

This pull request adds a `use()` method for the discord.Emoji class.

It makes it easier to send an emoji, so you don't need to manually format everything if you want to send an emoji in a chat from e.g. a reaction event.

One thing I wasn't sure of was emojis that has `require_colons` set to false. I wasn't sure how that would look, but I was told by a friend that bots can't use them anyway.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/emoji.py
+++ b/discord/emoji.py
@@ -249,6 +249,10 @@ class Emoji:
         """:class:`Guild`: The guild this emoji belongs to."""
         return self._state._get_guild(self.guild_id)
 
+    def use(self):
+        """:class:`str`: Returns the emoji in a usable state."""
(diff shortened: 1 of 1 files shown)


Rapptz:
Hello, thanks for the PR.

However, upon making this PR did you miss the `__str__` method which actually does this properly? Along side with the documentation that documents the behaviour. I see no reason to have two methods that do the same thing here.

Denayder:
> Hello, thanks for the PR.
> 
> However, upon making this PR did you miss the `__str__` method which actually does this properly? Along side with the documentation that documents the behaviour. I see no reason to have two methods that do the same thing here.

🤦‍♀ My bad, I completely missed that somehow. Apologies for the inconvenience.

PR #2367, opened 2019-10-08
Implement `TextChannel.follow()`

NCPlayz:
### Summary

This allows the client to follow a news (announcement) channel.

```py
>>> await my_news_channel.follow(destination=channel)
<Webhook id=:id:> # no token provided by discord
```

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -455,11 +455,11 @@ class TextChannel(discord.abc.Messageable, discord.abc.GuildChannel, Hashable):
         data = await self._state.http.create_webhook(self.id, name=str(name), avatar=avatar, reason=reason)
         return Webhook.from_state(data, state=self._state)
 
-    async def follow(self, channel):
+    async def follow(self, *, destination):
         """
         Follows a channel using a webhook.
 
-        You can only follow news channels.
+        Only news channels can be followed.
 
         .. note::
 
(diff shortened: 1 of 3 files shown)


CapnS:
I feel like this should be `await channel.follow(webhook_channel)` like how other methods are in d.py.

NCPlayz:
Ok! Reversed it so that the destination is provided for the method.

discord/webhook.py
```diff
@@ -495,6 +495,25 @@ def from_url(cls, url, *, adapter):
             raise InvalidArgument('Invalid webhook URL given.')
         return cls(m.groupdict(), adapter=adapter)
 
+    @classmethod
+    def as_follower(cls, data, *, channel, user):
```
Rapptz:
```suggestion
    def _as_follower(cls, data, *, channel, user):
```

discord/channel.py
```diff
@@ -455,6 +455,43 @@ def is_me(m):
...
+        --------
+        :class:`Webhook`
+            The created webhook.
+        """
+
+        if not self.is_news():
+            raise ClientException('The channel must be a news channel.')
+
+        data = await self._state.http.follow_webhook(self.id, webhook_channel_id=destination.id)
+        return Webhook.as_follower(data, channel=destination, user=self._state.user)
```
Rapptz:
```suggestion
        return Webhook._as_follower(data, channel=destination, user=self._state.user)
```

PR #2368, opened 2019-10-11
Refactor `name` kwarg for `Emoji.edit()`

### LINKED ISSUE #2259: discord.Emoji.edit's name kwarg is always required
### The Problem

`discord.Emoji.edit()` coro requires the `name` kwarg when used every time, even when the name is not the thing to change.

### The Ideal Solution

Doing something like `await emoji.edit(roles=[...])` should be possible - I already have the emoji, I just want to edit it's roles so I do that.

### The Current Solution

Since `name` is required every time, I have to do `await emoji.edit(name=emoji.name, roles=[...])` which is, well, clunky and unnecessary really.

### Summary

Make `name` kwarg be Optional / None by default, which will make it use the same, current name: https://github.com/Rapptz/discord.py/blob/ac00fcad771174090beacf2acfbd8af1a88108d0/discord/emoji.py#L275

This may make the `edit` method executable without any kwargs tho, so it could probably be slightly rewritten to match the `discord.Message`'s `edit` method: https://github.com/Rapptz/discord.py/blob/ac00fcad771174090beacf2acfbd8af1a88108d0/discord/message.py#L699

..., or `discord.Role`'s `edit` method: https://github.com/Rapptz/discord.py/blob/ac00fcad771174090beacf2acfbd8af1a88108d0/discord/role.py#L202

This change shouldn't break any currently existing code.

NCPlayz:
### Summary

This closes #2259. It refactors the fields for `Emoji.edit` so that the `name` parameter defaults to `Emoji.name`, which means it no longer throws a ``TypeError`` for not providing the keyword-only argument `name`.

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/emoji.py
+++ b/discord/emoji.py
@@ -281,7 +281,7 @@ class Emoji:
 
         await self._state.http.delete_custom_emoji(self.guild.id, self.id, reason=reason)
 
-    async def edit(self, *, name, roles=None, reason=None):
+    async def edit(self, *, name=None, roles=None, reason=None):
         r"""|coro|
 
         Edits the custom emoji.
@@ -306,6 +306,7 @@ class Emoji:
             An error occurred editing the emoji.
         """
 
+        name = name or self.name
         if roles:
             roles = [role.id for role in roles]
         await self._state.http.edit_custom_emoji(self.guild.id, self.id, name=name, roles=roles, reason=reason)


discord/emoji.py
```diff
@@ -291,7 +291,7 @@ def is_usable(self):
 
         Parameters
         -----------
-        name: :class:`str`
+        name: Optional[:class:`str`]
```
Rapptz:
```suggestion
        name: :class:`str`
```

As I've mentioned before, `Optional` = takes in `None` as a valid value. If anything, the docstring below it needs fixing.

discord/emoji.py
```diff
@@ -306,6 +306,9 @@ def is_usable(self):
             An error occurred editing the emoji.
         """
 
+        name = fields.get('name', self.name)
+
```
Rapptz:
```suggestion
```

discord/emoji.py
```diff
@@ -281,7 +281,7 @@ def is_usable(self):
 
         await self._state.http.delete_custom_emoji(self.guild.id, self.id, reason=reason)
 
-    async def edit(self, *, name, roles=None, reason=None):
+    async def edit(self, *, reason=None, **fields):
```
Rapptz:
```suggestion
    async def edit(self, *, name=None, roles=None, reason=None):
```

I don't get the pointless `**fields` transform since it's complicating the body.

discord/emoji.py
```diff
@@ -306,6 +306,9 @@ def is_usable(self):
             An error occurred editing the emoji.
         """
 
+        name = fields.get('name', self.name)
```
Rapptz:
```suggestion
        name = name or self.name
```

Following the signature change.

discord/emoji.py
```diff
@@ -306,6 +306,9 @@ def is_usable(self):
...
+
+        roles = fields.get('roles')
```
Rapptz:
```suggestion
```

Following signature change.

PR #2376, opened 2019-10-17
Prevent premium_subscription_count being None

### LINKED ISSUE #2331: discord.Guild.premium_subscription_count returns None instead of 0
### Summary

Small report, nothing really insteresting, but I had a case where `discord.Guild.premium_subscription_count` returned None. According to the doc, this parameter should always return an integer, which created a bug on my bot.
After some researchs, it was the case for 67 servers out of the 195 on which my bot is running.

Also according to this line:
https://github.com/Rapptz/discord.py/blob/ea3d119ca61e269f3d5bfdfe42178a82d2dba5e6/discord/guild.py#L388 I guess that the None value is provided at the creation of the Guild object, which maybe means (I am not sure, I didn't look further upstream) that Discord itself is returning None instead of 0. 
But I also have 115 servers where the same request return 0, so there's no certainly.

### Reproduction Steps

I am not sure how to reproduce it, but I guess if you have a bot with enough small servers you should find it with a simple `return [guild.id for guild in self.bot.guilds if guild.premium_subscription_count==None]`.

### Expected Results

Well either fix it in the lib so it returns 0 instead of None, or modify the documentation to explains that it can return None.

### Actual Results

I think I explained enough above.

But basically, on a server the bot was in, `guild.premium_subscription_count == None` instead of `guild.premium_subscription_count == 0`.

### Checklist

- [x] I have shown the entire traceback, if possible. (well there is no error, so...)

### System Information

- Python v3.7.4-final
- discord.py v1.3.0-alpha
    - discord.py pkg_resources: v1.2.3
- aiohttp v3.5.4
- websockets v6.0
- system info: Darwin 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64

XuaTheGrate:
### Summary
Fixes #2331 

`Guild.premium_subscription_count` can sometimes return None, for an unknown reason (as far as I've researched). This PR fixes this to never be None, and 0 instead.

### Checklist
- [x] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -385,7 +385,7 @@ class Guild(Hashable):
         self.max_presences = guild.get('max_presences')
         self.max_members = guild.get('max_members')
         self.premium_tier = guild.get('premium_tier', 0)
-        self.premium_subscription_count = guild.get('premium_subscription_count', 0)
+        self.premium_subscription_count = guild.get('premium_subscription_count') or 0
         self._system_channel_flags = guild.get('system_channel_flags', 0)
         self.preferred_locale = guild.get('preferred_locale')
 


Rapptz:
I still feel like this is a Discord bug but since they haven't and seemingly won't fix it then might as well do it client side.

PR #2385, opened 2019-10-21
Client: stop the loop before closing it.

iomintz:
According to the documentation for [loop.close()] it is not supposed to be called while the loop is still running.
It must be stopped first. Additionally this causes compatibility issues with trio-asyncio which unconditionally raises a RuntimeError
when a running loop is closed.

[loop.close()]: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.close

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -94,6 +94,7 @@ def _cleanup_loop(loop):
             loop.run_until_complete(loop.shutdown_asyncgens())
     finally:
         log.info('Closing the event loop.')
+        loop.stop()
         loop.close()
 
 class _ClientEventTask(asyncio.Task):


iomintz:
This change has been tested on both asyncio UnixSelectorEventLoop and trio-asyncio's asyncio-compatible event loop implementation.

Rapptz:
The loop is already stopped through multiple venues before that code should get hit:

https://github.com/Rapptz/discord.py/blob/4c3e53edf4ff90e3fcf195c429601a553d661cd8/discord/client.py#L613-L629

iomintz:
Upon further investigation, I am an idiot.

PR #2393, opened 2019-10-24
Reform message consistency

### LINKED ISSUE #1686: [rewrite] Adding reactions, but message.reactions is returning "[]"
So this is the code I have written:

`add_reactions_to = await main.send(user.mention, embed = e)`
`await add_reactions_to.add_reaction("↖")`
`await add_reactions_to.add_reaction("➡")`
`print(str(add_reactions_to.reactions))`

Adding the reactions to the message works just fine, I see them in the message. But it's like it isn't actually updating the message's reactions in context because the list of reactions is empty, both in the print method (always returning "[]") as well as an await to sense which reaction the _user_ selects (always returning False).

NCPlayz:
### Summary

This closes #1686. It basically caches the message from when the client sends the message, rather than when the ws receives it, at which point the cached message is just updated if it exists.

```py
In : add_reactions_to = await main.send(user.mention, embed = e)
...: await add_reactions_to.add_reaction("↖")
...: await add_reactions_to.add_reaction("➡")

In : add_reactions_to.reactions
Out: [<Reaction emoji='↖' me=True count=1>, <Reaction emoji='➡' me=True count=1>]
```

### Checklist

     -  [ ] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -835,7 +835,7 @@ class Messageable(metaclass=abc.ABCMeta):
         else:
             data = await state.http.send_message(channel.id, content, tts=tts, embed=embed, nonce=nonce)
 
-        ret = state.create_message(channel=channel, data=data)
+        ret = state.add_message(channel=channel, data=data)
         if delete_after is not None:
             await ret.delete(delay=delete_after)
         return ret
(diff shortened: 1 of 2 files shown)


NCPlayz:
Closing due to inactivity.

PR #2401, opened 2019-10-28
Update requirements.txt for future updates

Taewan-P:
### Summary

Since the latest version of aiohttp is 3.6.2, it says that it is incompatible with discord.py's latest version, 1.2.4. So I am suggesting an update on requirements.txt for future updates.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-aiohttp>=3.3.0,<3.6.0
+aiohttp>=3.5.4,<4.0.0
 websockets>=6.0,<7.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-aiohttp>=3.5.4,<4.0.0
+aiohttp>=3.6.0,<3.7.0
 websockets>=6.0,<7.0


Taewan-P:
Also attaching this [commit](https://github.com/Rapptz/discord.py/commit/e2478b17c998bb56fd96e183ddac053ba66363ae). It seems that he committed the aiohttps version prior to the update for future updates because aiohttp 3.6.0 was released on September, 2019, but as you see the commit date is Feb. I thought it was a good idea, so I updated one too.

Gelbpunkt:
I am using aiohttp 3.6.2 in several bots without any issues. +1 for this, dependencies should stay up to date

iomintz:
aiohttp.web from aiohttp 3.5 is also broken in python3.8 when using GunicornWebWorker. In my case I run a web server that imports my bot code, so I'm using aiohttp 3.5. I therefore can't use 3.8 for my web server.

RDIL:
@iomintz just out of interest, why did you choose that approach?

iomintz:
The API server performs the exact same functions as the bot, so I wrote my bot functions against bot.http and called them from the API server in order to minimize duplicate code.

requirements.txt
```diff
@@ -1,2 +1,2 @@
-aiohttp>=3.3.0,<3.6.0
+aiohttp>=3.5.4,<4.0.0
```
Rapptz:
```suggestion
aiohttp>=3.6.0,<3.7.0
```

Taewan-P: I just changed it...!

Taewan-P: @Rapptz Is it good?

PR #2414, opened 2019-11-05
Fix OggStream "invalid header magic" at end of stream

### LINKED ISSUE #2413: OggStream doesn't check for end of stream
### Summary

OggStream doesn't seem to check if the stream is finished or not. If the stream is finished, [the read](https://github.com/Rapptz/discord.py/blob/4c3e53edf4ff90e3fcf195c429601a553d661cd8/discord/oggparse.py#L77) reads 0 bytes, which doesn't match an 'OggS' header, which causes it to raise an exception even when the audio was played successfully.

### Reproduction Steps

```python
#!/usr/bin/python3.5
import discord
from BOT_TOKEN import TOKEN

def print_exception(e):
    if e is not None:
        print("Exception!")
        print(e)

class TestBot(discord.Client):
    async def on_ready(self):
        vc=await self.get_channel(640586707384795150).connect()
        vc.play(discord.FFmpegOpusAudio("test.mp3"), after=print_exception)

TestBot().run(TOKEN)
```

### Expected Results

When the audio finishes playing successfully, there are no exceptions.

### Actual Results

```
Exception!
invalid header magic
```

### Checklist

### System Information

- Python v3.5.3-final
- discord.py v1.3.0-alpha
    - discord.py pkg_resources: v1.3.0a2121+g4c3e53e
- aiohttp v3.5.4
- websockets v6.0
- system info: Linux 4.19.66-v7+ #1253 SMP Thu Aug 15 11:49:46 BST 2019

rkevin-arch:
### Summary

Fixes #2413 

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/oggparse.py
+++ b/discord/oggparse.py
@@ -77,7 +77,7 @@ class OggStream:
         head = self.stream.read(4)
         if head == b'OggS':
             return OggPage(self.stream)
-        elif len(head)==0:
+        elif not head:
             return None
         else:
             raise OggError('invalid header magic')
--- a/discord/oggparse.py
+++ b/discord/oggparse.py
@@ -77,6 +77,8 @@ class OggStream:
         head = self.stream.read(4)
         if head == b'OggS':
             return OggPage(self.stream)
+        elif len(head)==0:
+            return None
         else:
             raise OggError('invalid header magic')
 


discord/oggparse.py
```diff
@@ -77,6 +77,8 @@ def _next_page(self):
         head = self.stream.read(4)
         if head == b'OggS':
             return OggPage(self.stream)
+        elif len(head)==0:
```
iomintz:
```suggestion
        elif not head:
```

PR #2419, opened 2019-11-14
Return invites as https, various URL normalization

Gorialis:
### Summary

This pull request:
- Makes invites use `https://discord.gg/` instead of `http://discord.gg/`
- Normalizes URLs for the CDN in `Asset` using a `BASE` url (like Route)
- Normalizes URLs for invites using a `BASE` url (like Route)
- Fixes a minor documentation mistake related to `GuildChannel.overwrites`

This changes the behavior of Asset to natively refer to URLs under the CDN instead of absolute URLs like they used to be. `str(asset)` will still give you the full URL as normal, and constructors have been adjusted, but `Asset._url` and the repr will now show the URL relative to the root of the domain.

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -350,7 +350,7 @@ class GuildChannel:
         """Returns all of the channel's overwrites.
 
         This is returned as a dictionary where the key contains the target which
-        can be either a :class:`~discord.Role` or a :class:`~discord.Member` and the key is the
+        can be either a :class:`~discord.Role` or a :class:`~discord.Member` and the value is the
         overwrite as a :class:`~discord.PermissionOverwrite`.
 
         Returns
--- a/discord/asset.py
+++ b/discord/asset.py
@@ -63,6 +63,8 @@ class Asset:
     """
     __slots__ = ('_state', '_url')
 
+    BASE = 'https://cdn.discordapp.com/'
+
     def __init__(self, state, url=None):
         self._state = state
         self._url = url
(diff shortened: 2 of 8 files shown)


discord/abc.py
```diff
@@ -350,7 +350,7 @@ def overwrites(self):
         """Returns all of the channel's overwrites.
 
         This is returned as a dictionary where the key contains the target which
-        can be either a :class:`~discord.Role` or a :class:`~discord.Member` and the key is the
+        can be either a :class:`~discord.Role` or a :class:`~discord.Member` and the value is the
```
iomintz: This seems like an unrelated change. Separate PR?

discord/asset.py
```diff
@@ -132,15 +134,15 @@ def _from_guild_icon(cls, state, guild, *, format=None, static_format='webp', si
         if format is None:
             format = 'gif' if guild.is_icon_animated() else static_format
 
-        return cls(state, 'https://cdn.discordapp.com/icons/{0.id}/{0.icon}.{1}?size={2}'.format(guild, format, size))
+        return cls(state, '/icons/{0.id}/{0.icon}.{1}?size={2}'.format(guild, format, size))
 
 
     def __str__(self):
-        return self._url if self._url is not None else ''
+        return self.BASE + self._url if self._url is not None else ''
```
apple502j: is it really `self.BASE`?

iomintz: https://docs.python.org/3/tutorial/classes.html#class-and-instance-variables

PR #2315, opened 2019-08-17
Add message flags

diceroll123:
### Summary

Adds Message Flags to Message.

_If you look at the commits individually, you'll notice that I need to learn how to spell, but that's irrelevant._

### Checklist

- [x] If code changes were made then they have been tested.
    - [x] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.
- [x] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

PR #2433, opened 2019-11-20
Implement `MessageFlags`; Refactor flags placement

NCPlayz:
### Summary

This is an alternative implementation of `MessageFlags` with more up-to-date support. It mostly follows [Danny's Comment](https://github.com/Rapptz/discord.py/pull/2315#pullrequestreview-280015425) which suggests `MessageFlags` becoming more like `SystemChannelFlags`, and the accessibility of it. 

- Moves `SystemChannelFlags` to `discord.flags`.
- Creates `MessageFlags`.
- Removes `HTTPClient.suppress_message_embeds`.
- Sends the message flag for SUPPRESS_EMBED as part of `HTTPClient.edit_message`.
- Exposes `MessageFlags` a la `Message.flags`.
- Implements crossposting.

### Checklist

- [ ] This PR fixes an issue.


discord/flags.py
```diff
@@ -0,0 +1,237 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+"""
+
+__all__ = (
+    'SystemChannelFlags',
+    'MessageFlags'
```
Rapptz:
```suggestion
    'MessageFlags',
```

discord/flags.py
```diff
@@ -0,0 +1,237 @@
...
+    -----------
+    value: :class:`int`
+        The raw value. This value is a bit array field of a 53-bit integer
+        representing the currently available flags. You should query
+        flags via the properties rather than using this raw value.
+    """
+    __slots__ = ('value',)
+
+    def __init__(self, **kwargs):
+        self.value = self.ALL_OFF_VALUE
```
Rapptz: The default flag value should be 0 since it's not inverted. That's the entire point of ALL_OFF_VALUE.

```suggestion
        self.value = 0
```

discord/message.py
```diff
@@ -772,7 +779,9 @@ def system_content(self):
         except KeyError:
             pass
         else:
-            await self._state.http.suppress_message_embeds(self.channel.id, self.id, suppress=suppress)
+            if suppress:
+                self.flags._set_flag(4, True)
```
Rapptz: This entire implementation is weird. For starters, this is a boolean so you can do `suppress=True` and `suppress=False`. This code only handles if it's `suppress=True`. Likewise, we should probably support passing in a `MessageFlags` value directly.

discord/message.py
```diff
@@ -783,6 +792,27 @@ def system_content(self):
         if delete_after is not None:
             await self.delete(delay=delete_after)
 
+    async def publish(self):
```
Rapptz: I am unsure how you managed to put this into this PR.

PR #2445, opened 2019-11-23
Add support for YouTube Streaming

### LINKED ISSUE #2442: Update Streaming activity to support API changes
### The Problem
Currently, the library only supports retrieving a Streaming activity when it is from Twitch. As a result, all the properties are hard coded on the basis that Twitch is the only stream source, but that has since changed due to API updates.

### The Ideal Solution
Support all new properties from the API and allow for detection from YouTube.

### The Current Solution
See Ideal Solution

### Summary
Library only supports Twitch, so the activity.name doesn't include stream title any more, it just says Twitch because the .name property in the API is now the source site for a stream. .state and .details now include the data of Game Name and Stream Name as reported by those sites.

NCPlayz:
### Summary

Closes #2442.

### Checklist

- [x] This PR fixes an issue.


SuperSajuuk:
Can someone get a payload for this from the api so that the pr can be checked over more thoroughly? Original creator of this pr hasn’t appeared to answer the question from Danny since more than a week has passed and i fear not fixing this may be breaking bots relying upon this new payload info.

Rapptz:
Development discussion mainly takes place in the official server -- not in the github tracker. I received a payload soon after the question was asked but I didn't have time to check it over, truth is the data is not that different but it needs to be backwards compatible and that's hard to do.

Also this has a merge conflict now.

discord/activity.py
```diff
@@ -381,24 +381,29 @@ class Streaming(_ActivityTag):
...
 
     def __init__(self, *, name, url, **extra):
         super().__init__(**extra)
-        self.name = name
+        self.platform = name
+        self.name = extra.pop('details', None)
+        self.game = extra.pop('state', None)
         self.url = url
-        self.details = extra.pop('details', None)
+        self.details = self.name # compatibility
```
Rapptz:
```suggestion
        self.details = extra.pop('details', self.name) # compatibility
```

Issue #1996, opened 2019-03-18
[rewrite] Websocket issue:  asyncio.base_futures.InvalidStateError: invalid state


tigerism:
### Summary

There seems to be an issue with the websockets library which is causing this error. Even though this is a problem in a different library, I feel like discord.py should gracefully handle it and either retry the request or something, as it's completely crashing my bot.

Basically, my bot runs on clusters which are Docker containers. My bot uses about 10 clusters each responsible for 3 shards. On 1 cluster out of the 10, this problem constantly happens. It happens about once every 30 minutes, which takes down the process, and then the cluster respawns.

Something interesting to note: I edited the number of clusters and shards to spawn so it would force the distribution to change, and every time I change it, **this problem always happens on the cluster with the shard that is responsible for the support server of my bot**. Now, a few days ago, I bought the $25 "game developer fee" on my bot and picked my **support server as the target server to receive the perks,** so maybe this is related, but I'm not sure. I just know that when I change the cluster distribution, the 1 cluster which is affected happens to be where the game developer perks are on.

### Reproduction Steps

Explained above.

### Expected Results

No error

### Actual Results

Stacktrace:
[36mmaster               |[0m Traceback (most recent call last):
[36mmaster               |[0m 
[36mmaster               |[0m   File "bot.py", line 48, in <module>
[36mmaster               |[0m 
[36mmaster               |[0m     client.run(token)
[36mmaster               |[0m 
[36mmaster               |[0m   File "/usr/local/lib/python3.6/site-packages/discord/client.py", line 571, in run
[36mmaster               |[0m 
[36mmaster               |[0m     return task.result()
[36mmaster               |[0m 
[36mmaster               |[0m   File "/usr/local/lib/python3.6/site-packages/discord/client.py", line 479, in start
[36mmaster               |[0m 
[36mmaster               |[0m     await self.connect(reconnect=reconnect)
[36mmaster               |[0m 
[36mmaster               |[0m   File "/usr/local/lib/python3.6/site-packages/discord/client.py", line 402, in connect
[36mmaster               |[0m 
[36mmaster               |[0m     await self._connect()
[36mmaster               |[0m 
[36mmaster               |[0m   File "/usr/local/lib/python3.6/site-packages/discord/shard.py", line 271, in _connect
[36mmaster               |[0m 
[36mmaster               |[0m     f.result()
[36mmaster               |[0m 
[...]

PR #2453, opened 2019-11-26
Change websockets version constraint

Steffo99:
```diff
- websockets>=6.0,<7.0
+ websockets>=6.0,!=7,!=8.0.0,!=8.0.1,<9
```

### Summary

Since the bug preventing the upgrade [was fixed in `websockets==8.0`](https://github.com/Rapptz/discord.py/issues/1996#issuecomment-515787146), the version constraint can be safely changed to allow newer `websockets` versions.

Instead of changing the version constraint to `>=8.0`, i changed it to `>=6.0,!=7`, because as @Rapptz [said in the previous issue](https://github.com/Rapptz/discord.py/issues/1996#issuecomment-515816238), `websockets>=8.0` isn't compatible with Python 3.5.3, the minimum Python version required by `discord.py`.

I added `!=8.0.0,!=8.0.1`, as those two versions are bugged [[1]](https://websockets.readthedocs.io/en/stable/changelog.html#id4) [[2]](https://websockets.readthedocs.io/en/stable/changelog.html#id3) and would probably break the library.

I pinned the version to `8` with `<9`, as suggested by @Harmon758, so no breaking websocket changes will break `discord.py` in the future.

I've been testing using `websockets` ~~8.0~~ 8.0.2 for a while with my `discord.py` bot, and I didn't seem to encounter any problems.

This would make dependency management easier for some bots, as other packages they might use may have a `websockets>=8.0` dependency, which is currently incompatible with `discord.py`, even if it doesn't cause any problems. 

### Checklist

    - [ ] ~~I have updated the documentation to reflect the changes.~~ (Is there any documentation to update about this?)
- [ ] ~~This PR fixes an issue.~~
- [ ] ~~This PR adds something new (e.g. new method or parameters).~~
- [ ] ~~This PR is a breaking change (e.g. methods or parameters removed/renamed)~~
- [ ] ~~This PR is **not** a code change (e.g. documentation, README, ...)~~


Steffo99:
@Harmon758 thanks for your comment! I'll change the version requirements immediatly :)

requirements.txt
```diff
@@ -1,2 +1,2 @@
 aiohttp>=3.6.0,<3.7.0
-websockets>=6.0,<7.0
\ No newline at end of file
+websockets>=6.0,!=7,!=8.0.0,!=8.0.1,<9
```
Harmon758: To be consistent with websockets versioning:
```suggestion
websockets>=6.0,!=7.0,!=8.0,!=8.0.1,<9.0
```

Steffo99: Thanks!

PR #2463, opened 2019-12-05
Add support for xsalsa20_poly1305_lite

swarley:
### Summary

Adds support for `xsalsa20_poly1305_lite` and changes the protocol selection method to support ranked choice.

Also adds `lite_nonce` to `VoiceClient`, used for the incremental nonce.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -727,7 +727,7 @@ class DiscordVoiceWebSocket(websockets.client.WebSocketClientProtocol):
         log.debug('detected ip: %s port: %s', state.ip, state.port)
 
         # there *should* always be at least one supported mode (xsalsa20_poly1305)
-        modes = list(set(self._connection.supported_modes).intersection(data['modes']))
+        modes = [mode for mode in data['modes'] if mode in self._connection.supported_modes]
         log.debug('received supported encryption modes: %s', ", ".join(modes))
 
         mode = modes[0]
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -727,7 +727,7 @@ class DiscordVoiceWebSocket(websockets.client.WebSocketClientProtocol):
         log.debug('detected ip: %s port: %s', state.ip, state.port)
 
         # there *should* always be at least one supported mode (xsalsa20_poly1305)
-        modes = [mode for mode in data['modes'] if mode in self._connection.supported_modes]
+        modes = list(set(self._connection.supported_modes).intersection(data['modes']))
         log.debug('received supported encryption modes: %s', ", ".join(modes))
 
         mode = modes[0]
--- a/discord/voice_client.py
+++ b/discord/voice_client.py
@@ -111,7 +111,7 @@ class VoiceClient:
         self._runner = None
         self._player = None
         self.encoder = None
-        self.lite_nonce = 0
+        self._lite_nonce = 0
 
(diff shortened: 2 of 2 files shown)


imayhaveborkedit:
Turns out I already did this back when I did the vwsV4 pr and just never pr'd it here for some reason.  

https://github.com/imayhaveborkedit/discord.py/commit/5f15734b5528a1a6d5c54b04889ffb6ddc1b1886

discord/voice_client.py
```diff
@@ -331,6 +333,20 @@ def _encrypt_xsalsa20_poly1305_suffix(self, header, data):
 
         return header + box.encrypt(bytes(data), nonce).ciphertext + nonce
 
+    def _encrypt_xsalsa20_poly1305_lite(self, header, data):
+        box = nacl.secret.SecretBox(bytes(self.secret_key))
+        nonce = bytearray(24)
+
+        # Upper limit of unsigned 4 byte int
+        if self.lite_nonce >= 0xff_ff_ff_ff:
```
Rapptz: This is 3.6 syntax.

discord/gateway.py
```diff
@@ -727,7 +727,7 @@ def __init__(self, *args, **kwargs):
         log.debug('detected ip: %s port: %s', state.ip, state.port)
 
         # there *should* always be at least one supported mode (xsalsa20_poly1305)
-        modes = [mode for mode in data['modes'] if mode in self._connection.supported_modes]
+        modes = list(set(self._connection.supported_modes).intersection(data['modes']))
```
Rapptz: I don't see why this change was necessary.

swarley: I'm in the process of reverting it. In testing `xsalsa20_poly1305` was still being chosen over preferable modes. However more closely examining results from this change shows this also does not fix that consistently.

discord/voice_client.py
```diff
@@ -111,9 +111,11 @@ def __init__(self, state, timeout, channel):
         self._runner = None
         self._player = None
         self.encoder = None
+        self.lite_nonce = 0
```
imayhaveborkedit: The nonce is an impl detail and doesn't need to be public.

discord/voice_client.py
```diff
@@ -331,6 +333,20 @@ def _encrypt_xsalsa20_poly1305_suffix(self, header, data):
...
+    def _encrypt_xsalsa20_poly1305_lite(self, header, data):
+        box = nacl.secret.SecretBox(bytes(self.secret_key))
+        nonce = bytearray(24)
+
+        # Upper limit of unsigned 4 byte int
+        if self.lite_nonce >= 0xffffffff:
+            self.lite_nonce = 0
+
+        nonce[:4] = struct.pack('>I', self.lite_nonce)
+        self.lite_nonce += 1
```
imayhaveborkedit: This is exactly what the `checked_add` function is for.

swarley: I don't see how adding another function for this conditional is beneficial

imayhaveborkedit: The library already does this here:
https://github.com/Rapptz/discord.py/blob/c5f01eb66efb82fd1d2cf32b3658113322b37c01/discord/voice_client.py#L456

Issue #2469, opened 2019-12-06
Bot randomly disconnects from voice channel. 


mooniya:
### Summary

So this seems to happen 1~3 times per day, but the bot will randomly disconnect from the voice channel And I'm not sure why it disconnects as there are no logs.
After the bot disappears from the voice channel, it will need to be reconnected to function properly..

Is this problem related to Discord API response time?

### Checklist

- [x] I have searched the open issues for duplicates.
- [x] I have shown the entire traceback, if possible.
- [x] I have removed my token from display, if visible.

### System Information

Python v3.7.3
discord.py v1.2.4
websockets v6.0
system info: windows 10 64bit

SebbyLaw:
Refer to #2422 

In the future seek help in the [Official Support Server](https://discord.gg/r3sSKJJ).

Harmon758:
Without logs or more information, it's impossible to determine what the issue is.
Spamming multiple new issues about this also isn't going to help at all.

> there are no logs

Are you logging at the `DEBUG` level?

@SebbyLaw That issue is unrelated to simply disconnecting from voice channels.

PR #2470, opened 2019-12-06
UDP port is a big endian unsigned short

PendragonLore:
### Summary

This pull request makes the library interpret the voice UDP port as a big endian unsigned short as stated in https://github.com/discordapp/discord-api-docs/pull/1244

May be related to #2469

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -721,9 +721,7 @@ class DiscordVoiceWebSocket(websockets.client.WebSocketClientProtocol):
         ip_end = recv.index(0, ip_start)
         state.ip = recv[ip_start:ip_end].decode('ascii')
 
-        # the port is a little endian unsigned short in the last two bytes
-        # yes, this is different endianness from everything else
-        state.port = struct.unpack_from('<H', recv, len(recv) - 2)[0]
+        state.port = struct.unpack_from('>H', recv, len(recv) - 2)[0]
         log.debug('detected ip: %s port: %s', state.ip, state.port)
 
         # there *should* always be at least one supported mode (xsalsa20_poly1305)


Gorialis:
The referenced PR hasn't been reviewed by Discord developers yet (and is still unmerged), #2469 doesn't really have enough evidence to suggest it has anything to do with this, but yet this feels like a very breaking change (interpreting the port in the wrong endianness will give you a different number and you should fail to connect).

Are you sure this change is correct? I feel like there would be more documentation on this if it were the case (of course, if something has passed me by, please reference the contributing information in this thread).

PendragonLore:
I trusted the fact that the docs PR was correct due to many libraries quickly implementing this "fix" (see: https://github.com/bwmarrin/discordgo/issues/668, https://github.com/abalabahaha/eris/issues/592 and https://github.com/DV8FromTheWorld/JDA/commit/913210587da3304e0020b4c046a202b183e941f2). 

As mentioned in https://github.com/discordapp/discord-api-docs/pull/1243#issuecomment-562659896, this seems to be a change from over a year ago.

To be honest though, I'm very much confused by this as well as I've tested both little and big endianness with no issues.

Gorialis:
I can repro no issues - I performed a set of gateway tests (on both regular and voice websocket) and experienced no issues both with and without this change. How bizarre. Guess this is just another confusing gateway behavior.

Well, judging that it seems to be getting the OK from other library developers, I suppose this change is correct. I do wish Discord would make more of an effort to clear up these things ahead of time, though.

waynegerard:
Hey, just wanted to provide some insight:

If you *send* audio data first, discord will do the right thing and respond on the correct port after even if you send the LE port across the WS. 

If you are strictly receiving audio though, you'll notice that the UDP connection doesn't receive any data going forward.

Two quick gists illustrating: 
Little Endian: https://gist.github.com/waynegerard/0573c576892539738aa0adb741d6f59a
Big Endian: https://gist.github.com/waynegerard/4cbb6d87d4ce7e81db80e5a6d798d7c7

You can see that the port is clearly wrong if you interpret as LE

PR #2478, opened 2019-12-16
Add missing code in example.

alexandrianlibrarian:
### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -180,7 +180,7 @@ def find(predicate, seq):
     """A helper to return the first element found in the sequence
     that meets the predicate. For example: ::
 
-        member = find(lambda m: m.name == 'Mighty', channel.guild.members)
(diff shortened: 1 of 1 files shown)


Rapptz:
This isn't actually a typo or incorrect or even missing. It's a short example that assumes that you'd have done e.g. ```py
from discord.utils import find
```

At the top or somewhere. However I suppose since you got confused by it that being more explicit about this is helpful.

alexandrianlibrarian:
> This isn't actually a typo or incorrect or even missing. It's a short example that assumes that you'd have done e.g.
> 
> ```python
> from discord.utils import find
> ```
> 
> At the top or somewhere. However I suppose since you got confused by it that being more explicit about this is helpful.

That's pretty much what happened. I asked about it in the unofficial Discord API server (discord.py channel) and was told the same.

Would appreciate an edit in the docs making the assumption explicit; It's a potential stumbling block for someone new to the library.

PR #2485, opened 2019-12-22
Implement support for command argument separators.

NCPlayz:
### Summary

This is an adaptation of the now defunt PR - #645. It allows for two ways of command customization:

- Separator: Essentially lets you separate the arguments using something other than `` ``, requiring the use of quotes for multiple word arguments. `?foo a b, c, d, e` -> `('a', 'b', 'c', 'd', 'e')`
- Encapsulator: Another way of using quotes. `?bar a b [c d e] f g` -> `('a', 'b', 'c d e', 'f', 'g')`

As this changes quite a bit of the parsing of commands behind the scenes, I will try to test this as extensively as possible before marking it ready for review.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -37,6 +37,7 @@ from .cooldowns import Cooldown, BucketType, CooldownMapping
 from . import converter as converters
 from ._types import _BaseCommand
 from .cog import Cog
+from .view import Separator, Encapsulator
 
 __all__ = (
     'Command',
@@ -180,6 +181,9 @@ class Command(_BaseCommand):
         If ``True``\, cooldown processing is done after argument parsing,
         which calls converters. If ``False`` then cooldown processing is done
         first and then the converters are called second. Defaults to ``False``.
+    qualifier: Union[:class:`Separator`, :class:`Encapsulator`]
+        The qualifier for how arguments should be delimited. By default, it is
+        :class:`Separator`.
     """
 
     def __new__(cls, *args, **kwargs):
@@ -254,6 +258,7 @@ class Command(_BaseCommand):
         self.parent = parent if isinstance(parent, _BaseCommand) else None
         self._before_invoke = None
         self._after_invoke = None
+        self.qualifier = kwargs.pop('qualifier', Separator())
 
     @property
     def callback(self):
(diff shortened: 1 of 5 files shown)


discord/ext/commands/view.py
```diff
@@ -48,12 +48,73 @@
 }
 _all_quotes = set(_quotes.keys()) | set(_quotes.values())
 
+class Separator:
+    """An argument qualifier, which acts as the delimiter for arguments.
+    
+    .. code-block:: python3
+    
+        @bot.command(qualifier=Separator(',')
```
apple502j: Every doc is missing `)`

discord/ext/commands/core.py
```diff
@@ -180,6 +181,9 @@ class Command(_BaseCommand):
         If ``True``\, cooldown processing is done after argument parsing,
         which calls converters. If ``False`` then cooldown processing is done
         first and then the converters are called second. Defaults to ``False``.
+    qualifier: Union[:class:`Separator`, :class:`Encapsulator`]
```
Rapptz, line 184: Why can you only pass one instead of being able to pass both?

discord/ext/commands/core.py
```diff
@@ -908,20 +919,22 @@ def signature(self):
...
+                    result_params.append('[%s]' % name)
 
             elif param.kind == param.VAR_POSITIONAL:
-                result.append('[%s...]' % name)
+                result_params.append('[%s...]' % name)
             elif greedy:
-                result.append('[%s]...' % name)
+                result_params.append('[%s]...' % name)
             elif self._is_typing_optional(param.annotation):
                 result.append('[%s]' % name)
```
Rapptz, line 933: Shouldn't this also be `result_params.append`?

Likewise, what's the point of separating `result` and `result_params` if only one ends up being used?

discord/ext/commands/view.py
```diff
@@ -48,12 +48,77 @@
 }
 _all_quotes = set(_quotes.keys()) | set(_quotes.values())
 
+class Separator:
+    """An argument qualifier, which acts as the delimiter for arguments.
```
Rapptz, line 52: This documentation should probably show some output to show the difference between `strip_ws=False` and `strip_ws=True`.

PR #2494, opened 2020-01-02
Patch clean_prefix for new text editor

mikeshardmind:
### Summary

Makes clean_prefix handle both forms of mention.
The new editor seems to force mentions to have ! even without a nick.

### Checklist

- [X] This PR fixes an issue.


Rapptz:
After some discussion with the devs this is a bug on Discord's end that will be fixed. Unsure when though so this PR might get merged regardless, though this means I don't have to modify other stuff.

PR #2498, opened 2020-01-05
Add string check before conversion

hugopilot:
### Summary
My bot has a functionality to interact with people through both a server and DM. 
A command could look like `!interact <user>`. In a server, users are almost always passed by user mention. In Direct Messages however, the user is most commonly presented as `name#tag`. One of the problems however is that most users will copy an username directly from a server. This will paste as `@user#0000`. Conversion in this format will fail, as it will try to lookup `@user#0000`, rather than `user#0000`. To fix this, I've added a simple check where first character from the argument is truncated if this character is '@'. 

Something that is worth noting that I've only implemented this in `UserConverter`. _The `MemberConverter` is left untouched_

### Checklist

    - [ ] I have updated the documentation to reflect the changes. (not needed)
- [x] This PR fixes an issue.

### Testing
Before code changes:
image

After code changes:
image

Script used for testing available [here](https://gist.github.com/hugopilot/d04ad3fb6209a162aac7fcfeb2777aa3)

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -157,7 +157,13 @@ class UserConverter(IDConverter):
             result = ctx.bot.get_user(user_id) or _utils_get(ctx.message.mentions, id=user_id)
         else:
             arg = argument
-            # check for discriminator if it exists
+
+            # Remove the '@' character if this is the first character from the argument
+            if arg[:1] == '@':
+                # Remove first character
+                arg = arg[1:]
+
+            # check for discriminator if it exists,
             if len(arg) > 5 and arg[-5] == '#':
                 discrim = arg[-4:]
                 name = arg[:-5]
@@ -173,7 +179,6 @@ class UserConverter(IDConverter):
             raise BadArgument('User "{}" not found'.format(argument))
 
         return result
-
 class MessageConverter(Converter):
     """Converts to a :class:`discord.Message`.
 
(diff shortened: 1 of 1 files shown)


Phxntxm:
Users can have `@` in their name, likewise the first character in their name can be `@`.

hugopilot:
> 
> 
> Users can have `@` in their name, likewise the first character in their name can be `@`.

They cannot

image

Phxntxm:
Okay I guess the point in your difference of not touching `MemberConverter` I understand now, I was referring to a nickname when I said that but as you said you only changed `UserConverter`

hugopilot:
Yeah unless nicknames pass through the `UserConverter` too. I am will test this rq

hugopilot:
> Yeah unless nicknames pass through the `UserConverter` too


May be possible in theory, but Discord makes it really hard for me (and thus users) to just pass strings like strings:

image

hugopilot:
One other solution would be [checking a second time with the '@' truncated](https://gist.github.com/hugopilot/b1c79ecf9b0c55dfdd6544f073175ca7). It would be more fail-safe, but requires an extra try and it might make the function a lot slower.

discord/ext/commands/converter.py
```diff
@@ -157,7 +157,13 @@ class UserConverter(IDConverter):
             result = ctx.bot.get_user(user_id) or _utils_get(ctx.message.mentions, id=user_id)
         else:
             arg = argument
-            # check for discriminator if it exists
+
+            # Remove the '@' character if this is the first character from the argument
+            if arg[:1] == '@':
```
```suggestion
            if arg[0] == '@':
```

Rapptz:
Thanks for the PR. Sorry it took a while to merge but it's been rebased.

PR #2501, opened 2020-01-08
Implement `Guild.edit_role_positions`

### LINKED ISSUE #2143: [FR] Bulk role move
### The Problem

There is no user-exposed method to bulk-update role positions, which can result in behavior like #2142 

### The Ideal Solution

This method should obviously be part of the guild class. Possibly `guild.edit_role_positions`?
The interface for this is kind of tricky-perhaps an argument list of (role, position) tuples passed to the method? 

### The Current Solution

Using the internal `http.move_role_position` method. 

### Summary

A method to bulk-edit role positions.

NCPlayz:
### Summary

This closes #2143. The syntax for this method allows you to provide specific roles and their new role positions, however, I did think about providing an ordered list of all roles in their new role positions, and I felt like that would need more variables than providing specific role position changes. The syntax can still be changed if requested.

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -1591,6 +1591,68 @@ class Guild(Hashable):
         # TODO: add to cache
         return role
 
+    async def edit_role_positions(self, *, positions=None, reason=None):
+        """|coro|
+        
+        Changes the positions of a set of :class:`Role` for the guild.
+
+        You must have the :attr:`~Permissions.manage_roles` permission to
+        do this.
+
+        Example:
+
+        .. code-block:: python3
+
+            positions = {
+                bots_role: 1, # penultimate role
+                tester_role: 2,
+                admin_role: 6
+            }
+
+            await guild.edit_role_positions(positions=positions)
+
+        Parameters
+        -----------
+        positions
+            A :class:`dict` of :class:`Role` to :class:`int` to change the positions
+            of each given role.
+        reason: Optional[:class:`str`]
+            The reason for creating this channel. Shows up on the audit log.
+
+        Raises
+        -------
+        Forbidden
+            You do not have permissions to move the roles.
+        HTTPException
+            Moving the roles failed.
+        InvalidArgument
+            An invalid keyword argument was given.
+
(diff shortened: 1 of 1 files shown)


discord/guild.py
```diff
@@ -1591,6 +1591,68 @@ def fetch_members(self, *, limit=1, after=None):
         # TODO: add to cache
         return role
 
+    async def edit_role_positions(self, *, positions=None, reason=None):
+        """|coro|
+        
+        Changes the positions of a set of :class:`Role` for the guild.
```
Rapptz, line 1597:
```suggestion
        Bulk edits a list of :class:`Role` in the guild.
```

discord/guild.py
```diff
@@ -1591,6 +1591,68 @@ def fetch_members(self, *, limit=1, after=None):
         # TODO: add to cache
         return role
 
+    async def edit_role_positions(self, *, positions=None, reason=None):
```
Rapptz, line 1594: Why can you even pass `positions=None`?

discord/guild.py
```diff
@@ -1591,6 +1591,68 @@ def fetch_members(self, *, limit=1, after=None):
...
+                'id': role.id, 
+                'position': position
+            }
+
+            role_positions.append(payload)
+
+        data = await self._state.http.move_role_position(self.id, role_positions, reason=reason)
+        roles = [Role(guild=self, data=d, state=self._state) for d in data]
+
+        return roles
```
Rapptz, line 1654: Should probably update the internal cache as well.

discord/guild.py
```diff
@@ -1591,6 +1591,72 @@ async def create_role(self, *, reason=None, **fields):
         # TODO: add to cache
         return role
 
+    async def edit_role_positions(self, *, positions, reason=None):
```
Rapptz, line 1594:
```suggestion
    async def edit_role_positions(self, positions, *, reason=None):
```

I feel making this keyword-only is too strict.

PR #2502, opened 2020-01-09
Implement `commands.before/after_invoke`

### LINKED ISSUE #1986: [FR] Add a decorator for setting a command's before_invoke
### The Problem

At the moment, the way you would add a `before_invoke` to a command is by decorating a coro with `@yourcommand.before_invoke`
I think it would be nice to have a decorator similar to checks for adding a `before_invoke` to a command.
Mostly because using command checks for certain things isn't ideal.

### The Ideal Solution

Have a decorator for `before_invoke`. Something like:
```py
@commands.before_invoke(predicate)
# Or your own
@custom_before_invoke()
```
Which would set `_before_invoke` on a command.

### The Current Solution

Copy what a regular check does but change it to set the `_before_invoke` attribute of a command.
I don't know a clean solution for it though, since it's run with an extra cog argument in cog context, and it seems to set the command's cog during `call_before_hooks`, so while decorating `command.cog` would be `None`. 

### Summary

Add a decorator similar to command checks, but for setting a command's `before_invoke`.
Also applies for `after_invoke`.

NCPlayz:
### Summary

Closes #1986.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -707,7 +707,10 @@ class Command(_BaseCommand):
             try:
                 instance = self._after_invoke.__self__
             except AttributeError:
-                await self._after_invoke(ctx)
+                if self.cog:
+                    await self._after_invoke(cog, ctx)
+                else:
+                    await self._after_invoke(ctx)
             else:
                 await self._after_invoke(instance, ctx)
 
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -49,6 +49,8 @@ __all__ = (
     'has_any_role',
     'check',
     'check_any',
+    'before_invoke',
+    'after_invoke',
     'bot_has_role',
     'bot_has_permissions',
     'bot_has_any_role',
@@ -255,8 +257,16 @@ class Command(_BaseCommand):
         # bandaid for the fact that sometimes parent can be the bot instance
         parent = kwargs.get('parent')
         self.parent = parent if isinstance(parent, _BaseCommand) else None
-        self._before_invoke = None
-        self._after_invoke = None
+
+        try:
+            self.before_invoke(func.__before_invoke__)
+        except AttributeError:
+            self._before_invoke = None
+
+        try:
+            self.after_invoke(func.__after_invoke__)
+        except AttributeError:
+            self._after_invoke = None
 
     @property
     def callback(self):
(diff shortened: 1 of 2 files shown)


skyblockz:
You could just an on_message event, and place a function and the end of command function to do this

discord/ext/commands/core.py
```diff
@@ -255,8 +257,16 @@ def __init__(self, func, **kwargs):
         # bandaid for the fact that sometimes parent can be the bot instance
         parent = kwargs.get('parent')
         self.parent = parent if isinstance(parent, _BaseCommand) else None
-        self._before_invoke = None
-        self._after_invoke = None
+
+        try:
+            self.before_invoke(func.__before_invoke__)
```
Rapptz:
```suggestion
            before_invoke = func.__before_invoke__
```

I like to keep one attribute access per exception. You can work with it inside `else:`

discord/ext/commands/core.py
```diff
@@ -255,8 +257,16 @@ def __init__(self, func, **kwargs):
...
+        except AttributeError:
+            self._before_invoke = None
+
+        try:
+            self.after_invoke(func.__after_invoke__)
```
Rapptz:
```suggestion
            after_invoke = func.__after_invoke__
```

Same as above.

discord/ext/commands/core.py
```diff
@@ -1873,7 +1873,38 @@ def decorator(func):
...
+    
+    Example
+    ---------
+
+    .. code-block:: python3
+        
+        async def record_usage(ctx):
+            print(ctx.author, 'used', ctx.command, 'at', ctx.message.created_at)
+
+        @commands.before_invoke(record_usage)
```
Rapptz: General style-guide for these tends to have the bot command decorator on top, not on bottom.

discord/ext/commands/core.py
```diff
@@ -1873,7 +1873,38 @@ def decorator(func):
...
+        @bot.command()
+        async def who(ctx): # Output: <User> used who at <Time>
+            await ctx.send('i am a bot')
+
+        class What(commands.Cog):
```
Rapptz: This cog example can just be ripped entirely since it's not showcasing anything new that the previous example didn't and it's just bloating the page up.

discord/ext/commands/core.py
```diff
@@ -663,10 +677,14 @@ def __str__(self):
...
-            if cog is None:
+            try:
+                instance = self._before_invoke.__self__  
+                # should be cog unless the hook isn't in the cog
+            except AttributeError:
+                # __self__ only exists for methods, not functions
                 await self._before_invoke(ctx)
             else:
-                await self._before_invoke(cog, ctx)
+                await self._before_invoke(instance, ctx)
```
Rapptz: This is actually wrong. For regular cogs, e.g.

```py
class A(commands.Cog):
    @commands.command()
    async def foo(self, ctx):
        pass

    @foo.before_invoke
    async def before_foo(self, ctx):
        pass
```

Doing something like `bot.get_cog('A').foo._before_invoke.__self__` will be an attribute error -- because it's a function and it's not bound. The change-set suggested breaks that case. This is the same with `after_invoke`.

Rapptz:
Looks good to me.

PR #2513, opened 2020-01-19
Use the X-RateLimit-Bucket header provided by discord

Merlintor:
### Summary

It's recommended by the discord api docs that the `X-RateLimit-Bucket` is used for the rate limits. Due to the current implementation of the locking algorithm, we need a bucket before the first request is even made (we, therefore, don't have the bucket from the header). That's probably why the bucket is currently generated by discord.py with the following format: `{0.channel_id}:{0.guild_id}:{0.path}`. (isn't the method missing btw?)
This pull request is doing the same until we got the more accurate bucket from discord. This is done by mapping the endpoint (method + path) of the Route object (without the major parameters) to the bucket returned by discord (when we got it). The bucket with the major parameters is then generated in the `get_bucket` method of the Route.

This is not a perfect solution, because we are still using our generated bucket for the lock until we get the shared bucket from discord, but it's definitely better than ignoring the header completely.

Quote from the discord api docs:
`Per-route" rate limits may be shared across multiple, similar-use routes (or even the same route with a different HTTP method). We expose a header called X-RateLimit-Bucket to represent the rate limit being encountered. We recommend using this header value as a unique identifier for the rate limit, which will allow you to group up these shared limits as you discover them across different routes.`
### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -60,10 +60,17 @@ class Route:
         self.channel_id = parameters.get('channel_id')
         self.guild_id = parameters.get('guild_id')
 
-    @property
-    def bucket(self):
+    def get_bucket(self, shared_bucket=None):
+        # only used when the shared bucket returned by discord isn't available yet
         # the bucket is just method + path w/ major parameters
-        return '{0.channel_id}:{0.guild_id}:{0.path}'.format(self)
+        if shared_bucket is None:
+            return '{0.channel_id}:{0.guild_id}:{0.path}'.format(self)
+
+        return '{0.channel_id}:{0.guild_id}:{1}'.format(self, shared_bucket)
+
+    @property
+    def endpoint(self):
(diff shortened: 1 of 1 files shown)


Rapptz:
Thanks for the PR. This is however implemented locally in a different manner and due to a bug on Discord's part I wasn't sure how to proceed with it.

Merlintor:
Ah ok. 👌  I just think that it's better to stick to the discord docs as close as possible, because discord seems to like to change things without telling anyone. 
_Like the animated key in the reaction_add event._ 😄

alexsolex:
> Thanks for the PR. This is however implemented locally in a different manner and due to a bug on Discord's part I wasn't sure how to proceed with it.

I was looking for using the X-RateLimit-Limit as well. Can you point me to the way to know the rate limits with discord.py library ?

PR #2517, opened 2020-01-21
Add discord.utils.sleep_until helper function

bijij:
### Summary

Adds a helper function `discord.utils.sleep_until` to allow for sleeping until a specified datetime.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -338,6 +338,22 @@ async def sane_wait_for(futures, *, timeout):
 
     return done
 
+async def sleep_until(when):
+    """Sleep until a specified time.
+
+    If the time supplied is in the past this function will yield instantly.
+    
+    Parameters
+    -----------
+    when: :class:`datetime.datetime`
+        The timestamp in which to sleep until.
+
+    .. versionadded:: 1.3.0
+    """
+    now = datetime.datetime.now(datetime.timezone.utc)
+    delta = (when - now).total_seconds()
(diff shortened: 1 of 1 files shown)


discord/utils.py
```diff
@@ -338,6 +338,22 @@ def _parse_ratelimit_header(request, *, use_clock=False):
...
+    
+    Parameters
+    -----------
+    when: :class:`datetime.datetime`
+        The timestamp in which to sleep until.
+
+    .. versionadded:: 1.3.0
+    """
+    now = datetime.datetime.now(datetime.timezone.utc)
+    delta = (when - now).total_seconds()
```
Rapptz: Don't think this works if  `when` is naive.

```py
>>> import datetime
>>> x = datetime.datetime.utcnow() - datetime.timedelta(days=3)
>>> y = datetime.datetime.now(datetime.timezone.utc)
>>> y - x
  File "<stdin>", line 1, in <module>
TypeError: can't subtract offset-naive and offset-aware datetimes
```

bijij: Made changes to ensure naive datetimes will work by assuming they're UTC, should I document that do you think? Not sure what the precedent is for this.

Rapptz: All naive datetimes in the library have been UTC. e.g. `x.created_at` so that's fine. Likewise, this is what the task extension does.

PR #2520, opened 2020-01-22
Support subclasses in Enum instance check

Harmon758:
### Summary

This improves `isinstance` with `Enum`s by checking for instances of subclasses as well, matching `isinstance`'s default built-in behavior.

An example of how this would be useful is that it would allow subclassing of `BucketType`, so as to add custom buckets, such as channel *and* member. Currently, `isinstance` checks would fail when that custom subclass is used for a cooldown or maximum concurrency.

### Checklist

- [N/A] This PR fixes an issue.
- [N/A] This PR adds something new (e.g. new method or parameters).
- [N/A] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [N/A] This PR is **not** a code change (e.g. documentation, README, ...)

### DIFF
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -134,7 +134,8 @@ class EnumMeta(type):
         # isinstance(x, Y)
         # -> __instancecheck__(Y, x)
         try:
-            return instance._actual_enum_cls_ is self
+            cls = instance._actual_enum_cls_
+            return cls is self or issubclass(cls, self)
         except AttributeError:
             return False
 


Rapptz:
Custom buckets were planned to be supported through a different metric entirely, basically passing a custom class (see premature PR #2521). However recently through another refactoring (https://github.com/Rapptz/discord.py/commit/1a7b838d2adff881f825c8a16d14dab2207b311f) I ended up having a rather decent hook that would also work if this PR is merged.

I'm not sure where I stand on the matter.

Issue #2308, opened 2019-08-12
discord.py/ext/cooldowns.py - add overwrites for specific user ids


TheOneMusic:
So, my idea was to modify the cooldowns so that if a user is in a specific list of ids, the cooldown will be reduced or even disabled for that person. For example, let's create two tuples:
```python
donators = id_1, id_2
trusted = id_3, id_4
```
But now, if I add a command in this way:
```python
@client.command()
@commands.cooldown(1, 5, commands.BucketType.user)
async def my_command(ctx):
    pass
```
The cooldown is 5 seconds for everyone. Is it possible to make it to 3 if the context invoker's id is in the `donators` list and to disable it completely if the user is in the `trusted` list?

Rapptz:
This seems like a duplicate of #1999

Issue #1999, opened 2019-03-19
Allow for cooldown predicates and rollbacks.


EvieePy:
### The Problem

Currently, cooldowns lack flexibility in how they are processed. With no support for custom processing checks or state rollbacks.

### The Ideal Solution
**1.)** - Allow the cooldown to take in a predicate callable, which takes in a  sole parameter of `ctx`, which either returns a truth or false value. The predicate will be checked before the bucket is updated, allowing consumers to determine logically whether or not a cooldown bucket should be updated. 

Take for example a Whitelist example:
```py
async def my_cooldown_check(ctx):
    if ctx.author.id in [whitelistsed_users, ...]:
        return False
    return True

@commands.command(...)
@commands.cooldown(1, 60, commands.BucketType.user, predicate=my_cooldown_check)
async def my_command(...):
    ...
```
Here we return False if the author is in the whitelist, indicating that the cooldown should `not` be updated. True otherwise, indicating that the cooldown should be updated.

**2.)** - I haven't put too much thought into this one, but as it was brought up recently I'd thought I'd share a quick idea here. Currently there is no clean way of rolling back the cooldown by a single token in the case of a command failing during invocation, but after checks. 

My first thought was to add an optional kwarg, something along the lines of `update_on_failure=True/False`. I'm not sure what this means internally, I am yet to determine an appropriate solution to this, if one exists. My second thought was to expose a method on the cooldown which would allow the user to easily rollback a token. This would still need to be done in error handlers, but it would be much less ambiguous.

 
### The Current Solution

**1.)** - Currently the only way of achieving this sort of behaviour is to map the cooldown to a check.

**2.)** - Currently the only way of achieving this is by doing some messy workaround in error handlers.

### Summary

Allow flexibility in cooldowns by allowing optional predicates, and a way to clearly rollback the cooldown in case of a command failure.

Even though we briefly discussed on discord, I decided it was best to open this up and get feedback, ideas before actually making any sort of PR.

Rapptz:
Implemented by #6691 

The question so asynchronicity is still up in the air but returning `None` for the cooldown is effectively a bypass.

PR #2521, opened 2020-01-22
[commands] Allow passing a Cooldown class into the cooldown deco

4Kaylum:
### Summary

Small change to the `@cooldown` deco to allow users to pass in their own `Cooldown` and `CooldownMapping` classes. This is a similar change to ones that were made before, eg allowing custom classes for `Bot.get_context` and `@commands.command`.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1826,6 +1826,15 @@ def cooldown(rate, per, type=BucketType.default, *, cls=Cooldown, mapping=Cooldo
         The amount of seconds to wait for a cooldown when it's been triggered.
     type: :class:`.BucketType`
         The type of cooldown to have.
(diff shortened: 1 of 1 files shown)


Rapptz:
This PR is a bit premature.

These classes are not exactly exposed yet (rather, CooldownMapping isn't). This is intentional since the internal API is not meant to be relied on at this time. This PR assumes it is exposed and thus has a stable interface (though it does not actually say how). Ultimately I wanted to do this a bit differently later on, and this is a bit too premature for that. A few people have expressed different features that they want that would need to be accommodated such as #1999 and #2308. The asynchronicity of certain methods also needs to be discussed etc.

Overall, this PR is too early because in reality these things weren't exposed for a reason.

PR #2526, opened 2020-01-23
Add example on subclassing commands.Context

XuaTheGrate:
### Summary

Adds a new example, showcasing how to subclass commands.Context.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/custom_context.py
+++ b/examples/custom_context.py
@@ -1,3 +1,5 @@
+import random
+
 import discord
 from discord.ext import commands
 
@@ -6,9 +8,9 @@ class MyContext(commands.Context):
     async def tick(self, value):
         # reacts to the message with an emoji
         # depending on whether value is True or False
-        # if its True, itll add a green check mark
-        # otherwise, itll add a red cross mark
-        emoji = '\N{white heavy check mark}' if value else '\N{cross mark}'
+        # if its True, it'll add a green check mark
+        # otherwise, it'll add a red cross mark
+        emoji = '\N{WHITE HEAVY CHECK MARK}' if value else '\N{CROSS MARK}'
         try:
             # this will react to the command author's message
             await self.message.add_reaction(emoji)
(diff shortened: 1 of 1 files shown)


examples/custom_context.py
```diff
@@ -0,0 +1,43 @@
+import discord
+from discord.ext import commands
+
+
+class MyContext(commands.Context):
+    async def tick(self, value):
+        # reacts to the message with an emoji
+        # depending on whether value is True or False
+        # if its True, itll add a green check mark
```
Rapptz:
```suggestion
        # if its True, it'll add a green check mark
```

examples/custom_context.py
```diff
@@ -0,0 +1,43 @@
...
+        # otherwise, itll add a red cross mark
```
Rapptz:
```suggestion
        # otherwise, it'll add a red cross mark
```

examples/custom_context.py
```diff
@@ -0,0 +1,43 @@
...
+        emoji = '\N{white heavy check mark}' if value else '\N{cross mark}'
```
Rapptz:
```suggestion
        emoji = '\N{WHITE HEAVY CHECK MARK}' if value else '\N{CROSS MARK}'
```

examples/custom_context.py
```diff
@@ -0,0 +1,43 @@
...
+    """ Guess a random number from 1 to 6. """
+    # explained in a previous example, this gives you
+    # a random number from 1-6
+    value = random.randint(1, 6)
+    # with your new helper function, you can add a
+    # green check mark if the guess was correct,
+    # or a red cross mark if it wasnt
+    await ctx.tick(number == value)
+    
+bot.run("your token here")
```
Rapptz: Probably needs a comment saying that a token is like a password and shouldn't be hardcoded or shared. Maybe suggest a file etc.

examples/custom_context.py
```diff
@@ -0,0 +1,43 @@
...
+        
+
+bot = MyBot(command_prefix='!')
+
+@bot.command()
+async def guess(ctx, number: int):
+    """ Guess a random number from 1 to 6. """
+    # explained in a previous example, this gives you
+    # a random number from 1-6
+    value = random.randint(1, 6)
```
Rapptz: Oh almost forgot, this is missing `import random` at the top.

PR #2535, opened 2020-01-25
Add VoiceClient.latency and VoiceClient.average_latency Properties

Pytato:
### Summary

This PR adds the property `latency` to the `VoiceClient` object, similar to the property `Client.latency` but instead is `VoiceKeepAliveHandler`'s heartbeat and acknowledgement latency. `VoiceClient.latency` is `float('inf')` when `_keep_alive is None`, similar to behaviour of `Client.latency`.

In order to get reasonable resolution on latency measurements, the heartbeat interval has been set to a ceiling of 5000ms, meaning that if the `HELLO` heartbeat establishment requests a shorter interval, that will be used instead.

Example usage: this property may be used to monitor guild voice connection status to avoid voice latency issues plaguing the Europe nodes.

### Checklist

- [ ] This PR fixes an issue.

_Apologies for the past tense commit titles and unwarranted newline changes._

### DIFF
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -25,7 +25,7 @@ DEALINGS IN THE SOFTWARE.
 """
 
 import asyncio
-from collections import namedtuple
+from collections import namedtuple, deque
 import concurrent.futures
 import json
 import logging
@@ -34,7 +34,6 @@ import sys
 import time
 import threading
 import zlib
-import collections
 
 import websockets
 
@@ -133,7 +132,7 @@ class KeepAliveHandler(threading.Thread):
 class VoiceKeepAliveHandler(KeepAliveHandler):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.recent_ack_latencies = collections.deque(maxlen=20)
+        self.recent_ack_latencies = deque(maxlen=20)
         self.msg = 'Keeping voice websocket alive with timestamp %s.'
         self.block_msg = 'Voice heartbeat blocked for more than %s seconds'
         self.behind_msg = 'High socket latency, heartbeat is %.1fs behind'
@@ -149,9 +148,6 @@ class VoiceKeepAliveHandler(KeepAliveHandler):
         self._last_ack = ack_time
         self.latency = ack_time - self._last_send
         self.recent_ack_latencies.append(self.latency)
-
-        if self.latency > 10:
-            log.warning(self.behind_msg, self.latency)
 
 class DiscordWebSocket(websockets.client.WebSocketClientProtocol):
     """Implements a WebSocket for Discord's gateway v6.
@@ -751,6 +747,13 @@ class DiscordVoiceWebSocket(websockets.client.WebSocketClientProtocol):
         """:class:`float`: Latency between a HEARTBEAT and its HEARTBEAT_ACK in seconds."""
         heartbeat = self._keep_alive
         return float('inf') if heartbeat is None else heartbeat.latency
+
+    @property
+    def average_latency(self):
+        """:class:`list`: Average of last 20 HEARTBEAT latencies."""
+        heartbeat = self._keep_alive
+        average_latency = sum(heartbeat.recent_ack_latencies)/len(heartbeat.recent_ack_latencies)
+        return float('inf') if heartbeat is None else average_latency
 
     async def load_secret_key(self, data):
         log.info('received secret key for voice connection')
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -34,6 +34,7 @@ import sys
 import time
 import threading
 import zlib
+import collections
 
 import websockets
 
@@ -51,7 +52,6 @@ __all__ = (
     'DiscordVoiceWebSocket',
     'ResumeWebSocket',
 )
-
 
 class ResumeWebSocket(Exception):
     """Signals to initialise via RESUME opcode instead of IDENTIFY."""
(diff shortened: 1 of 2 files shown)


discord/gateway.py
```diff
@@ -129,9 +132,11 @@ def ack(self):
         if self.latency > 10:
             log.warning(self.behind_msg, self.latency)
 
+
 class VoiceKeepAliveHandler(KeepAliveHandler):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.recent_ack_latencies = []
```
Rapptz: A `deque` is the proper data type here since it handles automatically removing for you at the C level and has O(1) pop_front behaviour.

It resides in `collections`.

discord/gateway.py
```diff
@@ -52,13 +52,16 @@
     'ResumeWebSocket',
 )
 
+
```
Rapptz: Please don't add extraneous newlines in your PR.

discord/voice_client.py
```diff
@@ -183,7 +188,7 @@ def checked_add(self, attr, value, limit):
         endpoint = data.get('endpoint')
 
         if endpoint is None or self.token is None:
-            log.warning('Awaiting endpoint... This requires waiting. ' \
+            log.warning('Awaiting endpoint... This requires waiting. '
```
Rapptz: This change-set doesn't belong here either.

discord/gateway.py
```diff
@@ -702,7 +722,7 @@ def __init__(self, *args, **kwargs):
             await self.load_secret_key(data)
         elif op == self.HELLO:
             interval = data['heartbeat_interval'] / 1000.0
-            self._keep_alive = VoiceKeepAliveHandler(ws=self, interval=interval)
+            self._keep_alive = VoiceKeepAliveHandler(ws=self, interval=5.0)
```
Rapptz:
```suggestion
            self._keep_alive = VoiceKeepAliveHandler(ws=self, interval=min(interval, 5.0))
```

discord/gateway.py
```diff
@@ -142,6 +147,20 @@ def get_payload(self):
             'd': int(time.time() * 1000)
         }
 
+    def ack(self):
+        ack_time = time.perf_counter()
+        self._last_ack = ack_time
+        self.recent_ack_latencies.append(ack_time - self._last_send)
```
Rapptz: What is `_last_send`?

discord/gateway.py
```diff
@@ -142,6 +147,20 @@ def get_payload(self):
...
+        # Only the most recent 20 latencies are stored
+        if len(self.recent_ack_latencies) > 20:
+            del self.recent_ack_latencies[0]
+
+        self.latency = self.recent_ack_latencies
+
+        if self.latency[-1] > 10:
+            log.warning(self.behind_msg, self.latency)
```
Rapptz: This message doesn't really make sense for what it actually is.

discord/voice_client.py
```diff
@@ -207,6 +212,23 @@ def checked_add(self, attr, value, limit):
 
         self._handshake_complete.set()
 
+    @property
+    def latency(self):
```
Rapptz: I don't think exposing the list is good design. It should just return the latest latency value similar to `Client.latency`.

discord/gateway.py
Pytato: `_last_send` is the `time.perf_counter()` of the last `VoiceKeepAliveHandler` heartbeat send, inherited from parent: `KeepAliveHandler`.

discord/voice_client.py
Pytato: Changed functionality and exposure in b5f5e2b and a4a51fc.

discord/gateway.py
```diff
@@ -34,6 +34,7 @@
 import time
 import threading
 import zlib
+import collections
```
Rapptz: `collections` is already imported up there, just add `deque` to it.

discord/gateway.py
```diff
@@ -132,16 +133,26 @@ def ack(self):
 class VoiceKeepAliveHandler(KeepAliveHandler):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.recent_ack_latencies = collections.deque(maxlen=20)
```
Rapptz:
```suggestion
        self.recent_ack_latencies = deque(maxlen=20)
```

discord/voice_client.py
```diff
@@ -81,6 +80,10 @@ class VoiceClient:
         The voice connection token.
     endpoint: :class:`str`
         The endpoint we are connecting to.
+    latency: :class:`float`
```
Rapptz: If this is already documented as a property it doesn't need to be here.

discord/voice_client.py
Rapptz: What happened to the `average_latency`? I thought that was fine.

discord/gateway.py
Rapptz: Making it clearer:

A latency of 10 means that the ping is 10,000ms which probably means you're blocking somewhere.  For comparison the official client marks your connection red at 250ms. The regular client has a high throughput for events so 10 seconds being behind could just mean there's a lot to process. On the other hand, the voice client websocket is rather low maintenance spending most of its time just heartbeating.

Pytato:
Changes as requested have been made. Thanks

Rapptz:
Thanks for the PR and sorry this took so long.

PR #2540, opened 2020-01-27
Add is_running method to Loop

Gobot1234:
### Summary
Adds a property to the `ext.tasks.Loop` class enabling someone to check if a task is running more easily.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/tasks/__init__.py
+++ b/discord/ext/tasks/__init__.py
@@ -122,11 +122,6 @@ class Loop:
         elif self._task and self._task.done() or self._stop_next_iteration:
             return None
         return self._next_iteration
-    
-    @property
-    def is_running(self):
-        """:class:`bool`: Check if the task is currently running."""
-        return not self._task.done()
 
     def start(self, *args, **kwargs):
         r"""Starts the internal task in the event loop.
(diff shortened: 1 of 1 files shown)


discord/ext/tasks/__init__.py
```diff
@@ -276,6 +276,10 @@ def failed(self):
         """
         return self._has_failed
 
+    def is_running(self):
+        """:class:`bool`: Check if the task is currently running."""
+        return not self._task.done()
```
PikalaxALT: This needs to handle the case where `self._task is None`

Gobot1234: Will do

tylergibbs2:
I don't see the necessity of this. As documented, the `get_task()` method returns `None` if there is no task running. If anything, your code can be simplified to `return bool(self._task)`.

I believe returning only `True` or `False` in this case is required, rather than including `None` in the available return types, as this is an `is_` method.

Phxntxm:
You have misunderstood the docs... If the task has *not been ran* yet then yes, that will return `None`. However, if it has and the task has stopped, it will not return `None`... Instead it will return the task in its stopped/failed state

Rapptz:
LGTM.

Issue #2011, opened 2019-03-21
SIGINT handling on Windows


mdiller:
### The Problem

Windows does not seem to support doing `loop.add_signal_handler(signal.SIGINT, self._do_cleanup)` to handle SIGINT commands. 

### The Current Solution (and the issue with it)

Currently, we're just relying on catching the KeyboardInterrupt exception that gets thrown. ([line ~500 in client.py](https://github.com/Rapptz/discord.py/blob/d48a424f6f464044a59275d5e885b493a4a1aef7/discord/client.py#L551-L554))

This 'works,' but most of the time it does not react instantly, and I've observed it taking anywhere from 3 seconds to 30 seconds for the exception to be caught and the bot to close. This seems to be due to an issue with asyncio where the SIGINT gets blocked until something updates. I found [this issue](https://github.com/python/asyncio/issues/191) that describes the problem and says it was fixed in a later release of asyncio but it is still happening on my system, and for someone else in that thread.

### A Workaround

I've found [a workaround](https://stackoverflow.com/questions/24774980/why-cant-i-catch-sigint-when-asyncio-event-loop-is-running/24775107#24775107) that works for now. Adding the following lines to the run function in client.py solves the issue and the SIGINT is caught instantly.
```py
        def wakeup():
            loop.call_later(0.1, wakeup)

        try:
            loop.call_later(0.1, wakeup)
            loop.run_forever()
        except KeyboardInterrupt:
```
Is this an acceptable workaround to put in the main code or is it too hacky of a solution? If it sounds fine, I'll create a pull request to add it. Could also make it only run if it is on a windows machine.

### System Info

- discord.py version: 1.0.0a1781+gd48a424
- Python version: 3.6.5
- Operating system: Windows 10

Andrej730:
I'm encountering the same problem and my current solution is to add to client.py the following line:

```python
signal.signal(signal.SIGINT, lambda: bot.loop.stop())
```

Rapptz:
I feel like this should be fixed upstream via Python by now (and if not, then it should be handled by asyncio upstream) so I'm closing this for now.

sbourdeauducq:
You can fix it like this: https://github.com/m-labs/sipyco/commit/930ca5fc9c0ca79f31d538393c6238685ba53883

PR #2543, opened 2020-01-29
fix for slow shutdown

nfearnley:
### Summary

The existing code will stop the event loop on receiving a SIGINT or SIGTERM, but will not properly close the bot client. This means that the program will only exit once the current websocket connections have timed out, which takes approximately 10 seconds.

My change instead calls client.close(), which cleanly shuts down the bot. This means that the websockets are closed properly, and the program will exit without having to wait.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -611,8 +611,8 @@ class Client:
         loop = self.loop
 
         try:
-            loop.add_signal_handler(signal.SIGINT, lambda: loop.stop())
-            loop.add_signal_handler(signal.SIGTERM, lambda: loop.stop())
+            loop.add_signal_handler(signal.SIGINT, lambda: asyncio.create_task(self.close()))
+            loop.add_signal_handler(signal.SIGTERM, lambda: asyncio.create_task(self.close()))
         except NotImplementedError:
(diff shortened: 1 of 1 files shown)


nfearnley:
This may be related to #2011

Rapptz:
This PR seems misguided. The signal handlers calling `stop` is definitely what they should do since otherwise those tasks are just going to be cancelled. Note that the clean-up code *already* has `close()` in the `finally` block so it'll always run.

I'm not sure what the motivation was for this PR or what platform it's on but I think it's better suited as an issue.

nfearnley:
I'm definitely noticing a difference in shutdown times with and without this change. Maybe this should be implemented in a different way, but I recommend you try this with and without the change, and see the difference for yourself.

Rapptz:
This code is platform specific -- signal handlers are only on Linux. I don't know what platform you're on. As far as I know there was some bugs with the ProactorEventLoop (made the default on 3.8 on Windows) but on Linux this code definitely works fine since it uses the SelectorEventLoop.

PR #2554, opened 2020-02-05
Refactored and moved TextChannel purge function

XairGit:
### Summary

Refactored and moved the TextChannel purge function into the
discord.abc.Messageable class so that it can be used in DMChannel
instances and any other objects representing a messageable location.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -28,7 +28,7 @@ import abc
 import copy
 import asyncio
 import time
-from collections import namedtuple
+from collections import namedtuple, deque
 
 from .iterators import HistoryIterator
 from .context_managers import Typing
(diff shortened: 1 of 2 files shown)


Rapptz:
Thanks for the PR but `purge` is very much intentionally in `TextChannel` only since it's main use-case is for purging messages using bulk delete. It's easy to do bulk deleting without any sort of helper in the library:

```py
async for msg in messageable.history(limit=100):
   if msg.author == bot.user:
      await msg.delete()
```

Any other features can be added by keeping a buffer yourself.

In contrast, bulk delete requires certain handling such as chunking in groups of 100, checking for the time to ensure that it can be deleted, etc. Hence the need for an actual helper.

Your PR is using 3.6 features as well, this library is using 3.5.3+.

PR #2569, opened 2020-02-15
Add delay kwarg to TextChannel.delete_messages

wxMichael:
### Summary

Adds a delay kwarg to TextChannel.delete_messages, consistent with Message.delete.

### Checklist

- [ ] This PR fixes an issue.


Rapptz:
Thanks for the PR.

The `delay` and `delete_after` kwargs are added mainly because they're common operations that warrant them being in the library. This one doesn't seem to be as common or necessary so there are no arguments for their inclusion.

PR #2570, opened 2020-02-17
Fix invalid format specifier in PartialWebhookState.__getattr__

SebastiaanZ:
### Summary

The PartialWebhookState.\_\_getattr\_\_ tries to raise an `AttributeError`whenever it's called:

https://github.com/Rapptz/discord.py/blob/4cf5f792d402fe8f6fea054c87da7f460a594aa0/discord/webhook.py#L359-L360

However, the `!r` conversion flag is accidentally specified as a format specifier by a stray `:` causing a `ValueError` to be raised instead of the intended `AttributeError`:

```py
  File ".../lib/python3.8/site-packages/discord/webhook.py", line 358, in __getattr__
    raise AttributeError('PartialWebhookState does not support {0:!r}.'.format(attr))
ValueError: Invalid format specifier
```

This pull request simply removes that stray `:` to make it work as intended.

### Additional details

While it is unlikely to be an issue in regular bot use, this bug creates an issue when trying to use `unittest` and the `unittest.mock` from Python 3.8 onward. The reason is that `__getattr__` is used to determine whether or not something should be mocked with the new `AsyncMock` or a regular mock. Since the Python data model relies on that `AttributeError` to signal the non-existence of an attribute, the propagation of a `ValueError` instead of an `AttributeError` breaks things like `getattr(object, attribute)`.

### Checklist

- [x] This PR fixes an issue (an issue with the code, not a GH Issue)

### DIFF
--- a/discord/webhook.py
+++ b/discord/webhook.py
@@ -357,7 +357,7 @@ class _PartialWebhookState:
         return _FriendlyHttpAttributeErrorHelper()
 
     def __getattr__(self, attr):
-        raise AttributeError('PartialWebhookState does not support {0:!r}.'.format(attr))
+        raise AttributeError('PartialWebhookState does not support {0!r}.'.format(attr))
 
 class Webhook:
     """Represents a Discord webhook.


Rapptz:
Thanks for the PR.

Issue #2572, opened 2020-02-18
V1.2.5 throws KeyError when API is misbehaving, 1.3.1 shares the same code


nekokatt:
At the time of writing, the API is having latency issues. Responses are still coming back, but content-type headers are missing which is causing a key error to raise rather than an internal server error.

I am aware I am using an outdated library version, but inspecting the source for 1.3.1, no mitigation has been made for this issue, so this will impact 1.3.1 as well.

No clue what the body is causing it, but it is being triggered by editing messages:

```
Traceback (most recent call last):
  File "/neko3/.venv/lib/python3.8/site-packages/discord/ext/commands/core.py", line 79, in wrapped
    ret = await coro(*args, **kwargs)
  File "/neko3/neko3/features/ping_pong.py", line 124, in ping_command
    await msg.edit(content="", embed=embed)
  File "/neko3/.venv/lib/python3.8/site-packages/discord/message.py", line 744, in edit
    data = await self._state.http.edit_message(self.channel.id, self.id, **fields)
  File "/neko3/.venv/lib/python3.8/site-packages/discord/http.py", line 162, in request
    data = await json_or_text(r)
  File "/neko3/.venv/lib/python3.8/site-packages/discord/http.py", line 43, in json_or_text
    if response.headers['content-type'] == 'application/json':
KeyError: 'content-type'
```

Corresponds to [L164 on 4cf5f792d402fe8f6fea054c87da7f460a594aa0 in http.py](https://sourcegraph.com/github.com/Rapptz/discord.py@4cf5f792d402fe8f6fea054c87da7f460a594aa0/-/blob/discord/http.py#L164)

This has occurred three times within 10 minutes. Probably worth putting something in to check for this in the future and provide a meaningful exception instead to be safe.

Rapptz:
Cloudflare responses seem to no longer specify that it's HTML text via content type like it used to. Considering this has never come up in 4 years it must be a recent change on their end.

skyblockz:
dis.gd/status
Other than that 50k ping is api misbehavior 
For me discord is sending g empty websocket, it should be back soon

BudgetCode:
It might be due to the key value not being cased correctly.

NotSoSuper:
Had this issue yesterday, a few thousand errors..

nekokatt:
> It might be due to the key value not being cased correctly.

[...]

PR #2573, opened 2020-02-19
Fix Key 'Content-Type' not being cased correctly

BudgetCode:
### Summary

#2572 

- [x] This PR fixes an issue.


Rapptz:
This is a case insensitive dictionary.

PR #2582, opened 2020-03-01
Change package manager yum to dnf

tainn:
### Summary

Change of the `yum` package manager to the next-generation version called `dnf` in both `README` files. The use of `yum` is outdated on Red Hat-based Linux distributions and was replaced with `dnf`.

### Checklist

- [ ] This PR fixes an issue.


tainn1:
No idea why the PR name was changed from the commit's message to just `dnf`. Renamed.

skyblockz:
It should be an append to it, instead of direct overwriting, as they both works

Rapptz:
Thanks

PR #2591, opened 2020-03-09
Stop yielding duplicate commands from walk_commands due to aliases

MarkKoz:
### Summary

`GroupMixin.walk_commands` was yielding duplicate `Command` objects because it was processing aliases too. Aliases have the same `Command` object so it's redundant to yield them.

Documentation was not updated because it was never implied to begin with that it would yield aliases. If anything, it implied the opposite (i.e. the behaviour that will result from the changes I made).

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1110,7 +1110,7 @@ class GroupMixin:
 
     def walk_commands(self):
         """An iterator that recursively walks through all commands and subcommands."""
-        for command in tuple(self.all_commands.values()):
+        for command in self.commands:
             yield command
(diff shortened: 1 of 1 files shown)


MarkKoz:
The help command uses `cog.walk_commands` and that in turn uses `GroupMixin.walk_commands` (the one I have changed). As far as I can tell, nothing needs to be changed in what you mentioned. `__cog_commands__` won't have aliases and the help command just yields itself in addition to the usual.

Rapptz:
Oh that's true. Thanks for the PR.

PR #2634, opened 2020-04-05
A custom prefix takes a context instead of a message.

Fozar:
### Summary

This can add flexibility, especially in the case of custom Context. For example, it can help to avoid re-accessing the database if it stores the guild prefix and the custom Context has already received this information.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -42,7 +42,7 @@ from . import errors
 from .help import HelpCommand, DefaultHelpCommand
 from .cog import Cog
 
-def when_mentioned(bot, msg):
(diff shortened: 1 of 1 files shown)


Rapptz:
This change isn't going to happen. Here are the few reasons why:

1) This is a breaking change and I'm not accepting any of these without prior design discussion.
2) The created context at the time of this being created is *dead*. It's invalid and at a pending state. This function is called *as* the context is being created and is not done being created. In comparison, everything that takes a context parameter does it within 2 states -- valid or failed. This falls outside of these two states.

PR #2637, opened 2020-04-05
[commands] Add a converter for discord.Object

### LINKED ISSUE #2562: Use discord.Object annotation as an ID converter
### The Problem

It's quite common to have a command accept an arbitrary Discord ID value, in cases where obtaining such value (or it's respective model) via other converters would not be possible. The most iconic example of this would be a `hackban` command, where normally it's possible for the bot to not have any access to the member nor the user object for the ID provided. Similarly, an `unban` command would face the same issue. One could always write either their own custom ID converter, or accept an integer, check basic validity and wrap it in an object themselves, but I figured it'd be cool to have an already pre-made converter for this purpose.

### The Ideal Solution

Make it so that annotating an input argument in a command with `discord.Object` would use the already existing `IDConverter` to construct a `discord.Object` and return it as said argument's value. This has the benefit of already having basic input validation regarding the ID length, as well as obtaining an object that already meets the `Snowflake` abstract base class (some library methods can utilize this).

### The Current Solution

As explained, one has to make the argument accept an integer (possibly as a fallback at the end of a Union), and process the result themselves. Most of the time it's just boilerplate code, having this in the library itself would simplify this a lot, and I figured it wouldn't be too hard to implement given the already-existing `IDConverter`.

NCPlayz:
### Summary

Closes #2562. Also parses basic mentions for their IDs.

```py
@bot.command()
async def hackban(ctx, member: discord.Object):
    await ctx.guild.ban(member)
```

### Checklist

- [x] This PR fixes an issue.


Rapptz:
I'm unsure if I want this to parse mentions or not since it could be surprising. I'm kind of split between two opinions:

1) `discord.Object` handles the conversion of ID + mentions
2) `discord.Object` handles conversion of ID and a new type (name pending) handles ID + mentions.

This is not a review, merely a design discussion that should be thought about.

discord/ext/commands/converter.py
```diff
@@ -122,6 +122,30 @@ def _get_id_match(argument):
...
+    """Converts to a :class:`~discord.Object`.
+
+    The argument must follow the valid ID or mention formats (e.g. `<@80088516616269824>`).
+
+    .. versionadded:: 2.0
+
+    The lookup strategy is as follows (in order):
+
+    1. Lookup by ID.
+    2. Lookup by mention.
```
Rapptz, line 135:
```suggestion
    2. Lookup by member, role, or channel mention.
```

Rapptz (changes requested):
These need to be documented in the appropriate places in the restructured text files.

Issue #2567, opened 2020-02-14
on_ready doesn't fire when bot=False


pappabewar:
### Summary

When client is ran with bot=False, on_ready never fires

### Reproduction Steps

Run anything with bot=False

I ran
```
client = discord.AutoShardedClient(fetch_offline_members=False)
client.run('token', bot=False)
```

### Expected Results

Client connects and on_ready fires

### Actual Results

Log shows websocket events for messages, reactions etc. But on_ready never fires 

### Checklist

- [x] I have searched the open issues for duplicates.
- [x] I have shown the entire traceback, if possible.
- [x] I have removed my token from display, if visible.

### System Information

- Python v3.7.4-final
- discord.py v1.3.0-alpha
    - discord.py pkg_resources: v1.3.0a2132+g89bfd9c
- aiohttp v3.5.4
- websockets v6.0
- system info: Windows 10 10.0.18362

zomatree:
the ready event takes a long time for discord to send, also sharding for a user account is pointless and you should not selfbot as its against the discord TOS and can get you banned.

pappabewar:
Left it running for a couple of minutes, here's the tail of the log 
![image](https://user-images.githubusercontent.com/61059757/74550118-cdcc7b00-4f8b-11ea-8a26-90b932d25411.png)

Phxntxm:
If discord ain't sending READY for userbots that sucks ¯\_(ツ)_/¯

Reminder, as the last post mentions, userbots are against the TOS.

pappabewar:
> If discord ain't sending READY for userbots that sucks ¯_(ツ)_/¯
> 
> Reminder, as the last post mentions, userbots are against the TOS.

I am well aware, however, this library supports it

Phxntxm:
This library supports what the API supports. If the API isn't sending the READY event then....again, that sucks, but what do you want us to do about it? The library dispatches the events that are received, it's not gonna go "hey discord you forgot to send us READY for something that's not allowed! You should do that!"

pappabewar:
> This library supports what the API supports. If the gateway isn't sending the READY event then....again, that sucks, but what do you want us to do about it? The library dispatches the events that are received, it's not gonna go "hey discord you forgot to send us READY for something that's not allowed! You should do that!"

I don’t know what on_ready is, why do we even need it? If i’m getting messages and reactions etc over the websocket then i’m happy...

the-white-light:
[...]

PR #2645, opened 2020-04-07
Note that on_ready isn't called when logging in as a user

SirNapkin1334:
### Summary

This solves #2567, since there is nothing that can be done about that issue, as it is on Discord's end. So, I decided to update the documentation to reflect that.

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -113,7 +113,8 @@ to handle it, which defaults to print a traceback and ignoring the exception.
 .. function:: on_connect()
 
     Called when the client has successfully connected to Discord. This is not
-    the same as the client being fully prepared, see :func:`on_ready` for that.
+    the same as the client being fully prepared, see :func:`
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -136,6 +137,12 @@ to handle it, which defaults to print a traceback and ignoring the exception.
         Likewise, this function is **not** guaranteed to only be called
         once. This library implements reconnection logic and thus will
         end up calling this event whenever a RESUME request fails.
+        
+    .. warning::
+    
+        This function will  not be called if logging in with
```
numbermaniac, line 143:
```suggestion
        This function will not be called if logging in with
```
You've accidentally got 2 spaces here.

SirNapkin1334: Thank you, I did not notice that!

PR #3995, opened 2020-04-12
Fix an FFmpegOpusAudio documentation detail, and some capitalization

alasher:
### Summary

The documentation erroneously says that FFmpegOpusAudio returns "PCM bytes", when it should say it returns "Opus bytes". I also noticed other portions of the documentation typically have Opus capitalized, so I've corrected the appropriate instances within this file for consistency.

### Checklist


Rapptz:
Thanks

PR #4001, opened 2020-04-13
Removing from FAQ that there's not an event for invites being created

chickenmatty:
### Summary

Since on_invite_create is now a thing, the docs should be updated removing that Discord doesn't provide this information.

### Checklist

- [ ] This PR fixes an issue.


Rapptz:
Thanks

PR #4010, opened 2020-04-16
[commands] Add role mentions to when_mentioned prefixes.

mtimkovich:
### Summary

Adds the role mention `<@&{role_id}>` to the list of `when_mentioned` prefixes if the role's name is the same as the bot.

Note: This isn't escaped by `clean_prefix` in `DefaultHelpCommand`.

### Checklist

- [ ] This PR fixes an issue.


mtimkovich:
Seems like there's too many variables and unknowns to keep track of (renamed role name, which role corresponds to the bot, etc) that it should be left up to the user to implement by calling mentions_for themselves.

Issue #2617, opened 2020-03-27
Add `Cog.command_attrs` property.


webkaiyo:
### Summary

To access the `command_attrs` of a `Cog` class, it is needed to use `Cog.__cog_settings__`, which seems to be quite confusing. To fix this issue, I recommend to create a `command_attrs` property, which will probably look like this:
```py
@property
def command_attrs(self):
    return self.__cog_settings__
```

### Checklist

- [x] If code changes were made then they have been tested.
    - [ ] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.
- [x] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

PR #4013, opened 2020-04-17
[commands] Add Cog.command_attrs attribute

webkaiyo:
### Summary

This PR adds the attribute `Cog.command_attrs`. Currently, to access this attribute, it is necessary to use `Cog.__cog_settings__`. This feature solves the problem described in issue #2617.

### Checklist

- [x] This PR fixes an issue.


webkaiyo:
> This should probably be a read-only mapping if it's exposed.


Maybe I'm being ignorant but how do you propose this change, @Rapptz?

discord/ext/commands/cog.py
```diff
@@ -202,6 +202,11 @@ def qualified_name(self):
         """:class:`str`: Returns the cog's specified name, not the class name."""
         return self.__cog_name__
 
+    @property
+    def command_attrs(self):
+        """:class:`dict`: Returns the commands attributes applied for the cog."""
+        return self.__cog_settings__
```
Rapptz, line 208:
```suggestion
        return types.MappingProxyType(self.__cog_settings__)
```

Needs the import of course.

webkaiyo:
@Rapptz, did it.

PR #4018, opened 2020-04-18
member.joined_ago and user.created_ago

sergree:
### Summary

This pull request adds:
- `BaseUser` property `.created_ago` which returns the `timedelta` between `utcnow()` and `.created_at()`
- `Member` property `.joined_ago` which returns the *optional* `timedelta` between `utcnow()` and `.joined_at()`
- Helper function for calculating this difference (utils)

These properties allow users to get rid of the error-prone routine associated with calculating these values.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/member.py
+++ b/discord/member.py
@@ -33,6 +33,7 @@ from . import utils
 from .user import BaseUser, User
 from .activity import create_activity
 from .permissions import Permissions
+from .utils import get_timedelta
 from .enums import Status, try_enum
 from .colour import Colour
 from .object import Object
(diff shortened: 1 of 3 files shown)


Rapptz:
Thanks but these kind of trivial helpers aren't really something I want to do unless the operation is either really common or they actually buy you something that can be tricky to implement.

This operation is rather simple (it's one subtraction) so it's fairly trivial. The names are also not really descriptive or read very well (it sounds like broken English to a degree). Most people who end up working with this type of stuff tend to do something else with it such as getting the seconds and divmoding it up etc. So this doesn't really buy much other than renaming one line of code.

sergree:
How many people faced the problem of time zones using subtraction from `dt.now()` instead of `dt.utcnow()` unknowingly? Just a tiny time saver. But the truth is yours, of course.

PR #4038, opened 2020-04-28
[Commands] Remove "humanity" requirement for `process_commands`

RyanSchaefer:
### Summary

To better allow user choice and the ability to test with libraries such as https://github.com/JakeCover/distest, the requirement for the user being replied to to not be a bot should be removed. This would also allow for the `self_bot` parameter to actually work as intended as the bot will not disregard messages from itself because they came from a bot.

Since this technically [violates the TOS](https://www.reddit.com/r/discordapp/comments/a35txa/bots_setting_other_bots_commands/eb3man4?utm_source=share&utm_medium=web2x) it should be restricted to only testing, and thus is added along with a default global check that disallows the behavior. Thus, for most users of commands, nothing will change, but for people who actually want more control over the bot, they are allowed to do so.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -126,6 +126,9 @@ class BotBase(GroupMixin):
         else:
             self.help_command = help_command
 
+        if options.pop('bot_replies', True):
+            self._checks.append(lambda ctx: not ctx.message.author.bot)
+
     # internal helpers
 
     def dispatch(self, event_name, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


Rapptz:
Thanks but no thanks. Subclassing is the way to go here.

RyanSchaefer:
The [self_bot parameter ](https://github.com/Rapptz/discord.py/commit/20251c54a76810a030ea2604cedc88a6e66a11cc)is broken as it stands though because of [this later commit](https://github.com/Rapptz/discord.py/commit/aa7c5c3ec1b2df6c95c39b149639ebca9aa68859) which introduced the change.

Vexs:
The self_bot parameter is for userbots, ie, bots running on a user account. They allow the bot to respond to messages coming from itself. It is *not* to allow a bot to respond to other bots.

RyanSchaefer:
Ok, I guess I am just very new which is why I am bad at proposing stuff like this.

I am just confused about the design of this little part of it. I thought the design choice might be to prevent people from unwittingly violating the TOS. But, from what I can understand from your comment, this seems to be pushing people in a direction encouraging (bad?) that behavior. 

I will just fix it for myself with the suggested subclassing for now. Thanks for the explentation!

PR #4039, opened 2020-04-28
[commands] Change exception raised when cog not found

Genetical:
### Summary

When executing `commands.Bot.reload_extension` with an invalid extension name, it raises `ExtensionNotLoaded` when it seems like `ExtensionNotFound` would be a more useful exception to raise.

### Checklist
_Unchecked items are not applicable_
- [ ] This PR fixes an issue.

This is my first PR so I hope I didn't mess it up too much.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -712,7 +712,7 @@ class BotBase(GroupMixin):
 
         lib = self.__extensions.get(name)
         if lib is None:
-            raise errors.ExtensionNotLoaded(name)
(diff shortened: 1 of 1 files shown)


Rapptz:
This isn't going to change. The pre-condition for this is that the extension is already loaded. There is no good way to differentiate between the two cases so the best case is assuming the pre-condition is true.

Genetical:
Ah, I thought `ExtensionNotLoaded` meant that it failed to load, not that the condition of it being loaded wasn't met. Thanks anyway.

PR #4047, opened 2020-04-29
Fix `Guild.edit` suppressing `NotificationLevel` when editing

XuaTheGrate:
### Summary

Fixes a bug where passing a `NotificationLevel` into `Guild.edit` caused it to be suppressed.
This is due to `NotificationLevel` being an invalid argument to `int()`.

### Checklist

- [x] This PR fixes an issue.


Rapptz:
The other way to fix this would be to add `__int__` to the enum I feel. Would result in a smaller diff and wouldn't break users.

PR #4050, opened 2020-05-01
Documentation: Fix color properties being displayed as methods

kittenswolf:
### Summary

Currently, the documentation makes it look like the aliases for `color` and `colour` are methods. However, that is not the case and this PR fixes this.

### Checklist

- [ ] This PR fixes an issue.


Jekabs123:
> There is an alias for colour named meth.

 Hmmm

Issue #4051, opened 2020-05-01
Custom status not working


ldare373:
Hi, I am having some issues with setting a custom status for my bot. I have tried the following in on_ready()

`await client.change_presence(activity=discord.Activity(type=discord.ActivityType.playing, name="status"))`
Sets the status to "Playing status"

`await client.change_presence(activity=discord.Activity(type=discord.ActivityType.listening, name="status"))`
Sets the status to "Listening to status"

But `await client.change_presence(activity=discord.Activity(type=discord.ActivityType.custom, name="status"))`
Does not set a status.

Am I  doing something wrong?

Thanks in advance, Ldare373

webkaiyo:
Bots cannot have custom status.

helish88:
`# Setting `Playing ` status
await bot.change_presence(activity=discord.Game(name="a game"))

# Setting `Streaming ` status
await bot.change_presence(activity=discord.Streaming(name="My Stream", url=my_twitch_url))

# Setting `Listening ` status
await bot.change_presence(activity=discord.Activity(type=discord.ActivityType.listening, name="a song"))

# Setting `Watching ` status
await bot.change_presence(activity=discord.Activity(type=discord.ActivityType.watching, name="a movie"))`

ldare373:
Is there a reason they cannot have a custom status, or is is just not added yet?

Vexs:
Discord disallows bots to have a custom status. Watching/playing/etc statuses are the only ones allowed.

Issue #4049, opened 2020-04-30
Error when setting emoji in the CustomActivity


FightMan01:
### Summary

I'm trying to set a custom activity and getting an error.

### Reproduction Steps

```py
            cs = discord.CustomActivity(name="Custom status lol :P", emoji=discord.PartialEmoji(name='🤔'))
            await self.client.change_presence(activity=cs)
```

### Expected Results

The status will be set.

### Actual Results

First try with PartialEmoji (as seen in the docs):
```
Traceback (most recent call last):
  File "/Users/nemetharon/Desktop/FightMan01_bot/1.2.3/cogs/status.py", line 30, in on_status_ready
    cs = discord.CustomActivity(name="Custom status lol :P", emoji=discord.PartialEmoji(name='🤔'))
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/discord/activity.py", line 696, in __init__
    self.emoji = PartialEmoji.from_dict(emoji)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/discord/partial_emoji.py", line 80, in from_dict
    return cls(animated=data.get('animated', False), id=data.get('id'), name=data.get('name'))
AttributeError: 'PartialEmoji' object has no attribute 'get'
```

Second try with a simple string:
```
Traceback (most recent call last):
  File "/Users/nemetharon/Desktop/FightMan01_bot/1.2.3/cogs/status.py", line 30, in on_status_ready
    cs = discord.CustomActivity(name="Custom status lol :P", emoji='🤔')
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/discord/activity.py", line 696, in __init__
    self.emoji = PartialEmoji.from_dict(emoji)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/discord/partial_emoji.py", line 80, in from_dict
    return cls(animated=data.get('animated', False), id=data.get('id'), name=data.get('name'))
AttributeError: 'str' object has no attribute 'get'
```

### Checklist

- [x] I have searched the open issues for duplicates.
- [x] I have shown the entire traceback, if possible.
- [x] I have removed my token from display, if visible.

### System Information

- Python v3.8.2-final
- discord.py v1.3.3-final
- aiohttp v3.6.2
- websockets v6.0
[...]

Issue #2550, opened 2020-02-03
CustomActivity


jtrenaud1s:
When setting the client's (bot) presence to an Activity with type custom and a name, the presence is blank on the discord server, even though it looks to me like discord's added the api, and it's written in discord.py. Listening works fine, changing discord.ActivityType.listening to custom just makes the existing presence text disappear on discord.

numbermaniac:
I don't believe custom statuses for bots are supported yet.

> * Note that as of now, bots cannot send custom activities yet.

https://discordpy.readthedocs.io/en/latest/whats_new.html#v1-3-0

Harmon758:
Duplicate of #2400

Issue #2400, opened 2019-10-27
Custom Statuses


ThePiGuy24:
### The Problem

What I want is a way to set custom statuses as the feature has been added for users recently

### The Ideal Solution

Ideal solution would be using something like `client.change_presence(activity=discord.CustomStatus(status))`

### The Current Solution

None

### Summary

Just add or suggest some way to set custom statuses without the "Playing ...", "Watching ... " or "Streaming ..."

Rapptz:
The fields in the presence payload that actually give the custom status text are ignored for bots. 

Until the payload for it is finalized it's not going to be added. See here https://github.com/discordapp/discord-api-docs/issues/1160 for tracking it on the Discord side.

Shrubhog:
:tada: https://github.com/discordapp/discord-api-docs/issues/1160 was closed in https://github.com/discordapp/discord-api-docs/pull/1162!

LordBoos:
https://github.com/discordapp/discord-api-docs/issues/1160 was closed, can you check it and add the custom status to discord.py?

Harmon758:
Support for it should be coming soon.
> Custom status support is literally in my stash
I just have to pop it and that's it

\- [Danny, yesterday](https://discordapp.com/channels/336642139381301249/336642776609456130/646613363429736469)

MEXAHOTABOP:
https://discordapp.com/developers/docs/topics/gateway#activity-object-activity-types

> 4 | Custom | {emoji} {name} | ":smiley: I am cool"


this is it?

Phxntxm:
That is indeed what custom statuses are, if you were trying to inform others what it is though...I assure you Danny knows what custom statuses are, as the comment above yours showed he even has it ready in his stash.

MEXAHOTABOP:
sorry i dont know who is danny 

i just mention what this is appear in developer documentation but discordpy still not support this

MinerChAI:
> sorry i dont know who is danny
> 
> i just mention what this is appear in developer documentation but discordpy still not support this

Danny (aka Rapptz) is the creator of this lib lol

MEXAHOTABOP:
ok that makes sense

Harmon758:
This is likely delayed because:
> custom status [. . .] is gonna change in early 2020 apparently

\- [Danny, last Wednesday](https://discordapp.com/channels/336642139381301249/336642776609456130/656774207199510559)

Rapptz:
[...]

PR #4052, opened 2020-05-01
Note that bot accounts cannot set a CustomActivity

Vexs:
### Summary

This adds a doc note that bots cannot set a CustomActivity, a common misunderstanding seen in both the Discord and in #4051, #4049, #2550 and (kinda) #2400. 

Maybe clarifying as "customstatues are ignored" or similar might be better? 

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/activity.py
+++ b/discord/activity.py
@@ -672,6 +672,10 @@ class CustomActivity(BaseActivity):
 
             Returns the custom status text.
 
+    .. note::
+
+        Bot accounts cannot currently set a CustomActivity as their presence.
+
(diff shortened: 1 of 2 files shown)


Rapptz:
The note is technically already present in `BaseActivity` though that has visibility issues.

>  Note that although these types are considered user-settable by the library,
    Discord typically ignores certain combinations of activity depending on
    what is currently set. This behaviour may change in the future so there are
    no guarantees on whether Discord will actually let you set these types.

PR #4062, opened 2020-05-04
rename references from discordapp.com to discord.com

areyouokevan:
### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/webhook.py
+++ b/discord/webhook.py
@@ -53,7 +53,7 @@ class WebhookAdapter:
         The webhook that owns this adapter.
     """
 
-    BASE = 'https://discordapp.com/api/v7'
+    BASE = 'https://discord.com/api/v7'
 
     def _prepare(self, webhook):
         self._webhook_id = webhook.id
@@ -454,7 +454,7 @@ class Webhook:
     @property
     def url(self):
         """Returns the webhook's url."""
-        return 'https://discordapp.com/api/webhooks/{}/{}'.format(self.id, self.token)
+        return 'https://discord.com/api/webhooks/{}/{}'.format(self.id, self.token)
 
     @classmethod
     def partial(cls, id, token, *, adapter):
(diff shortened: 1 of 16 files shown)


discord/webhook.py
```diff
@@ -504,7 +504,7 @@ def from_url(cls, url, *, adapter):
             The URL is invalid.
         """
 
-        m = re.search(r'discordapp.com/api/webhooks/(?P<id>[0-9]{17,21})/(?P<token>[A-Za-z0-9\.\-\_]{60,68})', url)
+        m = re.search(r'discord.com/api/webhooks/(?P<id>[0-9]{17,21})/(?P<token>[A-Za-z0-9\.\-\_]{60,68})', url)
```
iomintz, line 507: this should be changed to make `app` optional, not to require the new URL.

areyouokevan: As per the announcement channel in Discord API
> We currently plan to start requiring the new domain later this year for third party developers on November 7th, 2020. We will be sending out more formal notices (system DMs and account emails) to announce this breaking change in the coming weeks.

iomintz: Just because Discord is gonna be breaking old URLs doesn't mean we have to do the same.

jacc: Make it `app` optional until November 7th, 2020, which at that point force users to use `discord.com` and not `discordapp.com`.

diceroll123: To put a time on `app` being there is to create a breaking change. It does not matter when. You have to consider people who set-and-forget bots with perfectly fine, yet old webhooks in their configs.

C0braD3v: All webhooks that are `discordapp.com` also work on the new `discord.com` domain, so any webhook ID/Token that is a valid webhook at `discordapp.com` will also be valid at `discord.com`

diceroll123: Not if they're not parsed here, they won't.

C0braD3v: In this case, could you not just add regex to check for either one?
```py
m = re.search(r'(discord|discordapp).com/api/webhooks/(?P<id>[0-9]{17,21})/(?P<token>[A-Za-z0-9\.\-\_]{60,68})', url)```

Comment (Rapptz) [inline discord/webhook.py:507]:
```suggestion
        m = re.search(r'discord(?:app)?.com/api/webhooks/(?P<id>[0-9]{17,21})/(?P<token>[A-Za-z0-9\.\-\_]{60,68})', url)
```

DleanJeans:
[MessageConverter](https://github.com/Rapptz/discord.py/blob/master/discord/ext/commands/converter.py#L197) is currently not working with `jump_url`. Removing `app` fixes it.

Rapptz:
Since this PR wasn't worked on I did it myself (along with the message converter issue). Thanks anyway.

PR #4068, opened 2020-05-06
Allow to clear author by passing `EmptyEmbed` to `name` arg in `set_author()`

### LINKED ISSUE #2610: Add a way to clear image/thumbnail/author in `discord.Embed`
### The Problem

It becomes easier to modify or reuse `discord.Embed` objects (with use of copy)

### The Ideal Solution

Make `name` kwarg of `set_author()` optional with default set to `EmptyEmbed`.
Make `url` kwarg of `set_image()` and `set_thumbnail()` optional with default set to `EmptyEmbed`.

### The Current Solution

The easiest but not supported one is accessing private attributes of `discord.Embed`.
Less easy one but supported would be using `to_dict()` and clearing image/thumbnail/author before using `from_dict()`

### Summary

I think there should be a way of clearing image/thumbnail/author of `discord.Embed` objects.

jack1142:
### Summary

This PR will allow to clear author in `Embed` by passing `EmptyEmbed` to `name` arg in `set_author()` (fixes #2610)

Alternatively, I could make a `clear_author()` function for clearing the author if you think that would be better.

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -374,7 +374,7 @@ class Embed:
         chaining.
 
         .. versionchanged:: 1.4
-            Passing :attr:`Empty` for `name` removes the author.
+            Passing :attr:`Empty` to all parameters removes the author.
 
         Parameters
         -----------
@@ -386,20 +386,19 @@ class Embed:
             The URL of the author icon. Only HTTP(S) is supported.
         """
 
-        if name is EmptyEmbed:
-            if url is not EmptyEmbed or icon_url is not EmptyEmbed:
-                raise ValueError("can't set author without name")
+        if all(param is EmptyEmbed for param in (name, url, icon_url)):
             del self._author
-        else:
-            self._author = {
-                'name': str(name)
-            }
+            return self
 
-            if url is not EmptyEmbed:
-                self._author['url'] = str(url)
+        self._author = {
+            'name': str(name)
+        }
 
-            if icon_url is not EmptyEmbed:
-                self._author['icon_url'] = str(icon_url)
+        if url is not EmptyEmbed:
+            self._author['url'] = str(url)
+
+        if icon_url is not EmptyEmbed:
+            self._author['icon_url'] = str(icon_url)
 
         return self
 
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -373,6 +373,9 @@ class Embed:
         This function returns the class instance to allow for fluent-style
         chaining.
 
+        .. versionchanged:: 1.4
+            Passing :attr:`Empty` for `name` removes the author.
+
         Parameters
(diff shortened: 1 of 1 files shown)


discord/embeds.py
```diff
@@ -383,15 +386,20 @@ def set_author(self, *, name, url=EmptyEmbed, icon_url=EmptyEmbed):
...
         """
 
-        self._author = {
-            'name': str(name)
-        }
+        if name is EmptyEmbed:
+            if url is not EmptyEmbed or icon_url is not EmptyEmbed:
+                raise ValueError("can't set author without name")
+            del self._author
+        else:
```
Rapptz, line 393: I don't like over-indenting too much. I suggest returning from here and leaving the rest unindented.

discord/embeds.py
```diff
@@ -383,15 +386,20 @@ def set_author(self, *, name, url=EmptyEmbed, icon_url=EmptyEmbed):
             The URL of the author icon. Only HTTP(S) is supported.
         """
 
-        self._author = {
-            'name': str(name)
-        }
+        if name is EmptyEmbed:
+            if url is not EmptyEmbed or icon_url is not EmptyEmbed:
+                raise ValueError("can't set author without name")
```
Rapptz, line 391: This restriction makes no sense. The API does the validation in this case and a name is not actually required.

jack1142: Should I then only clear if all of those 3 arguments are set to EmptyEmbed?
I only did the restriction cause API silently ignores author if it doesn't have a name, but it does make sense to leave handling of that to the API so I'll change that.

jack1142: I made it clear author only when all 3 params are `EmptyEmbed`, let me know if you wanted it to clear whenever just `name` is `EmptyEmbed`.

Also, do you want me to make the default of `name` be `EmptyEmbed` too? IMO, it would make sense but I noticed that's not what was done for `set_thumbnail()` and `set_image()` so I wanted to ask first.

Rapptz: Honestly it feels like if this were to be the case then the ergonomics of `clear_author()` are inherently better. Passing `EmptyEmbed` should only be if you want to clear specific attributes.

jack1142:
I did the changes and now author information can be cleared through separate method - `Embed.clear_author()`. While writing this, I figured there are few things that could be done differently so I'm listing them here:
- in current state, method returns the class instance for fluent-style chaining while `clear_fields()` and `remove_field()` don't
- method could be called `remove_author()` instead of `clear_author()`, not sure which fits better
- unlike `set_image()` and `set_thumbnail()`, this method catches `AttributeError` when clearing (for the cases when author information isn't set)

PR #4073, opened 2020-05-07
add support for allowed_mentions with message edit

StarrFox:
### Summary

This pull request adds support for passing allowed_mentions to message.edit like with messageable.send. It also uses the global allowed_mentions like with messageable.send.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -844,17 +844,13 @@ class Message:
         try:
             allowed_mentions = fields.pop('allowed_mentions')
         except KeyError:
-            allowed_mentions = self._state.allowed_mentions and self._state.allowed_mentions.to_dict()
+            pass
         else:
             if allowed_mentions is not None:
                 if self._state.allowed_mentions is not None:
                     allowed_mentions = self._state.allowed_mentions.merge(allowed_mentions).to_dict()
                 else:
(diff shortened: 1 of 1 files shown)


jack1142:
IMO, it's counter-intuitive to edit ``allowed_mentions`` in the message if nothing is passed to ``allowed_mentions``. And it makes it impossible to not edit the ``allowed_mentions`` that were already in the message before editing.

StarrFox:
@jack1142 If you don't pass any allowed_mentions to message edit then the message will default back to allowing all mentions, this might be an issue with discord itself I don't really know

jack1142:
> @jack1142 If you don't pass any allowed_mentions to message edit then the message will default back to allowing all mentions, this might be an issue with discord itself I don't really know

That's an issue with Discord API which should be reported to Discord - passing allowed_mentions object that might be completely different from message's when you don't pass anything at all is not a solution.

StarrFox:
It should no longer edit allowed_mentions if none are passed, it seems like discord is already aware of the issue (https://github.com/discord/discord-api-docs/issues/1419) and are either planning on fixing it or simply wont.

PR #4076, opened 2020-05-08
Use HTTP for Guild.query_members

iomintz:
## Summary

This changes Guild.query_members to use the new HTTP endpoint, which should alleviate some timeout issues.

## Checklist

- [ ] This PR fixes an issue.

## Remarks

I'm not sure what to do with the now-unused query_members functionality in the state class.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -1842,14 +1842,14 @@ class Guild(Hashable):
         Request members that belong to this guild whose username starts with
         the query given.
 
-        This is a websocket operation and can be slow.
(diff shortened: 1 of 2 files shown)


Rapptz:
No thanks. The http method is slower than the gateway method. In fact the http method internally on discord is just an rpc call to the gateway.

iomintz:
I created this PR because the current method times out for me whereas the HTTP call is instant. If there's something I should be doing differently please let me know.

PR #4080, opened 2020-05-08
Adjust `GuildChannel.permissions_for` to accept either a `Member` or `Role` object

LaughLax:
### Summary

This pull request adjusts the `GuildChannel.permissions_for` method to accept either a `Member` or `Role` object, rather than only a `Member` object.

This is my first contribution. I'm not sure how to test it and I don't expect it to be perfect on the first try; please be gentle 😅

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -37,6 +37,7 @@ from .role import Role
 from .invite import Invite
 from .file import File
 from .voice_client import VoiceClient
+from .member import Member
 from . import utils
 
 class _Undefined:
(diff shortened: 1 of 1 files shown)


LaughLax:
The use case for this is to avoid needing such a code snippet to see what access a role generally has for a channel. The required snippet would be larger than the three lines you suggested, for reasons that are handled within `permissions_for` but not within `overwrites_for`. For example, this ruleset:

> `@everyone` is denied `read_messages` for the channel `#hidden-channel`.
> `@ExampleRole` is granted the `Administrator` permission.

In this situation, a person with `@ExampleRole` has all permissions for `#hidden-channel`, but both code snippets would show "False" for the `view_channel` permission.

Likewise, this rule:

> `@ExampleRole` is denied `read_messages` for the channel `#hidden-channel`.

`@ExampleRole` should show no permissions of any kind for `#hidden-channel`. However, my testing of the first code snippet incorrectly shows certain permissions as true for that channel (for example, `read_history`).

These cases are all handled already within `permissions_for`, but not within `overwrites_for`. So, the use case for this is to avoid needing those suggested three lines *plus* implicit permissions handling (which takes 10 lines of code in `permissions_for`) any time I (or another lib user) want to check what access a role generally has for a channel.

Part of why it feels a little "tacked on" is because I'm not fully familiar which permission cases are handled elsewhere and which would still need to be handled within `permissions_for(role)`. I'm confident the code I submitted will handle them properly, but obviously it could be a bit more efficient if unnecessary code were bypassed. For the moment, I'll seek answers to these questions and report back:

* Are the `@everyone` permissions already built into the `Role`'s `Permissions` object?
* Are the `@everyone` channel overwrites already built into the `Role`'s channel overwrites?

sergree:
Faced the same problem. My temporary solution posted at Discord API guild: https://discordapp.com/channels/81384788765712384/381889733053251584/721445956607934544

PR #4090, opened 2020-05-11
implement AllowedMentions.__repr__

iomintz:
## Summary

Adds a repr to AllowedMentions for debugging.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/mentions.py
+++ b/discord/mentions.py
@@ -96,3 +96,6 @@ class AllowedMentions:
         users = self.users if other.users is default else other.users
         roles = self.roles if other.roles is default else other.roles
         return AllowedMentions(everyone=everyone, roles=roles, users=users)
+
+    def __repr__(self):
+        return '{0.__class__.__qualname__}(everyone={0.everyone}, users={0.users}, roles={0.roles})'.format(self)


discord/mentions.py
```diff
@@ -96,3 +96,6 @@ def merge(self, other):
         users = self.users if other.users is default else other.users
         roles = self.roles if other.roles is default else other.roles
         return AllowedMentions(everyone=everyone, roles=roles, users=users)
+
+    def __repr__(self):
+        return '{0.__class__.__qualname__}(everyone={0.everyone}, users={0.users}, roles={0.roles})'.format(self)
```
Rapptz, line 101: This doesn't really follow the format of the other `__repr__` in the library.

iomintz: If you're referring to the parentheses rather than angle brackets, I did it this way to follow the format of other simple data classes with no additional state in the library, such as EmbedProxy and PartialEmoji before the Asset redesign. Additionally it follows the Python convention of making eval(repr(x)) == x where possible, and using the angle brackets style where it's not.

PR #4097, opened 2020-05-15
Added on_socket_response to the events

applebee1558:
This adds the un-documented on_socket_response event to the api docs

### Summary

It adds the undocumented event on_socket_response to the api documentation

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -210,6 +210,21 @@ to handle it, which defaults to print a traceback and ignoring the exception.
                     WebSocket library. It can be :class:`bytes` to denote a binary
                     message or :class:`str` to denote a regular text message.
 
+.. function:: on_socket_response(payload)
+    
+    Called after the socket data is decoded. This data is decoded
+    from discord's ETF format.
+    
+    This is useful for debugging purposes if you do not want to manually
+    decode the ETF through the on_socket_raw_send event or if you want
+    to intercept payloads from the gateway and do something with it.
+    
+    .. note::
+        This is only for the messages received from the client
+        WebSocket. The voice WebSocket will not trigger this event.
+     
+    :param payload: The websocket payload decoded into a :class `dictionary`.
+
 .. function:: on_typing(channel, user, when)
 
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -210,6 +210,21 @@ to handle it, which defaults to print a traceback and ignoring the exception.
...
+    
+    This is useful for debugging purposes if you do not want to manually
+    decode the ETF through the on_socket_raw_send event or if you want
+    to intercept payloads from the gateway and do something with it.
+    
+    .. note::
+        This is only for the messages received from the client
+        WebSocket. The voice WebSocket will not trigger this event.
+     
+    :param payload: The websocket payload decoded into a :class `dictionary`.
```
ShineyDev, line 226:
```suggestion
    :param payload: The websocket payload decoded into a :class:`dictionary`.
```

Vexs:
This event is intentionally undocumented as the dictionary is modified by the time it is listened to, and largely useless/unreliable.

applebee1558:
What do you mean by modified by the time it is listened to? Based on looking in gateway.py, it seems to send this event right after the payload is loaded/decoded from json. Is this more or less supposed to be something internal?

Vexs:
The dictionary is passed as a reference, and event listens will happen at an indeterminate time, which is going to be after the library removes keys from the dict. Some things may not be removed because of the aforementioned indeterminate time of events happening.

applebee1558:
Would it be better if a new event was made and this one being deprecated? Such as something like on_socket_event with the raw string that isn't json-decoded but also already converted from the etf.

Rapptz:
This event is intentionally undocumented and there are no alternatives on purpose (producing a copy of a dictionary would be a high penalty for those who don't subscribe to it).

Vexs:
As nothing in the library actually listens to socket_response in the library, and it's an undocumented method (ie, safe to break), just moving it before the json parse seems most fitting.

Issue #4087, opened 2020-05-10
Member.top_role not updating when positions change


Silari:
### Summary

If a role position is changed, calling member.top_role does not update the top role if it has changed. It still shows whatever the old top position was.

Only similar issue still open is #2142 but that involved role creation and the cache, where these roles are being edited and the new position is set correctly. 

### Expected Results

The top_role property to always return the role with the highest position.

### Actual Results
The old top role is still returned as the top role, despite it being moved to a lower position or another role being moved higher. The output of the following code in an on_message(message) event shows the issue.

    msg = "Roles:"
    for myrole in message.guild.me.roles:
        msg += " Name: " + myrole.name + " Position:" + str(myrole.position)
    print(msg)
    msg = "Top Role: " + message.guild.me.top_role.name + " " str(message.guild.me.top_role.position)
    print(msg)

> Roles: Name: @everyone Position:0 Name: PicartoBot-Dev Position:2 Name: Trixie Position:5
> Top Role: Trixie 5

The highest role is Trixie, as it should be. After moving the Trixie role to just above @everyone (Position 1) and rerunning:

> Roles: Name: @everyone Position:0 Name: PicartoBot-Dev Position:3 Name: Trixie Position:1
> Top Role: Trixie 1

It still lists Trixie as the highest role, despite having updated the role with it's new position.

Looking at the code for Member.top_role, it seems like the list of roles should be sorted in ascending position order, but it's not being resorted when the role position is updated, thus causing top_role to return the wrong role.

### Checklist

- [x] I have searched the open issues for duplicates.
- [x] I have shown the entire traceback, if possible.
- [x] I have removed my token from display, if visible.

### System Information

- Python v3.7.2-final
- discord.py v1.3.3-final
- aiohttp v3.6.2
- websockets v6.0
- system info: Windows 10 10.0.18362

Rapptz:
I believe this happens because the cache for roles is not invalidated upon roles being updated. Invalidating the cache at this stage becomes kind of expensive so I may have to reconsider the cache layer here to properly fix this. Unsure at the moment.

PR #4105, opened 2020-05-17
Update Member.top_role

Jack073:
### Summary

Changed the behaviour of the Member top_role property, before it simply returned the role at the last index of self.roles, now it iterates over all the roles to find the role with the highest position and returns that instead, this seems more expensive and doesn't make sense initially, however there is some cases where self.roles isn't sorted correctly (presumably due to a caching issue) and in these cases the top_role function returned the wrong role, I've added a way to reproduce this behaviour below which seems to work for me.

I setup a test server, and added a bot with an integrated role, I then added another role and put it above the integrated role (in this case at the top of the roles list), and then started the bot (program below). I then sent `roles` to see what the bot considered its own roles were (their name and position), which should have been sorted by position and in this case they were. I then moved the new role (non integrated) below the integrated role and sent `roles` again, this time it had the positions of the roles correct again but the order of them wrong, resulting in top_role returning the wrong role.

Test program:
```python
import discord

class Client(discord.Client):
    async def on_ready(self):
        print("Ready")

    async def on_message(self, msg):
        if msg.content == "roles":
            roles = "\n".join([f"Name: {r.name} - Position: {r.position}" for r in msg.guild.me.roles])
            
            await msg.channel.send(f"top_role: {msg.guild.me.top_role.name}\n{roles}")

client = Client()
client.run("token")
```

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/member.py
+++ b/discord/member.py
@@ -434,7 +434,14 @@ class Member(discord.abc.Messageable, _BaseUser):
         This is useful for figuring where a member stands in the role
         hierarchy chain.
         """
-        return self.roles[-1]
+        highest_role = 0
+        highest_position = 0
+
+        for loc, r in enumerate(self.roles):
+            if r.position > highest_position:
+                highest_position = r.position
+                highest_role = loc
+        return self.roles[highest_role]
 
     @property
     def guild_permissions(self):


Rapptz:
Thanks for the PR but this is just a bandaid over a larger issue (see #4087)

PR #4109, opened 2020-05-18
Fix animated attribute type in PartialEmoji docstring

diceroll123:
### Summary

`animated` is optional in PartialEmoji

### Checklist
- [ ] This PR fixes an issue.


Rapptz:
This isn't true. It should be `False` and `True` only not `False`, `True` or `None`.

PR #4115, opened 2020-05-20
Fix KeyError on Member Channels not being in JSON

### LINKED ISSUE #4114: Widget members can have channels that aren't in the JSON
### Summary

The code assumed that if a widget member has a channel_id key set then it should be in the channels provided. This assumption is no longer true.

The proper solution in this case can either be:

1. Ignore it if the key isn't there
2. Use `discord.Object` if the channel isn't found. This requires a change to the signature.
3. Same as number 2 but fake a WidgetChannel to avoid the change to the signature.

### Reproduction Steps

Fetch a widget with someone in a voice channel.

### Expected Results

No error

### Actual Results

```py
  File "/usr/local/lib/python3.8/dist-packages/discord/ext/commands/core.py", line 83, in wrapped
  File "/home/getbeaned-discordbot/cogs/support.py", line 478, in doctor
    widget = await guild.widget()
  File "/usr/local/lib/python3.8/dist-packages/discord/guild.py", line 1827, in widget
    return Widget(state=self._state, data=data)
  File "/usr/local/lib/python3.8/dist-packages/discord/widget.py", line 208, in __init__
    connected_channel = channels[connected_channel]
KeyError: 71256824866275335
```

### Checklist

### System Information

- Python v3.7.3-final
- discord.py v1.4.0-alpha
    - discord.py pkg_resources: v1.3.3
- aiohttp v3.5.4
- websockets v6.0
- system info: Windows 8.1 6.3.9600

anurag-7:
### Summary

This PR fixes KeyError Caused by Member Channels not available in `channels`.
fixes #4114
### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/widget.py
+++ b/discord/widget.py
@@ -204,8 +204,10 @@ class Widget:
         channels = {channel.id: channel for channel in self.channels}
         for member in data.get('members', []):
             connected_channel = _get_as_snowflake(member, 'channel_id')
-            if connected_channel:
+            if connected_channel in channels:
                 connected_channel = channels[connected_channel]
+            else:
+                connected_channel = WidgetChannel(id=connected_channel, name=None, position=None)
 
             self.members.append(WidgetMember(state=self._state, data=member, connected_channel=connected_channel))
 
--- a/discord/widget.py
+++ b/discord/widget.py
@@ -206,7 +206,7 @@ class Widget:
             connected_channel = _get_as_snowflake(member, 'channel_id')
             if connected_channel in channels:
                 connected_channel = channels[connected_channel]
-            else:
+            elif connected_channel:
                 connected_channel = WidgetChannel(id=connected_channel, name=None, position=None)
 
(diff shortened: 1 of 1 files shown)


discord/widget.py
```diff
@@ -204,8 +204,10 @@ def __init__(self, *, state, data):
         channels = {channel.id: channel for channel in self.channels}
         for member in data.get('members', []):
             connected_channel = _get_as_snowflake(member, 'channel_id')
-            if connected_channel:
+            if connected_channel in channels:
                 connected_channel = channels[connected_channel]
+            elif connected_channel:
+                connected_channel = WidgetChannel(id=connected_channel, name=None, position=None)
```
Rapptz, line 210: Probably better to mock these so they don't change the typing. The current implementation as-is currently fails since `__str__` will return `None`.

anurag-7: Alright, I made the requested changes.

PR #4119, opened 2020-05-22
Allow webhooks to send emojis from other servers

Znunu:
### Summary

Currently webhooks only allow custom emojis from the server they belong to. Sending an auth header with each request to the webhook, makes the webhook behave more like normal messages coming from the bot, allowing emojis from other servers. I've only added this feature to webhooks made through `Webhook.from_state()` since those have the token readily available in their state attribute. An idea could be allowing any token to be passed to manually created webhooks.

### Checklist

    - [ ] I have updated the documentation to reflect the changes. (N/A?)
- [ ] This PR fixes an issue.

### DIFF
--- a/discord/webhook.py
+++ b/discord/webhook.py
@@ -113,7 +113,7 @@ class WebhookAdapter:
         finally:
             cleanup()
 
-    def execute_webhook(self, *, payload, wait=False, file=None, files=None):
+    def execute_webhook(self, *, payload, wait=False, file=None, files=None, headers=None):
         cleanup = None
         if file is not None:
             multipart = {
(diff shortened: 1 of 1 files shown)


Rapptz:
This behaviour is unsupported and a bug on Discord's end: https://github.com/discord/discord-api-docs/pull/1631

Znunu:
Indeed, but this bug hasn't been patched for a long time. This lib isn't affiliated with discord or anything, so I don't see what we stand to lose. Right now the lib is simply missing out. If the bug is patched, nothing is going to break. There's already a major bot, build around this very behavior.

Rapptz:
That mentality doesn't really work in this library. Maintenance burden for a Discord bot is not something I want to add.

Znunu:
Really don't see how this is a maintenance burden

PR #4121, opened 2020-05-22
Implement `Guild/TextChannel.edit(type=...)`

NCPlayz:
### Summary

This allows for text channels to be converted to news channels and vice versa.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -281,10 +281,14 @@ class GuildChannel:
                 perms.append(payload)
             options['permission_overwrites'] = perms
 
-        ch_type = options.get('type', self.type)
-        if not isinstance(ch_type, ChannelType):
-            raise InvalidArgument('type field must be of type ChannelType')
-        options['type'] = ch_type.value
+        try:
+            ch_type = options['type']
+        except KeyError:
+            pass
+        else:
+            if not isinstance(ch_type, ChannelType):
+                raise InvalidArgument('type field must be of type ChannelType')
+            options['type'] = ch_type.value
 
         if options:
             data = await self._state.http.edit_channel(self.id, reason=reason, **options)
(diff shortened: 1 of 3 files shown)


discord/abc.py
```diff
@@ -280,6 +281,11 @@ async def _edit(self, options, reason):
                 perms.append(payload)
             options['permission_overwrites'] = perms
 
+        ch_type = options.get('type', self.type)
+        if not isinstance(ch_type, ChannelType):
+            raise InvalidArgument('type field must be of type ChannelType')
+        options['type'] = ch_type.value
```
Rapptz, line 287: I think this implementation is too eager (it sends the `type` key all the time even if it doesn't actually change). Alternatively this implementation is better:

```suggestion
        try:
            ch_type = options['type']
        except KeyError:
            pass
        else:
            if not isinstance(ch_type, ChannelType):
                raise InvalidArgument('type field must be of type ChannelType')
            options['type'] = ch_type.value
```

discord/channel.py
```diff
@@ -216,6 +219,10 @@ async def edit(self, *, reason=None, **options):
         slowmode_delay: :class:`int`
             Specifies the slowmode rate limit for user in this channel, in seconds.
             A value of `0` disables slowmode. The maximum value possible is `21600`.
+        type: :class:`ChannelType`
+            Change the type of this text channel. Currently, only conversion between
+            :attr:`ChannelType.text` and :attr:`ChannelType.news` is supported. This 
+            is only available to guilds that contain `NEWS` in :attr:`Guild.features`.
```
Rapptz, line 225:
```suggestion
            is only available to guilds that contain ``NEWS`` in :attr:`Guild.features`.
```

PR #4123, opened 2020-05-23
Allow Lists and Tuples for checking roles

hugopilot:
# Summary

Lists and tuples can now be used to check for roles. 

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1599,8 +1599,13 @@ def has_any_role(*items):
             raise NoPrivateMessage()
 
         getter = functools.partial(discord.utils.get, ctx.author.roles)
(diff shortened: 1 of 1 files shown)


Vexs:
What exactly is the usecase for supporting nested lists? Further, this can be greatly improved by use of itertools, but I fail to see why.

hugopilot:
> What exactly is the usecase for supporting nested lists? 

Not having to add a role to every single line when using `@commands.has_any_role` but instead just referencing to a list and update roles there. It's a timesaver, really.

zomatree:
are you unable to do something like this? 
```py
@commands.has_any_role(*your_list)
```

hugopilot:
Ok apperantly that does work.

PR #4126, opened 2020-05-24
Clarify that TeamMember.id is the Discord user ID

monospacedmagic:
### Summary

This is just a very small improvement in the documentation that notes that a `TeamMember`'s ID is indeed the user's Discord ID.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/team.py
+++ b/discord/team.py
@@ -104,7 +104,7 @@ class TeamMember(BaseUser):
     name: :class:`str`
         The team member's username.
     id: :class:`int`
-        The team member's unique ID.
+        The team member's unique Discord user ID.
     discriminator: :class:`str`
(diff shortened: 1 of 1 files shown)


Rapptz:
I feel as if this is already implied and doesn't need to be explicitly mentioned. It looks a bit out of place as well since no other id documentation does this.

monospacedmagic:
I thought to add it because it wasn't obvious to me personally. I think it makes sense to add it here because it might not always be obvious to a user that this User-like class is actually a UserBase subclass. If there are other classes like that I'd be happy to clarify it there as well, it really is just a small thing that I stumbled upon and had to look up in the source code earlier.

PR #4128, opened 2020-05-25
[matrix] Dark Theme

bijij:
### Summary

Implements a dark theme for the docs
**THIS IS A WORK IN PROGRESS**

todo: 
 - optional toggle
 - fix the bits that aren't formatted right

### Checklist

- [ ] This PR fixes an issue.


docs/_static/custom.js
```diff
@@ -81,18 +92,11 @@ document.addEventListener('DOMContentLoaded', () => {
     parent.insertBefore(table, element.nextSibling);
   });
 
-  Object.entries(settings).forEach(([name, setter]) => {
+  Object.entries(settings).forEach(([name,]) => {
```
slice, line 95: In this case you'd be better of with `for (const key of Object.keys(settings))`.

PR #4131, opened 2020-05-25
[matrix] One-Page-Per-View Emulator

NCPlayz:
### Summary

> :warning: **This is extremely W.I.P. and needs to be optimised. Code reviews welcomed.**

Implements a discord.js-style documentation where each class is on a separate *page*.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/_static/custom.js
+++ b/docs/_static/custom.js
@@ -1,4 +1,4 @@
-'use-strict';
+"use-strict";
 
 let activeModal = null;
 let bottomHeightThreshold, sections;
@@ -6,6 +6,9 @@ let hamburgerToggle;
 let mobileSearch;
 let sidebar;
 let toTop;
+let apiReference;
+let enableScroll = false;
+let currentSection;
 
 class Modal {
   constructor(element) {
@@ -14,7 +17,7 @@ class Modal {
 
   close() {
     activeModal = null;
-    this.element.style.display = 'none'
+    this.element.style.display = "none";
   }
 
   open() {
@@ -22,17 +25,16 @@ class Modal {
       activeModal.close();
     }
     activeModal = this;
-    this.element.style.display = 'flex'
+    this.element.style.display = "flex";
   }
 }
 
 class SearchBar {
-
   constructor() {
-    this.box = document.querySelector('nav.mobile-only');
+    this.box = document.querySelector("nav.mobile-only");
     this.bar = document.querySelector('nav.mobile-only input[type="search"]');
-    this.openButton = document.getElementById('open-search');
-    this.closeButton = document.getElementById('close-search');
+    this.openButton = document.getElementById("open-search");
+    this.closeButton = document.getElementById("close-search");
   }
 
   open() {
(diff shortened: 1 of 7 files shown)


docs/_static/pages.js
```diff
@@ -0,0 +1,192 @@
...
+function collapseClasses() {
+  let sections = document.querySelectorAll(
+    "div.section:not(#api-reference)"
+  );
+  for (let sect of sections) {
+    let classes = Array.from(sect.children).filter((el) =>
+      el.classList.contains("section")
+    );
+
+    classes.forEach((v, i, a) => {
```
slice, line 110: You can drop `a` here.

slice:
I didn't review some instances of these:

- `Array.from(Node).filter` can usually be reduced into `Node.querySelector`.
- `Array.from(Node).forEach` can be reduced into `for (const Subnode of Node)`

docs/_static/custom.js
```diff
@@ -39,20 +41,36 @@ function getRootAttributeToggle(attributeName, valueName) {
...
+    }
+  });
+}
+
 const settings = {
-  useSansFont: getRootAttributeToggle('font', 'sans'),
-  useDarkTheme: getRootAttributeToggle('theme', 'dark')
+  useSansFont: [getRootAttributeToggle('font', 'sans'), true],
+  useDarkTheme: [getRootAttributeToggle('theme', 'dark'), true],
+  useClassicAPIRef: [toggleClassicAPIRef, false]
```
Rapptz, line 70: Definitely not a fan of this. It seems weird. I don't like random booleans.

docs/_static/custom.js
```diff
@@ -39,20 +41,36 @@ function getRootAttributeToggle(attributeName, valueName) {
...
+  if (apiReference) {
+    var event = new CustomEvent('APIRefStyle', { detail: { classic: add } });
+    window.dispatchEvent(event);
+    enableScroll = add;
+  } else {
+    enableScroll = true;
+  }
+}
+
+function loadSettings(settingsDict, pre=true) {
```
Rapptz, line 54: Why is thsi `pre` argument even here?

docs/_static/custom.js
```diff
@@ -39,20 +41,36 @@ function getRootAttributeToggle(attributeName, valueName) {
   return toggleRootAttribute;
 }
 
+function toggleClassicAPIRef(add) {
+  if (apiReference) {
+    var event = new CustomEvent('APIRefStyle', { detail: { classic: add } });
```
Rapptz, line 46:
```suggestion
    let event = new CustomEvent('APIRefStyle', { detail: { classic: add } });
```

What is this custom event for anyway?

docs/conf.py
```diff
@@ -237,6 +237,7 @@
 html_js_files = [
   'custom.js',
   'copy.js',
+  'pages.js'
```
Rapptz, line 240:
```suggestion
  'pages.js',
```

docs/_static/pages.js
```diff
@@ -0,0 +1,181 @@
...
+      expandAll();
+    }
+    obj = resolveClassicHref(window.location.href) || apiReference;
+  } else {
+    collapseSections();
+    collapseClasses();
+  
+    obj = resolveHref(window.location.href);
+  }
+  apiReference.style.display = "block";
```
Rapptz, line 16: Cross-script variables are a very very bad practice.

PR #4134, opened 2020-05-26
Show events as `Client.on_event` rather than `discord.on_event`

Gobot1234:
### Summary

This code change makes the documentation show dispatched events as methods of `discord.Client`, turning `discord.on_ready`, which isn't in line with the way the events are registered, into `discord.Client.on_ready`, which will also show that the methods can be registered on a subclass of `discord.Client`. 

This is achieved by documenting the events in `Client` and not that it's very important, reduces the chances of people mistyping event names in `Client` subclasses, which I have seen a fair few times.

A live version of the docs which I used for debugging with the changes can be viewed here https://notdiscordpy.readthedocs.io/en/latest/api.html#event-reference.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -142,7 +142,7 @@ class Client:
     shard_count: Optional[:class:`int`]
         The total number of shards.
     fetch_offline_members: :class:`bool`
-        Indicates if :func:`.on_ready` should be delayed to fetch all offline
+        Indicates if :meth`.on_ready` should be delayed to fetch all offline
         members from the guilds the client belongs to. If this is ``False``\, then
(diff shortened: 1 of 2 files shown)


jack1142:
Just an opinion, but considering you can add event listeners with `Client.event`, wouldn't it be better to have it out of the `discord` namespace in docs totally?

Also, adding all those methods for events is gonna make for a lot of additional spawned tasks, won't it?

Gobot1234:
I'm closing this due to the response on discord being overwhelmingly negative.

PR #4142, opened 2020-05-26
[matrix] Copy Codeblock

NCPlayz:
### Summary

TODO:
- use icons to avoid i18n issues
- scale properly on mobile

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/_static/copy.js
+++ b/docs/_static/copy.js
@@ -1,0 +1,22 @@
+const COPY = "Copy";
+const COPIED = "Copied!";
+
+var copy = (obj) => {
+  navigator.clipboard.writeText(obj.children[1].innerText);
+
+  obj.children[0].textContent = COPIED;
+  setTimeout(() => (obj.children[0].textContent = COPY), 2500);
+};
+
+document.addEventListener("DOMContentLoaded", () => {
+  let allCodeblocks = document.querySelectorAll("div[class='highlight']");
+
+  for (let codeblock of allCodeblocks) {
+      codeblock.parentNode.className += " relative-copy";
+      let copyEl = document.createElement("span");
+      copyEl.onclick = () => copy(codeblock);
+      copyEl.className = "copy";
+      copyEl.textContent = COPY;
+      codeblock.prepend(copyEl);
+  }
+});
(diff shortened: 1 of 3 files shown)


docs/_static/copy.js
```diff
@@ -0,0 +1,22 @@
+const COPY = "Copy";
+const COPIED = "Copied!";
+
+var copy = (obj) => {
```
Rapptz, line 4:
```suggestion
const copy = async (obj) => {
```

docs/_static/copy.js
```diff
@@ -0,0 +1,22 @@
...
+  navigator.clipboard.writeText(obj.children[1].innerText);
```
Rapptz, line 5:
```suggestion
  await navigator.clipboard.writeText(obj.children[1].innerText);
```

docs/_static/copy.js
```diff
@@ -0,0 +1,22 @@
+const COPY = "Copy";
+const COPIED = "Copied!";
+
+const copy = async (obj) => {
+  await navigator.clipboard.writeText(obj.children[1].innerText);
```
slice, line 5: This can fail, so you should probably catch it `alert` here so the user isn't confused when the button doesn't do anything.

docs/_static/copy.js
```diff
@@ -0,0 +1,22 @@
...
+  setTimeout(() => (obj.children[0].textContent = COPY), 2500);
+};
+
+document.addEventListener("DOMContentLoaded", () => {
+  let allCodeblocks = document.querySelectorAll("div[class='highlight']");
+
+  for (let codeblock of allCodeblocks) {
+      codeblock.parentNode.className += " relative-copy";
+      let copyEl = document.createElement("span");
+      copyEl.onclick = () => copy(codeblock);
```
slice, line 17: Nothing wrong with this, but it's more customary to use `addEventListener("click", ...)`.

docs/_static/style.css
```diff
@@ -618,3 +618,21 @@ div.code-block-caption {
...
+}
+
+.copy {
+  cursor: pointer;
+  position: absolute;
+  top: 0px;
+  right: 0px;
+  border: 1px solid #C6C9CB;
+  border-style: solid;
+  border-width: 1px;
```
slice, line 633: This property and the one above seems unnecessary?

PR #4143, opened 2020-05-27
[matrix] Add sans-serif font toggle to settings modal

bijij:
### Summary

Adds an option to switch to a sans-serif font.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/_static/custom.js
+++ b/docs/_static/custom.js
@@ -24,7 +24,7 @@ function updateSetting(element) {
 }
 
 function setFont(useSans) {
-  if (useSans === true) {
+  if (useSans) {
     document.body.classList.add('sans');
   } else {
     document.body.classList.remove('sans');
--- a/docs/_static/custom.js
+++ b/docs/_static/custom.js
@@ -21,19 +21,24 @@ function openModal(modal) {
 
 function updateSetting(element) {
   localStorage.setItem(element.name, element.checked);
-}
-
-function setFont(useSans) {
-  if (useSans) {
-    document.body.classList.add('sans');
-  } else {
-    document.body.classList.remove('sans');
+  if (element.name in settings) {
+    settings[element.name](element.checked);
   }
 }
 
-const settings = [['useSansFont', setFont]];
+function getBodyClassToggle(className) {
+  function toggleBodyClass(add) {
+    document.body.classList.toggle(className, add);
+  }
+  return toggleBodyClass;
+}
+
+const settings = {
+  useSansFont: getBodyClassToggle('sans')
+};
 
 document.addEventListener('DOMContentLoaded', () => {
+
   bottomHeightThreshold = document.documentElement.scrollHeight - 30;
   sections = document.querySelectorAll('div.section');
   settingsModal = document.querySelector('div#settings.modal')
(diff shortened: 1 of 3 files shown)


docs/_static/custom.js
```diff
@@ -19,10 +19,24 @@ function openModal(modal) {
   modal.style.removeProperty('display');
 }
 
+function updateSetting(element) {
+  localStorage.setItem(element.name, element.checked);
```
slice, line 23: I feel like we should prefix the storage keys here with something unique, maybe `discordpyDocs_` or something?

docs/_static/custom.js
```diff
@@ -19,10 +19,24 @@ function openModal(modal) {
...
+}
+
+function setFont(useSans) {
+  if (useSans === true) {
```
slice, line 27: I think you can remove the boolean comparison here?

```suggestion
  if (useSans) {
```

slice: Also, I think we should devise a more generalized system for applying classes to `<body>` according to a setting, as that'd probably be a common occurrence.

docs/_static/custom.js
```diff
@@ -19,10 +19,24 @@ function openModal(modal) {
...
+    document.body.classList.add('sans');
+  } else {
+    document.body.classList.remove('sans');
+  }
+}
+
+const settings = [['useSansFont', setFont]];
```
slice, line 34: Couldn't this be an object? You can iterate through an object's keys and values with `Object.entries`.

docs/_static/custom.js
```diff
@@ -31,6 +45,16 @@ document.addEventListener('DOMContentLoaded', () => {
     // insert ourselves after the element
     parent.insertBefore(table, element.nextSibling);
   });
+
+  settings.forEach(([settingName, settingFunction]) => {
+    let settingValue = JSON.parse(localStorage.getItem(settingName));
```
slice, line 50: You might want to wrap this in a `try`/`catch` block because any errors thrown will halt the entire `forEach`.

docs/_static/custom.js
```diff
@@ -64,4 +88,4 @@ document.addEventListener('keydown', (event) => {
   if (event.keyCode == 27 && activeModal) {
     closeModal(activeModal);
   }
-});
+});
```
slice, line 91: Reeeeeally minor, but your editor is nuking EOL newlines.

docs/_static/style.css
```diff
@@ -15,6 +15,10 @@ body {
   padding: 0;
 }
 
+body.sans {
+  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
```
slice, line 19: I'd like to point out the existence of [`system-ui`](https://caniuse.com/#feat=font-family-system-ui) here, although [it can be a little problematic](https://infinnie.github.io/blog/2017/systemui.html). This font stack is good, though.

docs/_templates/layout.html
```diff
@@ -25,11 +25,20 @@
...
+  <div id="settings" class="modal" style="display: none;" onclick="if (event.target == this){ closeModal(settingsModal); }">
     <div class="modal-content">
-      <span class="close" onclick="closeModal(settings);" title="Close">&times;</span>
+      <span class="close" onclick="closeModal(settingsModal);" title="Close">&times;</span>
       <h1>Settings</h1>
-      <!-- TODO: ADD OPTIONS HERE -->
+
+      <div class='setting'>
+        <h3>Use a sans-serif font:
+          <label class="toggle" title="Use a sans serif font?">
```
slice, line 35: The `title` attribute here is somewhat redundant. More detail could help:

```suggestion
          <label class="toggle" title="Use a sans serif font? Your system font will be used, falling back to `sans-serif`.">
```

docs/_static/custom.js
slice: This'll probably call for another function to grab settings though, since repeating the key prefix everywhere wouldn't be ideal.

PR #4147, opened 2020-05-29
Document that sleep until expects datetimes to be in UTC

Vexs:
### Summary

Improves documentation of sleep_until to note that it requires timestamps to be in UTC to function properly. 

### Checklist

- [ ] This PR fixes an issue.


Rapptz:
This documentation is wrong (also there are unrelated changes there for whatever reason).

PR #4178, opened 2020-06-05
Update intro.rst

jen140:
### Summary
As futher in the documentation we require to use python3 venv we should have the installation for it done(python3-venv).

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -44,7 +44,7 @@ For a Debian-based system, the following command will get these dependencies:
 
 .. code-block:: shell
 
-    $ apt install libffi-dev libnacl-dev python3-dev
+    $ apt install libffi-dev libnacl-dev python3-dev python3-venv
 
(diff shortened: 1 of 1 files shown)


Rapptz:
[`venv`](https://docs.python.org/3/library/venv.html) is a standard Python module. I am unsure what this is installing.

jen140:
For some reason it doesn't work on clean python3 installation on debian 10 as seen in the following picture:
error

Werseter:
Acutally, it is true some distros require installing python3-venv package for Virtual Environments to work8 cze 2020 01:14 Danny <notifications@github.com> napisał(a):
venv is a standard Python module. I am unsure what this is installing.

Rapptz:
Based on that error it seems to me that you don't have pip either.

jen140:
You were correct with that, but after installing it the issue is the same:
error2

PR #4182, opened 2020-06-07
Webhook extends Hashable

Znunu:
Webhooks can now properly be compared. Opened an issue about this a while ago, and was told it was an oversight, so yeah, now I'm suggesting to add this

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/webhook.py
+++ b/discord/webhook.py
@@ -402,15 +402,22 @@ class Webhook(Hashable):
         webhook.send('Hello World', username='Foo')
 
     .. container:: operations
+    
         .. describe:: x == y
+        
             Checks if two webhooks are equal.
+            
         .. describe:: x != y
+        
             Checks if two webhooks are not equal.
+            
         .. describe:: hash(x)
+        
             Returns the webhooks's hash.
-        .. describe:: str(x)
-            Returns the webhooks's name.
-
+            
+    .. versionchanged:: 1.4
+        Webhooks are now comparable and hashable.
+    
     Attributes
     ------------
     id: :class:`int`
@@ -462,9 +469,6 @@ class Webhook(Hashable):
     def __repr__(self):
         return '<Webhook id=%r>' % self.id
 
-    def __str__(self):
-        return self.name
-    
     @property
     def url(self):
         """:class:`str` : Returns the webhook's url."""
--- a/discord/webhook.py
+++ b/discord/webhook.py
@@ -54,7 +54,7 @@ class WebhookAdapter:
         The webhook that owns this adapter.
     """
 
-    BASE = 'https://discord.com/api/v7'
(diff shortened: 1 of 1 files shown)


Znunu:
I'm not _completely_ sure what documentation needs to be added, but I added the "supported operations container". My commits are a complete mess, with some changes that aren't actually mine but part of [this](https://github.com/Rapptz/discord.py/pull/5067) merge I believe. So sorry for that.

discord/webhook.py
```diff
@@ -400,6 +401,16 @@ async def foo():
         webhook = Webhook.partial(123456, 'abcdefg', adapter=RequestsWebhookAdapter())
         webhook.send('Hello World', username='Foo')
 
+    .. container:: operations
```
SebbyLaw, line 404: Should this container have a versionadded text?

Znunu: Sorry, I'm not familiar with what that is. Can I see an example or an explanation? ^_^"

Gobot1234: Something like 
```
.. versionchanged:: 1.4.0
     Webhooks now extend hashable.
```

discord/webhook.py
```diff
@@ -400,6 +401,16 @@ async def foo():
...
+        .. describe:: x == y
+            Checks if two webhooks are equal.
+        .. describe:: x != y
+            Checks if two webhooks are not equal.
+        .. describe:: hash(x)
+            Returns the webhooks's hash.
+        .. describe:: str(x)
+            Returns the webhooks's name.
```
Rapptz, line 412: These usually have newlines between them for rendering purposes. Take a look at the `Member` docstring for how it's done.

discord/webhook.py
```diff
@@ -451,17 +462,18 @@ def __init__(self, data, *, adapter, state=None):
     def __repr__(self):
         return '<Webhook id=%r>' % self.id
 
+    def __str__(self):
+        return self.name
```
Rapptz, line 466: If `self.name` is `None` (which is possible) then this will error out. Best to not provide this.

discord/webhook.py
```diff
@@ -400,6 +401,16 @@ async def foo():
...
+    .. container:: operations
+        .. describe:: x == y
+            Checks if two webhooks are equal.
+        .. describe:: x != y
+            Checks if two webhooks are not equal.
+        .. describe:: hash(x)
+            Returns the webhooks's hash.
+        .. describe:: str(x)
+            Returns the webhooks's name.
+
```
Rapptz, line 413: Somewhere here there should be 

```
.. versionchanged:: 1.4
    Webhooks are now comparable and hashable.
```

discord/webhook.py
Znunu: thx

PR #4183, opened 2020-06-07
Mixins is refractored

Znunu:
From what I could see __nq__ was just the inverse of __eq__ so I removed it. EqualityComparable wasn't used anywhere except for Hashable, and I can't imagine anything else, since they both rely on the same thing, object having an id. So I merged them.

### Checklist

- [ ] This PR fixes an issue.


Rapptz:
Thanks but I'll pass on this for now.

PR #4195, opened 2020-06-14
Update command_callback to make it prefix immune

Midorina:
### Summary

With this pull request, `command_callback()` removes the prefix from the beginning so that the help command can be used like this (assuming the prefix is `!`):

`!help !ping`

### Checklist

- [ ] This PR fixes an issue.


Rapptz:
Hello, thanks for the PR.

Unfortunately this isn't something I want to do nor is it something that makes sense. People tend to have custom prefixes set up and this makes it prime for breaking behaviour for no good reason. Consider a user setting a prefix to "hello" and a command named "helloworld". This scenario might seem outlandish but as a library with a general audience these are things that must be taken into consideration.

PR #4223, opened 2020-06-22
Fix error raised when using the roles parameter.

Taarek:
### Summary

Fix error raised when using the roles parameter. Also uses JSON fields now

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -529,7 +529,8 @@ class HTTPClient:
     def edit_channel(self, channel_id, *, reason=None, **options):
         r = Route('PATCH', '/channels/{channel_id}', channel_id=channel_id)
         valid_keys = ('name', 'parent_id', 'topic', 'bitrate', 'nsfw',
-                      'user_limit', 'position', 'permission_overwrites', 'rate_limit_per_user')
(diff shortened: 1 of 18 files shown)


discord/http.py
```diff
@@ -660,12 +660,12 @@ def get_member(self, guild_id, member_id):
         return self.request(Route('GET', '/guilds/{guild_id}/members/{member_id}', guild_id=guild_id, member_id=member_id))
 
     def prune_members(self, guild_id, days, compute_prune_count, roles, *, reason=None):
-        params = {
+        payload = {
             'days': days,
             'compute_prune_count': 'true' if compute_prune_count else 'false',
-            'include_roles': roles
+            'include_roles': ', '.join(roles)
```
Rapptz, line 666: Are we sure we can't just pass the array of snowflakes here as-is since it's JSON?

Taarek: I tried, there is no error returned, but at the same time, I am not sure if Discord is accepting it or not (since there is no indication whether it successfully pruned someone with the role)

Rapptz: Might be worthwhile to test on the computing endpoint and seeing if the numbers change w/ role and w/o role

PR #5068, opened 2020-06-27
Raise TypeError when Optional is used with Greedy converter

### LINKED ISSUE #2253: [Commands] Using a typing.Optional[] converter in a commands.Greedy[] one
### Summary

I tried making a command with an argument like
`cmd(ctx, arg: commands.Greedy[typing.Optional[int]])`
(actual command and types not shown, but they work on their own)
However, the bot gets stuck in an infinite loop in Command._transform_greedy_pos().

### Reproduction Steps

I had a typing.optional[] type in a command's commands.Greedy[] converter. Similar to
`cmd(ctx, arg: commands.Greedy[typing.Optional[int]])`

### Expected Results

For this example, if I had a command like
`>command str 1 2 str 3`
arg would be passed a value of
`[None, 1, 2, None, 3]`
since the strings would get converted to `None` under the typing.Optional converter

### Actual Results

The bot is stuck in an infinite loop between lines 460-476 of core.py, in Command._transform_greedy_pos() at the loop 'while not view.eof'

### Checklist

- [ ] I have shown the entire traceback, if possible.

### System Information

```
- Python v3.6.8-final
- discord.py v1.2.3-final
- aiohttp v3.5.4
- websockets v6.0
- system info: Windows 10 10.0.17134
```

SebbyLaw:
### Summary

This closes #2253. Having Union[..., `None`] makes no sense in Greedy, and should throw an error.

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -26,6 +26,7 @@ DEALINGS IN THE SOFTWARE.
 
 import re
 import inspect
+import typing
 
 import discord
 
@@ -554,6 +555,9 @@ class _Greedy:
 
         if converter is str or converter is type(None) or converter is _Greedy:
             raise TypeError('Greedy[%s] is invalid.' % converter.__name__)
+
+        if getattr(converter, '__origin__', None) is typing.Union and type(None) in converter.__args__:
+            raise TypeError('Greedy[Optional] is invalid.')
 
         return self.__class__(converter=converter)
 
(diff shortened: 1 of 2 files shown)


discord/ext/commands/converter.py
```diff
@@ -555,6 +556,9 @@ def __getitem__(self, params):
         if converter is str or converter is type(None) or converter is _Greedy:
             raise TypeError('Greedy[%s] is invalid.' % converter.__name__)
 
+        if getattr(converter, '__origin__', None) is typing.Union and type(None) in converter.__args__:
+            raise TypeError('Greedy[Optional] is invalid.')
```
Rapptz, line 560:
```suggestion
            raise TypeError('Greedy[%r] is invalid.' % converter)
```

PR #5069, opened 2020-06-28
implement Member.has_role

### LINKED ISSUE #2250: A Member.has_role method for speeded up checking if a member has a role
### The Problem

I want to check whether a member has a role. This is currently a little bit slow, as it involves accessing the `roles` property which could call `Guild.get_role` many times and using `list.__contains__` for the actual check. This would become faster.

### The Ideal Solution

A `Member.has_role` method that accepts integer snowflakes as well as `Role` instances would be nice. It would then use SnowflakeList.has.

### The Current Solution

I would have to get all roles of the member, which involves many dictionary accesses, and then many calls to `Role.__eq__`.

### Summary

Speedup for finding out if a member has a role

SebbyLaw:
### Summary

Implement Member.has_role, closes #2250 

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/member.py
+++ b/discord/member.py
@@ -382,6 +382,19 @@ class Member(discord.abc.Messageable, _BaseUser):
         if self.activities:
             return self.activities[0]
 
+    def has_role(self, role):
+        """
+        Checks if a member has the specified role.
+
+        .. versionadded:: 1.4
+
+        Parameters
+        -----------
+        role: :class:`Role`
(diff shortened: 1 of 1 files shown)


Rapptz:
This isn't really an approach I want to take as noted in another issue https://github.com/Rapptz/discord.py/pull/2243#issuecomment-504798984

PR #5088, opened 2020-07-01
Expose a way to call `Update Voice State` gateway command

### LINKED ISSUE #4215: Add a way to change `self_deaf` and `self_mute` in client's voice state
### The Problem

I would want the bot to be able to self-mute/self-deaf itself without using the guild mute/deaf.

### The Ideal Solution

I don't have anything specific in mind, a method that has `self_deaf` and `self_mute` kwargs is what I have in mind, but I don't really have a specific idea on which object it should be put. I'm guessing it would make some sense to be on `Guild` object since it's per-guild state but it doesn't really matter for me as long as it's there somewhere.

Edit: Considering we have info about current voice state, whatever the way to do will be, it should only require guild object (either by being method on guild object or by requiring guild parameter if it were to be a method on Client/ClientUser)

### The Current Solution

I can only use d.py's undocumented function: `client.ws.voice_state()`

### Summary

Add a method for self-mute/self-deafening the bot through [Update Voice State](https://discord.com/developers/docs/topics/gateway#update-voice-state)

jack1142:
### Summary

This PR adds a `Guild.change_voice_state()` method which exposes [Update Voice State](https://discord.com/developers/docs/topics/gateway#update-voice-state) command in the main gateway.

I decided to just make the PR and people can bikeshed it now with at least having *some* implementation.

1. Although I *can* get the `channel` from `guild.me.voice`, I figured that it can also be useful to connect the bot without using the `VoiceChannel.connect()` method which also creates a `VoiceClient` and establishes the connection with voice server. If you want to, I can add some sentinel as default value for `channel` parameter, so that d.py users could also use this method without passing the channel. (example for that [here](https://github.com/jack1142/discord.py/commit/96dfe1f87dde48bc1bcead24cda2739126429f5a))
2. I called this method `change_voice_state()` to make it consistent with `change_presence()` but other names could be used as well.
3. This method could possibly be put on a different object too, but I think this suits it rather well considering the data that this gateway command needs.

Fixes #4215 

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -2065,7 +2065,7 @@ class Guild(Hashable):
         limit = limit or 5
         return await self._state.query_members(self, query=query, limit=limit, user_ids=user_ids, cache=cache)
 
-    async def change_voice_state(self, *, self_mute, self_deaf):
+    async def change_voice_state(self, channel, *, self_mute, self_deaf):
         """|coro|
 
         Changes client's voice state in the guild.
@@ -2074,13 +2074,13 @@ class Guild(Hashable):
 
         Parameters
         -----------
+        channel: Optional[:class:`VoiceChannel`]
+            Channel the client wants to join. Use ``None`` to disconnect.
         self_mute: :class:`bool`
             Indicates if the client should be self-muted.
         self_deaf: :class:`bool`
             Indicates if the client should be self-deafened.
         """
         ws = self._state._get_websocket(self.id)
-        current_state = self._voice_state_for(self._state.self_id)
-        channel = current_state.channel if current_state else None
         channel_id = channel.id if channel else None
         await ws.voice_state(self.id, channel_id, self_mute, self_deaf)
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -2065,7 +2065,7 @@ class Guild(Hashable):
         limit = limit or 5
         return await self._state.query_members(self, query=query, limit=limit, user_ids=user_ids, cache=cache)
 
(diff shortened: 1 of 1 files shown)


RocketRace:
Should this return an optional VoiceClient?

jack1142:
@RocketRace no as this does not (nor should it) create a `VoiceClient` as explained in the description.

discord/guild.py
```diff
@@ -2064,3 +2064,23 @@ async def query_members(self, query=None, *, limit=5, user_ids=None, cache=True)
 
         limit = limit or 5
         return await self._state.query_members(self, query=query, limit=limit, user_ids=user_ids, cache=cache)
+
+    async def change_voice_state(self, channel, *, self_mute=False, self_deaf=False):
```
Rapptz, line 2068:
```suggestion
    async def change_voice_state(self, *, channel, self_mute=False, self_deaf=False):
```

jack1142: Since `channel` is required parameter, I thought that it might a little bit bothersome to have to use keyword arguments here; unless a sentinel is added that just uses the channel from current voice state?

jack1142:
> I'm unsure whether this should be here or inside `VoiceChannel`

I chose `Guild` over `VoiceChannel` because `channel` can be set to `None` too, but I was unsure where to put it too.

discord/guild.py
Rapptz: It's because `change_voice_state(None)` is a terrible read.

PR #5097, opened 2020-07-04
[docs] Update AuditLogDiff.type to be accurate

Artemis21:
### Summary

The documentation for `AuditLogDiff.type` is currently inaccurate and incomplete. This pull request fixes that, using data from [the Discord Developer Docs](https://discord.com/developers/docs/resources/channel#channel-object-channel-types).

### Checklist

- [x] If code changes were made then they have been tested (none have been made).
- [x] This PR fixes an issue.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -2170,8 +2170,23 @@ this goal, it must make use of a couple of data classes that aid in this goal.
 
         The type of channel or channel permission overwrite.
 
-        If the type is an :class:`int`, then it is a type of channel which can be either
-        ``0`` to indicate a text channel or ``1`` to indicate a voice channel.
+        If the type is an :class:`int`, then it is one of the following:
+
+        +---+---------------------------------+
+        | 0 | A guild text channel.           |
+        +===+=================================+
+        | 1 | A direct message channel.       |
+        +---+---------------------------------+
+        | 2 | A guild voice channel.          |
+        +---+---------------------------------+
+        | 3 | A group direct message channel. |
+        +---+---------------------------------+
+        | 4 | A guild category channel.       |
+        +---+---------------------------------+
+        | 5 | A guild news channel.           |
+        +---+---------------------------------+
+        | 6 | A guild store channel.          |
+        +---+---------------------------------+
 
         If the type is a :class:`str`, then it is a type of permission overwrite which
         can be either ``'role'`` or ``'member'``.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -2172,21 +2172,23 @@ this goal, it must make use of a couple of data classes that aid in this goal.
 
         If the type is an :class:`int`, then it is one of the following:
 
-        +-----+---------------------------------+
-        | `0` | A guild text channel.           |
-        +-----+---------------------------------+
-        | `1` | A direct message channel.       |
-        +-----+---------------------------------+
-        | `2` | A guild voice channel.          |
-        +-----+---------------------------------+
-        | `3` | A group direct message channel. |
-        +-----+---------------------------------+
-        | `4` | A guild category channel.       |
-        +-----+---------------------------------+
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -2170,8 +2170,23 @@ this goal, it must make use of a couple of data classes that aid in this goal.
 
         The type of channel or channel permission overwrite.
 
-        If the type is an :class:`int`, then it is a type of channel which can be either
-        ``0`` to indicate a text channel or ``1`` to indicate a voice channel.
+        If the type is an :class:`int`, then it is one of the following:
+
+        +---+---------------------------------+
+        | 0 | A guild text channel.           |
```
Gobot1234, line 2176: I think these digits should still be wrapped in ``

docs/api.rst
```diff
@@ -2170,8 +2170,23 @@ this goal, it must make use of a couple of data classes that aid in this goal.
 
         The type of channel or channel permission overwrite.
 
-        If the type is an :class:`int`, then it is a type of channel which can be either
-        ``0`` to indicate a text channel or ``1`` to indicate a voice channel.
+        If the type is an :class:`int`, then it is one of the following:
+
+        +---+---------------------------------+
```
NCPlayz, line 2175:
```suggestion
        +-------+---------------------------------+
        | value |     meaning                     |
        +-------+---------------------------------+
```
seems to make sense to have a table heading similar to this.

docs/api.rst
Artemis21: Sure, done

docs/api.rst
Artemis21: Accidentally outdated that but I will add.

docs/api.rst
```diff
@@ -2170,8 +2170,23 @@ this goal, it must make use of a couple of data classes that aid in this goal.
 
         The type of channel or channel permission overwrite.
 
-        If the type is an :class:`int`, then it is a type of channel which can be either
-        ``0`` to indicate a text channel or ``1`` to indicate a voice channel.
+        If the type is an :class:`int`, then it is one of the following:
+
+        +-----+---------------------------------+
+        | `0` | A guild text channel.           |
```
Gobot1234, line 2176: I meant like \``0``

docs/api.rst
Artemis21: Ok done, and as a header row too

docs/api.rst
Artemis21: Oh right sorry I am not familiar with rst

Rapptz:
This is already encapsulated and referenced under the ``ChannelType`` enum so re-typing it there feels... duplication heavy. I don't really like this approach because of it.

Artemis21:
> This is already encapsulated and referenced under the `ChannelType` enum so re-typing it there feels... duplication heavy. I don't really like this approach because of it.

True, though the documentation enum doesn't include the numbers, which are necessary. I suppose it would be nice to cast `AuditLogDiff.type` to an enum in the library, but that would be a breaking change.

Rapptz:
Well the easiest way would be to reference the numbers inside the enum documentation and then reference the enum within the AuditLogDiff.type

Artemis21:
The someone harder, but nicer (and still backwards compatible) way would be to cast `AuditLogDiff.type` to an `IntEnum` or equivalent.

Rapptz:
I don't use the `enum` module so that suggestion is out regardless.

Artemis21:
So I see, but an equivalent could work I think, by adding a `__int__` method around line 60 of `enums.py`...

Artemis21:
Actually, I think that would be too much of a breaking change - since the value can currently be either an `int` or a `str`, people will likely be checking if `isinstance(AuditLogDiff.type, int)`, which would be `False` - for it not to be breaking, it would have to actually inherit from `int` like `IntEnum` does.

PR #5102, opened 2020-07-07
Documentation Formatting

SkezzaB:
### Summary

Documentation formatting

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/iterators.py
+++ b/discord/iterators.py
@@ -173,16 +173,16 @@ class HistoryIterator(_AsyncIterator):
     """Iterator for receiving a channel's message history.
 
     The messages endpoint has two behaviours we care about here:
-    If `before` is specified, the messages endpoint returns the `limit`
-    newest messages before `before`, sorted with newest first. For filling over
-    100 messages, update the `before` parameter to the oldest message received.
+    If ``before`` is specified, the messages endpoint returns the `limit`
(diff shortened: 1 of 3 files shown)


discord/iterators.py
```diff
@@ -182,7 +182,7 @@ class HistoryIterator(_AsyncIterator):
     `after` parameter to the newest message received. If messages are not
     reversed, they will be out of order (99-0, 199-100, so on)
 
-    A note that if both before and after are specified, before is ignored by the
+    A note that if both `before` and `after` are specified, `before` is ignored by the
```
Gobot1234, line 185: This should be double base ticked right?

discord/iterators.py
```diff
@@ -471,7 +471,7 @@ class GuildIterator(_AsyncIterator):
     parameter to the newest guild received, If guilds are not reversed, they
     will be out of order (99-0, 199-100, so on)
 
-    Not that if both before and after are specified, before is ignored by the
+    Not that if both `before` and `after` are specified, `before` is ignored by the
```
Gobot1234, line 474: Same with this

discord/iterators.py
SkezzaB: Fixed in the latest commit

discord/iterators.py
SkezzaB: This too

PR #5107, opened 2020-07-08
Feature: Enable overriding default converters

james7132:
### Summary

Allows developers to globally alter default conversion behavior instead of hard-coding references to specific converters as the defaults. This is useful for making globally consistent conversion behavior that doesn't require importing a converter into every instance where the converter is required. An simple use case might be as parsing timedeltas using `pytimeparse` globally. A more a advanced use case might be changing the `discord.Member` converter to use a custom converter that uses `guild.query_members`  when the user is not in the immediate member cache, a potentially common use case for large bots with `guild_subscriptions` disabled.

### Checklist

- [ ] This PR fixes an issue.

I haven't directly tested these changes yet, but will soon. I might have missed a few points in the documentation.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -95,7 +95,8 @@ class _DefaultRepr:
 _default = _DefaultRepr()
 
 class BotBase(GroupMixin):
-    def __init__(self, command_prefix, help_command=_default, description=None, **options):
+    def __init__(self, command_prefix, help_command=_default, description=None,
+                 default_converters={}, **options):
         super().__init__(**options)
         self.command_prefix = command_prefix
         self.extra_events = {}
@@ -106,7 +107,9 @@ class BotBase(GroupMixin):
         self._before_invoke = None
         self._after_invoke = None
         self._help_command = None
-        self.default_converters = converters.DEFAULT_DISCORD_CONVERTERS.copy()
+        self._default_converters = {
+            **converters.DEFAULT_DISCORD_CONVERTERS.copy(), **default_converters
+        }
         self.description = inspect.cleandoc(description) if description else ''
         self.owner_id = options.get('owner_id')
         self.owner_ids = options.get('owner_ids', set())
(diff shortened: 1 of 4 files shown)


Vexs:
A concept like this has come up a lot in the discord- with some changes, it's pretty good to my mind. However, I've a number of issues with this particular implementation:

This change's capabilities far exceed it's naming- fundamentally, this allows you to link any class to a converter, which would be a far more used/useful capability than what it's currently named/intended for.

The user-exposed API is just a dict- this doesn't match anything in the library, and is a fairly poor interface in general for a lib. A better system might be to use decorators, a possible interface being:
```py
@bot.converter(class=someclass)
async def convert(......
```

james7132:
I'm not too fond of using decorators for something like this. I've changed it to use a constructor keyword argument instead with the intention that it the mapping be generally static throughout the lifetime of the client. Is this better?

Any suggestions for a better name for the parameter?

Vexs:
It's not unlikely for converters to be changed or added during the course of the bot's lifecycle as cogs are unloaded/loaded; This is a step backwards.
The only real option I see aside from exposing the dict directly is using a method (`bot.add/remove_converter(class, converter)`) which can tidily be wrapped in a decorator for ease of use.

james7132:
Thanks for bringing up cogs, I didn't think that far. However, I'm not sure how this feature interacts with cogs. My first intuition when I see such a declaration in a cog is that it's scoped only to that cog, whereas this current implementation is global. This would notably complicate the implementation and opens up multiple questions like "If two cogs implement the converter for the same type, and converters are global, which is used?" or "How do we ensure A's converter for a type is restored when B's converter for the same type is removed?". These questions and implementation, I think, are outside the scope of this PR, and should be addressed separately.

I kept the keyword argument for convenience, but added the client level add/remove methods and the decorator.

Rapptz:
I plan on doing a global type lookup kind of deal in a breaking version of the library near the future but not right now.

PR #5138, opened 2020-07-18
Use a subclass rather than monkey-patching for `ClientWebSocketResponse.close()`

jack1142:
### Summary

This PR makes use of `ws_response_class` argument in `ClientSession` so that d.py doesn't need to monkey-patch `ClientWebSocketResponse.close()` method.

### Checklist

- [ ] This PR fixes an issue.


jack1142:
I rebased the branch, it should be all good to go now.

PR #5167, opened 2020-07-29
Raise error when a cog with the same name is already registered

SebbyLaw:
# Summary

Currently, `Bot.add_cog` overwrites any existing cog in the internal cogs dictionary. The previously loaded cog is *not* ejected, and the listeners/commands added will indefinitely be registered to the bot. This change raises an error when a user tries to load a cog with the same name.

Alternative idea: Instead of always raising an error, add a kwarg (`override`? `force`?) that allows an already loaded cog to be overwritten in the internal dictionary and ejected from the bot properly.

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -489,7 +489,7 @@ class BotBase(GroupMixin):
 
         A cog is a class that has its own event listeners and commands.
 
-        .. versionchanged:: 1.5
+        .. versionchanged:: 2.0
 
             :exc:`.ClientException` is raised when a cog with the same name
             is already loaded.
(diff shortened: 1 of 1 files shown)


tripulse:
Yes, adding up a `override` or `force` kwarg or even a method, would've been helpful.

SebbyLaw:
Updated PR. Requesting review again.

discord/ext/commands/bot.py
```diff
@@ -515,8 +520,10 @@ def add_cog(self, cog):
         cog_name = cog.__cog_name__
         existing = self.__cogs.get(cog_name)
 
-        if existing:
-            raise discord.ClientException('a cog with name {0.__cog_name__} is already loaded'.format(existing))
+        if existing is not None:
+            if not override:
+                raise discord.ClientException('a cog with name %s is already loaded' % cog_name)
```
Rapptz, line 525:
```suggestion
                raise discord.ClientException('Cog named %r already loaded' % cog_name)
```

PR #5171, opened 2020-07-30
Added is_guild_owner check

TheOneMusic:
- Added the `is_guild_owner` check, a basic check that is used quite a lot.
- Added the `NotGuildOwner` error. This error is raised if `is_guild_owner` fails.


Vexs:
This check has been willingly not included due to the actually-pretty-uncommon nature of it. It's fairly rare that someone wants *just* an owner check and not a permissions-or-owner, etc.

Rapptz:
This has been proposed and denied before in the discord.py guild for the reasons Vex has given.

PR #5175, opened 2020-07-31
Small feature on permissions checks

TheOneMusic:
### Summary

Added a small and non-breaking change: `MissingPermissions` and `BotMissingPermissions` now have a `on_guild` attribute, to specify if the check was done via a channel permissions lookup or a guild wild permissions lookup. This is something I found missing and is quite useful when using `on_command_error`. Example:

```python
import discord, asyncio

bot = commands.Bot("!")

@bot.event
async def on_command_error(ctx, error):
    if isinstance(error, commands.MissingPermissions):
        missing = ", ".join(str(m) for m in error.missing_perms)
        if error.on_guild:
            await ctx.send("Missing guild permissions: {}".format(missing)
            return
        await ctx.send("Missing channel permissions: {}. Please check channel overwrites.".format(missing))
```

Another idea could be to have two separate classes for guild permissions lookup (`MissingGuildPermissions` and `BotMissingGuildPermissions`) instead.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1718,7 +1718,7 @@ def has_permissions(**perms):
         if not missing:
             return True
 
-        raise MissingPermissions(missing)
+        raise MissingPermissions(missing, on_guild=False)
 
     return check(predicate)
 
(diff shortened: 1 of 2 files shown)


Vexs:
Disregarding the core concept, the current implementation doesn't work. You're not checking if it's guild or channel permissions and then setting the flag, as someone could be denied channel permissions from guild (role) permissions and not overrides.
Additionally, on_guild is a terribly non-descriptive name.

Issue #5174, opened 2020-07-31
Change documentation to be more upfront about what uses cache and make use of API calls in type hints


asleep-cult:
### The Problem

Cache rarely being mentioned in the documentation creates ambiguity as to what's acceptable in getting discord objects, for example get_user uses cache which requires the user to have a mutual guild with the bot for it to be cached in the first place but nowhere in the documentation is this mentioned. The discord.User type hint also seemingly uses get_user and never fetch_user which means the user must be in in a mutual guild even though it is simply a user and not a member! This can get very confusing at times. 

### The Ideal Solution

The documentation for get could be changed to mention that it uses cache and that a mutual guild is required to use it and the discord.<user/member etc..> type hints could also be changed to make use of fetch (try looking through cache then trying to use fetch) rather than solely depending on cache .

### Summary

Change documentation to be more upfront about what uses cache and make use of API calls in discord.<user/member etc...> type hints when needed

numbermaniac:
For what it's worth, methods were standardised a while ago so that all `get_` methods rely on the cache whereas all `fetch_` methods call the API. Additionally, if `await` appears before a method name in the documentation then that can sometimes indicate that the method relies on an API call.

hamza1311:
Every method that makes an API call with an equivalent to get from cache mentions it as note. Everything that doesn't mention it uses the cache.
Example: 
![image](https://user-images.githubusercontent.com/47357913/89036292-dd7f2680-d355-11ea-973a-fdad1253bdf1.png)
from https://discordpy.readthedocs.io/en/latest/api.html#discord.Client.fetch_guilds

> The discord.User type hint also seemingly uses get_user and never fetch_user which means the user must be in in a mutual guild even though it is simply a user and not a member! This can get very confusing at times.

This is explicitly mentioned in the [docs](https://discordpy.readthedocs.io/en/latest/ext/commands/api.html#discord.ext.commands.UserConverter).
![image](https://user-images.githubusercontent.com/47357913/89036719-9cd3dd00-d356-11ea-8562-0221dfb2237a.png)

If you want to fetch the user, the solution to that has always been making your own converter

Vexs:
While the information is there, it's scattered all over the place with no centralized explanation. Considering the importance of the cache, and the recent reasons why things might not be in there, it seems wise to me to add some narrative docs on it.
Remember, documentation isn't just the API doc.

Rapptz:
[...]

PR #5178, opened 2020-08-01
API calls in default converters and more emphasis on what get does in documentation

asleep-cult:
### Summary

(reopening accidentally overwrote other branch don't ask)
I'm creating this PR on issue #5174 in hopes to give the default converters more ease of use (i've added an API call as a last resort for the discord.ext.commands converters because it doesn't make sense not to have them in my opinion although i'm not sure how others feel about it) I've also added a note next to get_user that states the following "This method uses cache meaning a mutual guild is required see fetch_user" to try to remain consistent with the fetch_user documentation... i'd hope to see this throughout the documentation for all get methods... didn't add it myself because i'm not too familiar with the formatting.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -51,6 +51,7 @@ __all__ = (
     'Greedy',
 )
 
+
 def _get_from_guilds(bot, getter, argument):
     result = None
     for guild in bot.guilds:
@@ -59,7 +60,9 @@ def _get_from_guilds(bot, getter, argument):
             return result
     return result
 
+
 _utils_get = discord.utils.get
+
 
 class Converter:
     """The base class of custom converters that require the :class:`.Context`
@@ -87,7 +90,6 @@ class Converter:
             The invocation context that the argument is being used in.
         argument: :class:`str`
             The argument that is being converted.
-
         Raises
         -------
         :exc:`.CommandError`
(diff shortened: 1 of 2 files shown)


Rapptz:
A few things to note here.

For starters, generally a PR should be limited in scope and this one's scope is a bit too high. Likewise, automatically doing linting formatting tends to get a PR rejected.

Now, aside from those minor nits there is actually a reason why the built-in converters historically didn't use API calls and that's because the rate limits for those endpoints are harsh and a user spamming these can consequently end up causing a bot to be rate limited if done in a naive way. I wanted the built-in converters to be as simple as possible for the most common use case without having too many foot-guns. There's also the element of consistency where, for example, a Member converter would only work with `Foo#1234` sometimes or with nicknames sometimes which would cause some issues. There is no API endpoint to fetch from nickname or username + discriminator, though there is a gateway call to do this which was exposed in 1.3.

Discord however has moved away from asking people to cache things. This causes a disconnect in the previous design and the current one though the things I am talking about up there as issues has not been resolved. A user who wants more control over this type of stuff has historically been told to provide their own converter -- and the reason why these converters are exposed the way they are is precisely to allow for ease of extension specifically for use-cases like this one.

Should this design decision be re-considered in the future is up in the air but for now I don't see this happening any time soon in the 1.x release line, so I'll have to reject this PR. Apologies for the trouble and I hope the reasons I provide at least make sense.

PR #5181, opened 2020-08-02
Check if self.guild is object as well as not None

tnoff:
### Summary

Catch AttributeError exception on user reactions iterator

It seems that sometimes the guild is a simple object file, and doesnt have get_member

File "/home/user/.envs/discord/lib/python3.6/site-packages/discord/iterators.py", line 166, in fill_users
    member = self.guild.get_member(member_id)
AttributeError: 'Object' object has no attribute 'get_member'

Add generic try/except to catch this case, and set member to zero.

Tested in my app and fixes issue

### Checklist

- [X ] If code changes were made then they have been tested.
- [X] This PR fixes an issue.

### DIFF
--- a/discord/iterators.py
+++ b/discord/iterators.py
@@ -157,7 +157,7 @@ class ReactionIterator(_AsyncIterator):
                 self.limit -= retrieve
                 self.after = Object(id=int(data[-1]['id']))
 
-            if self.guild is None:
+            if self.guild is None or isinstance(self.guild, Object):
                 for element in reversed(data):
                     await self.users.put(User(state=self.state, data=element))
             else:


Rapptz:
Not sure I like this solution. In what case was the `guild` an `discord.Object`?

tnoff:
Not too sure whats happening, this is what my code looks like if it helps

```
channel_id = '1234' # channel id
message_id = '234' # message id
channel = await client.fetch_channel(channel_id)
message = await channel.fetch_message(message_id)
for reaction in message.reactions:
    print(reaction)
    async for user in reaction.users():
        print(user)
```

If I add the try/except block locally it does work afterward

Rapptz:
I'm unsure why you're using `fetch_` instead of the cache lookup `get_` but I see why at least. The `fetch_` versions don't come with the full info that's being required here. However this isn't the proper solution for this issue.

discord/iterators.py
```diff
@@ -163,7 +163,10 @@ async def fill_users(self):
             else:
                 for element in reversed(data):
                     member_id = int(element['id'])
-                    member = self.guild.get_member(member_id)
+                    try:
+                        member = self.guild.get_member(member_id)
+                    except AttributeError:
+                        member = None
```
Rapptz, line 169:
```suggestion
                    member = self.guild.get_member(member_id)
```

Then in line 160 instead of `if self.guild is None` change it to `if self.guild is None or isinstance(self.guild, Object):`

tnoff:
Thanks!

PR #5184, opened 2020-08-03
Fix Member.joined_at documentation

### LINKED ISSUE #5183: Member.joined_at is incorrectly documented
Hello!

### Summary

The documentation for the `Member.joined_at` attribute appears to claim that it gives the timestamp of when the member joined the guild *for the first time*.

https://github.com/Rapptz/discord.py/blob/ae1ec8072f9ef26acec107824eed42fdbb897cc2/discord/member.py#L150-L152

This appears to be untrue; it looks like the attribute gives the timestamp of when the user in question has joined *most recently*.

Here's a short snippet to demonstrate that I can make it give distinct values for the same guild & users id combination.

image

The Discord API docs only annotate the attribute as ["when the user joined the guild"](https://discord.com/developers/docs/resources/guild#guild-member-object-guild-member-structure).

### Duplicates

When searching for duplicates I only found this old PR addressing this issue that got closed: #882.

I'd be happy to open a PR for this.

I hope it's ok that I didn't follow the markdown template, I felt it wasn't fitting for this kind of issue.

rodovia:
### Summary

Fixes the [Member.joined_at](https://discordpy.readthedocs.io/en/latest/api.html#discord.Member.joined_at) documentation.

Also, This PR closes #5183, and this is a potential duplicate of #882.

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/member.py
+++ b/discord/member.py
@@ -149,7 +149,7 @@ class Member(discord.abc.Messageable, _BaseUser):
     ----------
     joined_at: Optional[:class:`datetime.datetime`]
         A datetime object that specifies the date and time in UTC that the member joined the guild for
-        the first time. In certain cases, this can be ``None``.
+        the first time. If the member left and rejoined the guild, will be the latest date. In certain cases, this can be ``None``.
     activities: Tuple[Union[:class:`BaseActivity`, :class:`Spotify`]]
         The activities that the user is currently doing.
     guild: :class:`Guild`
--- a/discord/member.py
+++ b/discord/member.py
@@ -148,8 +148,8 @@ class Member(discord.abc.Messageable, _BaseUser):
     Attributes
     ----------
     joined_at: Optional[:class:`datetime.datetime`]
-        A datetime object that specifies the date and time in UTC that the member joined the guild for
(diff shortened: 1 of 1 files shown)


discord/member.py
```diff
@@ -149,7 +149,7 @@ class Member(discord.abc.Messageable, _BaseUser):
     ----------
     joined_at: Optional[:class:`datetime.datetime`]
         A datetime object that specifies the date and time in UTC that the member joined the guild for
-        the first time. In certain cases, this can be ``None``.
+        the first time. If the member left and rejoined the guild, this will be the latest date. In certain cases, this can be ``None``.
```
Gorialis, line 152: Adding this statement makes the docstring conflict with itself - either it's the time the member joined for the first time (it isn't), or it's the most recent join.

It would make more sense to just remove anything related to 'the first time the member joins' - since I'm not even sure if Discord themselves records such information.

Gorialis (changes requested):

discord/member.py
rodovia: oof, you are right. also, already fixed it.

PR #5188, opened 2020-08-04
[Bugfix]Remove 443 port of endpoint

jun50:
### Summary

I get the following error when Bot join a voice channel in the Japan region.
```
Task exception was never retrieved
future: <Task finished coro=<VoiceClient._create_socket() done, defined at D:\Program Files\Python37\lib\site-packages\discord\voice_c
  File "D:\Program Files\Python37\lib\site-packages\discord\voice_client.py", line 191, in _create_socket
    self.endpoint_ip = socket.gethostbyname(self.endpoint)
socket.gaierror: [Errno 11001] getaddrinfo failed
```
This Pull Request is to fix this bug.

### Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/voice_client.py
+++ b/discord/voice_client.py
@@ -187,7 +187,7 @@ class VoiceClient:
                         'If timeout occurred considering raising the timeout and reconnecting.')
             return
 
-        self.endpoint = endpoint.replace(':80', '')
+        self.endpoint = endpoint.replace(':80', '').replace(':443', '')
         self.endpoint_ip = socket.gethostbyname(self.endpoint)
 
         if self.socket:


discord/voice_client.py
```diff
@@ -187,7 +187,7 @@ async def _create_socket(self, server_id, data):
                         'If timeout occurred considering raising the timeout and reconnecting.')
             return
 
-        self.endpoint = endpoint.replace(':80', '')
+        self.endpoint = endpoint.replace(':80', '').replace(':443', '')
```
TurnrDev, line 190: Surely `endpoint.split(':')[0]` is more pythonic?

Rapptz, line 190:
```suggestion
        (self.endpoint, _, _) = endpoint.rpartition(':')
```

I think this breaks if `wss://` ever gets appended but if this happens there are bigger issues so no point in worrying about it.

iomintz: .split allocates a list. `.rpartition(':')[0]` would be more efficient (and support ipv6 hostnames with :80 in their addresses)

Rapptz:
I ended up fixing this differently because `getaddrinfo` shouldn't have been called at all but the previous PR addressing this did not touch that line.

Thanks for the PR though!

noblockhit:
I have a last question: Why does this error uccur at all and why does it send the 443 or 80 anyway if it gets filtered out anyway?! (or is that just convention!?)

discord/voice_client.py
noblockhit: But split would remove the error forever doesn't it? and what happens if this error occurs again but with 81 as an example?

Harmon758: This has already been fixed with https://github.com/Rapptz/discord.py/commit/a9824ffabf0b2c8031ac75245650711145aaa26c in v1.4.0 and newer.

PR #5195, opened 2020-08-05
[commands] Provide a way to retrieve time left for a cooldown

Phxntxm:
### Summary

Provides a way to get the current time left on a cooldown. Not sure on the name, sounds eh to me but couldn't come up with something better... also questioning what should be returned if the cooldown has not started yet (tokens left > 0), `None`? `self.per`? I went with `None` here but I could be convinced otherwise

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/cooldowns.py
+++ b/discord/ext/commands/cooldowns.py
@@ -91,6 +91,15 @@ class Cooldown:
             tokens = self.rate
         return tokens
 
+    def get_retry_after(self, current=None):
+        current = current or time.time()
+        tokens = self.get_tokens(current)
+
+        if tokens == 0:
+            return self.per - (current - self._window)
+
+        return 0.0
+
     def update_rate_limit(self, current=None):
         current = current or time.time()
         self._last = current
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -805,7 +805,8 @@ class Command(_BaseCommand):
             return False
 
         bucket = self._buckets.get_bucket(ctx.message)
-        return bucket.get_tokens() == 0
+        current = ctx.message.created_at.replace(tzinfo=datetime.timezone.utc).timestamp()
+        return bucket.get_tokens(current) == 0
 
     def reset_cooldown(self, ctx):
         """Resets the cooldown on this command.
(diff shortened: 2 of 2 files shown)


discord/ext/commands/cooldowns.py
```diff
@@ -91,6 +91,13 @@ def get_tokens(self, current=None):
             tokens = self.rate
         return tokens
 
+    def get_seconds_left(self, current=None):
+        current = current or time.time()
+        tokens = self.get_tokens(current)
+
+        if tokens == 0:
+            return self.per - (current - self._window)
+
```
Rapptz, line 100:
```suggestion
        return 0.0

```

discord/ext/commands/cooldowns.py
```diff
@@ -91,6 +91,13 @@ def get_tokens(self, current=None):
             tokens = self.rate
         return tokens
 
+    def get_seconds_left(self, current=None):
```
Rapptz, line 94:
```suggestion
    def get_retry_after(self, current=None):
```

discord/ext/commands/core.py
```diff
@@ -819,6 +819,23 @@ def reset_cooldown(self, ctx):
             bucket = self._buckets.get_bucket(ctx.message)
             bucket.reset()
 
+    def get_cooldown_seconds_left(self, ctx):
```
Rapptz, line 822:
```suggestion
    def get_cooldown_retry_after(self, ctx):
```

discord/ext/commands/core.py
```diff
@@ -819,6 +819,23 @@ def reset_cooldown(self, ctx):
...
+        """Retrieves the amount of seconds before this command can be tried again.
+
+        Parameters
+        -----------
+        ctx: :class:`.Context`
+            The invocation context to retrieve the cooldown from.
+
+        Returns
+        --------
+        Optional[:class:`float`]
```
Rapptz, line 832:
```suggestion
        :class:`float`
```

discord/ext/commands/core.py
```diff
@@ -819,6 +819,23 @@ def reset_cooldown(self, ctx):
             bucket = self._buckets.get_bucket(ctx.message)
             bucket.reset()
 
+    def get_cooldown_seconds_left(self, ctx):
+        """Retrieves the amount of seconds before this command can be tried again.
+
```
Rapptz, line 824:
```suggestion

        .. versionadded:: 1.4
```

discord/ext/commands/core.py
```diff
@@ -819,6 +819,23 @@ def reset_cooldown(self, ctx):
...
+            The invocation context to retrieve the cooldown from.
+
+        Returns
+        --------
+        Optional[:class:`float`]
+            The amount of time left on this command's cooldown in seconds
+        """
+        if self._buckets.valid:
+            bucket = self._buckets.get_bucket(ctx.message)
+            return bucket.get_seconds_left()
```
Rapptz, line 837:
```suggestion
            current = ctx.message.created_at.replace(tzinfo=datetime.timezone.utc).timestamp()
            return bucket.get_retry_after(current)
```

This made me realize that `is_on_cooldown` doesn't have this done either due to oversight. Would be good if this could be done as well there.

discord/ext/commands/core.py
```diff
@@ -819,6 +819,23 @@ def reset_cooldown(self, ctx):
...
+
+        Parameters
+        -----------
+        ctx: :class:`.Context`
+            The invocation context to retrieve the cooldown from.
+
+        Returns
+        --------
+        Optional[:class:`float`]
+            The amount of time left on this command's cooldown in seconds
```
Rapptz, line 833:
```suggestion
            The amount of time left on this command's cooldown in seconds. If this is ``0.0`` then the command isn't on cooldown.
```

discord/ext/commands/core.py
```diff
@@ -819,6 +819,23 @@ def reset_cooldown(self, ctx):
...
+        """
+        if self._buckets.valid:
```
Rapptz, line 835: This function should probably return `0.0` in the other branch as well.

Phxntxm:
No idea how to finish changes requested since I force pushed to keep it to one commit, but the requested change were implemented, including the fix for `is_on_cooldown`

Issue #2525, opened 2020-01-23
Improving examples


Rapptz:
Right now, we don't have a lot of examples that show how to work with various parts of the library. I'm hoping in the near future to revamp examples to be a bit more useful. Part of the on-going work in doing so is to document the current examples (or maybe even rewrite them) significantly through commenting most parts of the code as if someone is explaining it to you over your shoulder.

Hopefully by commenting the code extensively the examples are more helpful for those who learn primarily through examples.

Likewise, I want to add examples that showcase certain things:

- [ ] Subcommands
- [ ] Various types of converters
- [ ] Basic reaction roles
- [ ] HelpCommand overriding
- [ ] Cogs (not extensions)
- [ ] Extensions (using cogs)
- [ ] Custom contexts
- [ ] Working with permissions (reading and writing, e.g. roles)
- [ ] Working with permission overwrites (channel overwrites etc)
- [ ] How to make a secret channel
- [ ] Error handling via ext.commands
- [ ] How to use AutoShardedBot/AutoShardedClient
- [ ] An example showcasing how to use the task extension

More example suggestions are welcome. I also would want the previous examples to be heavily commented as well. We can draw inspiration from [serenity's examples](https://github.com/serenity-rs/serenity/tree/current/examples/).

One thing that remains is whether these examples should be linear, similar to a book with chapters where we build upon knowledge on the previous examples. That way we don't have to overly comment everything if it's been covered in a previous example.

XuaTheGrate:
Custom context example done at #2526

PendragonLore:
Error handler example done at #2529

Snaptraks:
Subcommands example done at #2530

Rapptz:
I should mention, for any future person who wants to offer help that although I appreciate the help given I would really appreciate if the contents of the examples were discussed beforehand in [the official discord.py server](https://discord.gg/dpy) so we can iterate through it together.

NCPlayz:
How to make a secret channel example done at #6448

nmsturcke:
#7963 Cogs + Extensions

ddjerqq:
I want to contribute to the `Cogs (not extensions)` and `Extensions (using cogs)`. make examples of dependency injection in cogs and so on.

pixeldeee:
I maked `Cogs (not extensions)` example in this pull - #9157

PR #5214, opened 2020-08-07
Add channel management example

JJgar2725:
### Summary

This adds a simple channel management example that showcases how to make open and private text / voice channels (as well as some usage with subcommands) to the repository under the examples folder. After reiterating over the example on the Discord server, I made edits and fixes as requested. If there are anymore changes that seem necessary, I'd be happy to edit it and get it fixed. 

Issue: #2525 

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/secret_channel.py
+++ b/examples/secret_channel.py
@@ -1,63 +1,27 @@
-# import discord and any other libraries
 import discord
 from discord.ext import commands
-from discord.utils import get
 
-# bot description
-description = "A simple example showcasing how to make text and voice channels (as well as private ones)."
+description = "A simple example showcasing how to make a secret text channel."
 
-# create a bot instance with prefix "?" and description
 bot = commands.Bot(command_prefix="?", description=description)
 
 @bot.event
 async def on_ready():
-    """Prints information to the console once the bot is ready."""
-
     print('Logged in as')
     print(bot.user.name)
     print(bot.user.id)
     print('Bot Online!')
 
-def can_create():
-    """A check to see if the user has the required manage_channels permission."""
-    
-    async def predicate(ctx):
-        member = ctx.author
-        return member.guild_permissions.manage_channels
-    return commands.check(predicate)
+@bot.command()
+@commands.has_permissions(manage_channels=True)
+async def create(ctx, *, channel_name: str):
+    """Command to create a private text channel."""
 
-@bot.group(invoke_without_command=True)
-@can_create()
-async def create(ctx, *, new_channel: str):
-    """Command to create an open text channel."""
-
-    # the passed argument new_channel is used to name the created channel
-    channel = await ctx.guild.create_text_channel(new_channel, overwrites=None, category=None, reason=None)
-    await channel.send(f"Text channel {new_channel} was created!") # sends a message into that channel
-
-@create.command()
-@can_create()
-async def voice(ctx, *, new_channel: str):
-    """Subcommand to create an open voice channel."""
-
-    # note: ctx.guild.categories will return a list of possible categories that the server has
-    # you can then select where the channel will be created in the category list
-    channel = await ctx.guild.create_voice_channel(new_channel, category=ctx.guild.categories[1])
-    await ctx.send(f"Voice channel {new_channel} was created!") # will send into channel that command was invoked from
-
-@create.command()
-@can_create()
-async def priv(ctx, *, new_channel: str):
-    """Subcommand to create a private text channel."""
-
-    # use discord.utils.get() for retrieving and storing a role into variables
-    # guild.roles is an iterable, and name is an attribute to search for
-    admin = get(ctx.guild.roles, name="YOUR_ADMIN_ROLE")
-    mods = get(ctx.guild.roles, name="YOUR_MOD_ROLE")
+    # find needed roles and store them in variables for later use
+    admin = discord.utils.get(ctx.guild.roles, name="YOUR_ADMIN_ROLE")
+    mods = discord.utils.get(ctx.guild.roles, name="YOUR_MOD_ROLE")
 
     # using a dictionary, permissions can be chosen for the new channel
-    # guild.default_role is @everyone, guild.me is the bot itself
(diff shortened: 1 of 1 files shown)


examples/secret_channel.py
```diff
@@ -0,0 +1,106 @@
+# import discord and any other libraries
+import discord
+from discord.ext import commands
+from discord.utils import get
```
Rapptz, line 4:
```suggestion
```

examples/secret_channel.py
```diff
@@ -0,0 +1,106 @@
...
+
+@create.command()
+@can_create()
+async def priv(ctx, *, new_channel: str):
+    """Subcommand to create a private text channel."""
+
+    # use discord.utils.get() for retrieving and storing a role into variables
+    # guild.roles is an iterable, and name is an attribute to search for
+    admin = get(ctx.guild.roles, name="YOUR_ADMIN_ROLE")
+    mods = get(ctx.guild.roles, name="YOUR_MOD_ROLE")
```
Rapptz, line 56:
```suggestion
    admin = discord.utils.get(ctx.guild.roles, name="YOUR_ADMIN_ROLE")
    mods = discord.utils.get(ctx.guild.roles, name="YOUR_MOD_ROLE")
```

examples/secret_channel.py
```diff
@@ -0,0 +1,106 @@
...
+    channel = await ctx.guild.create_text_channel(new_channel, overwrites=overwrites)
+    await channel.send(f"Private text channel {new_channel} was created!")
+
+@create.command()
+@can_create()
+async def priv_voice(ctx, *, new_channel: str):
+    """Subcommand to create a private voice channel."""
+    
+    admin = get(ctx.guild.roles, name="YOUR_ADMIN_ROLE")
+    mods = get(ctx.guild.roles, name="YOUR_MOD_ROLE")
```
Rapptz, line 78:
```suggestion
    admin = discord.utils.get(ctx.guild.roles, name="YOUR_ADMIN_ROLE")
    mods = discord.utils.get(ctx.guild.roles, name="YOUR_MOD_ROLE")
```

examples/secret_channel.py
```diff
@@ -0,0 +1,106 @@
...
+    await ctx.send(f"Private voice channel {new_channel} was created!")
+
+@bot.command()
+@can_create()
+async def delete(ctx, *, channel_name: str):
+    """Command to delete a given channel."""
+
+    # using discord.utils.get() and bot.get_all_channels(), you can specify an attribute
+    # to search through an iterable, in this case all the channels on a guild
+    channel = get(bot.get_all_channels(), name=channel_name)
```
Rapptz, line 99: This seems exceedingly dangerous for an example.

examples/secret_channel.py
```diff
@@ -0,0 +1,106 @@
...
+
+    print('Logged in as')
+    print(bot.user.name)
+    print(bot.user.id)
+    print('Bot Online!')
+
+def can_create():
+    """A check to see if the user has the required manage_channels permission."""
+    
+    async def predicate(ctx):
```
Rapptz, line 24:
```suggestion
    def predicate(ctx):
```

examples/secret_channel.py
```diff
@@ -0,0 +1,106 @@
...
+@bot.event
+async def on_ready():
+    """Prints information to the console once the bot is ready."""
+
+    print('Logged in as')
+    print(bot.user.name)
+    print(bot.user.id)
+    print('Bot Online!')
+
+def can_create():
```
Rapptz, line 21: This entire check already comes built-in, it's `commands.has_permissions(manage_channels=True)`

examples/secret_channel.py
```diff
@@ -0,0 +1,106 @@
...
+    """A check to see if the user has the required manage_channels permission."""
+    
+    async def predicate(ctx):
+        member = ctx.author
+        return member.guild_permissions.manage_channels
+    return commands.check(predicate)
+
+@bot.group(invoke_without_command=True)
+@can_create()
+async def create(ctx, *, new_channel: str):
```
Rapptz, line 31: The signature of a command is user-facing as well. This means that when someone runs a help command on it they can see the signature and be pointed in the right direction. In this case the parameter should be `channel_name` not `new_channel`, since the command name already implies that it's creating something.

As an aside, I feel like the command name doesn't really make much sense on its own but I guess for an example there can be leeway here.

examples/secret_channel.py
```diff
@@ -0,0 +1,106 @@
...
+    """Command to create an open text channel."""
+
+    # the passed argument new_channel is used to name the created channel
+    channel = await ctx.guild.create_text_channel(new_channel, overwrites=None, category=None, reason=None)
```
Rapptz, line 35:
```suggestion
    channel = await ctx.guild.create_text_channel(new_channel)
```

Don't pass parameters that don't need to be passed.

examples/secret_channel.py
```diff
@@ -0,0 +1,106 @@
...
+    await channel.send(f"Text channel {new_channel} was created!") # sends a message into that channel
+
+@create.command()
+@can_create()
+async def voice(ctx, *, new_channel: str):
+    """Subcommand to create an open voice channel."""
+
+    # note: ctx.guild.categories will return a list of possible categories that the server has
+    # you can then select where the channel will be created in the category list
+    channel = await ctx.guild.create_voice_channel(new_channel, category=ctx.guild.categories[1])
```
Rapptz, line 45: This code will fail if a server has no categories. Likewise, it's a bit odd to just fetch the second one in the list out of the blue.

examples/secret_channel.py
Phxntxm: Additionally, your comment suggests `in this case all the channels on a guild` but this is not the case with your code, you are using `.get_all_channels` which returns all the channels on *all* of the guilds the bot can see

JJgar2725:
Alright, thanks for the review! I'll get to making those changes soon.

PR #5217, opened 2020-08-09
Meta: Update Bot.command() decorator docs

skairunner:
### Summary

Bot.command() returns Command, but it is not documented. Add that documentation.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1222,6 +1222,11 @@ class GroupMixin:
     def command(self, *args, **kwargs):
         """A shortcut decorator that invokes :func:`.command` and adds it to
         the internal command list via :meth:`~.GroupMixin.add_command`.
+
+        Returns
+        --------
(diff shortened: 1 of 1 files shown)


discord/ext/commands/core.py
```diff
@@ -1197,6 +1197,11 @@ def get_command(self, name):
     def command(self, *args, **kwargs):
         """A shortcut decorator that invokes :func:`.command` and adds it to
         the internal command list via :meth:`~.GroupMixin.add_command`.
+
+        Returns
+        --------
+        :class:`Command`
```
Gobot1234, line 1203: It would be Callable[..., :class:\`Command\`] not just Command

skairunner: Ah, mb.

skairunner:
I think this should be mergeable now!

PR #5747, opened 2020-08-22
[commands] Add cls parameter to process_commands method.

sairam4123:
### Summary

Adds a cls parameter to the process_commands method in bot.py
So, you don't need to make a custom process_commands to pass in a new context.

Here's an example:
```py
@bot.event
async def on_message(message: discord.Message):
    await bot.process_commands(message, cls=InvitesContext)
```

Before this PR:
```py
class Bot(commands.Bot):
    async def get_context(self, message, *, cls=Context):
        return await super().get_context(message, cls=cls)
```
I feel the Bot subclassing unnecessary as it is very easy just to add a cls parameter to the bot.process_commands

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -911,7 +911,7 @@ class BotBase(GroupMixin):
             exc = errors.CommandNotFound('Command "{}" is not found'.format(ctx.invoked_with))
             self.dispatch('command_error', ctx, exc)
 
-    async def process_commands(self, message):
+    async def process_commands(self, message, cls=Context):
         """|coro|
 
         This function processes the commands that have been registered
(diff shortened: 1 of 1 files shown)


ChillFish8:
Rapptz has already explained to you why this got closed in the first place, for the sake of not subclassing `Bot` (Which anyone dealing with custom context probably already has done) it adds a lot of potential for added confusing about A) What the code is doing and B) to anyone using process_commands. If you're at the point of needing custom Context you're probably already subclassing Bot and gain no benifit by using this over [the current system.](https://github.com/Rapptz/discord.py/blob/master/examples/custom_context.py#L25)

jack1142:
And this would also break people who are replacing `get_context()`'s default of `cls` since the default from `process_commands()` would suddenly be used instead.

sairam4123:
Well, then, I'll mark this as a breaking change.

Rapptz:
This PR isn't getting merged. That's why it's closed.

Issue #5759, opened 2020-08-27
Add to Message.reply


iNaruse:
### The Problem

We used to use | await message.channel.send (f"{message.author.ment}, message").
When you add REPLY, await message.reply("HELLO") -> @SADZ#0000 HELLO
(Translator)

### The Ideal Solution

I was writing Discord.js and came over to Discord.py.
It's more functional and more relaxing than DIscord.js.
But it was a bit disappointing because there was no replay function. If you add this function, it will make it easier to make examples like ping pong with one await message.reply.
(Translator)
### The Current Solution

await message.channel.send(f"{message.author.mention}, HELLO")
### Summary

Send a reply to the person who entered the command.
@INIRU#0001, HELLO | Like this.
(Translator)

Rapptz:
Trivial helpers like these don't really gain you much, likewise this functionality conflicts with the actual discord reply feature coming soon.

PR #5760, opened 2020-08-27
Add new Message.reply method

bijij:
### Summary

Adds a new method `reply` to both Messsage and Context

Unsure if discord's implementation is final though so this implementation may need to change.

### Checklist

- [ ] This PR fixes an issue.


ChillFish8:
Pretty sure this was ruled out in [#5759](https://github.com/Rapptz/discord.py/issues/5759#issuecomment-681609438)
That Coroutine also seems rather pointless in the fact you:
- Cant change the reply message itself.
- Could just write that as a small helper function without adding extra things to the class.
- Really doesnt give much if any functionality over just using string formatting originally.

TurnrDev:
Inspired by this PR and the reasons why this PR is declined, I'm working on a `quote` parameter. #5761 

My PR adds a parameter which just formats message content as a quote. It doesn't send a message, just returns a string to be used in another string. Because of this, it doesn't check if the quote is too long, I've left that up to the end-user/end developer.

bijij:
@ChillFish8 unsure if this is the case as this is the actual discord reply feature Danny was referring to. It's currently locked behind a an experiment toggle however.

AtieP:
Why do we really need this feature?

numbermaniac:
This is a new feature being added to Discord. It already exists in Canary.

discord/message.py
```diff
@@ -1073,3 +1073,35 @@ async def ack(self):
...
+        content: :class:`str`
+            The content of the message to send.
+        mention: :class:`bool`
+            Indicates if the reply should mention the message author, defaults to True.
+        **kwargs:
+            All keyword arguments for :meth:`~discord.abc.Messageable.send` can be used here.
+
+        Raises
+        --------
+        Exceptions raised by :meth:`~discord.abc.Messageable.send` are propegated by this function.
```
jack1142, line 1095: A typo.
```suggestion
        Exceptions raised by :meth:`~discord.abc.Messageable.send` are propagated by this function.
```

AtieP:
> This is a new feature being added to Discord. It already exists in Canary.

Ah, okay, I didn't know that.

discord/ext/commands/context.py
```diff
@@ -322,3 +324,8 @@ async def send_help(self, *args):
                 return None
         except CommandError as e:
             await cmd.on_help_command_error(self, e)
+
+    async def reply(self, *args, **kwargs):
+        return self.message.reply(*args, **kwargs)
```
zephyrkul, line 329:
```suggestion
        return await self.message.reply(*args, **kwargs)
```

SebbyLaw: This change does nothing. The original code returned the un-awaited `Message.reply()` coroutine which is arguably better.

Phxntxm: > This change does nothing. The original code returned the un-awaited `Message.reply()` coroutine which is arguably better.

But it in itself is a coroutine. Which means reply must be awaited, `await ctx.reply()` would then provide you with a coroutine... it needs to either be a coroutine and await the other coroutine, or not be a coroutine and return the coroutine.

Herbstblatt:
This pr should probably be closed because discord decided to implement replies in other way and making new pr would be easier than updating that.

PR #5761, opened 2020-08-27
Add new Message.quote property

TurnrDev:
### Summary

Adds a `quote` parameter to the `Message` object which can be used like so

```python
# Example with ctx
ctx.send(f"{ctx.message.quote}\n Ha, no!")

# Example without ctx
message.channel.send(f"{message.quote}\n Ha, no!")
```

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -637,7 +637,13 @@ class Message:
     @property
     def quote(self):
         """:class:`str`: Returns a string formatted as a quote, without the author mention."""
-        return "\n".join(["> {line}".format(line) for line in self.content.splitlines()])
+        if self.type is MessageType.default:
+            text = self.clean_content
+        else:
+            text = self.system_content
+
+        return "\n".join(["> {line}".format(line) for line in text.splitlines()])
+
 
     def is_system(self):
         """:class:`bool`: Whether the message is a system message.
--- a/discord/message.py
+++ b/discord/message.py
@@ -634,6 +634,11 @@ class Message:
         guild_id = getattr(self.guild, 'id', '@me')
         return 'https://discord.com/channels/{0}/{1.channel.id}/{1.id}'.format(guild_id, self)
 
+    @property
+    def quote(self):
+        """:class:`str`: Returns a string formatted as a quote, without the author mention."""
+        return "\n".join(["> {line}".format(line) for line in self.content.splitlines()])
+
     def is_system(self):
         """:class:`bool`: Whether the message is a system message.
 
--- a/discord/message.py
+++ b/discord/message.py
@@ -28,6 +28,7 @@ import asyncio
 import datetime
 import re
 import io
+import textwrap
 
 from . import utils
 from .reaction import Reaction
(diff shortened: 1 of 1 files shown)


TurnrDev:
It might be a smart idea for me to check the message type _and/or_ use `Message.system_content` instead

TurnrDev:
> It might be a smart idea for me to check the message type _and/or_ use `Message.system_content` instead

Okay, so I'm now checking `Message.type`, if it's `MessageType.default`, I'm quoteing `Message.clean_content`, else I'm quoting `Message.system_content` to quote system messages.

discord/message.py
```diff
@@ -637,7 +637,13 @@ def jump_url(self):
     @property
     def quote(self):
         """:class:`str`: Returns a string formatted as a quote, without the author mention."""
-        return "\n".join(["> {line}".format(line) for line in self.content.splitlines()])
+        if self.type is MessageType.default:
+            text = self.clean_content
+        else:
+            text = self.system_content
+
+        return "\n".join(["> {line}".format(line) for line in text.splitlines()])
```
ChillFish8, line 645: if you're gonna do this in a fairly unnecessary way at least make sure you specify the kwarg in that format 😔

ChillFish8 (commented):
Lack of actual testing *before* committing, also pretty sure the splitlines and join method is rather Slow and long-winded instead of just doing .replace()

TurnrDev:
> Lack of actual testing _before_ committing, also pretty sure the splitlines and join method is rather Slow and long-winded instead of just doing .replace()

I didn't even think of doing `replace`. :facepalm:  It's in draft because I haven't tested yet. I have dodgy power where I live so I commit often.

discord/message.py
TurnrDev: Oops

Werseter:
You could use textwrap.indent, being a part of the stardard library, to achieve it in a clean(er) manner.27 sie 2020 19:22 Jay Turner <notifications@github.com> napisał(a):
@TurnrDev commented on this pull request.



In discord/message.py:
> @@ -637,7 +637,13 @@ def jump_url(self):
     @property
     def quote(self):
         """:class:`str`: Returns a string formatted as a quote, without the author mention."""
-        return "\n".join(["> {line}".format(line) for line in self.content.splitlines()])
+        if self.type is MessageType.default:
+            text = self.clean_content
+        else:
+            text = self.system_content
+
+        return "\n".join(["> {line}".format(line) for line in text.splitlines()])

Oops

TurnrDev:
I done a few tests:

>>> timeit.timeit(sl) # Splitlines, my example
1.1926177599998482
>>> timeit.timeit(rpl) # Replace, pointless since it misses the first line
0.20285339900146937
>>> timeit.timeit(tw) # Textwrap, marginally faster than Splitlines, cleaner code
1.452346336998744

Werseter:
Wouldn't removing the lambda predicate, which requires a redundant lambda function call (as it's always True), improve the performance, if we consider this an issue?

ChillFish8:
You forget that you can just add a `> ` on to the start of the string

Rapptz:
Discord is planning on replacing this with replies at some point so I'll close this. Also the quote feature takes up too many characters upon sending so it's a poor utility.

TurnrDev:
> Discord is planning on replacing this with replies at some point so I'll close this. Also the quote feature takes up too many characters upon sending so it's a poor utility.

Alright, no harm :)

PR #5767, opened 2020-08-29
Let `str(Message)` return the message content.

Arthurdw:
### Summary

Yesterday I was helping someone else who had the question on how to extract the `ctx.message`.
I knew that `ctx.message.content` returned the content but it seemed logical that `str(ctx.message)` also returns the content. (which is done splendidly by eg the `User` object.)

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -344,6 +344,9 @@ class Message:
             except KeyError:
                 continue
 
+    def __str__(self):
+        return self.content
+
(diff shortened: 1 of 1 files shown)


scragly:
Duplicate of #738

Quoting the lib developer:
> Message purposefully doesn't have `__str__` implemented.

PR #5768, opened 2020-08-29
Add convenience methods run_from_file and run_from_env

nitori:
### Summary

Fetching the token from alternative sources is an obvious thing to do, that I decided to add those two convenience methods. But the primary reason I made this change is because beginners new to python (or programming in general) tend to accidentally disclose their token when sharing/showing their code to get help.

So it doesn't really address any issue with the bot itself, but I thought it would be a nice thing to have.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -27,6 +27,7 @@ DEALINGS IN THE SOFTWARE.
 import asyncio
 from collections import namedtuple
 import logging
+import os
 import signal
 import sys
 import traceback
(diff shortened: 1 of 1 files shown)


parafoxia:
I see where you're coming from, but I personally don't think this is the best implementation. I think something like `self.run(tokenfile="path/to/token", tokenenv="ENVIRONVAR")` would be better if something like this were to be implemented. Plus, the file remains open for the entire time the bot is running at the moment.

Those are just my opinions, but I do think the idea has potential.

AbstractUmbra:
IMO this is not something discord.py should be handling on behalf of the end user.

This is a wrapper for the Discord API, and the handling of user secrets such as the token should be taken care of by the user. Not only because it's good to teach proper security practices (such as not committing your token to GitHub, etc), but because this isn't something an API wrapper should be handling.

It is not difficult to load an environment variable, or import a string from a `.gitignore`'d file, as such the end user should do this without this unnecessary hand holding.

Rapptz:
I understand where this PR is coming from but I'm not really a fan of things that read the environment or do things like this since it's saving a few lines of code at best. I think just having a python file for configuration with your secrets and ignoring it works out just fine, it's what the template maker does with `python -m discord newbot` as well (it even adds it to `.gitignore`).

PR #5780, opened 2020-08-31
Added proper hex support for Colour objects

parafoxia:
### Summary

Adds convenience methods to add better hexadecimal support, as well as an alias for one colour, and two factory methods.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/colour.py
+++ b/discord/colour.py
@@ -253,4 +253,18 @@ class Colour:
 
     grayple = greyple
 
+    @classmethod
+    def black(cls):
+        """A factory method that returns a :class:`Colour` with a value of ``0x000001``.
+        .. versionadded:: 1.5
+        """
+        return cls(1)
+
+    @classmethod
+    def white(cls):
+        """A factory method that returns a :class:`Colour` with a value of ``0xffffff``.
+        .. versionadded:: 1.5
+        """
+        return cls(0xffffff)
+
 Color = Colour
--- a/discord/colour.py
+++ b/discord/colour.py
@@ -263,6 +263,7 @@ class Colour:
     @classmethod
     def white(cls):
         """A factory method that returns a :class:`Colour` with a value of ``0xffffff``.
+
         .. versionadded:: 1.5
         """
         return cls(0xffffff)
--- a/discord/colour.py
+++ b/discord/colour.py
@@ -245,7 +245,12 @@ class Colour:
 
     @classmethod
     def greyple(cls):
-        """A factory method that returns a :class:`Colour` with a value of ``0x99aab5``."""
+        """A factory method that returns a :class:`Colour` with a value of ``0x99aab5``.
+        .. versionchanged:: 1.5
+            Added grayple alias.
+        """
         return cls(0x99aab5)
+
+    grayple = greyple
 
 Color = Colour
--- a/discord/colour.py
+++ b/discord/colour.py
@@ -101,6 +101,12 @@ class Colour:
         """Tuple[:class:`int`, :class:`int`, :class:`int`]: Returns an (r, g, b) tuple representing the colour."""
         return (self.r, self.g, self.b)
 
+    def to_hex(self):
+        """:class:`str`: Returns the hex format for the colour.
+        .. versionadded:: 1.5
+        """
+        return '#{:0>6x}'.format(self.value)
+
     @classmethod
     def from_rgb(cls, r, g, b):
         """Constructs a :class:`Colour` from an RGB tuple."""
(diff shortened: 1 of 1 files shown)


discord/colour.py
```diff
@@ -101,6 +101,12 @@ def to_rgb(self):
         """Tuple[:class:`int`, :class:`int`, :class:`int`]: Returns an (r, g, b) tuple representing the colour."""
         return (self.r, self.g, self.b)
 
+    def to_hex(self):
+        """:class:`str`: Returns the hex format for the colour.
+        .. versionadded:: 1.5
+        """
+        return '#{:0>6x}'.format(self.value)
```
Rapptz, line 108: This already exists, it's `__str__`.

discord/colour.py
```diff
@@ -112,6 +118,13 @@ def from_hsv(cls, h, s, v):
         rgb = colorsys.hsv_to_rgb(h, s, v)
         return cls.from_rgb(*(int(x * 255) for x in rgb))
 
+    @classmethod
+    def from_hex(cls, hexa):
+        """Constructs a :class:`Colour` from hex format.
+        .. versionadded:: 1.5
+        """
+        return cls(int(hexa.strip("#"), base=16))
```
Rapptz, line 126: This method doesn't really feel like it belongs? Some people will begin to expect other formats like `0xFFFFFF` or people will confuse passing in `int` and expecting it to work etc. There's a bit too much confusion that can arise from this method compared ot the value it brings, which is shaving off one line where you do `int(string[1:], base=16)`

discord/colour.py
```diff
@@ -232,7 +245,26 @@ def blurple(cls):
...
     @classmethod
     def greyple(cls):
-        """A factory method that returns a :class:`Colour` with a value of ``0x99aab5``."""
+        """A factory method that returns a :class:`Colour` with a value of ``0x99aab5``.
+        .. versionchanged:: 1.5
+            Added grayple alias.
+        """
         return cls(0x99aab5)
 
+    grayple = greyple
```
Rapptz, line 254: The `greyple` name is the branding one: https://discord.com/branding

discord/colour.py
```diff
@@ -232,7 +245,26 @@ def blurple(cls):
...
+
+    @classmethod
+    def black(cls):
+        """A factory method that returns a :class:`Colour` with a value of ``0x000001``.
+        .. versionadded:: 1.5
+        """
+        return cls(1)
```
Rapptz, line 261: Colours in the factory methods correspond to either the branding guidelines or the role colour picker, adding colours to work around Discord's weird reservation of a specific colour seems a bit off to me so I'll have to reject it.

If you want to have one added here then it could be `not_quite_black` which would correspond to `0x23272A` but I'm not sure I've seen anyone request this.

discord/colour.py
```diff
@@ -232,7 +245,26 @@ def blurple(cls):
...
+
+    @classmethod
+    def white(cls):
+        """A factory method that returns a :class:`Colour` with a value of ``0xffffff``.
+        .. versionadded:: 1.5
```
Rapptz, line 266:
```suggestion

        .. versionadded:: 1.5
```

discord/colour.py
parafoxia: To me it seemed a bit weird to me that it didn't exist seen as the hex format is the standard used in the GUI colour picker, and there's a way to convert the object into a hex string. It was more of a consistency thing to fit in with that, plus the existing RGB and HSV methods if that makes sense, but I could understand the functionality not necessarily being needed.

The `to_hex()` method is just there as a counter to the `from_hex()`, as I feel that should be available if the `from_hex()` ended up being implemented.

discord/colour.py
parafoxia: Ah, was not aware of that.

discord/colour.py
parafoxia: Fairs. Thought I'd propose these since I was in the file anyways.

I'm happy to add the `not_quite_black` one in if you felt like it would fit. Having looked at the branding link you sent in an above comment, it does appear, so might as well I spose?

discord/colour.py
Phxntxm: To me this doesn't make sense as a method because "from hex" is the default. Creating an instance of a colour takes in an integer value, and in python you can provide `0xFFFFFF`. This is hex, this is an integer, and is already supported as the only argument in the constructor. Due to this, I don't see a purpose for a class method that does the same thing, but from a string instead.

sairam4123: I think, there's no use of `from_hex` and `to_hex`.

PR #5789, opened 2020-09-04
Added __eq__ to Message class in message.py

TurretAA12:
### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -347,6 +347,9 @@ class Message:
     def __repr__(self):
         return '<Message id={0.id} channel={0.channel!r} type={0.type!r} author={0.author!r} flags={0.flags!r}>'.format(self)
 
+    def __eq__(self, other):
+        return isinstance(other, self.__class__) and self.id == other.id
+
     def _try_patch(self, data, key, transform=None):
         try:
             value = data[key]
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -903,6 +903,15 @@ of :class:`enum.Enum`.
     Specifies the type of :class:`Message`. This is used to denote if a message
     is to be interpreted as a system message or a regular message.
 
+    .. container:: operations
+    
+      .. describe:: x == y
+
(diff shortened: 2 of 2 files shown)


TurretAA12:
I am not sure how I would go about updating the documentation.  Please let me know!

BeatButton:
Duplicate of #772, unfortunately

TurretAA12:
O I did not see that, my bad. Would you like me to close?

Vexs:
I think it's worth re-approaching again because after literally three years of every single mention of this coming up, PR being made, or expectation of users, every single instance I can recall has had the exact same id=id class=class comparison. Not just because it falls inline with every other object in the library both logically and philosophically, but also because it's the bloody expected behavior.

Rapptz:
> I am not sure how I would go about updating the documentation. Please let me know!

There's a supported operations table, you can just copy it from everywhere else.

TurretAA12:
Ya I think that the current implementation (which is just the default `object.__eq__`) is not the most intuitive functionality.  If you are also concerned about comparing content why not also add a `compare_content()` method. Maybe something along the lines of...
```python
def compare_content(self, other):
    return isinstance(other, self.__class__) and self.id == other.id and self.content == other.content
```

TurretAA12:
So what's the status of this PR? Should I close it or is this change up for debate?

TurnrDev:
> Ya I think that the current implementation (which is just the default `object.__eq__`) is not the most intuitive functionality. If you are also concerned about comparing content why not also add a `compare_content()` method. Maybe something along the lines of...
> 
> ```python
> def compare_content(self, other):
>     return isinstance(other, self.__class__) and self.id == other.id and self.content == other.content
> ```

Doesn't Python treat `==` and `is` differently? Maybe we can make use of that.

muddyfish:
It's not treated differently in a way that would help this particular problem

TurnrDev:
Ah ok

PR #5799, opened 2020-09-08
[commands] Change exception raising behaviour of Command before / after invoke hooks.

bijij:
### Summary

Before if an exception was raised in a `Command.before_invoke` or `Command.after_invoke` callable in a `Cog` it would be raised as

`AttributeError: 'function' object has no attribute '__self__'`

This is due to the way python exception handling is done, and the use of a `try: except:` block to determine whether the `Command` was part of a `Cog` or not.

This PR changes this behaviour by instead raising the original exception in the `Command.before_invoke` / `Command.after_invoke` callable.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -716,7 +716,7 @@ class Command(_BaseCommand):
         cog = self.cog
         if self._before_invoke is not None:
             # should be cog if @commands.before_invoke is used
-            instance = getattr(self.before_invoke, '__self__', self.cog)
+            instance = getattr(self._before_invoke, '__self__', self.cog)
             # __self__ only exists for methods, not functions
             # however, if @command.before_invoke is used, it will be a function
             if instance:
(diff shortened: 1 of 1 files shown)


discord/ext/commands/core.py
```diff
@@ -715,18 +715,14 @@ async def call_before_hooks(self, ctx):
...
-            except AttributeError:
-                # __self__ only exists for methods, not functions
-                # however, if @command.before_invoke is used, it will be a function
-                if self.cog:
-                    await self._before_invoke(cog, ctx)
-                else:
-                    await self._before_invoke(ctx)
-            else:
+            # should be cog if @commands.before_invoke is used
+            instance = getattr(self._before_invoke, '__self__', self.cog)
```
Rapptz, line 719:
```suggestion
            instance = getattr(self._before_invoke, '__self__', cog)
```

Defined above.

Also this probably needs to apply to `after_invoke` as well.

Rapptz (changes requested):

discord/ext/commands/core.py
bijij: Both suggested changges have been made.

Rapptz (approved):

PR #5810, opened 2020-09-10
Fix by_category in guild.py overwriting channels

CrunchyDuck:
### Summary

Fixes by_category. Currently, by_category will almost always not return all of the channels, as they're overwritten due to a needless line of code.

### Checklist

- [x] This PR fixes an issue.


Rapptz:
This doesn't override anything. https://docs.python.org/3/library/stdtypes.html#dict.setdefault

PR #5821, opened 2020-09-15
Update README.rst

rogervs:
Added line to indicate how to install the voice package if the user is using zsh

### Summary

The [voice] section in the pip install does not work with zsh, added line to show how to install with zsh by adding quotation marks.

### Checklist

- [ ] This PR fixes an issue.


TheRockettek:
Why not just change the README to use quotes on all commands as that is also accepted?

PR #5826, opened 2020-09-16
Command instead of "on_message" commands (README.rst)

12944qwerty:
### Summary

Changes the content of README.rst section `Quick Example` so that it uses the commands format instead of on_message.

My logic behind this is so that people new to this library won't immediately use the `on_message` way as it is a lot more confusing to switch to commands. This will encourage newer users to start with commands instead of parsing the messages manually.

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -81,11 +81,16 @@ Quick Example
             # don't respond to ourselves
             if message.author == self.user:
                 return
-
-            if message.content == 'ping':
-                await message.channel.send('pong')
+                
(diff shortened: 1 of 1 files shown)


Gobot1234:
You are aware this is going to raise an AttributeError upon running it right?

12944qwerty:
Oh oops. I just realized that the bot example has this anyway. ~~i'll close this then~~

PR #5827, opened 2020-09-16
[RFC] Run on_ready on module load if bot is ready

timschumi:
### Summary

An `on_ready` might do some specific setup on the extension that is only doable once the bot is ready,
but it isn't being run if the extension is loaded after the Client has finished initializing.

Run all the `on_ready` handlers from that extension once the extension has been loaded
(provided that the bot is in a ready state at that point).

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -625,6 +625,12 @@ class BotBase(GroupMixin):
         else:
             self.__extensions[key] = lib
 
+        # Call on_ready if module has been loaded after the bot is ready
+        if self.is_ready():
+            for event in self.extra_events.get('on_ready', []):
+                if event.__module__ is not None and _is_submodule(key, event.__module__):
+                    self._schedule_event(event, 'on_ready')
+
     def load_extension(self, name):
         """Loads an extension.
 


timschumi:
I do realize that this might have been a deliberate design decision, so I marked it as a "kind-of-RFC" for now. I'm also unsure what boxes to tick, since this doesn't really add anything new according to the included definition, nor does it explicitly fix an "issue".

At last, both the code quality and the place where I added the functionality might be subpar, since this is much more of a prototype assembled from lines that I found somewhere else in the Bot.

Vexs:
Its trivial to do what you want with the following code-

```
class mycog(commands.Cog):
	def __init__(self, bot):
	    self.bot.loop.create_task(myready())

	async def myready(self):
	    await self.bot.wait_until_ready()
	    do_thing```

timschumi:
I know that I could have solved this in a different way. In fact, I'm doing it like you suggested at the moment (although with a different and horribly long asyncio command, so I'll keep that snippet around).

I just figured that this _might_ be a thing that could be useful as something that is already included in the library.

Vexs:
The subject has been brought up before but it isn't implemented because notably, the on_ready event isn't actually happening and there are conceivably situations in which you wouldn't want that- making this a breaking change notably. Further, it isn't done because it's very easy to implement yourself.

timschumi:
> The subject has been brought up before

I haven't been able to find any previous discussion on this, apologies for that.

> and there are conceivably situations in which you wouldn't want that- making this a breaking change

I can't think of any cases there, especially since the documentation says that the function might be called multiple times during the lifetime of a bot.

> Further, it isn't done because it's very easy to implement yourself.

I've rarely ever seen a case where that's a good reason to _not_ have something. Especially if the "very easy to implement" thing grazes "internal bot stuff" territory. Also, we'd have to remove `User.mention` then, wouldn't we?

Rapptz:
This isn't a particularly bad idea I suppose though it breaks my expectations on how events are done. To me, an event (barring command extension ones) is something that must come from within Discord and is dispatched that way. There are many related events outside of ready that essentially deal with connection that aren't addressed in this PR:

- `on_shard_ready`
- `on_shard_connect`
- `on_connect`
- `on_shard_disconnect`
- `on_disconnect`
- `on_resumed`
- `on_shard_resumed`

These all fall in the same umbrella of "connection related events", though this one special cases merely `on_ready`. It's simple to make the argument that the library itself is what classifies the importance of the `on_ready` event -- since there are things like `wait_until_ready` and `is_ready`. These however are provided merely for convenience since at the time this library was written, waiting for the cache to be ready was a desirable feature. In the future the line of this convenience is not as black and white due to changes in the API making it harder to maintain a complete cache -- further blurring the line of usefulness that this event provides.

It's a bit long winded and I've gone on a too far tangent at this point but essentially I don't really like giving this extra special treatment when I might remove the other special treatment this event already has in the future. It's a bit weird to me that this event will be called multiple times for reasons outside of connection related reasons which is what people expect this event to be. Reloading a cog now constantly calls this and would break those who use `on_ready` as a sentinel for keeping track of their connections.

Ultimately I'll have to pass on this PR. Sorry.

timschumi:
Understandable, especially if you are considering changing the event in the future. From the comments by Vexs I understood that this might not be a desired change, but I still kept it open in the hope to get a better answer than a vague "this breaks stuff" and "this isn't included because it's trivial to implement oneself".

Have a great day!

PR #5832, opened 2020-09-18
Add support for message_reference on Message object

### LINKED ISSUE #5754: Add a method to access the "message_reference" object on messages
### The Problem

Currently when a message is pinned, you have no way of knowing what message was pinned. Discord sends a [message_reference](https://discord.com/developers/docs/resources/channel#message-object-message-reference-structure) object on the "User pinned a message to this channel" messages and also for crossposted messages either denoting the message that was pinned or the original message that was crossposted.

### The Ideal Solution

Ideally, messages could have a "reference" attribute that would allow you to access the data provided in the message_reference object, either allowing access to the id itself or the object linked to that id if possible (e.g. if the bot is in the guild and has access to the channel or if the pinned/original message is cached)

### The Current Solution

There is no good solution to this at the moment.

### Summary

Add a way to access a message's reference object :)

SebbyLaw:
### Summary

This PR resolves #5754. Only the `message_id` attribute is exposed since the guild and channel id can be found on the original message object. A `referenced_message` attribute is also added for convenience.

Note: for some reason the [message_id key is optional](https://discord.com/developers/docs/resources/channel#message-object-message-reference-structure).

### Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -236,10 +236,10 @@ class MessageReference:
     @property
     def cached_message(self):
         """Optional[:class:`Message`]: The cached message, if found in the internal message cache."""
-        return self._state._get_message(self._message_id)
+        return self._state._get_message(self.message_id)
 
     def __repr__(self):
-        return '<MessageReference message_id={0._message_id!r} channel_id={0._channel_id!r} guild_id={0._guild_id!r}>'.format(self)
+        return '<MessageReference message_id={0.message_id!r} channel_id={0.channel_id!r} guild_id={0.guild_id!r}>'.format(self)
 
 def flatten_handlers(cls):
     prefix = len('_handle_')
--- a/discord/message.py
+++ b/discord/message.py
@@ -229,7 +229,7 @@ class MessageReference:
     def __init__(self, **kwargs):
         self.message_id = utils._get_as_snowflake(kwargs, 'message_id')
         self.channel_id = int(kwargs.pop('channel_id'))
-        self.guild_id = int(kwargs.pop('guild_id', 0)) or None
+        self.guild_id = utils._get_as_snowflake(kwargs, 'guild_id')
         self.cached_message = None
 
     def __repr__(self):
--- a/discord/message.py
+++ b/discord/message.py
@@ -212,29 +212,26 @@ class MessageReference:
     """Represents a reference to a :class:`Message`.
 
     .. versionadded:: 1.5
+
+    Attributes
+    -----------
+    message_id: Optional[:class:`int`]
+        The id of the message referenced.
+    channel_id: :class:`int`
+        The channel id of the message referenced.
+    guild_id: Optional[:class:`int`]
+        The guild id of the message referenced.
+    cached_message: Optional[:class:`Message`]
+        The cached message, if found in the internal message cache.
     """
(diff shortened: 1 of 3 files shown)


Rapptz:
I'm pretty sure this is also filled in other cases (such as crossposting) so full support is better than partial support for the pin case.

SebbyLaw:
Ready for review.

discord/message.py
```diff
@@ -208,6 +208,33 @@ async def to_file(self, *, use_cached=False, spoiler=False):
...
+    cached_message: Optional[:class:`Message`]
+        The cached message, if found in the internal message cache.
+    """
+    __slots__ = ('message_id', 'channel_id', 'guild_id', 'cached_message')
+
+    def __init__(self, **kwargs):
+        self.message_id = int(kwargs.pop('message_id', 0)) or None
+        self.channel_id = int(kwargs.pop('channel_id'))
+        self.guild_id = int(kwargs.pop('guild_id', 0)) or None
+        self.cached_message = None
```
Rapptz, line 233: I'd prefer if this was a dynamic lookup property rather than a strong reference.

discord/message.py
```diff
@@ -208,6 +208,33 @@ async def to_file(self, *, use_cached=False, spoiler=False):
...
+        The channel id of the message referenced.
+    guild_id: Optional[:class:`int`]
+        The guild id of the message referenced.
+    cached_message: Optional[:class:`Message`]
+        The cached message, if found in the internal message cache.
+    """
+    __slots__ = ('message_id', 'channel_id', 'guild_id', 'cached_message')
+
+    def __init__(self, **kwargs):
+        self.message_id = int(kwargs.pop('message_id', 0)) or None
```
Rapptz, line 230:
```suggestion
        self.message_id = utils._get_as_snowflake(kwargs, 'message_id')
```

discord/message.py
```diff
@@ -208,6 +208,33 @@ async def to_file(self, *, use_cached=False, spoiler=False):
...
+        self.channel_id = int(kwargs.pop('channel_id'))
+        self.guild_id = int(kwargs.pop('guild_id', 0)) or None
```
Rapptz, line 232:
```suggestion
        self.guild_id = utils._get_as_snowflake(kwargs, 'guild_id')
```

Rapptz (changes requested):

discord/message.py
SebbyLaw: I'm unclear on the reasoning behind dynamic lookup on properties aside from the cached message.
Resolved.

discord/message.py
```diff
@@ -208,6 +208,39 @@ async def to_file(self, *, use_cached=False, spoiler=False):
...
+    Attributes
+    -----------
+    message_id: Optional[:class:`int`]
+        The id of the message referenced.
+    channel_id: :class:`int`
+        The channel id of the message referenced.
+    guild_id: Optional[:class:`int`]
+        The guild id of the message referenced.
+    cached_message: Optional[:class:`Message`]
+        The cached message, if found in the internal message cache.
```
Rapptz, line 225:
```suggestion
```

Already documented.

Rapptz (changes requested):
Mostly fine

Rapptz (approved):

PR #5834, opened 2020-09-18
Remove namedtuples

Taarek:
## Summary

It removes the usage of namedtuples to avoid a future ``deny_new`` issue.

## Checklist

- [ ] This PR fixes an issue.

I was not able to test integrations because I don't have access to said guild, but the other stuff works fine.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -25,7 +25,6 @@ DEALINGS IN THE SOFTWARE.
 """
 
 import asyncio
-from collections import namedtuple
 import logging
 import signal
 import sys
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -48,20 +48,10 @@ from .asset import Asset
 from .flags import SystemChannelFlags
 from .integrations import Integration
 
+
+BanEntry = namedtuple('BanEntry', 'reason user')
 _GuildLimit = namedtuple('_GuildLimit', 'emoji bitrate filesize')
 
-class BanEntry:
-    __slots__ = ('reason', 'user')
-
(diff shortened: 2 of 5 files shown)


SebbyLaw:
Could something like this be used to cut down on repetition?
```py
class FakeNamedTuple:
    def __init__(self, **kwargs):
        for k in self.__slots__:
            setattr(self, k, kwargs.get(k))

    def _asdict(self):
        return {k: getattr(self, k) for k in self.__slots__}

    def __repr__(self):
        return '%s(%s)' % (self.__class__.__name__, ', '.join('%s=%r' % (k, getattr(self, k)) for k in self.__slots__))
```
ex:
```py
class BanEntry(FakeNamedTuple):
    __slots__ = ('reason', 'user')
```

iomintz (changes requested):
This kills the repr.

jack1142:
This breaks tuple unpacking (among other things) and BanEntry is specifically documented as a namedtuple. Maybe change for d.py v2.0?

mikeshardmind:
Tuple unpacking is *already* version unsafe if discord ever adds anything to what they include in a ban entry and this needs to be updated for that.

Rapptz (changes requested):
A few nits.

1) This can't be removed from `BanEntry` since it's explicitly documented to be a namedtuple there and it doesn't have any of the issues that made `_Overwrites` annoying.
2) This is missing `__repr__` for those classes.
3) If `_asdict` is not used then don't bother defining it.

Rapptz (approved):

PR #5839, opened 2020-09-20
[commands] Added optional arguments to load_extension

FrostByte266:
## Summary

I have added the ability to pass extra positional and keyword arguments to an extension's setup method to provide a nice way to share data with an extension and the main bot code. I have tested this and code that doesn't pass extra arguments works the same as before, and passing positional and keyword arguments to the setup function also works.
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -599,7 +599,7 @@ class BotBase(GroupMixin):
                 if _is_submodule(name, module):
                     del sys.modules[module]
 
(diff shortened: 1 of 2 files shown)


Rapptz:
This isn't really a design space I want to explore like this. Indiscriminately passing parameters like this tends to block the `load_extension` from being changed (e.g. if I wanted to pass a package path to `importlib` in the future). There is another method being explored to allow to pass configuration in an extension but that'll be explored once the library moves on to Python 3.6.

PR #5858, opened 2020-09-26
Move attributes and methods from BaseUser into abc.User

MarkKoz:
## Summary

All concrete implementations of `discord.abc.User` have the following attributes in common

* avatar_url
* avatar_url_as
* color
* colour
* default_avatar
* default_avatar_url
* is_avatar_animated
* mentioned_in
* permissions_in
* public_flags
* system

Therefore, it makes sense for them to be defined in the abstract class. When working with an `abc.User` rather than a concrete type, it was misleading to only see a handful of attributes defined. This change will clearly communicate which attributes, properties, and methods are actually available.

## Checklist

- [x] This PR fixes an issue.


discord/member.py
```diff
@@ -86,7 +86,9 @@ def __repr__(self):
         return '<VoiceState self_mute={0.self_mute} self_deaf={0.self_deaf} self_stream={0.self_stream} channel={0.channel!r}>'.format(self)
 
 def flatten_user(cls):
-    for attr, value in itertools.chain(BaseUser.__dict__.items(), User.__dict__.items()):
+    # Despite being inherited, discord.abc.User's attributes are redirected because
+    # some (e.g. public_flags) may rely on attributes only defined in BaseUser
+    for attr, value in itertools.chain(discord.abc.User.__dict__.items(), BaseUser.__dict__.items(), User.__dict__.items()):
```
MarkKoz, line 91: It may be fine to use `dir(BaseUser), User.__dict__.items()` instead.

discord/abc.py
```diff
@@ -124,7 +263,11 @@ def __subclasshook__(cls, C):
                 return NotImplemented
 
             mro = C.__mro__
-            for attr in ('display_name', 'mention', 'name', 'avatar', 'discriminator', 'bot'):
+            attrs = ('display_name', 'mention', 'name', 'avatar', 'discriminator', 'bot',
+                     'avatar_url', 'avatar_url_as', 'color', 'colour', 'default_avatar',
+                     'default_avatar_url', 'is_avatar_animated', 'mentioned_in', 'permissions_in',
+                     'public_flags', 'system')
```
MarkKoz, line 269: This could probably shortened by using `__dict__` for most of these, but I honestly don't see why this check is even useful in the first place.

discord/member.py
MarkKoz: Never mind, dir() only gives names.

Rapptz (changes requested):
I am confused why this PR was done this way.

The original abc.User class was made not with concrete inheritance in mind but as an actual abc following the module. This changeset throws this stuff away along with the `@abc.abstractmethod` properties.

I would have much preferred for the scope of these changes to be significantly reduced to be similar to the previous implementation.

MarkKoz:
It seems redundant to define abstract methods and properties when they are all going to be funnelled through `BaseUser` anyway. In fact, `BaseUser` as a whole seems redundant. Other classes in `discord.abc` are more concrete, which makes more sense to me, despite being in the "abc" module. Why the inconsistency? Am I missing something? Is it conceivable for a class unrelated to `BaseUser` (i.e. one that doesn't benefit from any of the default implementations) to make use of `abc.User`?

Rapptz:
A lot of the things in the abc module are not concrete (I can only think of `abc.Messageable` and `abc.GuildChannel` that are concrete) and whether they are is mostly a stylistic choice. `BaseUser` was because there needed to be a tag type to bring back comparison between `User` and `Member` which was previously removed during the transitional beta period from 0.16 to 1.0.

discord/member.py
```diff
@@ -121,7 +121,11 @@ def general(self, *args, **kwargs):
 _BaseUser = discord.abc.User
 
 @flatten_user
-class Member(discord.abc.Messageable, _BaseUser):
+class _BaseMember:
+    pass
+
+@flatten_user
+class Member(_BaseMember, discord.abc.Messageable, _BaseUser):
```
MarkKoz, line 128: Had to do this hack because the `User` attributes were being added too late, so not all abstract methods were considered defined. Not sure of a better way to solve this.

discord/abc.py
```diff
@@ -102,29 +105,143 @@ class User(metaclass=abc.ABCMeta):
...
+        Returns
+        -------
+        :class:`bool`
+            Indicates if the user is mentioned in the message.
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def permissions_in(self, channel):
+        """An alias for :meth:`abc.GuildChannel.permissions_for<discord.abc.GuildChannel.permissions_for>`.
```
MarkKoz, line 209: I don't know of a better way to get rid of only the `discord.` prefix.

PR #5862, opened 2020-09-27
Set default values and document them

PythonCoderAS:
## Summary

This PR makes it so that I can update the value of a field without having to provide the name. It also sets some methods (the ones that accept a `discord.Embed.Empty` object) to have a default of `discord.Embed.Empty`.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
These are staying as explicit arguments since they were designed with it in-mind. Removing it from the calling code makes it hard to read.

PR #5885, opened 2020-10-04
Added before_cached_message and after_cached_message attributes to RawMessageUpdateEvent

asleep-cult:
## Summary

So we can access both the old and new versions of the message from RawMessageUpdateEvent's if it was cached
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/raw_models.py
+++ b/discord/raw_models.py
@@ -96,17 +96,20 @@ class RawMessageUpdateEvent(_RawReprMixin):
 
     data: :class:`dict`
         The raw data given by the `gateway <https://discord.com/developers/docs/topics/gateway#message-update>`_
-    cached_message: Optional[:class:`Message`]
-        The cached message, if found in the internal message cache.
+    before_cached_message: Optional[:class:`Message`]
+        The cached message before it was updated, if found in the internal message cache.
+    after_cached_message: Optional[:class:`Message`]
+        The cached message after it was updated, if found in the internal message cache.
     """
 
-    __slots__ = ('message_id', 'channel_id', 'data', 'cached_message')
+    __slots__ = ('message_id', 'channel_id', 'data', 'before_cached_message', 'after_cached_message')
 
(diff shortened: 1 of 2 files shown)


Zomatree:
This is a breaking change as you're removing a attribute.

Rapptz:
On top of being a breaking change, there is no real reason to have the `after` since it can be derived from the data already given. These are meant to be low level events. If you want the higher level interface use the other one.

asleep-cult:
So we're expected to fetch the message to get the newer version?

SebbyLaw:
The data from the updated message can be gotten from [RawMessageUpdateEvent.data](https://discordpy.readthedocs.io/en/latest/api.html#discord.RawMessageUpdateEvent.data), which is just a [message object](https://discord.com/developers/docs/resources/channel#message-object).

This is meant to be a low level event, hence the "raw". If you want a higher level interface, [on_message_edit](https://discordpy.readthedocs.io/en/latest/api.html#discord.on_message_edit) already exists.

asleep-cult:
My main problem with this is that there's no "simple" interface to get to the newer version of the message without fetching it when it's already in the message cache. There should at least be a `client.get_message` method (which can be done through `client._connection._get_message` but shouldn't be used).

Gobot1234:
You can use `utils.get` on `client.cached_messages` to get messages.

asleep-cult:
Ok didn't know `client.cached_messages` existed although I don't like the idea of iterating over possibly thousands of messages twice to get the same message.

Rapptz:
You don't have to iterate. The update data *is already given to you*. You can check for its existence there. Alternatively, use the other event. 

At this point this is going in circles.

SebbyLaw:
Iterating over possibly thousands of messages is [exactly how discord.py does it](https://github.com/Rapptz/discord.py/blob/master/discord/state.py#L347).

Issue #2159, opened 2019-05-16
[tasks] Allow running a function at exact times.


Rapptz:
### The Problem

Some people want tasks to run at specified times (e.g. `9PM` every day). Support for `datetime.time` objects for these cases is probably a good idea. Supporting anything else is outside of the scope of this extension since it involves an actual scheduler functionality but for basic "on X time every day" kind of deal this should be supported since it's the most common.

### The Ideal Solution

Something similar to

```py
@tasks.loop(on=datetime.time(21, 0, 0))
```

Design pending.

### The Current Solution

None

scj643:
There should be some check to make sure that the task ran. My logic behind this is that if the bot is blocked at the scheduled time the task might not run.

Rapptz:
If the event loop is blocked there is nothing that I can do about that, this falls under user responsibility.

BluePhoenixGame:
I feel like it also should support `datetime.timedelta` so you can specify one time each hour for example. Otherwise a work around i think users would start doing is adding a task with every hour, which i dont think is positive because it would be easier for the library to handle it as one task?

bmintz:
I could also imagine `on` being a list so that you could specify the times it'll run every day.

bmintz:
> If the event loop is blocked there is nothing that I can do about that, this falls under user responsibility.

The event loop may be blocked at any particular time, non deterministically, right? So shouldn't the lib try and wait until next context switch to run the task?

Rapptz:
>  I feel like it also should support datetime.timedelta so you can specify one time each hour for example. Otherwise a work around i think users would start doing is adding a task with every hour, which i dont think is positive because it would be easier for the library to handle it as one task?

I don't know what you're suggesting here that isn't already possible. `@tasks.loop(hours=1.0)`.

> The event loop may be blocked at any particular time, non deterministically, right? So shouldn't the lib try and wait until next context switch to run the task?

Like I mentioned earlier, this is impossible to do on my end and falls on user responsibility. Consider:

```py
async def my_awful_coroutine():
    await foo()
    # we get to run at this point at 5:59:58 PM
    # and we have a task set to run at 6:00 PM

    # uh oh
    time.sleep(5)
    return 10
```

DevilXD:
[...]

PR #5891, opened 2020-10-04
[tasks] Added basic support for datetime.time

MrHedmad:
## Summary
This pull adds very basic support for the `datetime.time` objects in the Tasks extension. It's a very simple change, which just shifts the time of execution to align to the 'at' parameter (if any is specified) and then resumes normal loop behavior. 

Do note that from internal testing the loop seems to shift by about 2 seconds when the task is initiated, and the reason is unknown. However, this seems irrelevant to me for everyday usage.

This tries to amend issue #2159 

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/discord/ext/tasks/__init__.py
+++ b/discord/ext/tasks/__init__.py
@@ -37,12 +37,17 @@ from discord.backoff import ExponentialBackoff
 
 log = logging.getLogger(__name__)
 
+def time_diff(x, y):
+    """Returns the timedelta between two times"""
+    now = datetime.datetime.now().date()
+    return datetime.datetime.combine(now, x) - datetime.datetime.combine(now, y)
+
 class Loop:
     """A background task helper that abstracts the loop and reconnection logic for you.
 
     The main interface to create this is through :func:`loop`.
     """
-    def __init__(self, coro, seconds, hours, minutes, count, reconnect, loop):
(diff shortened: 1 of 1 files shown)


feelingnothing:
Great work, would be cool to see this approved

xernphoton:
OMG i needed this

Technerder:
Looks great!

xernphoton:
i want this

Rapptz:
Hello, thanks for the PR and sorry for the late review.

I am not overly a fan of this implementation for a few reasons.

1. It is more complicated than it should be. The loop class already has all the tools necessary to support this by configuring `_get_next_sleep_time` as the hook function to change if an explicit time is passed.
2. The time delta utility function seems to be using local time despite the times being passed using UTC time. The original implementation tried to be timezone aware, and this seems to toss all of that away.
3. The original specification called for a `times=[...]` parameter that allowed you to pass multiple times if needed, which was a superset of allowing a single time. The specification also referred to this as explicit time, rather than relative time. This means that the loop will only be called at the specified times given. This PR however adds a single parameter named `at` that supports a single time point to start the initial loop .
4. The original implementation blocked both relative times and explicit times, this PR does not due to it being fundamentally different.

MrHedmad:
Thanks for the review Danny.

PR #5909, opened 2020-10-08
Add markdown-inspired newline handling for help messages obtained from wrapped Command docstrings.

Hawkpath:
## Summary

Command help messages obtained from function docstrings now handle newlines similarly to Markdown. Single newlines will be replaced with a space, and double newlines will be replaced with a single newline. Before, newlines in docstrings were preserved (not affected by `inspect.cleandoc`/`inspect.getdoc`), and it caused issues long wrapped docstrings.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1401,9 +1401,10 @@ def command(name=None, cls=None, **attrs):
     By default the ``help`` attribute is received automatically from the
     docstring of the function and is cleaned up with the use of
     ``inspect.cleandoc``. Additionally, like in Markdown, single newlines in
-    the cleaned docstring will be transformed into spaces, and double newlines
-    will be transformed into single newlines. If the docstring is ``bytes``,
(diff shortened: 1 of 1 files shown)


iomintz:
I like the idea, but the roundtrip through `'\r'` is troublesome. I think a single-pass approach that goes over the string and emits a StringIO as it goes would be more efficient.

Hawkpath:
@iomintz I took your advice and replaced the `\r` hack with a loop that counts consecutive newlines and writes to a StringIO stream. Thank you for your suggestion!

Rapptz:
I don't particularly like this idea too much.

1. It's a bit computationally heavy since it's now transforming the docstring into a different representation.
2. I think it doesn't really meet my expectations on how docstrings work. 
3. The implementation seems a bit too complicated.

I think if you were to want to do this, you can have a helper method do it for you within your help command class.

PR #5975, opened 2020-10-27
Add linesep kwarg to Paginator

PikalaxALT:
## Summary

Extends the commands.Paginator class to allow custom line separators, for example if you want to
- add individual words rather than whole lines (useful if your input material is all on one line)
- make an Emoji paginator with no spaces between Emoji objects
- clap 👏 at 👏 your 👏 friends

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/help.py
+++ b/discord/ext/commands/help.py
@@ -81,6 +81,7 @@ class Paginator:
         The maximum amount of codepoints allowed in a page.
     linesep: :class:`str`
         The character string inserted between lines. e.g. a newline character.
+            .. versionadded:: 1.6
     """
     def __init__(self, prefix='```', suffix='```', max_size=2000, linesep='\n'):
         self.prefix = prefix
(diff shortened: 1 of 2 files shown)


discord/ext/commands/help.py
```diff
@@ -162,8 +169,8 @@ def pages(self):
         return self._pages
 
     def __repr__(self):
-        fmt = '<Paginator prefix: {0.prefix} suffix: {0.suffix} max_size: {0.max_size} count: {0._count}>'
-        return fmt.format(self)
+        fmt = '<Paginator prefix: {0.prefix} suffix: {0.suffix} linesep: {1} max_size: {0.max_size} count: {0._count}>'
+        return fmt.format(self, repr(self.linesep))
```
Rapptz, line 173:
```suggestion
        fmt = '<Paginator prefix: {0.prefix} suffix: {0.suffix} linesep: {0.linesep!r} max_size: {0.max_size} count: {0._count}>'
        return fmt.format(self)
```

discord/ext/commands/help.py
```diff
@@ -79,18 +79,21 @@ class Paginator:
         The suffix appended at the end of every page. e.g. three backticks.
     max_size: :class:`int`
         The maximum amount of codepoints allowed in a page.
+    linesep: :class:`str`
+        The character string inserted between lines. e.g. a newline character.
```
Rapptz, line 83: Needs a versionadded string here.

Rapptz (changes requested):
Truthfully I'm a bit unsure whether there's much use case that makes this PR worth merging. The interface in general assumes that things are line separated both in naming and in the original code and I don't know how common this operation is.

discord/ext/commands/help.py
PikalaxALT: Might as well repr the prefix and suffix too while I'm at it.

Rapptz (approved):
I guess this is fine. Though admittedly I'm still not sure I get the use cases.

PR #6002, opened 2020-11-03
Update MessageConverter regexes to support message links with "www" and remove redundant parts

JonathanFeenstra:
## Summary

This updates the `MessageConverter` link regex to support message links with "www", which are valid but previously not recognised by the `convert` method. In addition, some redundant parts in both the `id_regex` and the `link_regex` are removed:
- the `^` at the start of both regexes, which is unnecessary since `match` only matches at the beginning of the string anyway
- the parentheses around the guild ID and `@me`, these groups are not used for anything

## Checklist

- [x] This PR fixes an issue.


Rapptz (approved):
LGTM. Thanks.

PR #6006, opened 2020-11-04
Add support for css-style shorthand colours

alvesvaren:
## Summary

Adds support for css shorthand colour codes like `#abc` (which would be expanded to `#aabbcc`) to the colour converter in the commands extension.
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -429,7 +429,7 @@ class ColourConverter(Converter):
         if arg[0] == '#':
             arg = arg[1:]
 
-        if len(arg) == 3:
+        if "#" in argument and len(arg) == 3:
             arg = ''.join(i * 2 for i in arg)
         try:
             value = int(arg, base=16)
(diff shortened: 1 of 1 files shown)


avayert:
This breaks the existing conversion of the `discord.Colour` classmethod `red`

apple502j:
I think it should be good if the if is inside try. Not tested yet though

alvesvaren:
> This breaks the existing conversion of the `discord.Colour` classmethod `red`

Oh, didn't think of that, will fix

alvesvaren:
Now it only works if you have a `#` in the colour code too

oliver-ni:
There's a lot of different ways to do it, but maybe something like this is more consistent with previous behavior..

```py

    # ...

    try:
        if len(arg) <= 3:
            value = int(''.join(i * 2 for i in arg), base=16)
        else:
            value = int(arg, base=16)
        # ...

    # ...
```

Rapptz:
I don't think I really like this behaviour, sorry.

PR #6018, opened 2020-11-07
Modifies documentation for wait_for to handle emojis

mikeshardmind:
## Summary

This modifies documentation regarding wait_for surrounding emoji use which may not be obvious if people are unaware of the specifics of unicode. I suspect the docs may also benefit from a section on emojis, unicode, and discord but that is not part of this change set.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -920,17 +920,21 @@ class Client:
             async def on_message(message):
                 if message.content.startswith('$thumb'):
                     channel = message.channel
-                    await channel.send('Send me that \N{THUMBS UP SIGN} reaction, mate')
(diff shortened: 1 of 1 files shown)


mikeshardmind:
Closing this, thumbs up appears to be a special case where discord isn't respecting the spec, and I'm too lazy to test each character that should have this according to unicode to see if discord does it right or not for each.

mikeshardmind:
Since I didn't get around to explaining what was found yesterday.

1. Slightly overzealous use of Variation Selector-16 with it being valid but unnecessary in this case.
2. Twemoji (which discord uses) doesn't recognize thumbs up with variation selector-16 (specifically, any character with default emoji presentation) (see twemoji#363 if curious on the specifics)

None of the examples in the docs need adjustment, but there may still be a reason to add a section on emoji handling later on.

PR #6028, opened 2020-11-10
add description kwarg in CogMeta

### LINKED ISSUE #6024: Custom description for Cogs
## The Problem

Cogs don't seem to allow their description to be overridden except for providing a docstring on the class. This becomes a problem if we want to put technical documentation on the docstring or otherwise change the description during program execution.

## The Ideal Solution

Allow the description to be set during instantiation (in the metaclass `CogMeta`), similar to how name and command_attrs is set (https://github.com/Rapptz/discord.py/blob/abe54f5bdc48b2a92fd2a0d158ac15655049174d/discord/ext/commands/cog.py#L94). 

In addition, allow description to be assignable (currently it is a read-only property) on the cog.

## The Current Solution

Set docstring (but we lose the option to configure the description later + are forced to limit what is in it)

## Summary

Allow cog description to be better customisable

SebbyLaw:
## Summary

Resolves #6024

Not sure if this is considered a breaking change or not.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/cog.py
+++ b/discord/ext/commands/cog.py
@@ -95,9 +95,9 @@ class CogMeta(type):
         attrs['__cog_name__'] = kwargs.pop('name', name)
         attrs['__cog_settings__'] = command_attrs = kwargs.pop('command_attrs', {})
 
-        attrs['description'] = description = kwargs.pop('description', None)
+        attrs['__cog_description__'] = description = kwargs.pop('description', None)
         if description is not None:
-            attrs['description'] = inspect.getdoc(cls)
+            attrs['__cog_description__'] = inspect.cleandoc(attrs.get('__doc__', ''))
 
         commands = {}
         listeners = {}
@@ -163,10 +163,6 @@ class Cog(metaclass=CogMeta):
     When inheriting from this class, the options shown in :class:`CogMeta`
     are equally valid here.
 
-    Attributes
-    -----------
-    description: :class:`str`
-        Returns the cog's description, typically the cleaned docstring.
     """
 
     def __new__(cls, *args, **kwargs):
--- a/discord/ext/commands/cog.py
+++ b/discord/ext/commands/cog.py
@@ -98,9 +98,10 @@ class CogMeta(type):
         attrs['__cog_name__'] = kwargs.pop('name', name)
         attrs['__cog_settings__'] = command_attrs = kwargs.pop('command_attrs', {})
 
-        attrs['__cog_description__'] = description = kwargs.pop('description', None)
+        description = kwargs.pop('description', None)
         if description is None:
-            attrs['__cog_description__'] = inspect.cleandoc(attrs.get('__doc__', ''))
+            description = inspect.cleandoc(attrs.get('__doc__', ''))
+        attrs['__cog_description__'] = description
 
         commands = {}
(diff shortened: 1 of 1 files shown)


ShineyDev:
To resolve the linked issue this PR should also make Cog.description a writable property.

> Not sure if this is considered a breaking change or not.

Adding a new optional kwarg is not a breaking change.

discord/ext/commands/cog.py
```diff
@@ -93,6 +95,10 @@ def __new__(cls, *args, **kwargs):
         attrs['__cog_name__'] = kwargs.pop('name', name)
         attrs['__cog_settings__'] = command_attrs = kwargs.pop('command_attrs', {})
 
+        attrs['description'] = description = kwargs.pop('description', None)
```
Rapptz, line 98:
```suggestion
        attrs['__cog_description__'] = description = kwargs.pop('description', None)
```

discord/ext/commands/cog.py
```diff
@@ -93,6 +95,10 @@ def __new__(cls, *args, **kwargs):
...
+        if description is not None:
+            attrs['description'] = inspect.getdoc(cls)
```
Rapptz, line 100:
```suggestion
            attrs['__cog_description__'] = inspect.cleandoc(attrs.get('__doc__', ''))
```

discord/ext/commands/cog.py
```diff
@@ -156,6 +162,11 @@ class Cog(metaclass=CogMeta):
 
     When inheriting from this class, the options shown in :class:`CogMeta`
     are equally valid here.
+
+    Attributes
+    -----------
+    description: :class:`str`
+        Returns the cog's description, typically the cleaned docstring.
```
Rapptz, line 169:
```suggestion
```

Rapptz (changes requested):
I prefer having the property rather than a plain class attribute since this makes it less likely for the user to shoot themselves on the foot when they write e.g. `cog.description = "hi"` which should error out since it did before. I guess it's up to debate whether having it read-only is necessary though.

This probably also needs a versionadded.

discord/ext/commands/cog.py
```diff
@@ -93,6 +98,10 @@ def __new__(cls, *args, **kwargs):
         attrs['__cog_name__'] = kwargs.pop('name', name)
         attrs['__cog_settings__'] = command_attrs = kwargs.pop('command_attrs', {})
 
+        attrs['__cog_description__'] = description = kwargs.pop('description', None)
+        if description is None:
+            attrs['__cog_description__'] = inspect.cleandoc(attrs.get('__doc__', ''))
```
Rapptz, line 103:
```suggestion
        description = kwargs.pop('description', None)
        if description is None:
            description = inspect.cleandoc(attrs.get('__doc__', ''))
        attrs['__cog_description__'] = description
```

Rapptz (changes requested):
If this is meant to be modifiable then adding a property setter would probably work at this point.

Rapptz (approved):

Harmon758:
This was merged with https://github.com/Rapptz/discord.py/commit/4adbe03d7ca03fce3411bea7bcd9e3619513ad94 and can be closed now.

PR #6035, opened 2020-11-12
HTTPClient: set a real user-agent string for selfbots

iomintz:
## Summary

With the current discord.py user-agent string, selfbots get instantly flagged for email re-verification as soon as they make a request, and are subsequently prevented from making any further HTTP requests. This PR fixes that by setting a "real" user-agent string in static_login() if the user agent has not already been set. The `None` check is there to preserve the current behavior where the user can write:

```py
client = discord.Client(self_bot=True)
client.http.user_agent = SOME_USER_AGENT_STRING
```

**Yes, I am aware that self-bots are against Discord TOS.** However, discord.py supports them, and so I believe that if they do not work it is a bug in the library.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -108,9 +108,7 @@ class HTTPClient:
         self.proxy = proxy
         self.proxy_auth = proxy_auth
         self.use_clock = not unsync_clock
-
-        user_agent = 'DiscordBot (https://github.com/Rapptz/discord.py {0}) Python/{1[0]}.{1[1]} aiohttp/{2}'
-        self.user_agent = user_agent.format(__version__, sys.version_info, aiohttp.__version__)
+        self.user_agent = None
 
     def recreate(self):
         if self.__session.closed:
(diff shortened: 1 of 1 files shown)


muddyfish:
This causes issues in any system which runs bots in a Discord proxy service that changes the auth headers.

SebbyLaw:
@muddyfish Can you provide an example of this? The `None` check in this PR should prevent it from breaking existing code.

muddyfish:
I know a number of large bots out there which proxy their requests to Discord though services such as nginx.
My own system authenticates to nginx by preventing to be a selfbot on dpy's side to a avoid the added Bot header and providing my own Auth credentials which get swapped out. 
This helps prevent token leaks in the case of a  security issue as well as adds the capacity for additional logging and shared ratelimit handling between instances.

iomintz:
I'm having trouble understanding this use-case. Are you saying that on the d.py side, your code says it's a real bot, and then nginx proxies these requests to make them look like a selfbot? But this PR only changes the User-Agent string for bot=True code, so if I'm understanding your post correctly, this PR would not affect that.

muddyfish:
No, I'm saying that dpy thinks it's a selfbot, and thus this code would break it by choosing alternate headers which aren't valid for bots.

bijij:
I feel as if it would be more sensible to take a user_agent as a kwarg somewhere in the chain.
And if not supplied use the default.

Allows for userbots to set their own agent, and removes the hardcoded browser user agent from the codebase.

Rapptz:
I don't necessarily agree with the idea that this is a breaking change. Sure, it's a breaking change if you're relying on an internal behaviour that is not documented but this isn't something I've ever guaranteed to not mess with, since otherwise it would preclude me from refactoring code.

To that end, this PR isn't something I want to accept either since it's just playing a cat and mouse game. I am unsure on how long the client code is going to live in the repo in the future either since it's something I want to think about when 2.0 comes along.

Issue #1719, opened 2018-11-11
PyNaCl warning issued, even though VoiceClient not being used


sqldiablo:
I'm using discord.py via Home Assistant, and I'm receiving the warning that PyNaCl isn't installed, and voice will not be available, but all Home Assistant is using is the text features. Can a feature be added to allow disocrd.py to disable voice altogether or to just suppress the PyNaCl warning when voice features aren't being actively used? It seems silly to fire a warning every time discord.py is initialized, even though the voice features aren't being used.

Phxntxm:
Installing `discord.py` should install the package without voice support. Installing `discord.py[voice]` will include the voice requirements. I'm fairly certain that it does not try to install pynacl with the first option.

sqldiablo:
Correct, but what pip installs or doesn't isn't the issue I'm reporting. The issue is that discord.py still logs a warning that PyNaCl isn't installed, even when I've only installed the non-voice version and have made no attempt to use voice features. Just by importing/initializing discord.py, discord.py does a check for PyNaCl and logs a warning if it's not present.

sqldiablo:
Basically it's logging noise that I'm trying to get rid of. It shouldn't warn you until you try to initialize\use voice features.

sqldiablo:
Or if there's some way I could initialize it with a parameter telling it to disable or ignore voice features, that would work too.

Rapptz:
You are free to do `discord.VoiceClient.warn_pynacl = False`.

sqldiablo:
Thanks @Rapptz, that took care of it.

PR #6051, opened 2020-11-17
Use warnings instead of logging for the PyNaCl warning

iomintz:
## Summary

According to https://docs.python.org/3/howto/logging.html#when-to-use-logging
logging.warning() should only be used if the situation cannot be resolved.

This is a draft PR because I would like a review on the documentation changes first.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/__init__.py
+++ b/discord/__init__.py
@@ -54,7 +54,7 @@ from .mentions import AllowedMentions
 from .shard import AutoShardedClient, ShardInfo
 from .player import *
 from .webhook import *
-from .voice_client import VoiceClient, VoiceProtocol
+from .voice_client import VoiceClient, VoiceProtocol, NaClWarning
 from .audit_logs import AuditLogChanges, AuditLogEntry, AuditLogDiff
 from .raw_models import *
 from .team import *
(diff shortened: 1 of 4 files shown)


iomintz:
I just noticed https://github.com/Rapptz/discord.py/issues/1719#issuecomment-437703581. I did not realize that this was possible, but I still believe the current code to be a misuse of `logging.warning`. I will update the metadata for this change set to reflect this.

Rapptz:
I don't think these guidelines are strictly true. The standard library even offers a way to convert `warning` module calls into `logging` calls.

iomintz:
I don't see that in the docs, would you mind linking to that?

Rapptz:
https://docs.python.org/3/library/logging.html#logging.captureWarnings

iomintz:
I don't see how this is a counterpoint to the guidelines employed in this PR, as it's a conversion in the opposite direction.

Rapptz:
I don't really see much of a point in this PR. The warning mechanism is quite bad, most go ignored by default and while that's true from a logging perspective as well this is remedied by the fact that its usage is documented and there is a way for the library to aid the user into seeing the logs.

Anyway a better suggestion would be to just change this warning completely or remove it but that's for another patch set.

PR #6061, opened 2020-11-18
Implement discord.Message.reply

### LINKED ISSUE #6054: Support the new Reply feature
## The Problem

Discord recently added support for inline replies to the client: https://support.discord.com/hc/en-us/articles/360057382374-Replies-FAQ

Right now, discord.py only provides partial support for this functionality.

## The Ideal Solution

I should be able to know when a given message is a reply to another. The existing `Message.reference` field gives me a `MessageReference`, to the source message, but the documentation needs to be updated to reveal that it will also be populated for replies. `MessageType` should have a `reply` member so I can tell it is a reply (looks like #5200 implements this part) - right now on a reply the `type` is `MessageType.default`.

I should also be able to create a new reply to an existing message. Perhaps the `Message` object could have a `reply` coroutine for this purpose.

## The Current Solution

Only partial functionality exists today. To determine if a message is a reply, I would need to check that the message is neither a cross-post nor a `MessageType.pins_add`. There is currently no solution for sending replies (other than perhaps interacting with `client.http` directly).

## Summary

Please add support for the new replies feature. Thank you!

PikalaxALT:
## Summary

Implements the ability to reply to messages, which makes use of the MessageReference class. This is a new feature in Discord that will soon be available in all servers. Resolves #6054.

## Checklist

- [x] This PR fixes an issue.


discord/message.py
```diff
@@ -1114,3 +1148,74 @@ async def ack(self):
         if state.is_bot:
             raise ClientException('Must not be a bot account to ack messages.')
         return await state.http.ack_message(self.channel.id, self.id)
+
+    async def reply(self, content=None, *, tts=False, embed=None, file=None,
+                                          files=None, delete_after=None, nonce=None,
+                                          allowed_mentions=None):
+        """|coro|
```
NCPlayz, line 1155: this whole docstring can be simplified to refer to `Messageable.send`. E.G.: https://github.com/Rapptz/discord.py/blob/93fa46713a198baf08ab1e760be1adfdcb852c97/discord/channel.py#L838-L848

discord/abc.py
```diff
@@ -845,6 +845,11 @@ async def send(self, content=None, *, tts=False, embed=None, file=None,
 
             .. versionadded:: 1.4
 
+        message_reference: :class:`discord.MessageReference`
+            Message to which you are replying. If this is passed, then you should
```
Willy-C, line 849: Looks like you are missing some words here

discord/message.py
PikalaxALT: Simplified call signature and inserted docs reference to `Messageable.send`.

discord/abc.py
PikalaxALT: Fixed, and made other aesthetic repairs where new features are documented.

discord/message.py
```diff
@@ -1114,3 +1148,30 @@ async def ack(self):
         if state.is_bot:
             raise ClientException('Must not be a bot account to ack messages.')
         return await state.http.ack_message(self.channel.id, self.id)
+
+    async def reply(self, content=None, **kwargs):
```
NCPlayz, line 1152: Might be worth adding a shorthand to `Context`?

discord/mentions.py
```diff
@@ -101,6 +107,9 @@ def to_dict(self):
         elif self.roles != False:
             data['roles'] = [x.id for x in self.roles]
 
+        if self.replied_user == True:
```
numbermaniac, line 110: Could this be simplified to just `if self.replied_user:`?

discord/message.py
```diff
@@ -232,6 +232,24 @@ def __init__(self, state, **kwargs):
         self.guild_id = utils._get_as_snowflake(kwargs, 'guild_id')
         self._state = state
 
+    @classmethod
+    def from_message(cls, message):
+        """Creates a :class:`MessageReference` from an existing :class:`Message`
```
numbermaniac, line 237:
```suggestion
        """Creates a :class:`MessageReference` from an existing :class:`Message`.
```

discord/message.py
```diff
@@ -232,6 +232,24 @@ def __init__(self, state, **kwargs):
...
+
+        Parameters
+        ----------
+        message: :class:`Message`
+            The message to be converted into a reference.
+
+        Returns
+        -------
+        :class:`MessageReference`
+            A reference to the message
```
numbermaniac, line 249:
```suggestion
            A reference to the message.
```

discord/message.py
nihaals: This can be marked as resolved

nihaals (approved):

discord/message.py
PikalaxALT: Already done

scrazzz (approved):

wasi-master (approved):

alyssadev:
Sorry, I forgot that mentioning issues tags them to the PR. please ignore :)

pikaninja (approved):

marzeq (approved):

discord/abc.py
```diff
@@ -845,6 +845,13 @@ async def send(self, content=None, *, tts=False, embed=None, file=None,
 
             .. versionadded:: 1.4
 
+        message_reference: :class:`~discord.MessageReference`
```
Rapptz, line 848: I don't really like the design of this parameter and its typing. Having the user explicitly pass in a wrapped class is verbose. It's probably better to allow an implicit conversion so a user can just pass in a `Message` object as well.

```suggestion
        reference: Union[:class:`~discord.Message`, :class:`~discord.MessageReference`]
```

There is however, an issue with the way this is laid out. Currently either approach requires a full `Message` object, which might not be desirable. I think `MessageReference` should be reconstructed in such a way that it functions similar to `PartialEmoji` where the state parameter injected is optional and only done if within the library -- that way a user can construct their own `MessageReference` without relying on any sort of message caching at all and just operating on IDs. Right now the `state` is only used in a few scenarios and it needs to be guarded.

discord/message.py
```diff
@@ -240,6 +258,29 @@ def cached_message(self):
...
+        -------
+        specify_channel: Optional[:class:`bool`]
+            Whether to include the channel ID in the returned object.
+            Defaults to False.
+
+        Returns
+        -------
+        :class:`dict`
+            The reference as a dict.
+        """
```
Rapptz, line 276:
```suggestion
```

Please don't document internal functions.

docs/whats_new.rst
```diff
@@ -11,6 +11,16 @@ Changelog
...
+.. _vp2p0p0:
+
+v1.6.0
+-------
+
+API Changes
+~~~~~~~~~~~
+
+- Add support for replying to a :class:`Message` (:issue:`6054`)
+
```
Rapptz, line 23:
```suggestion
```

Changelogs are done by hand by me when release time comes.

Rapptz (changes requested):
The UX of this PR needs a lot of work, I think. I've left comments explaining why in detail.

Lost1419 (approved):

docs/whats_new.rst
PikalaxALT: Ah, gotcha.

discord/ext/commands/context.py
```diff
@@ -322,3 +322,41 @@ async def send_help(self, *args):
...
+
+        Returns
+        ---------
+        :class:`~discord.Message`
+            The message that was sent.
+        """
+        allowed_mentions = kwargs.pop('allowed_mentions', discord.AllowedMentions())
+        if mention_author is not None:
+            allowed_mentions.replied_user = mention_author
+        return await self.message.reply(content, allowed_mentions=allowed_mentions, **kwargs)
```
bijij, line 362: Why is this not a duplicate of `Message.reply`?
Would lower redundancy and allow you to do `reply.__doc__ = Message.reply.__doc__`.

discord/message.py
```diff
@@ -1114,3 +1140,43 @@ async def ack(self):
         if state.is_bot:
             raise ClientException('Must not be a bot account to ack messages.')
         return await state.http.ack_message(self.channel.id, self.id)
+
+    async def reply(self, content=None, **kwargs):
```
bijij, line 1144: `Context.reply` has a `mention_author` bool why doesn't this? (ties in with my above comment)

bijij (changes requested):
Confused about differences between `Message.reply` and `Context.reply`.

discord/abc.py
```diff
@@ -875,6 +882,12 @@ async def send(self, content=None, *, tts=False, embed=None, file=None,
         else:
             allowed_mentions = state.allowed_mentions and state.allowed_mentions.to_dict()
 
+        if reference is not None:
+            try:
+                reference = reference.to_dict()
```
NCPlayz, line 887: this seems a bit shaky and probably makes sense to check if it's a message or messagereference, and raise if neither

discord/message.py
```diff
@@ -1114,3 +1140,43 @@ async def ack(self):
...
+
+        Returns
+        ---------
+        :class:`Message`
+            The message that was sent.
+        """
+
+        return await self.channel.send(content, reference=self, **kwargs)
+
+    def make_reference(self):
```
NCPlayz, line 1170:
```suggestion
    def to_reference(self):
```

the rest of the library generally uses `to_x` and `from_x`

discord/message.py
```diff
@@ -1114,3 +1140,43 @@ async def ack(self):
...
+
+        .. versionadded:: 1.6
+
+        Returns
+        ---------
+        :class:`MessageReference`
+            The reference to this message.
+        """
+
+        return MessageReference(self._state, message_id=self.id, channel_id=self.channel.id, guild_id=self.guild and self.guild.id)
```
bijij, line 1182: Why not `MessageReference.from_message(self)`

discord/message.py
```diff
@@ -1114,3 +1140,43 @@ async def ack(self):
...
+        ---------
+        :class:`Message`
+            The message that was sent.
+        """
+
+        return await self.channel.send(content, reference=self, **kwargs)
+
+    def make_reference(self):
+        """
+        Creates a :class:`MessageReference` from the current message.
```
NCPlayz, line 1172:
```suggestion
        """Creates a :class:`MessageReference` from the current message.
```

discord/message.py
```diff
@@ -232,6 +232,24 @@ def __init__(self, state, **kwargs):
...
+        ----------
+        message: :class:`Message`
+            The message to be converted into a reference.
+
+        Returns
+        -------
+        :class:`MessageReference`
+            A reference to the message.
+        """
+        return cls(message._state, message_id=message.id, channel_id=message.channel.id, guild_id=message.guild and message.guild.id)
```
bijij, line 251: Prefer `getattr(message.guild, 'id', None)` over  `message.guild and message.guild.id`

discord/ext/commands/context.py
```diff
@@ -322,3 +322,41 @@ async def send_help(self, *args):
...
+        ~discord.InvalidArgument
+            The ``files`` list is not of the appropriate size or
+            you specified both ``file`` and ``files``.
+
+        Returns
+        ---------
+        :class:`~discord.Message`
+            The message that was sent.
+        """
+        allowed_mentions = kwargs.pop('allowed_mentions', discord.AllowedMentions())
```
bijij, line 359: this should make a copy of the bots allowed_mentions object instead of making a new one. conscider:
`await ctx.reply('@everyone')` may have unexpected consequences.

NCPlayz (changes requested):
looked over some of the docstrings.

discord/abc.py
PikalaxALT: Can't import from message.py due to circular reference, would have to use `hasattr` or let the `AttributeError` be raised.

Rapptz: There are a few methods I use to resolve circular imports.

1) Shared private attribute. So, for example there'd be a tag property such as `self._can_cast_ref` that can be used without importing anything. I don't exactly like this approach here.
2) Private tag type that can be imported standalone that the types inherit from. Basically the same as #1 except using inheritance and `isinstance` instead of flimsy attributes. See, e.g. `BaseActivity` or `_EmojiTag` for examples in the library.

discord/message.py
Rapptz: Seconding this. Casts are done using `to_x`/`from_x`.

Rapptz (changes requested):
The previous review nits by NC still seem to be valid as well.

discord/ext/commands/context.py
PikalaxALT: This gets resolved in the `abc.Messageable.send` call. The new AllowedMentions has everything set to `default` which is a _FakeBool.

discord/message.py
PikalaxALT: Benchmarking tests concur, `getattr` is significantly faster.

discord/mentions.py
```diff
@@ -59,30 +59,37 @@ class AllowedMentions:
         roles are not mentioned at all. If a list of :class:`abc.Snowflake`
         is given then only the roles provided will be mentioned, provided those
         roles are in the message content.
+    replied_user: :class:`bool`
+        Whether to mention the author of the message being replied to. Defaults
+        to ``True``. This is different from the client behavior which defaults
```
NCPlayz, line 64:
```suggestion
        to ``True``. This is different from the client behaviour which defaults
```

NCPlayz (changes requested):
Just another doc inconsistency. Also, since `MessageReference` is now in its own file, the import in `__init__.py` should reference it from the module rather than through `message.py`.

discord/abc.py
```diff
@@ -848,15 +849,24 @@ async def send(self, content=None, *, tts=False, embed=None, file=None,
 
             .. versionadded:: 1.4
 
+        reference: Union[:class:`~discord.Message`, :class:`~discord.MessageReference`]
+            A reference to the :class:`~discord.Message` to which you are replying, this can be created using
+            :meth:`~discord.MessageReference.from_message` or passed directly as a :class:`~discord.Message`. You can control
+            whether this mentions the author of the referenced Message using :attr:`~discord.AllowedMentions.replied_user`.
```
NCPlayz, line 855:
```suggestion
            whether this mentions the author of the referenced message using :attr:`~discord.AllowedMentions.replied_user`.
```

discord/abc.py
```diff
@@ -878,6 +888,12 @@ async def send(self, content=None, *, tts=False, embed=None, file=None,
         else:
             allowed_mentions = state.allowed_mentions and state.allowed_mentions.to_dict()
 
+        if reference is not None:
+            if isinstance(reference, _MessageType):
+                reference = (reference if isinstance(reference, MessageReference) else reference.to_reference()).to_dict()
```
NCPlayz, line 893: this inline if statement isn't exactly readable; something like this could suffice it.
```suggestion
                if not isinstance(reference, MessageReference):
                    reference = reference.to_reference()
                reference = reference.to_dict()
```

NCPlayz (changes requested):

discord/message.py
```diff
@@ -1127,3 +1099,55 @@ async def ack(self):
...
+
+        Returns
+        ---------
+        :class:`Message`
+            The message that was sent.
+        """
+
+        allowed_mentions = kwargs.pop('allowed_mentions', AllowedMentions())
+        if mention_author is not None:
+            allowed_mentions.replied_user = mention_author
```
Rapptz, line 1139: This implementation modifies what the user passes in which is undesirable.

discord/message_reference.py
```diff
@@ -0,0 +1,87 @@
...
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+"""
+
+from . import utils
+
+class _MessageType:
+    __slots__ = ()
+
+class MessageReference(_MessageType):
```
Rapptz, line 32: I don't really understand why this was moved to a new file.

discord/message_reference.py
```diff
@@ -0,0 +1,87 @@
...
+
+    @property
+    def cached_message(self):
+        """Optional[:class:`~discord.Message`]: The cached message, if found in the internal message cache."""
+        return self._state._get_message(self.message_id)
+
+    def __repr__(self):
+        return '<MessageReference message_id={0.message_id!r} channel_id={0.channel_id!r} guild_id={0.guild_id!r}>'.format(self)
+
+    def to_dict(self, specify_channel=True):
```
Rapptz, line 81: I don't get the point of this `specify_channel` parameter.

discord/mentions.py
```diff
@@ -59,30 +59,37 @@ class AllowedMentions:
         roles are not mentioned at all. If a list of :class:`abc.Snowflake`
         is given then only the roles provided will be mentioned, provided those
         roles are in the message content.
+    replied_user: :class:`bool`
+        Whether to mention the author of the message being replied to. Defaults
+        to ``True``. This is different from the client behaviour which defaults
+        to ``False`` in some contexts such as DMs.
```
Rapptz, line 65: I don't think it's worth mentioning this.

Rapptz (changes requested):

discord/message_reference.py
PikalaxALT: To avoid circular imports in abc.py

discord/message_reference.py
PikalaxALT: The API allows you to omit this, in which case the channel ID is filled in using the ID of the channel in which the request is made. This does affect the behavior resulting from replying to a message in a different channel.

- `specify_channel=True`: 400 Bad Request: cannot reply to a message from a different channel
- `specify_channel=False`: 400 Bad Request: unknown message

Since the former behavior is desired (in that it lays out explicitly why the request is bad), I have opted to remove the kwarg entirely.

Rapptz (approved):
I still have issues with this pull request in terms of code quality but since there's too much back and forth I can just handle these cases locally.

Thanks for the PR.

PR #6063, opened 2020-11-19
Properly document that categories throw 404's on create_invite.

webtax-gh:
## Summary

Categories will throw 404's when channel.create_invite is used. This was a bit confusing for me since
```py
;jsk py
guild = _ctx.guild
guildchannels = [channel for channel in guild.channels]
channel = guildchannels[0]
invite_link = await channel.create_invite()
return await _ctx.send(invite_link)
```
threw NotFound, which was because the first channel was a category, it would be nice if the docs properly showed that categories will throw this error.
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -739,7 +739,7 @@ class GuildChannel:
             Invite creation failed.
 
         ~discord.NotFound
-            The channel that was passed is a category or a invalid channel.
+            The channel that was passed is a category or an invalid channel.
 
         Returns
         --------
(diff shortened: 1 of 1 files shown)


discord/abc.py
```diff
@@ -738,6 +738,9 @@ async def create_invite(self, *, reason=None, **fields):
         ~discord.HTTPException
             Invite creation failed.
 
+        ~discord.NotFound
+            The channel that was passed is a category or a invalid channel.
```
WickedTree, line 742:
```suggestion
            The channel that was passed is a category or an invalid channel.
```
Fixed Grammar

Rapptz (approved):
Seems fine. Thanks.

PR #6067, opened 2020-11-19
Add Colour.random factory method

webkaiyo:
## Summary

Add a helper method to generate random colors.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/colour.py
+++ b/discord/colour.py
@@ -25,6 +25,7 @@ DEALINGS IN THE SOFTWARE.
 """
 
 import colorsys
+import random
 
 class Colour:
     """Represents a Discord role colour. This class is similar
@@ -116,6 +117,11 @@ class Colour:
     def default(cls):
         """A factory method that returns a :class:`Colour` with a value of ``0``."""
         return cls(0)
+
+    @classmethod
+    def random(cls):
+        """A factory method that returns a :class:`Colour` with a random value."""
+        return cls.from_hsv(random.random(), 1, 1)
 
     @classmethod
     def teal(cls):


Rapptz (commented):
I don't dislike the idea of this (and it is using proper random rather than the naive one), however this currently changes the `discord.Colour` converter behaviour. i.e. you can now type `random` as an input and get a random colour. Maybe this is fine though.

webkaiyo:
I particularly found it interesting to be able to use "random" as an input and receive a random color, if you want I can change the behavior of the converter.

Rapptz:
I think it's fine after giving it some thought.

averwhy:
Should there be an alias for this like `random_color`/`random_colour`?

webkaiyo:
> I think it's fine after giving it some thought.

Nice! Thanks for taking the time to review.

discord/colour.py
```diff
@@ -117,6 +118,11 @@ def default(cls):
         """A factory method that returns a :class:`Colour` with a value of ``0``."""
         return cls(0)
 
+    @classmethod
+    def random(cls):
+        """A factory method that returns a :class:`Colour` with a random value."""
```
mikeshardmind, line 123: This probably doesn't matter and would only be a point of clarity and preventing future pedantry, but having this say that it returns a color with a random hue (and either saying nothing of the saturation or value, or saying that those are both maxed in this) rather than using the word value which has specific meaning in discussing colors different to the use here may make some sense. (even though most people would realize that a random value without mentioning hue is a much less likely interpretation)

Rapptz (approved):
Thanks. I'll merge and do some fix ups in the docstrings.

joebyjo:
how many possible random colors can u get using the function?

diceroll123:
However many numbers are between `[0.0, 1.0)`

...sorta.

mikeshardmind:
https://stackoverflow.com/questions/13539420/in-rgb-model-how-many-distinct-hues-are-available#13539845


119,694

PR #6068, opened 2020-11-19
Add MessageReference.fetch_message

nihaals:
## Summary

This adds a convenient way to *fetch* the reference's message. Currently, you can only easily *get* the message.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -341,6 +341,8 @@ class MessageReference:
 
         Retrieves the :class:`Message` from the destination.
 
+        The channel or guild may be a partial object if it is not found.
+
         This can only be used by bot accounts.
 
         .. versionadded:: 2.0
--- a/discord/message.py
+++ b/discord/message.py
@@ -44,6 +44,7 @@ from .utils import escape_mentions
 from .guild import Guild
 from .mixins import Hashable
 from .sticker import Sticker
+from .object import Object
 
 __all__ = (
     'Attachment',
@@ -361,10 +362,20 @@ class MessageReference:
             The message asked for.
         """
 
-        channel = self._state.get_channel(self.channel_id)
+        state = self._state
+        if self.guild_id:
+            guild = state.get_guild(self.guild_id)
+            if guild:
+                channel = guild.get_channel(self.channel_id)
+        else:
+            channel = state.get_channel(self.channel_id)
+
+        data = await state.http.get_message(self.channel_id, self.message_id)
         if channel is None:
-            raise ChannelNotFound(self.channel_id)
-        return await channel.fetch_message(self.message_id)
+            channel = Object(id=self.channel_id)
+            if self.guild_id:
+                channel.guild = Object(id=self.guild_id)
+        return state.create_message(channel=channel, data=data)
 
     def __repr__(self):
         return '<MessageReference message_id={0.message_id!r} channel_id={0.channel_id!r} guild_id={0.guild_id!r}>'.format(self)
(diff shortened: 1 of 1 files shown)


discord/message.py
```diff
@@ -237,6 +237,42 @@ def cached_message(self):
...
+        Parameters
+        ------------
+        id: :class:`int`
+            The message ID to look for.
+
+        Raises
+        --------
+        ~discord.NotFound
+            The message was not found.
+        ~discord.ChannelNotFound
```
nihaals, line 258: I'm unsure about this. `abc.Messageable.fetch_message` raises `discord.NotFound` due to the 404 so it seems retuning `None` doesn't make sense and this seems like the best exception based on the name, but `BadArgument` doesn't make much sense

discord/message.py
```diff
@@ -237,6 +237,42 @@ def cached_message(self):
...
+        ~discord.HTTPException
+            Retrieving the message failed.
+
+        Returns
+        --------
+        :class:`Message`
+            The message asked for.
+        """
+
+        channel = self._state.get_channel(self.channel_id)
```
nihaals, line 271: This could be more efficient but `ConnectionState` doesn't publicly expose any methods that make this simple. I considered adding a `guild_id` parameter to `ConnectionState.get_channel` but iterating over every guild should still be fast as there are no API requests

muddyfish: Personally I avoid iterating over every guild like the plague as it's way too easy to not notice it does this and block the bot for a noticeable period when you have a lot of guilds and call it relatively often

nihaals: I could either implement a `guild_id` parameter or just fetch the channel (but have to do possibly an unnecessary API request)
@Rapptz which would you prefer?

Rapptz: There's already a `MessageReference.guild_id`

discord/message.py
```diff
@@ -35,7 +35,7 @@
 from .partial_emoji import PartialEmoji
 from .calls import CallMessage
 from .enums import MessageType, try_enum
-from .errors import InvalidArgument, ClientException, HTTPException
+from .errors import InvalidArgument, ClientException, HTTPException, ChannelNotFound
```
Rapptz, line 38: I don't know where `ChannelNotFound` came from but that's not a standard error.

Rapptz (changes requested):

discord/message.py
nihaals: I meant add `guild_id` to `ConnectionState.get_channel`

discord/message.py
```diff
@@ -237,6 +238,49 @@ def cached_message(self):
...
+            else:
+                channel = None
+        else:
+            channel = state.get_channel(self.channel_id)
+
+        data = await state.http.get_message(self.channel_id, self.message_id)
+        if channel is None:
+            channel = Object(id=self.channel_id)
+            if self.guild_id:
+                channel.guild = Object(id=self.guild_id)
```
nihaals, line 281: Wouldn't this cause issues as it is assumed that `TextChannel.guild` is `Guild` not just an `Object`, so if a user were to access attributes other than `id` they would get an error?

Rapptz: It's a valid complaint. In v2.0 this would be swapped out with a different type named `UnavailableGuild` since the guild model would be split into two. Though that idea is not exactly finalised either.

Rapptz:
I think due to the few gotchas presented here this feature seems non-desirable.

PR #6076, opened 2020-11-22
Allow arguments to be passed to extension setup functions

bijij:
## Summary

This PR allows for `*args` and `**kwargs` to be passed to extension `setup` functions. 

Some notes about this:
 - Reduces the ability to add new features to `Bot.load_extension` and `Bot.reload_extension` without breaking changes.
 - When reloading an extension new arguments would need to be passed each time.
 - The same functionality could be acheived with the use of a botvar.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -647,6 +647,10 @@ class BotBase(GroupMixin):
             The extension name to load. It must be dot separated like
             regular Python imports if accessing a sub-module. e.g.
             ``foo.test`` if you want to import ``foo/test.py``.
+        args:
+            The position arguments to be passed to the extension setup function.
+        kwargs:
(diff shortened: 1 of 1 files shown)


discord/ext/commands/bot.py
```diff
@@ -641,6 +647,10 @@ def load_extension(self, name):
             The extension name to load. It must be dot separated like
             regular Python imports if accessing a sub-module. e.g.
             ``foo.test`` if you want to import ``foo/test.py``.
+        args:
```
NCPlayz, line 650: maybe provide an example?

Rapptz:
I've already mentioned I'm against this idea a few times already. Sorry.

I think there's a better way to do this that is more beneficial for 3rd party extensions than this fire and forget absorb everything method this uses.

PR #6089, opened 2020-11-26
Add Role Reaction Example

Kippiii:
## Summary

Adds a new example, which showcases how to give and remove roles from users through reactions.
Issue #2525

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/reaction_roles.py
+++ b/examples/reaction_roles.py
@@ -20,13 +20,13 @@ async def on_raw_reaction_add(payload):
     # Make sure that the message the user is reacting to is the one we care about
     if payload.message_id != role_message_id:
         return
-    
+
     try:
         role_id = emoji_to_role[str(payload.emoji)]
     except KeyError:
         # If the emoji isn't the one we care about then exit as well.
         return
-    
+
     guild = bot.get_guild(payload.guild_id)
     if guild is None:
         # Check if we're still in the guild and it's cached.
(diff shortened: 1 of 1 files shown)


examples/reaction_roles.py
```diff
@@ -0,0 +1,38 @@
...
+                guild = bot.get_guild(payload.guild_id)
+                role = guild.get_role(emoji_to_role[payload.emoji.name])
+                await payload.member.add_roles(role)
+
+@bot.event
+async def on_raw_reaction_remove(payload):
+    """Removes a role based on a reaction emoji."""
+    if payload.guild_id:
+        if payload.emoji.name in emoji_to_role:
+            if payload.message_id == role_message_id:
```
BluePhoenixGame, line 32: See other review ^^

examples/reaction_roles.py
```diff
@@ -0,0 +1,38 @@
+"""Uses two messages to add and remove roles through reactions."""
```
BluePhoenixGame, line 1: This docstring is no longer correct

BluePhoenixGame (changes requested):

examples/reaction_roles.py
```diff
@@ -0,0 +1,34 @@
+"""Uses a messages to add and remove roles through reactions."""
+
+import discord
+from discord.ext import commands
+
+# This bot requires the members and reactions intensions.
+intents = discord.Intents.default()
+intents.members = True
+bot = commands.Bot(command_prefix=commands.when_mentioned_or("!"), intents=intents)
```
Rapptz, line 9: This is a `Bot` instance yet it has no commands. Why?

examples/reaction_roles.py
```diff
@@ -0,0 +1,34 @@
...
+
+role_message_id = 0  # ID of message that can be reacted to to add role
+emoji_to_role = {
+    "👍": 0,  # ID of role associated with thumbs up emoji
+    "test": 0  # ID of role associated with custom emoji 'test'
+}
```
Rapptz, line 15: This should probably use the full emoji form not just the plain name, including the ID. Alternatively you can use `discord.PartialEmoji` as the key but this might seem too much for the passer-by.

examples/reaction_roles.py
```diff
@@ -0,0 +1,34 @@
...
+
+@bot.event
+async def on_raw_reaction_add(payload):
+    """Gives a role based on a reaction emoji."""
+    if payload.message_id == role_message_id and payload.emoji.name in emoji_to_role:
+        guild = bot.get_guild(payload.guild_id)
+        role = guild.get_role(emoji_to_role[payload.emoji.name])
+        await payload.member.add_roles(role)
```
Rapptz, line 23:
```suggestion
    # Make sure that the message the user is reacting to is the one we care about
    if payload.message_id != role_message_id:
        return
    
    try:
        role_id = emoji_to_role[str(payload.emoji)]
    except KeyError:
        # If the emoji isn't the one we care about then exit as well.
        return
    
    guild = bot.get_guild(payload.guild_id)
    if guild is None:
        # Check if we're still in the guild and it's cached.
        return

    role = guild.get_role(role_id)
    if role is None:
        # Make sure the role still exists and is valid.
        return

    try:
        # Finally add the role
        await payload.member.add_roles(role)
    except discord.HTTPException:
        # If we want to do something in case of errors we'd do it here.
        pass
```

Generally this is the style I would consider good for the examples. I left the other pair as an exercise.

Rapptz (changes requested):
This example could use some work.

1. Most examples in here tend to prefer subclassing over globals since it contains the state into one object rather than leaking it.
2. Nothing is really explained in here.
3. There is no error checking code in here, for example whether something is `None`.

examples/reaction_roles.py
```diff
@@ -0,0 +1,83 @@
...
+            return
+
+        try:
+            # Finally, remove the role
+            await member.remove_roles(role)
+        except discord.HTTPException:
+            # If we want to do something in case of errors we'd do it here.
+            pass
+
+# This bot requires the members and reactions intensions.
```
Rapptz, line 78:
```suggestion
# This bot requires the members and reactions intents.
```

Rapptz (approved):
Looks good to me

PR #6091, opened 2020-11-27
Create new_replies.py

cdgamedev:
added a simple example for how the new reply system works


Rapptz:
Hi, sorry this example is a bit too low quality.

PR #6132, opened 2020-12-08
Fix embed suppression when using client-wide allowed_mentions

SnowyLuma:
## Summary

Fixes being unable to suppress embeds on other users' messages when the client was created with an initial `allowed_mentions` value, as including the `allowed_mentions` field will return a 403 from the API. 

## Checklist

- [x] This PR fixes an issue.


Rapptz:
I think a better solution would be to revert this logic entirely since it doesn't really make sense and is ultimately too complicated here. For now, I think this is fine though.

Issue #5962, opened 2020-10-23
Include an __eq__ method for Embeds (v.2)


dob9601:
## The Problem

2 embeds that are identical won't necessarily be flagged as equal when passed through an equivalence check. However, the dictionaries returned by their to_dict methods do.

## The Ideal Solution

Add the following method or something to the same effect to the Embed class:
```py
def __eq__(self, other):
    return self.to_dict() == other.to_dict()
```

## The Current Solution

Check the equivalence of the dictionaries of the embeds instead of the overall objects:
```py
self.to_dict() == other.to_dict()
```

## Summary

TL;DR aren't necessarily equivalent even if their dictionaries are identical

Rapptz:
I'm not sure there's a way to make this equality work without it being overly expensive so I'm against it purely on those grounds

PR #6148, opened 2020-12-10
Add __eq__ method to Embed object

NotTani:
Resolves issue #5962, compare Embeds by dictionary rather than object

## Summary
Currently, comparing two Embeds with the same attributes will not work because the objects are different.
This change fixes that odd behavior by comparing the dict representations of the two objects in the `__eq__` method.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -123,6 +123,9 @@ class Embed:
         else:
             self.timestamp = timestamp
 
+    def __eq__(self, other):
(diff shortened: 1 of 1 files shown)


Rapptz:
This implementation seems expensive.

NotTani:
If you think so! Is there a way you'd prefer it to be implemented?

Rapptz:
I don't really think deep equality is desirable.

PR #6190, opened 2020-12-23
[commands] Correct concurrency never releasing during prepare call

Phxntxm:
## Summary

This corrects an issue where concurrency is never released when the `prepare` method is called. This test case shows the issue:

```
bot = commands.Bot(command_prefix=">")


@bot.command()
@commands.max_concurrency(1, per=commands.BucketType.user)
async def test(ctx, argument):
    await ctx.send("test")
```

During this if you use `>test` then MissingRequiredArgument is raised. Since this happens during `prepare` (before this PR) MaxConcurrency.release is never called (since that happens during [the hooked_wrapped_callback](https://github.com/Rapptz/discord.py/blob/eb11079569570ce991d9468ca40eb8fcf129e0be/discord/ext/commands/core.py#L95-L97) call which happens [later in the invoke method](https://github.com/Rapptz/discord.py/blob/eb11079569570ce991d9468ca40eb8fcf129e0be/discord/ext/commands/core.py#L858)), leading to any subsequent calls to this raising MaxConcurrencyReached.

Also, I do realize that if a CheckFailure happens... this will end up releasing before ever acquiring a semaphore. However, since release [catches KeyError already](https://github.com/Rapptz/discord.py/blob/master/discord/ext/commands/cooldowns.py#L285-L287) this is fine. 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -848,7 +848,12 @@ class Command(_BaseCommand):
         return 0.0
 
     async def invoke(self, ctx):
-        await self.prepare(ctx)
+        try:
+            await self.prepare(ctx)
+        except:
+            if self._max_concurrency is not None:
+                await self._max_concurrency.release(ctx)
+            raise
 
         # terminate the invoked_subcommand chain.
         # since we're in a regular command (and not a group) then
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -781,14 +781,19 @@ class Command(_BaseCommand):
         if self._max_concurrency is not None:
             await self._max_concurrency.acquire(ctx)
 
-        if self.cooldown_after_parsing:
-            await self._parse_arguments(ctx)
-            self._prepare_cooldowns(ctx)
-        else:
-            self._prepare_cooldowns(ctx)
-            await self._parse_arguments(ctx)
+        try:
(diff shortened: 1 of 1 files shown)


Rapptz (changes requested):
The idea behind this changeset is good.

However, the current implementation uses `invoke` which is an injection point for groups as well. Doing this changeset in both places is not desirable to me and it should instead be entirely self-contained somewhere else.

Phxntxm:
(Oops I completely forgot about groups)

For the sake of sense it feels like acquiring should be the very first thing that happens and releasing the very last thing that happens. Which then feels to me like it should properly be here: <https://github.com/Rapptz/discord.py/blob/master/discord/ext/commands/bot.py#L900-L908> (the release in an added finally) but current implementation has `_max_concurrency` as a private attribute. 

Would changing that to a public attribute and moving it here be an acceptable fix? Or if not, including methods in the Command class for acquiring and releasing be instead? If not either of those, I guess it can be thrown in `prepare` since Group does not override that method

Rapptz:
Changing it in `prepare` is probably for the best. Another entry point is probably `Command.dispatch_error` but it feels out of place and can potentially lead to multiple release calls.

Rapptz (approved):

PR #6217, opened 2020-12-28
[commands] fully remove command when CommandRegistrationError is raised for alias

SebbyLaw:
## Summary

Currently, `GroupMixin.add_command` will keep the command registered under the original name and any previously registered aliases if `CommandRegistrationError` is raised for an alias. This may sometimes leave the command in a "half-registered" state where only some of the aliases are registered to the Group. This change makes `add_command` atomic by fully removing the command if an alias fails to register.

Example:
```py

bot = commands.Bot('$')

@bot.command()
async def baz(ctx):
    pass

@commands.command(aliases=['foo', 'bar', 'baz', 'this_will_not_be_registered'])
async def example(ctx):
    pass

try:
    bot.add_command(example)
except:
    pass

print([c for c in bot.all_commands])  # ['help', 'baz', 'example', 'foo', 'bar']
# "example" is registered under 'example', 'foo', and 'bar' still, but not any other aliases.
# This is unexpected behavior.
```
## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1143,13 +1143,10 @@ class GroupMixin:
             raise CommandRegistrationError(command.name)
 
         self.all_commands[command.name] = command
-        to_undo = [command.name]
         for alias in command.aliases:
             if alias in self.all_commands:
-                for name in to_undo:
-                    self.all_commands.pop(name)
+                self.remove_command(command.name)
                 raise CommandRegistrationError(alias, alias_conflict=True)
-            to_undo.append(alias)
             self.all_commands[alias] = command
 
     def remove_command(self, name):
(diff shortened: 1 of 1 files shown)


Rapptz (changes requested):
`remove_command` already does this logic. It's better to reuse it instead of doing this.

SebbyLaw:
`remove_command` will remove the original `baz` command in the example as well. I have tested this.

Rapptz:
While that's an interesting edge case, it should probably be fixed in the code that removes the command rather than the one that adds the command.

SebbyLaw:
> While that's an interesting edge case, it should probably be fixed in the code that removes the command rather than the one that adds the command.

I've modified the logic in `remove_command` to support doing this.

Rapptz (approved):

PR #6218, opened 2020-12-28
Replace "page" with "section" in documentation

Steffo99:
## Summary

This pull request improves (I think it does, at least) a bit the wording on the Event Reference section of the documentation.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -155,7 +155,7 @@ Opus Library
 Event Reference
 ---------------
 
-This page outlines the different types of events listened by :class:`Client`.
+This section outlines the different types of events listened by :class:`Client`.
 
 There are two ways to register an event, the first way is through the use of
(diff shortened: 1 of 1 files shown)


Rapptz:
Of all these changes the only one I agree with is the first one, which is basically turning "page" to "section". The rest don't seem to benefit much to me.

Steffo99:
Ok, I'll update it.

Is there a way to make the `|coroutine_link|_` substitution into a plural? 

"All the events must be a coroutine" sounds a bit weird to me...

kaylynn234:
That part of the wording seems fine to me - I think changing it in this regard would only serve to make it more difficult to read, or less correct in terms of grammar.

Steffo99:
Ok, then. I'll update the pull request.

Steffo99:
Done.

Rapptz (approved):
LGTM

PR #6272, opened 2021-01-07
Update state.py

aschenkuttel:
## Summary

if member intents are off on_reaction_add gets still dispatched because discord is sending the members data.
on_reaction_remove doesnt receive that data / doesnt get dispatched because of that.
either dispatch it with an user object or update documentation that on_reaction_remove needs the members intent enabled

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/state.py
+++ b/discord/state.py
@@ -1074,7 +1074,7 @@ class ConnectionState:
             self.dispatch('relationship_remove', old)
 
     def _get_reaction_user(self, channel, user_id):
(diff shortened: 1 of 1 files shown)


Rapptz:
I don't agree with this change.

I'm pretty sure this is already mentioned in the documentation in the intents page where it's relevant though it wouldn't hurt to document it in the event itself if it's not already.

PR #6275, opened 2021-01-08
Updated permissions class methods

TheOneMusic:
## Summary

The Discord UI has recently changed its permission categories, which are as follows:
- **General Permissions** no longer have the `administrator`, `create_instant_invite`, `change_nickname`, `manage_nicknames`, `kick_members` and `ban_members` permissions. It now has the `view_channel`/`read_messages` permission.
- New **Membership Permissions** which contain `create_instant_invite`, `change_nickname`, `manage_nicknames`, `kick_members` and `ban_members` permissions
- **Text Permissions** no longer have the `view_channel`/`read_messages` permission (this has been moved to the **General Permissions**).
- New **Advanced Permissions** which only contain the `administrator` permission.

The commit is therefore reflecting these changed. More specifically:
- [`Permissions.general()`](https://github.com/Rapptz/discord.py/blob/master/discord/permissions.py#L162) has been updated according to the Discord UI general permissions.
- Added `Permissions.membership()` which contains the `create_instant_invite`, `change_nickname`, `manage_nicknames`, `kick_members` and `ban_members` permissions.
- [`Permissions.text()`](https://github.com/Rapptz/discord.py/blob/master/discord/permissions.py#L168) no longer contain the `read_messages`/`view_channel` permission.
- Added `Permissions.advanced()` which is an alias for `Permission.administrator`.

This commit also modifies the [`Permissions.all_channel()`](https://github.com/Rapptz/discord.py/blob/master/discord/permissions.py#L147) class method, by adding the `stream` and `priority_speaker` permissions which were missing.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/permissions.py
+++ b/discord/permissions.py
@@ -149,32 +149,49 @@ class Permissions(BaseFlags):
         ``True`` and the guild-specific ones set to ``False``. The guild-specific
         permissions are currently:
 
+        - manage_emojis
+        - view_audit_log
+        - view_guild_insights
         - manage_guild
+        - change_nickname
+        - manage_nicknames
         - kick_members
         - ban_members
         - administrator
-        - change_nickname
-        - manage_nicknames
         """
-        return cls(0b00110011111101111111110001010001)
+        return cls(0b00110011111101111111111101010001)
 
     @classmethod
     def general(cls):
         """A factory method that creates a :class:`Permissions` with all
         "General" permissions from the official Discord UI set to ``True``."""
-        return cls(0b01111100000010000000000010111111)
+        return cls(0b01110000000010000000010010110000)
+
+    @classmethod
+    def membership(cls):
+        """A factory method that creates a :class:`Permissions` with all
+        "Membership" permissions from the official Discord UI set to ``True``."""
+        return cls(0b00001100000000000000000000000111)
 
     @classmethod
     def text(cls):
         """A factory method that creates a :class:`Permissions` with all
         "Text" permissions from the official Discord UI set to ``True``."""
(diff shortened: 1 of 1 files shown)


TheOneMusic:
Also, I was wondering why there is no `Permissions.all_guild()` factory method, it could be useful.

Zomatree:
no versionadded, plus this is a breaking change so should be set for v2.0.

TheOneMusic:
Thank you. Just wanted to know, why is this a breaking change? It doesn't remove any methods or parameters, nor does it change method/parameters names (it simply modifies the return value).

bijij:
It's been debated as to whether it is or not.
On one hand, it's changing a value which was pretty much treated as a constant.
On the other hand the value is tied to the discord UI and as such would make sense to match what the UI displays.

Rapptz:
I don't have a strong lean towards this being a breaking change or not.

As was mentioned above, the documentation does not explicitly mention what the constant value is and ties it to the UI. It makes sense to change it if the UI changes however some people might be relying on the old value. It's a tricky subject.

TheOneMusic:
I guess it makes sense to update. I mean discord is updating and we can't stay with the old UI design forever. The advantage of this one is also to split up general, membership and advanced permissions into separate categories for more control, and also update `Permissions.all_channel()` which were missing recent permissions.

TheOneMusic:
So, will this be part of update 1.6 or 2.0? What do you think is best?

bijij:
Personally I'd say it's non breaking and should be in 1.7

But others probably disagree.

PR #6279, opened 2021-01-08
Added GuildChannel.add_permissions method

TheOneMusic:
## Summary

Added `GuildChannel.add_permissions` method. This is the same as [`GuildChannel.set_permissions`](https://github.com/Rapptz/discord.py/blob/master/discord/abc.py#L567), except that it "keeps" the permissions that have not been overwritten.

For example, if a [`text_channel`](https://github.com/Rapptz/discord.py/blob/master/discord/channel.py#L52) has the `send_messages` permission set to `False` for **target**, calling `await text_channel.add_permissions(target, add_reactions=False)` will now result in this channel having the `send_messages` and `add_reactions`  permissions set to `False` for **target**, whereas `await text_channel.set_permissions(target, add_reactions=False)` will result only in the `add_reactions` permission set to `False` for **target**.

This can be very useful, for example, to lock a text channel by setting `send_messages` to `False` for `@everyone`, without changing the other channel permissions.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -659,6 +659,25 @@ class GuildChannel:
         else:
             raise InvalidArgument('Invalid overwrite type provided.')
 
+    async def add_permissions(self, target, *, overwrite=_undefined, reason=None, **permissions):
+        """|coro|
+        
+        Same as `Channel.set_permissions` except it adds the channel overwrites instead of overwriting them.
+        """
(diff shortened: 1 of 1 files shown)


Rapptz:
I don't think this helper method is worth adding considering the confusion it'll bring and the fact you can already do this simply by using the pre-existing tools.

Issue #1814, opened 2019-01-02
Discord Bot Rewrite - add_roles


florianojunior:
Hello,

I'm now starting to develop a Bot, using Discord.py rewrite.

When I try to use the command "await client.add_roles (member, role)" it is informed that there is no add_roles.

Error: discord.ext.commands.errors.CommandInvokeError: Command raised an exception: AttributeError: 'Bot' object has no attribute 'add_roles'

```
...
client = commands.Bot (command_prefix = "!")
...

```

```
@ client.event
async def on_member_join (member):
     ...
     role = discord.utils.get (member.guild.roles, name = "role_name")
     await client.add_roles (member, role)
     ...
```

If anyone can assist me in what I am doing wrong, thank you.

RainbowLegend:
On the rewrite branch, more methods have been moved to their corresponding class in order to shift to a more object oriented programming style. Instead of `Client.add_roles`, you’ll want to use `Member.add_roles`.

See the [migrating guide](https://discordpy.readthedocs.io/en/rewrite/migrating.html) for more details on the differences between the async branch and rewrite.

florianojunior:
Thanks so much for the help, I had not tried that. Now, I no longer have this compile error.

But now I'm getting the exception: discord.errors.Forbidden: FORBIDDEN (status code: 403): Missing Permissions

```
@client.event
async def on_member_join(member):
      ...
      role = discord.utils.get (member.guild.roles, name = "role_name")
      await member.add_roles (role, reason = 'member_join', atomic = True)
      ...
```

When I add the Bot to my server, I have used the URL below, which requires Administrator permission.

https://discordapp.com/oauth2/authorize?client_id=CLIENT_ID&scope=bot&permissions=8

Any idea what it could be?

Thank you

RainbowLegend:
Most likely the role you are trying to give is above the bot’s top role.

You should join the [official discord.py help server](https://discord.gg/r3sSKJJ), as GitHub issues aren’t the right place to ask for help.

PR #6284, opened 2021-01-10
Update add_roles docstring to mention role order.

perrycate:
## Summary

Adds a quick mention of the relevance of discord's role ordering. When I tried to use this function, I had no idea that that mattered until I eventually found [this comment](https://github.com/Rapptz/discord.py/issues/1814#issuecomment-451047619) in issue #1814.

## Checklist

- [ ] This PR fixes an issue.
- [x ] This PR is **not** a code change (e.g. documentation, README, ...)

### DIFF
--- a/discord/member.py
+++ b/discord/member.py
@@ -644,7 +644,7 @@ class Member(discord.abc.Messageable, _BaseUser):
 
         You must have the :attr:`~Permissions.manage_roles` permission to
         use this, and the added :class:`Role`\s must appear lower in the list
-        of roles than the highest role of the caller.
+        of roles than the highest role of the member.
 
         Parameters
         -----------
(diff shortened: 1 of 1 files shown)


discord/member.py
```diff
@@ -643,7 +643,8 @@ async def add_roles(self, *roles, reason=None, atomic=True):
         Gives the member a number of :class:`Role`\s.
 
         You must have the :attr:`~Permissions.manage_roles` permission to
-        use this.
+        use this, and the added :class:`Role`\s must appear lower in the list
+        of roles than the highest role of the caller.
```
Rapptz, line 647:
```suggestion
        of roles than the highest role of the member.
```

Rapptz (changes requested):
I think in the future this might be better suited with a page describing the intricacies of role hierarchy.

For now though, could this also be mentioned in `remove_roles`? It's basically the same concept -- you can't remove roles higher than the member.

discord/member.py
perrycate: Done.

Rapptz (approved):

PR #6295, opened 2021-01-12
Update cog.walk_commands when removing a command from a Cog.

InterStella0:
## Summary

When `bot.remove_command` removes a command from a Cog, `cog.__cog_commands__` does not get updated. This lead to `Cog.walk_commands` not updated.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1176,6 +1176,11 @@ class GroupMixin:
             # we're removing an alias so we don't want to remove the rest
             return command
 
+        if command.cog is not None:
+            # removes the command instance from the cog.
+            cog = command.cog
+            cog.__cog_commands__ = tuple(c for c in cog.__cog_commands__ if c is not command)
+
         # we're not removing the alias so let's delete the rest of them.
         for alias in command.aliases:
             cmd = self.all_commands.pop(alias, None)


discord/ext/commands/core.py
```diff
@@ -1176,6 +1176,11 @@ def remove_command(self, name):
             # we're removing an alias so we don't want to remove the rest
             return command
 
+        if command.cog is not None:
+            # removes the command instance from the cog.
+            cog = command.cog
```
SebbyLaw, line 1181:
```suggestion
        cog = command.cog
        if cog is not None:
            # removes the command instance from the cog.
            
```

SebbyLaw (commented):
What happens when a removed command is added back? The tuple isn't updated.

Rapptz:
Cog introspection is different from injected introspection. Ergo, this isn't a bug. That command still exists within the cog it's just not "hooked" into an object. This is the same reason why you can do `Cog.walk_commands()` even before you add the cog to the bot.

InterStella0:
> What happens when a removed command is added back? The tuple isn't updated.

I was thinking of doing that too where add_command supports adding into a cog.
```python
def add_command(self, command):
   ...
   cog = commands.cog
   if cog is not None:
       if not isinstance(cog, Cog):
           raise TypeError('The cog passed into command.cog attribute must be a subclass of Cog')
       cog.__cog_commands__ += (command,)
```
 But I guess after reading what Danny said, I guess I will scrap that.

PR #6329, opened 2021-01-18
[commands] make built-in converters use classmethods

SebbyLaw:
## Summary

None of the built-in object converters (aside from `clean_content`) currently utilise state. As such there is no reason for any of the built-in converters to use an instance method when a classmethod does the same.

This *can* be a breaking change in the case of converters being used as such:
```py
class CustomConverter(Converter):
        try:
            return MemberConverter.convert(self, ctx, argument)
        except BadArgument:
            ...
```

This is a draft waiting on the status of #6328.

This PR also removed the members of the autoclass, since the docstring is simply repeated from `commands.Converter`

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -100,12 +100,11 @@ class Converter:
         raise NotImplementedError('Derived classes need to implement this.')
 
 class IDConverter(Converter):
-    def __init__(self):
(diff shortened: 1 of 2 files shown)


Rapptz:
There is no benefit to doing this.

PR #6332, opened 2021-01-20
Added empty string attribute

TheOneMusic:
## Summary

A very straightforward pull request which is just meant to add an `EmptyString` attribute in embeds. This is equal to `"\u200b"` and I think it is useful to define it here, considering that this technique is pretty commonly used to make empty field names/values. Example:

```python3
import discord
embed = discord.Embed()
embed.add_field(name="->", value=discord.Embed.EmptyString)
```

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -102,6 +102,7 @@ class Embed:
                  '_fields', 'description')
 
     Empty = EmptyEmbed
+    EmptyString = "\u200b"
 
     def __init__(self, **kwargs):
         # swap the colour/color aliases


xFGhoul:
Looks good to me, tho typing `zero width space` or just `\u200b` may be easier to memorize of just keep in mind this is pretty good for convenience, just my thoughts 👍

oliver-ni:
I don't think this is easier nor better than just writing `"\u200b"`

TheOneMusic:
Well, it avoids searching for it every time. No big difference but I think it's nice to have it.

oliver-ni:
Something like this should not be a part of the library. It's neither shorter nor easier to use, and you would still have to search for or remember something—namely, the name of the attribute—every time you use it. It'd also be easily confused with other things like `discord.Embed.Empty`.

Most of all, it's not a good idea to hide something like this in an attribute. It covers what's actually happening for no benefit at all. When reading code, you should be able to easily tell what a piece of code is doing—not this. With this change, if you've never heard of what `EmptyString` is, you would easily assume that Discord provides some way to make embed fields empty, or something like that.

TheOneMusic:
I see, well `discord.EmptyString` could do the work but ok

Phxntxm:
This to me feels like something that should realistically defined in your *own* scripts IE ```
zws = "\u200b"
```

I don't see a benefit of having it in the library.

Rapptz:
Hey,

Thanks for the PR. Unfortunately the reason this hasn't been added historically isn't because of its simplicity to define yourself but rather because fundamentally this works as an undocumented hack that while probably won't change in the foreseeable future, still has the capacity of messing up eventually. There's no guarantee that the library can make it this is added and documented because ultimately the format of embeds is up to the discord client. If embeds end up changing their validation to strip then this becomes baggage to the library (though this seems unlikely since it'd break pre-existing bots, this hasn't really stopped discord before). The design of embeds has changed before many times over as well and the behavior of this in the mobile clients has changed as well. 

Some libraries define an abstraction for an "empty field" but I don't for the same reasons above.

PR #6353, opened 2021-01-26
Make the style of external and internal cross-references consistent

jack1142:
## Summary

The cross-references have inconsistent formatting:
image

## Checklist

- [x] This PR fixes an issue.


Rapptz:
I think this is better off written differently. The `reference` is just a regular `a` tag, the `internal` and the `external` classes are what separate their location. We already have a few external resources outside of PEP and RFCs, such as github issues or things exported through the new `resourcelinks` extension.

I think it might be worth exploring where bold is expected and whether bold should even continue to exist.

PR #6365, opened 2021-01-30
Add discord.Guild converter and GuildNotFound error

AbstractUmbra:
## Summary

This adds a `discord.Guild` converter as well as the resulting error if a match is not found.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -531,7 +531,7 @@ class GuildConverter(IDConverter):
     The lookup strategy is as follows (in order):
 
     1. Lookup by ID.
-    2. Lookup by name.
+    2. Lookup by name. (There is no disambiguation for Guilds with multiple matching names).
 
     .. versionadded:: 1.7
     """
@@ -543,14 +543,14 @@ class GuildConverter(IDConverter):
         if match is not None:
             guild_id = int(match.group(1))
             result = ctx.bot.get_guild(guild_id)
-            return result
 
-        predicate = lambda g: g.name == argument
-        result = discord.utils.find(predicate, ctx.bot.guilds)
-        if result is not None:
-            return result
+        if result is None:
+            predicate = lambda g: g.name == argument
+            result = discord.utils.find(predicate, ctx.bot.guilds)
 
-        raise GuildNotFound(argument)
+            if result is None:
+                raise GuildNotFound(argument)
+        return result
 
 class EmojiConverter(IDConverter):
     """Converts to a :class:`~discord.Emoji`.
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -545,8 +545,7 @@ class GuildConverter(IDConverter):
             result = ctx.bot.get_guild(guild_id)
 
         if result is None:
-            predicate = lambda g: g.name == argument
-            result = discord.utils.find(predicate, ctx.bot.guilds)
+            result = discord.utils.get(ctx.bot.guilds, name=argument)
 
             if result is None:
                 raise GuildNotFound(argument)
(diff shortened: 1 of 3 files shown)


discord/ext/commands/converter.py
```diff
@@ -524,6 +525,33 @@ async def convert(self, ctx, argument):
...
+    """
+
+    async def convert(self, ctx, argument):
+        match = self._get_id_match(argument)
+        result = None
+
+        if match is not None:
+            guild_id = int(match.group(1))
+            result = ctx.bot.get_guild(guild_id)
+            return result
```
z03h, line 546: This could return None instead of raising GuildNotFound

Gobot1234: That wouldn't match the behaviour of any of the current converters

z03h: That's what I mean? I didn't change anything and it currently can return None for an wrong id

discord/ext/commands/converter.py
```diff
@@ -524,6 +525,33 @@ async def convert(self, ctx, argument):
         except Exception as exc:
             raise BadInviteArgument() from exc
 
+class GuildConverter(IDConverter):
+    """Converts to a :class:`~discord.Guild`.
+
+    The lookup strategy is as follows (in order):
```
Rapptz, line 531: This should probably mention there's no disambiguation if guilds have multiple matching names.

discord/ext/commands/converter.py
Rapptz, line 546:
```suggestion
```

Rapptz (changes requested):
Historically I'm against such a converter but I guess if the shortcomings are documented it's okay, even though I think the use cases for it aren't that big.

Zomatree (changes requested):

discord/ext/commands/converter.py
```diff
@@ -524,6 +525,33 @@ async def convert(self, ctx, argument):
...
+        match = self._get_id_match(argument)
+        result = None
+
+        if match is not None:
+            guild_id = int(match.group(1))
+            result = ctx.bot.get_guild(guild_id)
+            return result
+
+        predicate = lambda g: g.name == argument
+        result = discord.utils.find(predicate, ctx.bot.guilds)
```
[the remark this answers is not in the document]
bijij: Afaik `uitlls.find` was used to be concistent with other convereters.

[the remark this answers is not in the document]
Zomatree: I can only find two other use of `utils.find` where `utils.get` can be used instead (both in `UserConverter`), most other uses of it have more of a "custom" check that cant be done with `utils.get`, having consistency here doesn't really matter anyway.

discord/ext/commands/converter.py
AbstractUmbra: Fixed in a manner similar to the other converters.

discord/ext/commands/converter.py
```diff
@@ -524,6 +525,33 @@ async def convert(self, ctx, argument):
...
+        match = self._get_id_match(argument)
+        result = None
+
+        if match is not None:
+            guild_id = int(match.group(1))
+            result = ctx.bot.get_guild(guild_id)
+
+        if result is None:
+            predicate = lambda g: g.name == argument
+            result = discord.utils.find(predicate, ctx.bot.guilds)
```
Rapptz, line 549: `get` is probably better (it is faster than `find` for the single attribute case).

Rapptz (changes requested):
This is missing the documentation changes required in api.rst

Mainly, adding the class to the list of converters and adding the exception to the exception hierarchy and the list of exceptions.

Rapptz (approved):

PR #6369, opened 2021-02-01
ensure encoder is set in send_audio_packet

bradmorgz:
## Summary

encoder is not always set in send_audio_packet. this makes sure it gets set if it's not already

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/voice_client.py
+++ b/discord/voice_client.py
@@ -630,6 +630,8 @@ class VoiceClient(VoiceProtocol):
 
         self.checked_add('sequence', 1, 65535)
         if encode:
+            if not self.encoder:
+                self.encoder = opus.Encoder()
             encoded_data = self.encoder.encode(data, self.encoder.SAMPLES_PER_FRAME)
         else:
             encoded_data = data


Rapptz:
This code path is deliberately thin. What case are you hitting where the encoder isn't set?

bradmorgz:
> This code path is deliberately thin. What case are you hitting where the encoder isn't set?

connecting to a channel and then calling send_audio_packet i find that the encoder is 'None'. encoder seems to only be set if there's an audio source in the current code

Rapptz:
If you want to send audio then you should use the AudioSource interface not the lower level function. There's already a raw PCM AudioSource so I'm unsure why there's a need to use this function.

bradmorgz:
I was under the impression that audio sources could only be file-like but I'm probably wrong. I'm trying to play a stream from an audio input device

Rapptz:
The audio source requirements are pretty thin. You just have to poll 20ms of PCM audio. There's nothing inherently file like about it, though even a stream tends to have "file-like" properties such as reading.

bradmorgz:
I'll have a go at that. thanks :)

PR #6390, opened 2021-02-06
Ensure Message.call is None by default

peterdelevoryas:
`Message` has an attribute `call` which is claimed to have type
`Optional[CallMessage]`. From
https://docs.python.org/3/library/typing.html#typing.Optional:

    Optional[X] is equivalent to Union[X, None].

But `Message` doesn't actually ensure that `call` is initialized to a value in
`__init__`. This commit fixes that inconsistency.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -504,6 +504,7 @@ class Message(Hashable):
         self.application = data.get('application')
         self.activity = data.get('activity')
         self.channel = channel
+        self.call = None
         self._edited_timestamp = utils.parse_time(data['edited_timestamp'])
         self.type = try_enum(MessageType, data['type'])
         self.pinned = data['pinned']


Rapptz:
In what cases did you find this was not true? The attribute is set further down when the individual handlers are called.

peterdelevoryas:
If the message is not related to a call, then `_handle_call` is never executed, right? My point is that the class documentation specifies the type for all the attributes in the class, but some of these attributes may not even exist, depending on the kind of `Message` it is. This is allowed in Python (and people end up doing this), but it's confusing if you're expecting something similar to static typing. In a statically typed language, I would expect to be able to access `message.call` no matter what kind of message it is, and get a value. That's why I'm referencing the documentation for Python 3 type annotations: `Optional[T]` is  not saying that the attribute may-or-may-not exist, it's saying that the value of the attribute may be `None`.

I ran into this while debug printing `Message`'s, i.e. `print(message.call)`.

Rapptz:
> If the message is not related to a call, then _handle_call is never executed, right?

Right, I see what you mean now. I missed this on the first read.

You don't have to educate me on type systems though.

Thanks for the PR.

Issue #6427, opened 2021-02-16
add_reaction event not working properly on DMChannels


ScriptHosT12:
**Summary**

add_reaction event is not dispatched on direct messages

**Reproduction Steps**

run this code with the the necessary changes
```py
import asyncio

import discord

client = discord.Client()

@client.event
async def on_message(message):
    if isinstance(message.channel, discord.DMChannel) and message.content == "hey":
        target_message = await message.channel.send("Hey there buddy react this message please")
        await target_message.add_reaction("✅")

        def check(r, u):
            return u.id == message.author.id and target_message.id == r.message.id

        try:
            await client.wait_for("reaction_add", timeout=30, check=check)
        except asyncio.exceptions.TimeoutError:
            await message.channel.send("You didn't react...")
        else:
            await message.channel.send("Thanks!")

client.run("<token>")
```

**Expected Results**

The bot should respond after the reaction
(this behaviour happens on discord.py 1.5)

**Actual Results**

The bot does not respond to the reaction, and will blame it on you for not reacting after the timeout
(this behaviour happens on latest)

**Intents**

default intents

**System Information**

- Python v3.8.3-final
- discord.py v1.6.0-final
- aiohttp v3.6.3
- system info: Windows 10 10.0.18362

**Checklist**

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.

Zomatree:
You require the members intent to be able to use the `on_reaction_add` inside DMs, either [enable the members intent](https://discordpy.readthedocs.io/en/latest/intents.html) or use the `on_raw_reaction_add` event.

Issue #6280, opened 2021-01-08
Regression or other impact to wait_for reaction in 1.6.0


DefiDebauchery:
## Summary

In version 1.6.0, the `wait_for` routine does not appear to return in the same manner as 1.5.1. 
While the check may or may not fire with the bot's `add_reaction` (just prior to the check), it never seems to handle subsequent user interaction.

## Reproduction Steps

```py
import asyncio
import discord
from discord.ext import commands

class TestCog(commands.Cog):
    def __init__(self, bot):
        self.bot = bot
        self.prompt = ['✅', '❌']

    @commands.command()
    @commands.dm_only()
    async def test(self, ctx, *, test_input):
        embed = discord.Embed(title='Your input', description='Please confirm the value below.')
        embed.add_field(name='Input', value=test_input)

        prompt = await ctx.channel.send(embed=embed)
        for emoji in self.prompt:
            await prompt.add_reaction(emoji=emoji)

        def check(reaction, user):
            print('Checking user interaction for', user, emoji)
            return user == ctx.author and str(reaction.emoji) in self.prompt

        try:
            print('pre-reaction')
            reaction, user = await self.bot.wait_for('reaction_add', timeout=5.0, check=check)
            print('post-reaction')

            await ctx.channel.send('Post-reaction!')
        except asyncio.TimeoutError:
            await ctx.channel.send("We did not receive a response")

def setup(bot):
    bot.add_cog(TestCog(bot))
```

## Expected Results

As with 1.5.1, we would expect the following output:

```
❯ python3 main.py
pre-reaction
Checking user interaction for theBot#9876 ❌
Checking user interaction for MyUser#1234 ✅
post-reaction
```

<img width="273" alt="Screen Shot 2021-01-08 at 4 37 59 PM" src="https://user-images.githubusercontent.com/75273961/104066748-ec54e980-51cf-11eb-96cc-e309c2b1544b.png">

## Actual Results

When changed to 1.6.0, this is no longer handled in the same way

```
❯ pip3 install discord.py==1.6.0
❯ python3 main.py
pre-reaction
Checking user interaction for theBot#9876 ❌
```

And after 5 seconds, the bot states that it did not receive any input

<img width="257" alt="Screen Shot 2021-01-08 at 4 40 44 PM" src="https://user-images.githubusercontent.com/75273961/104066983-51104400-51d0-11eb-8b4c-1d2839daa672.png">

## Intents

No intents specifically defined, but do not believe changes would affect the outcome

[...]

PR #6429, opened 2021-02-16
Update docs to be clearer about discord limitations.

mikeshardmind:
## Summary

This clarifies that reaction add events may not function outside of guilds without the member cache, while correctly pointing to this being a [limitation made by discord](https://discord.com/developers/docs/topics/gateway#message-reaction-add-message-reaction-add-event-fields)

This has been a point of confusion for a few users (see #6280 , #6427).

The related [discussion about a stale cache issue](https://github.com/Rapptz/discord.py/discussions/6034) (and [commit fixing this](https://github.com/Rapptz/discord.py/commit/7a3a571e0a641aaeb061e86d8a6e851c5cd4c381)) for why this doesn't use the user cache indicate that this just needs to be made clearer for users about what they need and what they should use and that the library should not be changed to work around this limitation.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -478,6 +478,14 @@ to handle it, which defaults to print a traceback and ignoring the exception.
 
     This requires :attr:`Intents.reactions` to be enabled.
 
+    .. note::
+
+        This doesn't require :attr:`Intents.members` within a guild context,
+        but due to Discord not providing updated user information in a direct message
+        it's required for direct messages to receive this event.
+        Consider using :func:`on_raw_reaction_add` if you need this and do not otherwise want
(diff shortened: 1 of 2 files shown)


Rapptz:
Maybe there should also be a warning in the corresponding event reference page?

mikeshardmind:
I added it as a note to the corresponding event reference. I'm not thrilled by the specific wording, but I'm not sure of a better way to be technically accurate about the situation without explaining more than is needed in this note either.

mikeshardmind:
updated per discussion in discord, rebased to a single commit.

PR #6431, opened 2021-02-16
Handle positional-only parameters in bot commands

bryanforbes:
## Summary

Currently, the library is unable to parse command functions that use [positional-only parameters](https://www.python.org/dev/peps/pep-0570/):

```python
@bot.command()
async def add(ctx, left: int, right: int, /):
    """Adds two numbers together."""
    await ctx.send(left + right)
```

This PR updates the parameter parsing in the commands extension to handle positional-only parameters the same as the standard parameter type (`POSITIONAL_OR_KEYWORD`).

## Checklist

- [x] This PR fixes an issue.


Rapptz:
Oh I almost missed this one. Thanks.

I was curious if this attribute existed in Python 3.5 and it does, so this seems good.

PR #6441, opened 2021-02-20
added User.mutual_guilds() as per 2539

### LINKED ISSUE #2539: Add mutual guild method to user class
### There is no simple way to quickly get the servers you share with a Member

Bot which helps users identify people who they share lots in common with on a server.

### The Ideal Solution

Adding a method to the Member class:
https://discordpy.readthedocs.io/en/latest/api.html#discord.Member
This would be the best and most obvious place to put it.

### Summary

Add a method to a User object to return a list of guilds you share in common

sudosnok:
## Summary
Closes #2539 
Adds a mutual_guilds method to the User class

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -2,8 +2,8 @@ name: Bug Report
 about: Report broken or incorrect behaviour
 labels: bug
 issue_body: true
-inputs:
-  - type: description
+body:
+  - type: markdown
     attributes:
       value: >
         Thanks for taking the time to fill out a bug.
@@ -14,48 +14,54 @@ inputs:
     attributes:
       label: Summary
       description: A simple summary of your bug report
+    validations:
       required: true
   - type: textarea
     attributes:
       label: Reproduction Steps
-      required: true
       description: >
          What you did to make it happen.
          Ideally there should be a short code snippet in this section to help reproduce the bug.
+    validations:
+      required: true
   - type: textarea
     attributes:
       label: Expected Results
-      required: true
       description: >
         What did you expect to happen?
+    validations:
+      required: true
   - type: textarea
     attributes:
       label: Actual Results
-      required: true
       description: >
         What actually happened?
+    validations:
+      required: true
   - type: input
     attributes:
       label: Intents
-      required: true
       description: >
         What intents are you using for your bot?
         This is the `discord.Intents` class you pass to the client.
+    validations:
+      required: true
   - type: textarea
     attributes:
       label: System Information
-      required: true
       description: >
         Run `python -m discord -v` and paste this information below.
 
         This command required v1.1.0 or higher of the library. If this errors out then show some basic
         information involving your system such as operating system and Python version.
+    validations:
+      required: true
   - type: checkboxes
     attributes:
       label: Checklist
       description: >
         Let's make sure you've properly done due dilligence when reporting this issue!
-      choices:
+      options:
         - label: I have searched the open issues for duplicates.
           required: true
         - label: I have shown the entire traceback, if possible.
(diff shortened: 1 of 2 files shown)


discord/user.py
```diff
@@ -698,6 +698,16 @@ async def _get_channel(self):
         ch = await self.create_dm()
         return ch
 
+    def mutual_guilds(self):
```
Rapptz, line 701: This should be a property, not a method.

discord/user.py
```diff
@@ -698,6 +698,16 @@ async def _get_channel(self):
...
+        """List[:class:`~discord.Guild`]: Return a list of :class:`~discord.Guild` that the :class:`~User` shares with the bot.
+
+        .. note::
+
+            This can only return guilds shared on an individual shard, if applicable.
```
Rapptz, line 706:
```suggestion
            This can only return mutual guilds within the client's internal cache.
```

discord/user.py
```diff
@@ -698,6 +698,16 @@ async def _get_channel(self):
...
+        """
+        _guilds = list(self._state._guilds.values())
```
Rapptz, line 708: No need for the copy here.

discord/user.py
```diff
@@ -698,6 +698,16 @@ async def _get_channel(self):
         ch = await self.create_dm()
         return ch
 
+    def mutual_guilds(self):
+        """List[:class:`~discord.Guild`]: Return a list of :class:`~discord.Guild` that the :class:`~User` shares with the bot.
+
+        .. note::
+
+            This can only return guilds shared on an individual shard, if applicable.
+        """
```
Rapptz, line 707: This is missing `.. versionadded:: 1.7` somewhere.

discord/user.py
```diff
@@ -698,6 +698,16 @@ async def _get_channel(self):
...
+        _guilds = list(self._state._guilds.values())
+        return [guild for guild in _guilds if guild.get_member(self.id)]
```
Rapptz, line 709:
```suggestion
        return [guild for guild in self._state._guilds.values() if guild.get_member(self.id)]
```

Rapptz (changes requested):
Good starting point, just needs a few improvements.

discord/user.py
```diff
@@ -698,6 +698,16 @@ async def _get_channel(self):
...
 
+    @property
+    def mutual_guilds(self):
+        """List[:class:`~discord.Guild`]: Return a list of :class:`~discord.Guild` that the :class:`~User` shares with the bot.
+        .. versionadded:: 1.7
+        .. note::
+
+            This can only return mutual guilds within the client's internal cache.
+        """
+        return [guild for guild in self._state._guilds if guild.get_member(self.id)]
```
Gobot1234, line 709: This is going to AttributeError you need to .values() this

sudosnok: Thanks for pointing that out, didnt noticed when i removed the copy

discord/user.py
```diff
@@ -698,6 +698,16 @@ async def _get_channel(self):
         ch = await self.create_dm()
         return ch
 
+    @property
+    def mutual_guilds(self):
+        """List[:class:`~discord.Guild`]: Return a list of :class:`~discord.Guild` that the :class:`~User` shares with the bot.
```
NCPlayz, line 703:
```suggestion
        """List[:class:`Guild`]: The guilds that the user shares with the client.
```

discord/user.py
```diff
@@ -698,6 +698,16 @@ async def _get_channel(self):
...
+        .. versionadded:: 1.7
+        .. note::
+
+            This can only return mutual guilds within the client's internal cache.
```
NCPlayz, line 707:
```suggestion
        .. note::

            This will only return mutual guilds within the client's internal cache.

        .. versionadded:: 1.7
```

NCPlayz (changes requested):

PR #6442, opened 2021-02-20
New/examples converters

AbstractUmbra:
## Summary

This adds an example to the repository for library converter methods and their applicable uses.
This was wanted in #2525 and contributes to that.

## Checklist

- [ ] This PR fixes an issue.


examples/converters.py
```diff
@@ -0,0 +1,78 @@
...
+    return await ctx.send("No member or channel matching this ID was found.")
+
+# Builtin type converters
+@bot.command()
+async def trial_converter(ctx: commands.Context, number: int, maybe: bool):
+    # We want and int and a bool parameter here.
+    # Bool is a slightly special case, as shown here:
+    # https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html#bool
+
+    return await ctx.send("Number: {} -- Bool: {}.".format(number, maybe))
```
Zomatree, line 75:
```suggestion
    await ctx.send("Number: {} -- Bool: {}.".format(number, maybe))
```

examples/converters.py
```diff
@@ -0,0 +1,78 @@
...
+    # Do the same for TextChannel...
+    textchannel_converter = commands.TextChannelConverter()
+    try:
+        channel = await textchannel_converter.convert(ctx, argument)
+    except commands.ChannelNotFound:
+        pass
+    else:
+        return await ctx.send("Channel found: {}".format(channel))
+
+    return await ctx.send("No member or channel matching this ID was found.")
```
Zomatree, line 66:
```suggestion
    await ctx.send("No member or channel matching this ID was found.")
```

examples/converters.py
```diff
@@ -0,0 +1,78 @@
...
+    # parameter is typehinted to `discord.User`. This means that
+    # during command invocation we will attempt to convert
+    # the value passed as `user` to a `discord.User` instance.
+
+    # If the conversion is successful we will have a User instance
+    # and can do the following:
+    return await ctx.send(user.name)
+
+@userinfo.error
+async def userinfo_error(ctx: commands.Context, error: Exception):
```
Zomatree, line 23: should have an `else` clause to catch stray errors

examples/converters.py
```diff
@@ -0,0 +1,78 @@
...
+@bot.command()
+async def userinfo(ctx: commands.Context, user: discord.User):
+    # In the command signature about, you can see that the `user`
+    # parameter is typehinted to `discord.User`. This means that
+    # during command invocation we will attempt to convert
+    # the value passed as `user` to a `discord.User` instance.
+
+    # If the conversion is successful we will have a User instance
+    # and can do the following:
+    return await ctx.send(user.name)
```
Zomatree, line 20:
```suggestion
    await ctx.send(user.name)
```

examples/converters.py
```diff
@@ -0,0 +1,78 @@
...
+async def userinfo_error(ctx: commands.Context, error: Exception):
+    # if the conversion above fails for any reason, it will raise `commands.errors.BadArgument`
+    # so we handle this in this error handler
+    if isinstance(error, commands.BadArgument):
+        return await ctx.send("Failed to convert the argument for `user` to `discord.User`.")
+
+# Manual use of converters will follow
+
+@bot.command()
+async def channel_or_member(ctx: commands.Context, argument: int):
```
Zomatree, line 32: converters take strings not int so the typehint should not be `int`

Zomatree (changes requested):

examples/converters.py
AbstractUmbra: This is intending to use the `int` converter. This is explained in the comments just below.

Zomatree: this will typeerror from memberconverter.convert not accepting ints

examples/converters.py
```diff
@@ -0,0 +1,80 @@
...
+# We will need the members intents at the minimum to use the Member converter.
+intents = discord.Intents.all()
+
+# Construct the Bot instance.
+bot = commands.Bot("!", intents=intents)
+
+
+@bot.command()
+async def userinfo(ctx: commands.Context, user: discord.User):
+    # In the command signature about, you can see that the `user`
```
jonasbohmann, line 13:
```suggestion
    # In the command signature above, you can see that the `user`
```

examples/converters.py
```diff
@@ -0,0 +1,80 @@
+import discord
+from discord.ext import commands
+
+# We will need the members intents at the minimum to use the Member converter.
+intents = discord.Intents.all()
+
+# Construct the Bot instance.
+bot = commands.Bot("!", intents=intents)
```
NCPlayz, line 10:
```suggestion
# This example requires the 'members' privileged intent to use the Member converter.

import discord
from discord.ext import commands

intents = discord.Intents.default()
intents.members = True

bot = commands.Bot(command_prefix="!", intents=intents)
```

examples/converters.py
```diff
@@ -0,0 +1,80 @@
...
+@bot.command()
+async def trial_converter(ctx: commands.Context, number: int, maybe: bool):
+    # We want and int and a bool parameter here.
+    # Bool is a slightly special case, as shown here:
+    # https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html#bool
+
+    await ctx.send("Number: {} -- Bool: {}.".format(number, maybe))
+
+token = "your token here"
+bot.run(token)
```
NCPlayz, line 94:
```suggestion
bot.run('token')

```

NCPlayz (changes requested):

examples/converters.py
```diff
@@ -61,6 +63,24 @@ async def channel_or_member(ctx: commands.Context, argument: str):
...
+    # See: https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html#typing-union
+    # NOTE: If a Union typehint converter fails it will raise `commands.errors.BadUnionArgument`
+    # instead of `commands.errors.BadArgument`.
+
+    # This is a more user friendly alternative to the above command.
+
+    # Let's check the type we actually got...
+    if isinstance(target, discord.User):
+        return await ctx.send("User found: {}".format(target))
+    elif isinstance(target, discord.TextChannel): # this could be an `else` but for completeness' sake.
```
mesub7, line 81: Would it be better to mention the user/channel?

examples/converters.py
```diff
@@ -0,0 +1,94 @@
...
+        return await ctx.send("Channel found: {}".format(channel))
+
+    await ctx.send("No member or channel matching this ID was found.")
+
+@bot.command()
+async def alternative_channel_or_member(ctx: commands.Context, target: typing.Union[discord.Member, discord.TextChannel]):
+    # This command signature utilises the `typing.Union` typehint.
+    # In discord.py this attempts a conversion of each type in this Union *in order*.
+    # So we will attempt to convert whatever is passed to `target` to a `discord.User` and if that fails
+    # we attempt to convert it to a `discord.TextChannel`
```
NCPlayz, line 72:
```suggestion
    # The `commands` framework attempts a conversion of each type in this Union *in order*.
    # So, it will attempt to convert whatever is passed to `target` to a `discord.User` instance.
    # If that fails, it will attempt to convert it to a `discord.TextChannel` instance.
```

examples/converters.py
```diff
@@ -0,0 +1,94 @@
...
+    # See: https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html#typing-union
+    # NOTE: If a Union typehint converter fails it will raise `commands.errors.BadUnionArgument`
+    # instead of `commands.errors.BadArgument`.
+
+    # This is a more user friendly alternative to the above command.
```
NCPlayz, line 77:
```suggestion
    # This is a more user-friendly alternative to the previous command.
```

examples/converters.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # Let's check the type we actually got...
+    if isinstance(target, discord.User):
+        return await ctx.send("User found: {}".format(target))
+    elif isinstance(target, discord.TextChannel): # this could be an `else` but for completeness' sake.
+        return await ctx.send("Channel found: {}".format(target))
+
+# Builtin type converters
```
NCPlayz, line 85:
```suggestion
# Built-in type converters
```

examples/converters.py
```diff
@@ -0,0 +1,94 @@
...
+@bot.command()
+async def trial_converter(ctx: commands.Context, number: int, maybe: bool):
+    # We want and int and a bool parameter here.
```
NCPlayz, line 88:
```suggestion
    # We want an `int` and a `bool` parameter here.
```

examples/converters.py
```diff
@@ -0,0 +1,94 @@
...
+    # Bool is a slightly special case, as shown here:
```
NCPlayz, line 89:
```suggestion
    # `bool` is a slightly special case, as shown here:
```

examples/converters.py
```diff
@@ -0,0 +1,94 @@
...
+        return await ctx.send("Failed to convert the argument for `user` to `discord.User`.")
+    else:
+        return await ctx.send("Unhandled error: {}".format(error))
+
+# Manual use of converters will follow
+@bot.command()
+async def channel_or_member(ctx: commands.Context, argument: str):
+    # NOTE: command parameters are `str` type by default, the typehint above is just for completeness' sake.
+
+    # If you are doing this kind of thing commonly, consider making a Custom Converter
```
NCPlayz, line 39:
```suggestion
    # If you use this pattern commonly, consider making a Custom Converter.
```

examples/converters.py
```diff
@@ -0,0 +1,94 @@
...
+    # https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html#advanced-converters
```
NCPlayz, line 40:
```suggestion
    # See: https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html#advanced-converters
```

examples/converters.py
```diff
@@ -0,0 +1,94 @@
...
+        return await ctx.send("User found: {}".format(target))
+    elif isinstance(target, discord.TextChannel): # this could be an `else` but for completeness' sake.
+        return await ctx.send("Channel found: {}".format(target))
+
+# Builtin type converters
+@bot.command()
+async def trial_converter(ctx: commands.Context, number: int, maybe: bool):
+    # We want and int and a bool parameter here.
+    # Bool is a slightly special case, as shown here:
+    # https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html#bool
```
NCPlayz, line 90:
```suggestion
    # See: https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html#bool
```

NCPlayz (changes requested):

examples/converters.py
```diff
@@ -0,0 +1,94 @@
...
+    await ctx.send(user.mention)
+
+@userinfo.error
+async def userinfo_error(ctx: commands.Context, error: Exception):
+    # if the conversion above fails for any reason, it will raise `commands.errors.BadArgument`
+    # so we handle this in this error handler
+    if isinstance(error, commands.BadArgument):
+        return await ctx.send("Failed to convert the argument for `user` to `discord.User`.")
+    else:
+        return await ctx.send("Unhandled error: {}".format(error))
```
Bluenix2, line 32:
```suggestion
    if isinstance(error, commands.BadArgument):
        return await ctx.send("Couldn't find that user.")
    else:
        # Raise the error to the global error handler
        raise error
```
I think that the example should be more accurate to how users will use it.

examples/converters.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # This is a more user-friendly alternative to the previous command.
+
+    # Let's check the type we actually got...
+    if isinstance(target, discord.User):
+        return await ctx.send("User found: {}".format(target.mention))
+    elif isinstance(target, discord.TextChannel): # this could be an `else` but for completeness' sake.
+        return await ctx.send("Channel found: {}".format(target.mention))
+
+# Built-in type converters
```
mesub7, line 86:
```suggestion
# Built-in type converters.
```

examples/converters.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # If for any reason you have an ID, but it may be of one of more
+    # items, like a TextChannel or Member, you can manually call the converters
+    # like the following:
+    member_converter = commands.MemberConverter()
+    try:
+        # Try and convert to a Member instance.
+        member = await member_converter.convert(ctx, argument)
+    except commands.MemberNotFound:
+        # Could not convert to a Member instance
```
mesub7, line 51:
```suggestion
        # Could not convert to a Member instance.
```

examples/converters.py
```diff
@@ -0,0 +1,94 @@
...
+    # the value passed as `user` to a `discord.User` instance.
+
+    # If the conversion is successful we will have a User instance
+    # and can do the following:
+    await ctx.send(user.mention)
+
+@userinfo.error
+async def userinfo_error(ctx: commands.Context, error: Exception):
+    # if the conversion above fails for any reason, it will raise `commands.errors.BadArgument`
+    # so we handle this in this error handler
```
mesub7, line 28:
```suggestion
    # so we handle this in this error handler:
```

examples/converters.py
```diff
@@ -0,0 +1,94 @@
...
+    if isinstance(error, commands.BadArgument):
+        return await ctx.send("Failed to convert the argument for `user` to `discord.User`.")
+    else:
+        return await ctx.send("Unhandled error: {}".format(error))
+
+# Manual use of converters will follow
```
mesub7, line 35:
```suggestion
# Manual use of converters will follow:
```

mesub7 (changes requested):
One or two punctuation improvements.

fwrs (approved):

examples/converters.py
```diff
@@ -0,0 +1,95 @@
...
+
+intents = discord.Intents.default()
+intents.members = True
+
+bot = commands.Bot("!", intents=intents)
+
+
+@bot.command()
+async def userinfo(ctx: commands.Context, user: discord.User):
+    # In the command signature above, you can see that the `user`
```
Phxntxm, line 16: This should maybe mention that type hinting means converting **in discord.py** since I have seen some users assume this means it happens in normal Python

Phxntxm (changes requested):

NCPlayz (changes requested):
Unsure on whether calling it the "`commands` framework" or "`commands` extension" is better.

examples/converters.py
```diff
@@ -0,0 +1,98 @@
...
+    # parameter is typehinted to `discord.User`. This means that
+    # during command invocation we will attempt to convert
+    # the value passed as `user` to a `discord.User` instance.
+
+    # NOTE: typehinting acts as a converter within discord.py only.
```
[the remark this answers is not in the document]
mesub7: I would actually go further and say the `discord.ext.commands` framework to make it super clear.

examples/converters.py
```diff
@@ -0,0 +1,99 @@
...
+
+import typing
+
+import discord
+from discord.ext import commands
+
+intents = discord.Intents.default()
+intents.members = True
+
+bot = commands.Bot("!", intents=intents)
```
Rapptz, line 11:
```suggestion
bot = commands.Bot('!', intents=intents)
```

examples/converters.py
```diff
@@ -0,0 +1,99 @@
...
+    # during command invocation, the `commands` framework will
+    # attempt to convert the value passed for the `user` parameter
+    # to a `discord.User` instance.
+
+    # NOTE: typehinting acts as a converter within the `commands` framework only.
+    # In standard Python, it is use for documentation and IDE assistance purposes.
+
+    # If the conversion is successful, we will have a `discord.User` instance
+    # and can do the following:
+    await ctx.send(user.mention)
```
Rapptz, line 27: This should probably do something a bit more meaningful since this is kind of a terrible command.

examples/converters.py
```diff
@@ -0,0 +1,99 @@
...
+bot = commands.Bot("!", intents=intents)
+
+
+@bot.command()
+async def userinfo(ctx: commands.Context, user: discord.User):
+    # In the command signature above, you can see that the `user`
+    # parameter is typehinted to `discord.User`. This means that
+    # during command invocation, the `commands` framework will
+    # attempt to convert the value passed for the `user` parameter
+    # to a `discord.User` instance.
```
Rapptz, line 20: This should mention an example of a value being converted, e.g. IDs, mentions, and names.

examples/converters.py
```diff
@@ -0,0 +1,99 @@
...
+    # NOTE: typehinting acts as a converter within the `commands` framework only.
+    # In standard Python, it is use for documentation and IDE assistance purposes.
+
+    # If the conversion is successful, we will have a `discord.User` instance
+    # and can do the following:
+    await ctx.send(user.mention)
+
+@userinfo.error
+async def userinfo_error(ctx: commands.Context, error: Exception):
+    # if the conversion above fails for any reason, it will raise `commands.errors.BadArgument`
```
Rapptz, line 31:
```suggestion
    # if the conversion above fails for any reason, it will raise `commands.BadArgument`
```

The explicit namespace is not part of the contract of the library.

examples/converters.py
```diff
@@ -0,0 +1,99 @@
...
+
+    # NOTE: typehinting acts as a converter within the `commands` framework only.
+    # In standard Python, it is use for documentation and IDE assistance purposes.
+
+    # If the conversion is successful, we will have a `discord.User` instance
+    # and can do the following:
+    await ctx.send(user.mention)
+
+@userinfo.error
+async def userinfo_error(ctx: commands.Context, error: Exception):
```
Rapptz, line 30:
```suggestion
async def userinfo_error(ctx: commands.Context, error: commands.CommandError):
```

examples/converters.py
```diff
@@ -0,0 +1,99 @@
...
+    # if the conversion above fails for any reason, it will raise `commands.errors.BadArgument`
+    # so we handle this in this error handler:
+    if isinstance(error, commands.BadArgument):
+        return await ctx.send("Couldn't find that user.")
+    else:
+        # Raise the error to the global error handler
+        raise error
```
Rapptz, line 37: Don't do this.

It gives a false impression of how error handlers work. You don't propagate errors to them or have the ability to stop the propagation. 

I'd just remove it.

examples/converters.py
```diff
@@ -0,0 +1,99 @@
...
+    # Do the same for TextChannel...
+    textchannel_converter = commands.TextChannelConverter()
+    try:
+        channel = await textchannel_converter.convert(ctx, argument)
+    except commands.ChannelNotFound:
+        pass
+    else:
+        return await ctx.send("Channel found: {}".format(channel))
+
+    await ctx.send("No member or channel matching this ID was found.")
```
Rapptz, line 70: I think this entire snippet is bad practice and against the spirit of what I wanted this example to show -- which is defining custom converters. Doing it within the body of the command is just not a good one.

examples/converters.py
```diff
@@ -0,0 +1,99 @@
...
+
+@bot.command()
+async def alternative_channel_or_member(ctx: commands.Context, target: typing.Union[discord.Member, discord.TextChannel]):
```
Rapptz, line 73: Personally I'm against toy names for commands like these since the point of examples is also to be ran and used. No one will really use or see a command name like this.

Fundamentally examples should still show something inherently useful (even though I know it's hard to do so). It shouldn't be hard to come up with a concept command that does use either a member or a channel and naming it appropriately.

examples/converters.py
```diff
@@ -0,0 +1,99 @@
...
+
+    # Let's check the type we actually got...
+    if isinstance(target, discord.User):
+        return await ctx.send("User found: {}".format(target.mention))
+    elif isinstance(target, discord.TextChannel): # this could be an `else` but for completeness' sake.
+        return await ctx.send("Channel found: {}".format(target.mention))
+
+# Built-in type converters.
+@bot.command()
+async def trial_converter(ctx: commands.Context, number: int, maybe: bool):
```
Rapptz, line 92: The comment I made above about toy examples applies here as well.

Rapptz (changes requested):
I think there are decent ideas here but it could be improved.

AbstractUmbra:
Requested changes have been actioned.

examples/converters.py
```diff
@@ -0,0 +1,116 @@
...
+# Built-in type converters.
+@bot.command()
+async def multiply(ctx: commands.Context, number: int, maybe: bool):
+    # We want an `int` and a `bool` parameter here.
+    # `bool` is a slightly special case, as shown here:
+    # See: https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html#bool
+
+    if maybe is True:
+        return await ctx.send(str(number * 2))
+    return await ctx.send(str(number * 5))
```
Zomatree, line 114:
```suggestion
        return await ctx.send(number * 2)
    await ctx.send(number * 5)
```

examples/converters.py
```diff
@@ -0,0 +1,116 @@
...
+    # instead of `commands.BadArgument`.
+
+    # This is a an example of a Union converter to native discord.py types. For more flexibility please see the
+    # previous Custom Converter example.
+
+    # Let's check the type we actually got...
+    if isinstance(target, discord.User):
+        return await ctx.send("User found: {}, adding them to the ignore list.".format(target.mention))
+    elif isinstance(target, discord.TextChannel): # this could be an `else` but for completeness' sake.
+        return await ctx.send("Channel found: {}, adding it to the ignore list.".format(target.mention))
```
Zomatree, line 103:
```suggestion
        await ctx.send('Channel found: {}, adding it to the ignore list.'.format(target.mention))
```

examples/converters.py
```diff
@@ -0,0 +1,116 @@
...
+        textchannel_converter = commands.TextChannelConverter()
+        try:
+            channel = await textchannel_converter.convert(ctx, argument)
+        except commands.ChannelNotFound:
+            pass
+        else:
+            return channel
+
+        # In the case that it was not converted we should return None for expliciness' sake.
+        return None
```
Zomatree, line 71: i feel like it would be better to raise an error than to return `None` here

examples/converters.py
```diff
@@ -0,0 +1,116 @@
...
+    # See: https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html#typing-union
+    # NOTE: If a Union typehint converter fails it will raise `commands.BadUnionArgument`
+    # instead of `commands.BadArgument`.
+
+    # This is a an example of a Union converter to native discord.py types. For more flexibility please see the
+    # previous Custom Converter example.
+
+    # Let's check the type we actually got...
+    if isinstance(target, discord.User):
+        return await ctx.send("User found: {}, adding them to the ignore list.".format(target.mention))
```
Zomatree, line 101:
```suggestion
        await ctx.send('Member found: {}, adding them to the ignore list.'.format(target.mention))
```

examples/converters.py
```diff
@@ -0,0 +1,116 @@
...
+    # If that fails, it will attempt to convert it to a `discord.TextChannel` instance.
+    # See: https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html#typing-union
+    # NOTE: If a Union typehint converter fails it will raise `commands.BadUnionArgument`
+    # instead of `commands.BadArgument`.
+
+    # This is a an example of a Union converter to native discord.py types. For more flexibility please see the
+    # previous Custom Converter example.
+
+    # Let's check the type we actually got...
+    if isinstance(target, discord.User):
```
Zomatree, line 100:
```suggestion
    if isinstance(target, discord.Member):
```

examples/converters.py
```diff
@@ -0,0 +1,116 @@
...
+
+    if argument is None:
+        return await ctx.send("No channel or member found with this argument.")
+    await ctx.send("Locking down {}!".format(argument.mention))
+
+@bot.command()
+async def ignore(ctx: commands.Context, target: typing.Union[discord.Member, discord.TextChannel]):
+    # This command signature utilises the `typing.Union` typehint.
+    # The `commands` framework attempts a conversion of each type in this Union *in order*.
+    # So, it will attempt to convert whatever is passed to `target` to a `discord.User` instance.
```
Zomatree, line 90:
```suggestion
    # So, it will attempt to convert whatever is passed to `target` to a `discord.Member` instance.
```

examples/converters.py
```diff
@@ -0,0 +1,116 @@
...
+@bot.command()
+async def lockdown(ctx: commands.Context, argument: ChannelOrMemberConverter):
+    # So from the command signature, you can see that for `argument` we have typehinted
+    # the custom converter we defined previously.
+    # What will happen during command invocation is that the `argument` above will be passed to
+    # `ChannelOrMemberConverter.convert` and the conversion will go through the process defined there.
+
+    if argument is None:
+        return await ctx.send("No channel or member found with this argument.")
+    await ctx.send("Locking down {}!".format(argument.mention))
```
Zomatree, line 84:
```suggestion
    await ctx.send('Locking down {}!'.format(argument.mention))
```

examples/converters.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # NOTE: typehinting acts as a converter within the `commands` framework only.
+    # In standard Python, it is use for documentation and IDE assistance purposes.
+
+    # If the conversion is successful, we will have a `discord.User` instance
+    # and can do the following:
+    user_id = user.id
+    username = user.name
+    avatar = user.avatar_url
+    await ctx.send("User found: {} -- {}\n{}".format(user_id, username, avatar))
```
Zomatree, line 32:
```suggestion
    await ctx.send('User found: {} -- {}\n{}'.format(user_id, username, avatar))
```

Zomatree (changes requested):

examples/converters.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+@bot.command()
+async def lockdown(ctx: commands.Context, argument: ChannelOrMemberConverter):
+    # So from the command signature, you can see that for `argument` we have typehinted
+    # the custom converter we defined previously.
+    # What will happen during command invocation is that the `argument` above will be passed to
+    # `ChannelOrMemberConverter.convert` and the conversion will go through the process defined there.
+
+    await ctx.send('Locking down {}!'.format(argument.mention))
```
NCPlayz, line 83:
```suggestion
@bot.command()
async def notify(ctx: commands.Context, target: ChannelOrMemberConverter):
    # So from the command signature, you can see that for `argument` we have typehinted
    # the custom converter we defined previously.
    # What will happen during command invocation is that the `argument` above will be passed to
    # `ChannelOrMemberConverter.convert` and the conversion will go through the process defined there.

    await target.send('Hello, {}!'.format(target.name))
```

examples/converters.py
```diff
@@ -0,0 +1,115 @@
...
+        return await ctx.send('Couldn\'t find that user.')
+
+# Custom Converter here
+class ChannelOrMemberConverter(commands.Converter):
+    async def convert(self, ctx: commands.Context, argument: str):
+        # In this example we have made a custom converter.
+        # Specifically - to check if an input is convertable to a
+        # `discord.Member` or `discord.TextChannel` instance from a
+        # single input.
+        # E.g. we have an ID we want to add to a list of ignored IDs, or similar
```
Rapptz, line 48:
```suggestion
        # This checks if an input is convertible to a
        # `discord.Member` or `discord.TextChannel` instance from the
        # input the user has given us using the pre-existing converters
        # that the library provides.
```

examples/converters.py
```diff
@@ -0,0 +1,115 @@
...
+        member_converter = commands.MemberConverter()
+        try:
+            # Try and convert to a Member instance.
+            member = await member_converter.convert(ctx, argument)
+        except commands.MemberNotFound:
+            # Could not convert to a Member instance.
+            pass
+        else:
+            # We have our `member` so lets return here.
+            return member
```
Rapptz, line 59:
```suggestion
        try:
            # Try and convert to a Member instance.
            # If this fails, then an exception is raised.
            # Otherwise, we just return the converted member value.
            member = await member_converter.convert(ctx, argument)
        except commands.MemberNotFound:
            pass
        else:
            return member
```

examples/converters.py
```diff
@@ -0,0 +1,115 @@
...
+        try:
+            channel = await textchannel_converter.convert(ctx, argument)
+        except commands.ChannelNotFound:
+            pass
+        else:
+            return channel
+
+        # In the case that it was not converted we can return None or raise an error.
+        # I will raise an error.
+        raise commands.BadArgument('No Member or TextChannel could be converted from "{}"'.format(argument))
```
Rapptz, line 72:
```suggestion
        # If the value could not be converted we can raise an error
        # so our error handlers can deal with it in one place.
        # The error has to be CommandError derived, so BadArgument works fine here.
        raise commands.BadArgument('No Member or TextChannel could be converted from "{}"'.format(argument))
```

examples/converters.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+
+@bot.command()
+async def notify(ctx: commands.Context, target: ChannelOrMemberConverter):
+    # So from the command signature, you can see that for `target` we have typehinted
+    # the custom converter we defined previously.
+    # What will happen during command invocation is that the `target` above will be passed to
+    # `ChannelOrMemberConverter.convert` and the conversion will go through the process defined there.
```
Rapptz, line 81:
```suggestion
    # This command signature utilises the custom converter written above
    # What will happen during command invocation is that the `target` above will be passed to
    # the `argument` parameter of the `ChannelOrMemberConverter.convert` method and 
    # the conversion will go through the process defined there.
```

examples/converters.py
```diff
@@ -0,0 +1,115 @@
...
+    # This command signature utilises the `typing.Union` typehint.
+    # The `commands` framework attempts a conversion of each type in this Union *in order*.
+    # So, it will attempt to convert whatever is passed to `target` to a `discord.Member` instance.
+    # If that fails, it will attempt to convert it to a `discord.TextChannel` instance.
+    # See: https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html#typing-union
+    # NOTE: If a Union typehint converter fails it will raise `commands.BadUnionArgument`
+    # instead of `commands.BadArgument`.
+
+    # This is a an example of a Union converter to native discord.py types. For more flexibility please see the
+    # previous Custom Converter example.
```
Rapptz, line 96:
```suggestion
```

examples/converters.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # Let's check the type we actually got...
```
Rapptz, line 98:
```suggestion
    # To check the resulting type, `isinstance` is used
```

Rapptz (changes requested):

Rapptz (approved):

PR #6443, opened 2021-02-20
Add note in member.Activity about API limitation

kaylynn234:
## Summary

Adds a note about the behaviour outlined in #1738. I'm not entirely sure about the wording, but I can't really think of a better way to phrase it.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/member.py
+++ b/discord/member.py
@@ -406,6 +406,9 @@ class Member(discord.abc.Messageable, _BaseUser):
         activity the user is currently doing. Could be ``None`` if no activity is being done.
 
         .. note::
+            Due to a Discord API limitation, this may be ``None`` if the user is listening to a song on spotify with a long title.
+
+        .. note::
 
             A user may have multiple activities, these can be accessed under :attr:`activities`.
         """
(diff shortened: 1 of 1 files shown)


discord/member.py
```diff
@@ -405,6 +405,9 @@ def activity(self):
         """Union[:class:`BaseActivity`, :class:`Spotify`]: Returns the primary
         activity the user is currently doing. Could be ``None`` if no activity is being done.
 
+        .. note::
+            Due to a Discord API limitation, this may be ``None`` if the user is listening to a song on spotify with a long title.
```
NCPlayz, line 409:
```suggestion
        .. note::
        
            Due to a Discord API limitation, this may be ``None`` if the user is listening to a song on spotify with a long title.
```

kaylynn234: Ah whoops, I'd meant to add a blank line there but must've missed it.

NCPlayz (changes requested):

SnowyLuma:
This should probably also be documented under `Member.activities`, so it can't be missed.

Rapptz (changes requested):
I think @SnowyLuma 's point is good as well.

PR #6444, opened 2021-02-20
adds mutual_guilds method to the User class

### LINKED ISSUE #2539: Add mutual guild method to user class

sudosnok:
## Summary
Apologies for this again
I made a mistake with my vcs, so this is the same as when it was last… reviewed
Closes #2539 

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/user.py
+++ b/discord/user.py
@@ -713,7 +713,8 @@ class User(BaseUser, discord.abc.Messageable):
 
         .. note::
 
-            This can only return mutual guilds within the client's internal cache.
+            This will only return mutual guilds within the client's internal cache.
+
         .. versionadded:: 1.7
         """
         return [guild for guild in self._state._guilds.values() if guild.get_member(self.id)]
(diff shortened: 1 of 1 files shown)


discord/user.py
```diff
@@ -707,6 +707,17 @@ def dm_channel(self):
...
         return self._state._get_private_channel_by_user(self.id)
 
+    @property
+    def mutual_guilds(self):
+        """List[:class:`Guild`]: The guilds that the user shares with the client.
+
+        .. note::
+
+            This can only return mutual guilds within the client's internal cache.
+        .. versionadded:: 1.7
```
NCPlayz, line 717:
```suggestion
        .. note::

            This will only return mutual guilds within the client's internal cache.

        .. versionadded:: 1.7
```

Rapptz (approved):

PR #6445, opened 2021-02-20
[commands] Allow relative paths when handling extensions

### LINKED ISSUE #2465: Allow loading extensions via relative paths
### The Problem

I would like to use relative import paths to describe my extensions, e.g., `.extensions.foo` rather than requiring an absolute path like `my_bot.extensions.foo`. I use relative paths for package-internal imports, so it seems logical that extensions would follow the same pattern.

This does not currently work. I get the following error:

```
  File "/usr/lib64/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib64/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/chris/Projects/discord-fate-bot/discord_fate_bot/__main__.py", line 15, in <module>
    bot.load_extension('.extensions.error_handling')
  File "/home/chris/.local/share/virtualenvs/discord-fate-bot-MUSnTVTS/lib/python3.7/site-packages/discord/ext/commands/bot.py", line 621, in load_extension
    lib = importlib.import_module(name)
  File "/home/chris/.local/share/virtualenvs/discord-fate-bot-MUSnTVTS/lib64/python3.7/importlib/__init__.py", line 122, in import_module
    raise TypeError(msg.format(name))
TypeError: the 'package' argument is required to perform a relative import for '.extensions.error_handling'
```

It would seam this library is using `importlib` internally, which can handle relative paths, but is not passing enough parameters.

### The Ideal Solution

I would like `Bot.load_extension` to take an optional second parameter, which would specify the package to resolve against. This follows the example of `importlib.import_module`, which takes a package parameter in the same way. If no package is provided, the path is looked up as if from a top-level script (the current behavior).

This would allow us to say

```python
bot.load_extension('.extensions.foo', __package__)
```

which would load the extension relative to the current package.

### The Current Solution

Currently the solution is to only use absolute paths. I use `importlib.utils.resolve_name` to resolve my relative paths to absolute ones:

```python
from importlib.util import resolve_name
# ...
bot.load_extension(resolve_name('.extensions.foo', __package__))
```

### Summary

Please allow us to load extensions using relative paths via an optional package parameter in `Bot.load_extension`.

This feature request looks similar to #1703, but that request wants something similar old Python 2.x-style relative paths (just concatenating package and name with a dot). I would like something compatible with Python 3 and `importlib`.

I would be willing to work on a merge request. I just wanted to make an issue first in case there was a conscious decision not to support relative paths. (I would still ask you to reconsider.)

kaylynn234:
## Summary

Fixes #2465.
Adds an optional parameter to `load_extension`, `unload_extension` and `reload_extension` for resolving relative paths.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -659,7 +659,11 @@ class BotBase(GroupMixin):
             The extension or its setup function had an execution error.
         """
 
-        name = importlib.util.resolve_name(name, package)
+        try:
+            name = importlib.util.resolve_name(name, package)
+        except ImportError:
+            raise errors.ExtensionNotFound(name)
+
         if name in self.__extensions:
             raise errors.ExtensionAlreadyLoaded(name)
 
@@ -695,11 +699,17 @@ class BotBase(GroupMixin):
 
         Raises
         -------
+        ExtensionNotFound
+            The name of the extension could not be resolved using the provided ``package`` parameter.
         ExtensionNotLoaded
             The extension was not loaded.
         """
 
-        name = importlib.util.resolve_name(name, package)
+        try:
+            name = importlib.util.resolve_name(name, package)
+        except ImportError:
+            raise errors.ExtensionNotFound(name)
+
         lib = self.__extensions.get(name)
         if lib is None:
             raise errors.ExtensionNotLoaded(name)
@@ -740,7 +750,11 @@ class BotBase(GroupMixin):
             The extension setup function had an execution error.
         """
 
-        name = importlib.util.resolve_name(name, package)
+        try:
+            name = importlib.util.resolve_name(name, package)
+        except ImportError:
+            raise errors.ExtensionNotFound(name)
+
         lib = self.__extensions.get(name)
         if lib is None:
             raise errors.ExtensionNotLoaded(name)
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -624,7 +624,7 @@ class BotBase(GroupMixin):
         else:
             self.__extensions[key] = lib
 
-    def _resolve_name(name, package):
+    def _resolve_name(self, name, package):
         try:
             name = importlib.util.resolve_name(name, package)
         except ImportError:
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -624,7 +624,7 @@ class BotBase(GroupMixin):
         else:
             self.__extensions[key] = lib
 
-    def load_extension(self, name):
+    def load_extension(self, name, package=None):
         """Loads an extension.
 
         An extension is a python module that contains commands, cogs, or
@@ -640,6 +640,12 @@ class BotBase(GroupMixin):
             The extension name to load. It must be dot separated like
             regular Python imports if accessing a sub-module. e.g.
             ``foo.test`` if you want to import ``foo/test.py``.
+        package: Optional[:class:`str`]
+            The package name to resolve relative imports with.
+            This is required when loading an extension using a relative path, e.g ``.foo.test``.
+            Defaults to ``None``.
+
+            .. versionadded:: 1.7
 
         Raises
         --------
@@ -653,6 +659,7 @@ class BotBase(GroupMixin):
             The extension or its setup function had an execution error.
         """
 
+        name = importlib.util.resolve_name(name, package)
         if name in self.__extensions:
             raise errors.ExtensionAlreadyLoaded(name)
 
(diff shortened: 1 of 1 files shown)


Rapptz (changes requested):
According to the documentation, [`importlib.util.resolve_name`](https://docs.python.org/3/library/importlib.html#importlib.util.resolve_name) can raise `ImportError` which is undesirable. 

This should probably be wrapped and return `ExtensionNotFound` instead. Some functions aren't documented to raise that exception so it needs to be updated as well.

kaylynn234:
I've made the requested changes, though I'm somewhat unsure on wording in regards to the documentation. Is it worth detailing the case where ``ExtensionNotFound`` can be raised in ``load_extension`` and ``reload_extension`` by the wrapped ``resolve_name`` call?

Rapptz:
> I've made the requested changes, though I'm somewhat unsure on wording in regards to the documentation. Is it worth detailing the case where `ExtensionNotFound` can be raised in `load_extension` and `reload_extension` by the wrapped `resolve_name` call?

Yes.

I think if this code is being copy pasted around everywhere it should probably be in a helper method as well.

kaylynn234:
Alright.

discord/ext/commands/bot.py
```diff
@@ -624,7 +624,13 @@ def _load_from_module_spec(self, spec, key):
         else:
             self.__extensions[key] = lib
 
-    def load_extension(self, name):
+    def _resolve_name(name, package):
```
apple502j, line 627: Missing self?

kaylynn234: Oh whoops. My bad.

kaylynn234:
Should be all good now. Apologies for all the hassle.

discord/ext/commands/bot.py
```diff
@@ -624,7 +624,13 @@ def _load_from_module_spec(self, spec, key):
...
             self.__extensions[key] = lib
 
-    def load_extension(self, name):
+    def _resolve_name(self, name, package):
+        try:
+            name = importlib.util.resolve_name(name, package)
+        except ImportError:
+            raise errors.ExtensionNotFound(name)
+
+    def load_extension(self, name, package=None):
```
NCPlayz, line 633: Seems better to make `package` a keyword argument.

discord/ext/commands/bot.py
```diff
@@ -640,11 +646,18 @@ def load_extension(self, name):
...
+            This is required when loading an extension using a relative path, e.g ``.foo.test``.
+            Defaults to ``None``.
+
+            .. versionadded:: 1.7
 
         Raises
         --------
         ExtensionNotFound
             The extension could not be imported.
+            This is also raised if the name of the extension could not be resolved using the provided ``package`` parameter.
```
Rapptz, line 660:
```suggestion
            This is also raised if the name of the extension could 
            not be resolved using the provided ``package`` parameter.
```

discord/ext/commands/bot.py
Rapptz: Yeah not a bad idea to make them keyword only.

discord/ext/commands/bot.py
```diff
@@ -707,19 +730,27 @@ def reload_extension(self, name):
...
+
+            .. versionadded:: 1.7
 
         Raises
         -------
         ExtensionNotLoaded
             The extension was not loaded.
         ExtensionNotFound
             The extension could not be imported.
+            This is also raised if the name of the extension could not be resolved using the provided ``package`` parameter.
```
Rapptz, line 746:
```suggestion
            This is also raised if the name of the extension could not 
            be resolved using the provided ``package`` parameter.
```

discord/ext/commands/bot.py
```diff
@@ -679,21 +693,30 @@ def unload_extension(self, name):
...
+            The package name to resolve relative imports with.
+            This is required when unloading an extension using a relative path, e.g ``.foo.test``.
+            Defaults to ``None``.
+
+            .. versionadded:: 1.7
 
         Raises
         -------
+        ExtensionNotFound
+            The name of the extension could not be resolved using the provided ``package`` parameter.
```
Rapptz, line 706:
```suggestion
            The name of the extension could not
            be resolved using the provided ``package`` parameter.
```

Rapptz (changes requested):
Some minor doc nits.

kaylynn234:
Alright.

Rapptz (approved):
LGTM. Thanks

PR #6446, opened 2021-02-20
added tribool behaviour to HelpCommand.verify_checks

### LINKED ISSUE #2008: HelpCommand.VerifyChecks Tribool - Always, only in guild context, never
### The Problem

I would like it so I can achieve the following without having to make my own help implementation if possible:

 - If a user triggers help from a guild, the bot will verify checks when generating help (send to DMs or context dependant on option still)
 - If a user triggers help from DMs, the bot will ignore checks and show all commands

### The Ideal Solution

Make [HelpCommand.verify_checks](https://discordpy.readthedocs.io/en/rewrite/ext/commands/api.html#discord.ext.commands.HelpCommand.verify_checks) a Tribool with the following occurring:

 - True: Always verify checks
 - None: Only verify in guild context
 - False: Never verify

### The Current Solution

I'm trying to avoid using guild only checks and instead, writing them as part of the command so that they will still show in help if triggered from DMs.

### Summary

HelpCommand.VerifyChecks Tribool - Always, only in guild context, never

sudosnok:
## Summary
Allows verify_checks to be None, giving finer control over when check are run
Closes #2008 

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/help.py
+++ b/discord/ext/commands/help.py
@@ -272,7 +272,7 @@ class HelpCommand:
     show_hidden: :class:`bool`
         Specifies if hidden commands should be shown in the output.
         Defaults to ``False``.
-    verify_checks: Union[:class:`bool`, :class:`None`]
+    verify_checks: Optional[:class:`bool`]
         Specifies if commands should have their :attr:`.Command.checks` called
         and verified. Defaults to ``True``.
     command_attrs: :class:`dict`
--- a/discord/ext/commands/help.py
+++ b/discord/ext/commands/help.py
@@ -274,9 +274,9 @@ class HelpCommand:
         Defaults to ``False``.
     verify_checks: Optional[:class:`bool`]
         Specifies if commands should have their :attr:`.Command.checks` called
-        and verified. If True; always calls :attr:`.Commands.checks`.
-        If None; only calls :attr:`.Commands.checks` in a guild setting.
-        If False; never calls :attr:`.Commands.checks`. Defaults to ``True``.
+        and verified. If ``True``, always calls :attr:`.Commands.checks`.
+        If ``None``, only calls :attr:`.Commands.checks` in a guild setting.
+        If ``False``, never calls :attr:`.Commands.checks`. Defaults to ``True``.
 
         ..versionchanged:: 1.7
     command_attrs: :class:`dict`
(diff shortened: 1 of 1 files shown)


discord/ext/commands/help.py
```diff
@@ -272,7 +272,7 @@ class HelpCommand:
     show_hidden: :class:`bool`
         Specifies if hidden commands should be shown in the output.
         Defaults to ``False``.
-    verify_checks: :class:`bool`
+    verify_checks: Union[:class:`bool`, :class:`None`]
```
Rapptz, line 275:
```suggestion
    verify_checks: Optional[:class:`bool`]
```

The ``None`` behaviour is completely undocumented.

Rapptz (changes requested):

discord/ext/commands/help.py
sudosnok: No worries, was uhming and ahing about how to document a tribool

NCPlayz:
I generally feel like this tribool should be an enum as it is easier to document the behaviour in a readable manner, as well as being easy to understand the behaviour internally.

discord/ext/commands/help.py
```diff
@@ -272,7 +272,7 @@ class HelpCommand:
     show_hidden: :class:`bool`
         Specifies if hidden commands should be shown in the output.
         Defaults to ``False``.
-    verify_checks: :class:`bool`
+    verify_checks: Optional[:class:`bool`]
         Specifies if commands should have their :attr:`.Command.checks` called
         and verified. Defaults to ``True``.
```
Rapptz, line 277: The `None` behaviour is still not documented.

Rapptz (changes requested):

discord/ext/commands/help.py
```diff
@@ -274,7 +274,11 @@ class HelpCommand:
         Defaults to ``False``.
     verify_checks: Optional[:class:`bool`]
         Specifies if commands should have their :attr:`.Command.checks` called
-        and verified. Defaults to ``True``.
+        and verified. If True; always calls :attr:`.Commands.checks`.
+        If None; only calls :attr:`.Commands.checks` in a guild setting.
+        If False; never calls :attr:`.Commands.checks`. Defaults to ``True``.
```
Rapptz, line 279:
```suggestion
        and verified. If ``True``, always calls :attr:`.Commands.checks`.
        If ``None``, only calls :attr:`.Commands.checks` in a guild setting.
        If ``False``, never calls :attr:`.Commands.checks`. Defaults to ``True``.
```

Rapptz (changes requested):
Sorry, some other minor nits.

Rapptz (approved):

PR #6448, opened 2021-02-20
"Secret" example

NCPlayz:
## Summary

Shows how to use permission overwrites to create secret voice/text channels, as well as secret role-locked emojis.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/secret.py
+++ b/examples/secret.py
@@ -1,0 +1,98 @@
+import typing
+
+import discord
+from discord.ext import commands
+
+bot = commands.Bot(command_prefix=commands.when_mentioned, description="Nothing to see here!")
+
+@bot.event
+async def on_command(ctx):
+    try:
+        await ctx.message.delete(delay=5)
+    except discord.HTTPException:
+        # Oh well, looks like we don't have the ability
+        # to hide the secret.
+        pass
+
+# the `hidden` keyword argument hides it from the help command. 
+@bot.group(invoke_without_command=True, hidden=True)
+async def secret(ctx: commands.Context):
+    """What is this "secret" you speak of?"""
+    if ctx.invoked_subcommand is None:
+        await ctx.send('Shh!', delete_after=5)
+
+def create_overwrites(ctx, *objects):
+    """This is just a helper function that creates the overwrites for the 
+    voice/text channels.
+
+    A `discord.PermissionOverwrite` allows you to determine the permissions
+    of an object, whether it be a `discord.Role` or a `discord.Member`.
+
+    In this case, the `read_messages` permission is being used to hide the channel
+    from being viewed by whoever does not meet the criteria, thus creating a
+    secret channel.
+    """
+
+    overwrites = {
+        obj: discord.PermissionOverwrite(read_messages=True) for obj in objects
+    }
+ 
+    overwrites.setdefault(ctx.guild.default_role, discord.PermissionOverwrite(read_messages=False))
+    # prevents the default role (@everyone) from viewing the channel
+    # if it isn't already allowed to view the channel.
+    
+    overwrites[ctx.guild.me] = discord.PermissionOverwrite(read_messages=True)
+    # makes sure the client is always allowed to view the channel.
+
+    return overwrites
+
+@commands.guild_only()
+@secret.command()
+async def text(ctx: commands.Context, name: str, *objects: typing.Union[discord.Role, discord.Member]):
+    """This may or may not make a text channel with a specified name 
+    that is only visible to roles or members that are specified.
+    """
+    
+    overwrites = create_overwrites(ctx, *objects)
+
+    await ctx.guild.create_text_channel(
+        name,
+        overwrites=overwrites,
+        topic='Top secret text channel. Any leakage of this channel may result in serious trouble.',
+        reason='Very secret business.',
+    )
+
+@commands.guild_only()
+@secret.command()
+async def voice(ctx: commands.Context, name: str, *objects: typing.Union[discord.Role, discord.Member]):
+    """This may or may not do the same thing as the `text` subcommand
+    but instead creates a voice channel.
+    """
+
+    overwrites = create_overwrites(ctx, *objects)
+
+    await ctx.guild.create_voice_channel(
+        name,
+        overwrites=overwrites,
+        reason='Very secret business.'
+    )
+
+@commands.guild_only()
+@secret.command()
+async def emoji(ctx: commands.Context, emoji: discord.PartialEmoji, *roles: discord.Role):
+    """There is a slight chance this could very well clone an emoji
+    that only specified roles are allowed to use.
+    """
+
+    emoji_bytes = await emoji.url.read()
+    # fetch the emoji asset and read it as bytes
+
+    await ctx.guild.create_custom_emoji(
+        name=emoji.name,
+        image=emoji_bytes,
+        roles=roles,
+        reason='Very secret business.'
+    )
+
+
+bot.run('token')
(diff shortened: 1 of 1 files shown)


examples/secret.py
```diff
@@ -0,0 +1,98 @@
...
+bot = commands.Bot(command_prefix=commands.when_mentioned, description="Nothing to see here!")
+
+@bot.event
+async def on_command(ctx):
+    try:
+        await ctx.message.delete(delay=5)
+    except discord.HTTPException:
+        # Oh well, looks like we don't have the ability
+        # to hide the secret.
+        pass
```
Bluenix2, line 15: I think this can have unexpected outcomes for users building on this.

examples/secret.py
```diff
@@ -0,0 +1,98 @@
...
+    of an object, whether it be a `discord.Role` or a `discord.Member`.
+
+    In this case, the `read_messages` permission is being used to hide the channel
+    from being viewed by whoever does not meet the criteria, thus creating a
+    secret channel.
+    """
+
+    overwrites = {
+        obj: discord.PermissionOverwrite(read_messages=True) for obj in objects
+    }
```
Bluenix2, line 38: I think the dict comprehension can be confusing for some lesser experienced users. At least add a comment saying it's a dict comprehension so that they can look it up?

examples/secret.py
```diff
@@ -0,0 +1,98 @@
...
+ 
+    overwrites.setdefault(ctx.guild.default_role, discord.PermissionOverwrite(read_messages=False))
+    # prevents the default role (@everyone) from viewing the channel
+    # if it isn't already allowed to view the channel.
```
Bluenix2, line 42:
```suggestion
    # prevents the default role (@everyone) from viewing the channel
    # if it isn't already allowed to view the channel.
    overwrites.setdefault(ctx.guild.default_role, discord.PermissionOverwrite(read_messages=False))
```
Comments should go above

examples/secret.py
```diff
@@ -0,0 +1,98 @@
...
+    
+    overwrites[ctx.guild.me] = discord.PermissionOverwrite(read_messages=True)
+    # makes sure the client is always allowed to view the channel.
```
Bluenix2, line 45:
```suggestion
    # makes sure the client is always allowed to view the channel.
    overwrites[ctx.guild.me] = discord.PermissionOverwrite(read_messages=True)
```

examples/secret.py
```diff
@@ -0,0 +1,98 @@
+import typing
+
+import discord
+from discord.ext import commands
+
+bot = commands.Bot(command_prefix=commands.when_mentioned, description="Nothing to see here!")
+
+@bot.event
+async def on_command(ctx):
```
Bluenix2, line 9:
```suggestion
async def on_command(ctx: commands.Context):
```
If this is going to be kept, be consistent with typing ctx.

examples/secret.py
```diff
@@ -0,0 +1,98 @@
...
+    try:
+        await ctx.message.delete(delay=5)
+    except discord.HTTPException:
+        # Oh well, looks like we don't have the ability
+        # to hide the secret.
+        pass
+
+# the `hidden` keyword argument hides it from the help command. 
+@bot.group(invoke_without_command=True, hidden=True)
```
Bluenix2, line 18: Add comments about using groups perhaps. I don't think groups and subcommands have been used in many other examples.

examples/secret.py
```diff
@@ -0,0 +1,98 @@
...
+    overwrites.setdefault(ctx.guild.default_role, discord.PermissionOverwrite(read_messages=False))
+    # prevents the default role (@everyone) from viewing the channel
+    # if it isn't already allowed to view the channel.
+    
+    overwrites[ctx.guild.me] = discord.PermissionOverwrite(read_messages=True)
+    # makes sure the client is always allowed to view the channel.
+
+    return overwrites
+
+@commands.guild_only()
```
Bluenix2, line 49: Maybe comment on the guild_only check.

examples/secret.py
```diff
@@ -0,0 +1,98 @@
...
+@secret.command()
+async def emoji(ctx: commands.Context, emoji: discord.PartialEmoji, *roles: discord.Role):
+    """There is a slight chance this could very well clone an emoji
+    that only specified roles are allowed to use.
+    """
+
+    emoji_bytes = await emoji.url.read()
+    # fetch the emoji asset and read it as bytes
```
Bluenix2, line 88:
```suggestion
    # fetch the emoji asset and read it as bytes
    emoji_bytes = await emoji.url.read()
```

examples/secret.py
Rapptz, line 15: There is no point to this event being there other than a gimmick.

examples/secret.py
Rapptz, line 18: `invoke_without_command=True` changes the invocation order in such a way that the check being used in the body (`ctx.invoked_subcommand is None`) makes no sense and would always be triggered.

examples/secret.py
```diff
@@ -0,0 +1,98 @@
...
+    A `discord.PermissionOverwrite` allows you to determine the permissions
+    of an object, whether it be a `discord.Role` or a `discord.Member`.
+
+    In this case, the `read_messages` permission is being used to hide the channel
+    from being viewed by whoever does not meet the criteria, thus creating a
+    secret channel.
+    """
+
+    overwrites = {
+        obj: discord.PermissionOverwrite(read_messages=True) for obj in objects
```
Rapptz, line 37:
```suggestion
        obj: discord.PermissionOverwrite(read_messages=True) 
        for obj in objects
```

examples/secret.py
```diff
@@ -0,0 +1,98 @@
...
+        await ctx.send('Shh!', delete_after=5)
+
+def create_overwrites(ctx, *objects):
+    """This is just a helper function that creates the overwrites for the 
+    voice/text channels.
+
+    A `discord.PermissionOverwrite` allows you to determine the permissions
+    of an object, whether it be a `discord.Role` or a `discord.Member`.
+
+    In this case, the `read_messages` permission is being used to hide the channel
```
Rapptz, line 31: `read_messages` as a permission for hiding the channel seems a bit unintuitive. The library has permission aliases now so maybe `view_channel` should be used instead.

examples/secret.py
Rapptz, line 42: Please place comments above the thing it's explaining, not below.

examples/secret.py
```diff
@@ -0,0 +1,98 @@
...
+    # prevents the default role (@everyone) from viewing the channel
+    # if it isn't already allowed to view the channel.
+    
+    overwrites[ctx.guild.me] = discord.PermissionOverwrite(read_messages=True)
+    # makes sure the client is always allowed to view the channel.
+
+    return overwrites
+
+@commands.guild_only()
+@secret.command()
```
Rapptz, line 50: Checks go below the command decorator.

examples/secret.py
```diff
@@ -0,0 +1,98 @@
...
+async def text(ctx: commands.Context, name: str, *objects: typing.Union[discord.Role, discord.Member]):
+    """This may or may not make a text channel with a specified name
```
Rapptz, line 52: This type of comment is cheeky but I think it can also be confusing to a reader. Best to just be explicit when explaining.

Rapptz (changes requested):
A lot of the comments I mentioned apply to various parts of the example, I didn't feel like re-iterating them.

Rapptz (approved):
LGTM.

PR #6453, opened 2021-02-21
Adds an invoked_parents to context

### LINKED ISSUE #1874: Rewrite: please add a `invoked_parent` feature.
I know it's not a must, I can do it regularly, with `{ctx.message.content[0:(len(ctx.message.content)-(len(ctx.invoked_with)+1))]}` inside the subcommand, but is that really how you wanna do it? is that good code? I just thought that `invoked_parent` would be good to add, thank you.

sudosnok:
## Summary
This adds an invoked_parents property allowing easy access to the aliases of the parent commands to be accessed in a mapping of {Group : str}

Closes #1874 
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/context.py
+++ b/discord/ext/commands/context.py
@@ -212,7 +212,7 @@ class Context(discord.abc.Messageable):
 
     @property
     def invoked_parents(self):
-        """Mapping[:class:`Group`, :class:`str`] A readonly mapping of parent command to the alias used for that parent.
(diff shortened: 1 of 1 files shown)


Rapptz:
I'm definitely not a fan of the way this implementation was done.

The way I imagined this attribute was similar to how `invoked_with` and `parents` are done. A rolling list that's populated as an attribute within the context. As it stands right now this walks over the view a second time to do its operation.

PR #6454, opened 2021-02-21
deprecate non-bot methods

NCPlayz:
## Summary

This PR aims to deprecate the non-bot methods in preparation for their removal at v2.0. 
I'm sure I missed something so critique welcome.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/calls.py
+++ b/discord/calls.py
@@ -36,6 +36,8 @@ class CallMessage:
     This is only received in cases where the message type is equivalent to
     :attr:`MessageType.call`.
 
+    .. deprecated:: 1.7
+
     Attributes
     -----------
     ended_timestamp: Optional[:class:`datetime.datetime`]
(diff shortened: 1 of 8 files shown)


Rapptz:
This is missing `GroupCall` and `Message.call` which also deal with user bots. I'm not really sure what else there is.

Note however that "deprecated" is already a built-in Sphinx directive, [it works similarly to versionchanged](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated) and should be what is used here, not a note directive.

Rapptz (changes requested):
Need to deprecate the `Message.call` attribute as well.

Zomatree:
Will we also deprecate the `bot` kwarg in `Client.run/start/login` for logging into a user account?

Rapptz (approved):

PR #6456, opened 2021-02-22
 [tasks] Add support for explicit time parameter

### LINKED ISSUE #2159: [tasks] Allow running a function at exact times.

SebbyLaw:
## Summary

This pull request does a few things:
- properly document types such as `Union[int, float]` for relative times
- make `Loop.seconds`, `Loop.minutes`, and `Loop.hours` read-only properties
- document properties `Loop.seconds`, `Loop.minutes`, and `Loop.hours`
- **Add support for explicit time parameter, resolves #2159**

This PR is mostly inspired by 9f822a1 and smooths out some of the issues that caused it to be reverted.

One issue that may arise with this implementation is long running tasks causing drift that skips entire days.

For example:
```py
import asyncio, datetime
from discord.ext import tasks

@tasks.loop(time=[datetime.time(23, 59, 0), datetime.time(23, 59, 30), datetime.time.max])
async def my_horrible_coroutine():
    await asyncio.sleep(600)
```

This would
1. calculate the first and second iterations properly on day 1
2. run the first iteration at 23:59 on day 1
3. calculate the third iteration to run "tomorrow" (wrong day)
4. run the second iteration "late" early into day 2
4. "skip" the first two iterations, and run the third iteration at midnight on day 2

I'm not sure if there's any way around this issue, but feel free to suggest solutions.

This PR was not tested. If you'd like to try it out, you can install the branch with pip:
```
pip install git+https://github.com/SebbyLaw/discord.py.git@feature/tasks-explicit-time
```

## Checklist

- [x] This PR fixes an issue.


discord/ext/tasks/__init__.py
```diff
@@ -143,6 +149,40 @@ def __get__(self, obj, objtype):
...
+        """
+        return self._hours
+
+    @property
+    def time(self):
+        """
+        Optional[List[:class:`datetime.time`]]: Read-only list for the exact times this loop runs at.
+        ``None`` if relative times were passed instead.
+
+        .. versionadded:: 1.7.0
```
NCPlayz, line 182:
```suggestion
        .. versionadded:: 1.7
```

I think it's generally preferred to hide the patch for minor bumps.

discord/ext/tasks/__init__.py
```diff
@@ -143,6 +149,40 @@ def __get__(self, obj, objtype):
...
+        """
+        Optional[Union[:class:`int`, :class:`float`]]: Read-only value for the number of hours
+        between each iteration. ``None`` if an explicit ``time`` value was passed instead.
+        """
+        return self._hours
+
+    @property
+    def time(self):
+        """
+        Optional[List[:class:`datetime.time`]]: Read-only list for the exact times this loop runs at.
```
NCPlayz, line 179:
```suggestion
        """Optional[List[:class:`datetime.time`]]: Read-only list for the exact times this loop runs at.
```

discord/ext/tasks/__init__.py
```diff
@@ -143,6 +149,40 @@ def __get__(self, obj, objtype):
...
+        """
+        Optional[Union[:class:`int`, :class:`float`]]: Read-only value for the number of minutes
+        between each iteration. ``None`` if an explicit ``time`` value was passed instead.
+        """
+        return self._minutes
+    
+    @property
+    def hours(self):
+        """
+        Optional[Union[:class:`int`, :class:`float`]]: Read-only value for the number of hours
```
NCPlayz, line 171:
```suggestion
        """Optional[Union[:class:`int`, :class:`float`]]: Read-only value for the number of hours
```

discord/ext/tasks/__init__.py
```diff
@@ -143,6 +149,40 @@ def __get__(self, obj, objtype):
...
+        """
+        Optional[Union[:class:`int`, :class:`float`]]: Read-only value for the number of seconds
+        between each iteration. ``None`` if an explicit ``time`` value was passed instead.
+        """
+        return self._seconds
+    
+    @property
+    def minutes(self):
+        """
+        Optional[Union[:class:`int`, :class:`float`]]: Read-only value for the number of minutes
```
NCPlayz, line 163:
```suggestion
        """Optional[Union[:class:`int`, :class:`float`]]: Read-only value for the number of minutes
```

discord/ext/tasks/__init__.py
```diff
@@ -143,6 +149,40 @@ def __get__(self, obj, objtype):
         setattr(obj, self.coro.__name__, copy)
         return copy
 
+    @property
+    def seconds(self):
+        """
+        Optional[Union[:class:`int`, :class:`float`]]: Read-only value for the number of seconds
```
NCPlayz, line 155:
```suggestion
        """Optional[Union[:class:`int`, :class:`float`]]: Read-only value for the number of seconds
```

discord/ext/tasks/__init__.py
```diff
@@ -442,40 +526,72 @@ def change_interval(self, *, seconds=0, minutes=0, hours=0):
...
         The number of minutes between every iteration.
-    hours: :class:`float`
+    hours: Union[:class:`int`, :class:`float`]
         The number of hours between every iteration.
+    time: Union[:class:`datetime.time`, Sequence[:class:`datetime.time`]]
+        The exact times to run this loop at. Either a non-empty list or a single
+        value of :class:`datetime.time` should be passed.
+        This cannot be used in conjunction with the relative time parameters.
+
+        .. versionadded:: 1.7.0
```
NCPlayz, line 589:
```suggestion
        .. versionadded:: 1.7
```

NCPlayz (changes requested):

discord/ext/tasks/__init__.py
```diff
@@ -428,9 +468,53 @@ def error(self, coro):
...
+                break
+        else:
+            self._time_index = 0
+
+    def _get_time_parameter(self, time, *, inst=isinstance, dt=datetime.time, utc=datetime.timezone.utc):
+        if inst(time, dt):
+            ret = time if time.tzinfo is not None else time.replace(tzinfo=utc)
+            return [ret]
+        if not inst(time, Sequence):
+            raise TypeError('time parameter must be datetime.time or a sequence of datetime.time')
```
NCPlayz, line 500:
```suggestion
            raise TypeError('Expected datetime.time or a sequence of datetime.time for ``time``, received {0.__name__!r} instead.'.format(type(time)))
```

discord/ext/tasks/__init__.py
```diff
@@ -428,9 +468,53 @@ def error(self, coro):
...
+        if not time:
+            raise ValueError('time parameter must not be an empty sequence.')
+
+        ret = []
+        for index, t in enumerate(time):
+            if not inst(t, dt):
+                raise TypeError('index %d of time sequence expected %r not %r' % (index, dt, type(t)))
```
NCPlayz, line 507:
```suggestion
                raise TypeError('Expected a sequence of {0!r} for ``time``, received {1.__name__!r} at index {2} instead.'.format(dt, type(t), index))
```

diceroll123:
> One issue that may arise with this implementation is long running tasks causing drift that skips entire days.

Running the coroutine as a task internally would rectify that, no? Even if the previous iteration is still running, whatever, run the next one 24h after the previous started as a new task, that kind of thing...

Maybe a `wait=True` to run it "synchronously" *for lack of better terms*?

This implementation will have layers. 🧅

diceroll123:
Also, did you mean to commit this line?

`await sleep_until(self._next_iteration) # allows canceling in before_loop`

The same line (without the comment) to sleep happens later as well. That line originally was just a hack to be able to cancel the loop in before_loop

SebbyLaw:
> Also, did you mean to commit this line?
> 
> 
> 
> `await sleep_until(self._next_iteration) # allows canceling in before_loop`
> 
> 
> 
> The same line (without the comment) to sleep happens later as well. That line originally was just a hack to be able to cancel the loop in before_loop

Yep. It's still there as a hack to be able to cancel the loop in before_loop in the case of relative times being passed: it'll sleep for zero. It just has a little bit more functionality now when an explicit time is passed.

fwrs (approved):

SebbyLaw:
> One issue that may arise with this implementation is long running tasks causing drift that skips entire days.

Latest changes (14f5b21) should "fix" this, or at least make it consistent with current behavior. Overrunning the time for next_iteration doesn't skip anything since iteration times are now calculated based off previous iteration times.

SebbyLaw:
Ready for final round of review/testing

discord/ext/tasks/__init__.py
```diff
@@ -143,6 +149,36 @@ def __get__(self, obj, objtype):
...
+        return self._hours
+
+    @property
+    def time(self):
+        """Optional[List[:class:`datetime.time`]]: Read-only list for the exact times this loop runs at.
+        ``None`` if relative times were passed instead.
+
+        .. versionadded:: 1.7
+        """
+        return self._time.copy()
```
Zomatree, line 180: this will error if `self._time` is None

discord/ext/tasks/__init__.py
```diff
@@ -143,6 +149,36 @@ def __get__(self, obj, objtype):
         setattr(obj, self.coro.__name__, copy)
         return copy
 
+    @property
+    def seconds(self):
+        """Optional[:class:`float`]: Read-only value for the number of seconds
+        between each iteration. ``None`` if an explicit ``time`` value was passed instead.
+        """
```
Zomatree, line 156:
```suggestion
        """Optional[:class:`float`]: Read-only value for the number of seconds
        between each iteration. ``None`` if an explicit ``time`` value was passed instead.
        
        .. versionadded:: 1.7
        """
```

discord/ext/tasks/__init__.py
```diff
@@ -143,6 +149,36 @@ def __get__(self, obj, objtype):
...
+        return self._seconds
+    
+    @property
+    def minutes(self):
+        """Optional[:class:`float`]: Read-only value for the number of minutes
+        between each iteration. ``None`` if an explicit ``time`` value was passed instead.
+        """
```
Zomatree, line 163:
```suggestion
        """Optional[:class:`float`]: Read-only value for the number of minutes
        between each iteration. ``None`` if an explicit ``time`` value was passed instead.
        
        .. versionadded:: 1.7
        """
```

discord/ext/tasks/__init__.py
```diff
@@ -143,6 +149,36 @@ def __get__(self, obj, objtype):
...
+        return self._minutes
+    
+    @property
+    def hours(self):
+        """Optional[:class:`float`]: Read-only value for the number of hours
+        between each iteration. ``None`` if an explicit ``time`` value was passed instead.
+        """
```
Zomatree, line 170:
```suggestion
        """Optional[:class:`float`]: Read-only value for the number of hours
        between each iteration. ``None`` if an explicit ``time`` value was passed instead.
        
        .. versionadded:: 1.7
        """
```

Zomatree (changes requested):

SebbyLaw:
@Zomatree I'm not sure about this versionadded on seconds, minutes, hours. They've existed since v1.1 ext.tasks release, just never documented. Although now they're coerced into floats, they can be `None`, and they're read-only properties.

discord/ext/tasks/__init__.py
```diff
@@ -134,15 +140,52 @@ def __get__(self, obj, objtype):
...
         copy._after_loop = self._after_loop
         copy._error = self._error
         setattr(obj, self.coro.__name__, copy)
         return copy
 
+    @property
+    def seconds(self):
+        """Optional[:class:`float`]: Read-only value for the number of seconds
+        between each iteration. ``None`` if an explicit ``time`` value was passed instead.
+
```
NCPlayz, line 156:
```suggestion

```

discord/ext/tasks/__init__.py
```diff
@@ -476,6 +590,17 @@ def loop(*, seconds=0, minutes=0, hours=0, count=None, reconnect=True, loop=None
         The number of minutes between every iteration.
     hours: :class:`float`
         The number of hours between every iteration.
+    time: Union[:class:`datetime.time`, Sequence[:class:`datetime.time`]]
+        The exact times to run this loop at. Either a non-empty list or a single
+        value of :class:`datetime.time` should be passed.
+        This cannot be used in conjunction with the relative time parameters.
+
+        .. versionadded:: 1.7
+
```
NCPlayz, line 599:
```suggestion

```

discord/ext/tasks/__init__.py
```diff
@@ -134,15 +140,52 @@ def __get__(self, obj, objtype):
...
+        
+        .. versionadded:: 1.7
+        """
+        return self._minutes
+    
+    @property
+    def hours(self):
+        """Optional[:class:`float`]: Read-only value for the number of hours
+        between each iteration. ``None`` if an explicit ``time`` value was passed instead.
+
```
NCPlayz, line 174:
```suggestion

```

discord/ext/tasks/__init__.py
```diff
@@ -134,15 +140,52 @@ def __get__(self, obj, objtype):
...
+        
+        .. versionadded:: 1.7
+        """
+        return self._seconds
+    
+    @property
+    def minutes(self):
+        """Optional[:class:`float`]: Read-only value for the number of minutes
+        between each iteration. ``None`` if an explicit ``time`` value was passed instead.
+
```
NCPlayz, line 165:
```suggestion

```

NCPlayz (changes requested):

discord/ext/tasks/__init__.py
```diff
@@ -134,15 +140,49 @@ def __get__(self, obj, objtype):
...
         copy._after_loop = self._after_loop
         copy._error = self._error
         setattr(obj, self.coro.__name__, copy)
         return copy
 
+    @property
+    def seconds(self):
+        """Optional[:class:`float`]: Read-only value for the number of seconds
+        between each iteration. ``None`` if an explicit ``time`` value was passed instead.
+        .. versionadded:: 1.7
```
Rapptz, line 156:
```suggestion
        """Optional[:class:`float`]: Read-only value for the number of seconds
        between each iteration. ``None`` if an explicit ``time`` value was passed instead.

        .. versionadded:: 1.7
```

Same for the other ones.

discord/ext/tasks/__init__.py
```diff
@@ -428,9 +468,55 @@ def error(self, coro):
...
 
-    def change_interval(self, *, seconds=0, minutes=0, hours=0):
+        if self._current_loop == 0:
+            self._time_index += 1
+            return datetime.datetime.combine(datetime.datetime.now(datetime.timezone.utc), next_time)
+
+        next_date = self._last_iteration
+        if self._time_index == 0:
+            # we can assume that the earliest time should be scheduled for "tomorrow"
+            next_date += datetime.timedelta(days=1)
```
Rapptz, line 486: So how does this assumption work if it's 6PM and we have the following times set?

```py
times=[
    time(19, 0),
    time(20, 0),
    time(22, 0),
]
```

Is this only fired on 7PM the next day?

discord/ext/tasks/__init__.py
```diff
@@ -448,23 +534,48 @@ def change_interval(self, *, seconds=0, minutes=0, hours=0):
...
+                # prepare the next time index starting from after the next iteration
+                self._prepare_time_index(now=self._next_iteration)
 
-        self._sleep = sleep
-        self.seconds = seconds
-        self.hours = hours
-        self.minutes = minutes
 
-def loop(*, seconds=0, minutes=0, hours=0, count=None, reconnect=True, loop=None):
+def loop(*, seconds=0.0, minutes=0.0, hours=0.0, count=None, time=None, reconnect=True, loop=None):
```
Rapptz, line 578:
```suggestion
def loop(*, seconds=0, minutes=0, hours=0, count=None, time=None, reconnect=True, loop=None):
```

No point in using float literals.

discord/ext/tasks/__init__.py
```diff
@@ -476,6 +587,16 @@ def loop(*, seconds=0, minutes=0, hours=0, count=None, reconnect=True, loop=None
...
         The number of hours between every iteration.
+    time: Union[:class:`datetime.time`, Sequence[:class:`datetime.time`]]
+        The exact times to run this loop at. Either a non-empty list or a single
+        value of :class:`datetime.time` should be passed.
+        This cannot be used in conjunction with the relative time parameters.
+
+        .. versionadded:: 1.7
+        .. note::
+
+            Duplicate times will be ignored, and only run once.
```
Rapptz, line 598:
```suggestion
        The exact times to run this loop at. Either a non-empty list or a single
        value of :class:`datetime.time` should be passed. Timezones are supported.
        If no timezone is given for the times, it is assumed to represent UTC time. 

        This cannot be used in conjunction with the relative time parameters.

        .. note::

            Duplicate times will be ignored, and only run once.

        .. versionadded:: 1.7
```

discord/ext/tasks/__init__.py
```diff
@@ -134,15 +140,49 @@ def __get__(self, obj, objtype):
         if obj is None:
             return self
 
-        copy = Loop(self.coro, seconds=self.seconds, hours=self.hours, minutes=self.minutes,
-                               count=self.count, reconnect=self.reconnect, loop=self.loop)
+        copy = Loop(self.coro, seconds=self._seconds, hours=self._hours, minutes=self._minutes,
+                               count=self._count, time=self._time, reconnect=self.reconnect, loop=self.loop)
```
Rapptz, line 144: Might as well format this in a way to minify the chance of conflicts in the future.

```suggestion
        copy = Loop(
            self.coro, 
            seconds=self._seconds, 
            hours=self._hours, 
            minutes=self._minutes,
            count=self._count, 
            time=self._time, 
            reconnect=self.reconnect, 
            loop=self.loop,
        )
```

discord/ext/tasks/__init__.py
SebbyLaw: Good question! I actually caught and fixed this bug during my own testing.

When the loop is started, the first time is slept to: https://github.com/Rapptz/discord.py/pull/6456/files#diff-d612a21a8825c4e1754417b8840eed107ac96e41def4c0c92cf6394300921b51R94-R101 (lines 94-101)

In the case of an already running loop, the timedelta is added to the previous iteration time, so it will always be the "next day" relative to the previous time since `self._time` is sorted.

SebbyLaw:
Postponing target version to `2.0` due to required breaking changes to `Loop.change_interval`

SebbyLaw:
Changed target version to 2.0, and change_interval now changes the next iteration instead of the iteration after the next iteration if the tasks Loop is running

Rapptz (approved):

PR #6467, opened 2021-02-23
[docs] fix inconsistencies within the documentation

SebbyLaw:
## Summary

The successor of #5067, but only touches on docstrings, not the actual rendered documentation.

- Standardizes header indicators for docstring headers (`Attributes`, `Parameters`, `Raises`, `Returns`, `Yields`) to be `len(header) + 1` (as it is in the majority of places)
- ~~Removes unnecessary :exc:`.ExceptionName` from docstrings `Raises`.~~

- [x] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -370,7 +370,7 @@ class GuildChannel:
             whose overwrite to get.
 
         Returns
-        ---------
+        --------
         :class:`~discord.PermissionOverwrite`
             The permission overwrites for this object.
         """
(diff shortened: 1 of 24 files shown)


Rapptz:
> Standardizes header indicators for docstring headers (Attributes, Parameters, Raises, Returns, Yields) to be len(header) + 1 (as it is in the majority of places)

This isn't really that crucial.

> Removes unnecessary :exc:.ExceptionName from docstrings Raises.

These aren't unnecessary (well, maybe Sphinx has changed since then). The way these work is that it by prefixing with the period the lookup for it succeeds regardless of where it is. So in `commands.Bot` and `discord.Client` it'll render properly since those two are in different modules. If the period wasn't there then the `commands.Bot` one would look for e.g. `commands.Exception` instead of `discord.Exception`.

PR #6471, opened 2021-02-24
Code optimisations

NCPlayz:
## Summary

Mostly refactored code (via [Sourcery](https://sourcery.ai)). Also removes unused variables.
May need to be tested for accidental regressions, although seems highly unlikely.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -754,7 +754,7 @@ class Client:
 
     @allowed_mentions.setter
     def allowed_mentions(self, value):
-        if value is None and isinstance(value, AllowedMentions):
+        if value is None or isinstance(value, AllowedMentions):
             self._connection.allowed_mentions = value
         else:
             raise TypeError('allowed_mentions must be AllowedMentions not {0.__class__!r}'.format(value))
--- a/discord/client.py
+++ b/discord/client.py
@@ -754,9 +754,7 @@ class Client:
 
     @allowed_mentions.setter
     def allowed_mentions(self, value):
-        if value is None:
-            self._connection.allowed_mentions = value
-        elif isinstance(value, AllowedMentions):
+        if value is None and isinstance(value, AllowedMentions):
             self._connection.allowed_mentions = value
         else:
             raise TypeError('allowed_mentions must be AllowedMentions not {0.__class__!r}'.format(value))
@@ -1227,15 +1225,13 @@ class Client:
         if icon is not None:
             icon = utils._bytes_to_base64_data(icon)
 
-        if region is None:
-            region = VoiceRegion.us_west.value
-        else:
-            region = region.value
+        region = region or VoiceRegion.us_west
+        region_value = region.value
 
         if code:
-            data = await self.http.create_from_template(code, name, region, icon)
+            data = await self.http.create_from_template(code, name, region_value, icon)
         else:
-            data = await self.http.create_guild(name, region, icon)
+            data = await self.http.create_guild(name, region_value, icon)
         return Guild(data=data, state=self._connection)
 
     # Invite management
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -715,8 +715,7 @@ class Command(_BaseCommand):
                     except RuntimeError:
                         break
 
-        if not self.ignore_extra:
-            if not view.eof:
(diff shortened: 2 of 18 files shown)


discord/client.py
```diff
@@ -754,9 +754,7 @@ def allowed_mentions(self):
 
     @allowed_mentions.setter
     def allowed_mentions(self, value):
-        if value is None:
-            self._connection.allowed_mentions = value
-        elif isinstance(value, AllowedMentions):
+        if value is None and isinstance(value, AllowedMentions):
```
alkatar21, line 757: This condition is now never true.
```suggestion
        if value is None or isinstance(value, AllowedMentions):
```

discord/ext/commands/core.py
```diff
@@ -715,8 +715,7 @@ async def _parse_arguments(self, ctx):
                     except RuntimeError:
                         break
 
-        if not self.ignore_extra:
-            if not view.eof:
+        if not self.ignore_extra and not view.eof:
                 raise TooManyArguments('Too many arguments passed to ' + self.qualified_name)
```
SebbyLaw, line 719: Indentation error for some reason.
```suggestion
            raise TooManyArguments('Too many arguments passed to ' + self.qualified_name)
```

discord/opus.py
```diff
@@ -276,17 +276,14 @@ class _OpusStruct:
...
+            raise OpusNotLoaded()
 
         return _lib.opus_get_version_string().decode('utf-8')
 
 class Encoder(_OpusStruct):
     def __init__(self, application=APPLICATION_AUDIO):
-        if not is_loaded():
-            if not _load_default():
-                raise OpusNotLoaded()
+        _OpusStruct.get_opus_version()
```
SebbyLaw, line 286: Just use `and`

discord/opus.py
```diff
@@ -342,9 +339,7 @@ def encode(self, pcm, frame_size):
 
 class Decoder(_OpusStruct):
     def __init__(self):
-        if not is_loaded():
-            if not _load_default():
-                raise OpusNotLoaded()
+        _OpusStruct.get_opus_version()
```
SebbyLaw, line 342: Just use `and`

discord/state.py
```diff
@@ -928,8 +928,7 @@ def parse_guild_ban_add(self, data):
 
     def parse_guild_ban_remove(self, data):
         guild = self._get_guild(int(data['guild_id']))
-        if guild is not None:
-            if 'user' in data:
+        if guild is not None and 'user' in data:
                 user = self.store_user(data['user'])
                 self.dispatch('member_unban', guild, user)
```
SebbyLaw, line 933: Indentation error here for some reason
```suggestion
            user = self.store_user(data['user'])
            self.dispatch('member_unban', guild, user)
```

discord/template.py
```diff
@@ -168,12 +168,10 @@ async def create_guild(self, name, region=None, icon=None):
...
 
-        if region is None:
-            region = VoiceRegion.us_west.value
-        else:
-            region = region.value
+        region = region or VoiceRegion.us_west
+        region_value = region.value
 
-        data = await self._state.http.create_from_template(self.code, name, region, icon)
+        data = await self._state.http.create_from_template(self.code, name, region_value, icon)
```
SebbyLaw, line 174: Same logic error as above.

SebbyLaw (changes requested):

discord/opus.py
Gobot1234: It's not necessary as get_opus_version checks this

SebbyLaw: You could argue that this is now slower (and not optimized, going against the entire point of the PR) since get_opus_version calls another two methods unnecessarily.

discord/member.py
```diff
@@ -398,7 +398,7 @@ def display_name(self):
         if they have a guild specific nickname then that
         is returned instead.
         """
-        return self.nick if self.nick is not None else self.name
+        return self.nick or self.name
```
Rapptz, line 401: Technically these aren't equivalent, if e.g. `self.nick` is an empty string. As far as I know though this isn't possible at the API layer so it should be fine.

Rapptz (approved):
Looks fine.

PR #6481, opened 2021-02-25
A (small) fix for 4098.

### LINKED ISSUE #4098: on_guild_channel_update edge case
### Summary

`on_guild_channel_update` does not work as expected if you move a channel from one category to another using code. Moving the channel by hand in the client works properly.

### Reproduction Steps

Move a channel from one category to another, using code, with an `on_guild_channel_update` listener checking the `before.category.id` and `after.category.id`. Compare them, they're the same.
`await ctx.channel.edit(category=other_category)`

### Expected Results

`on_guild_channel_update` to have the proper `before.category`

### Actual Results

`before.category.id` is the same as `after.category.id`

---

`GuildChannel._edit` updates the channel with the new category as you run it, and when `ConnectionState.parse_channel_update` is called, the channel already has the new category, so `before.category` is the new category. Given the other possible parameters that may affect it, I don't see a solution to this that isn't hacky.

NCPlayz:
## Summary

This essentially resolves #4098. It maintains that this issue may only happen when the category is being edited, or in terms of the API when `parent_id` is set. When this condition is met, it will wait for the channel update matching the edited channel's id to confirm that the properties have been updated and are consistent, before returning back to the user code. As discussed in the issue, there is the potential that there may be a failure in eventual consistency where the event is never received by the client. From my testing, it seems that the event is received fairly instantaneously, but for the benefit of the doubt, a generous 2-second timeout is provided before it will fallback to the regular method of updating itself. 

It may be beneficial to extensively test this with larger bots to make sure that it works as consistently as possible.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -308,13 +308,16 @@ class GuildChannel:
 
         if options:
             data = await self._state.http.edit_channel(self.id, reason=reason, **options)
-            # this will look ugly but it works i'm sure of it
-            client = self._state.dispatch.__self__
-            try:
-                await client.wait_for('guild_channel_update', check=lambda b, a: b.id == a.id, timeout=2)
-            except asyncio.TimeoutError:
-                # fallback, unfortunately we didn't receive the event in 2s
-                self._update(data)
+            # see issue #4098
+            if 'parent_id' in options:
+                client = self._state._get_client()
+                try:
+                    await client.wait_for('guild_channel_update', check=lambda b, a: b.id == a.id, timeout=2)
+                    return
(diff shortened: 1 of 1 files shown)


Rapptz:
This issue would happen with any model update that does an in place update. I've mentioned in the past that waiting for a corresponding websocket event is not the way to go due to added perceived slowness and race conditions, on top of not working well for the HTTP only use cases. Special casing this behaviour for channel update seems a bit silly, doubly so due to the event spam that it does.

PR #6484, opened 2021-02-26
Add support for fail_if_not_exists in MessageReference

SnowyLuma:
## Summary

Allows replying to a message which may have been deleted without needing to re-try if this is the case.
Also adds a shortcut to `Messageable.send` to set this field, similarly to `mention_author` for allowed mentions.

## Checklist

- [x] This PR fixes an issue.

# Other

I'm unsure whether all of the docstrings are correct, please let me know if they aren't.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -279,6 +279,12 @@ class MessageReference:
         The channel id of the message referenced.
     guild_id: Optional[:class:`int`]
         The guild id of the message referenced.
+    fail_if_not_exists: Optional[:class:`bool`]
+        Whether replying to the referenced message should raise :class:`HTTPException`
+        if the message has been deleted or reply without the message reference attached.
+
+        .. versionadded:: 1.7
+
     resolved: Optional[Union[:class:`Message`, :class:`DeletedReferencedMessage`]]
         The message that this reference resolved to. If this is ``None``
         then the original message was not fetched either due to the Discord API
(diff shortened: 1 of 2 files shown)


CyberFlameGO (approved):

Rapptz (commented):
Sorry this took so long.

I was thinking about the naming for this. I think maybe `fail_if_reply_deleted` might be more explicit as a parameter name for `send`? Maybe we can keep `fail_if_not_exists` in other contexts but in the `send` context it seems a little bizarre. Though, right now message references only refer to replying maybe in the future there'll be other types of message references that will be possible.

Nothing I'm saying here is really final, just trying to get some bikeshedding going cause something about the current name just strikes me as being bad.

SnowyLuma:
> I think maybe `fail_if_reply_deleted` might be more explicit as a parameter name for send?

I agree that the name is not ideal, I could however not think of anything better when I originally opened the PR.
Another option might be `fail_if_original_deleted` as the reply is the message we're currently sending, not responding to?

Rapptz:
I gave this a lot of thought and I think the name is fine if it's confined only to `MessageReference`. I think by adding it to `send` as a shortcut helper it kind of blurs the line when you read it and I don't think the operation is common enough to warrant it. So maybe just remove the shortcut and it's fine?

discord/message.py
```diff
@@ -291,14 +297,15 @@ class MessageReference:
         .. versionadded:: 1.6
     """
 
-    __slots__ = ('message_id', 'channel_id', 'guild_id', 'resolved', '_state')
+    __slots__ = ('message_id', 'channel_id', 'guild_id', 'fail_if_not_exists', 'resolved', '_state')
 
-    def __init__(self, *, message_id, channel_id, guild_id=None):
+    def __init__(self, *, message_id, channel_id, guild_id=None, fail_if_not_exists=None):
```
Rapptz, line 302:
```suggestion
    def __init__(self, *, message_id, channel_id, guild_id=None, fail_if_not_exists=True):
```

A regular `bool` might be better than an `Optional[bool]`? I'm not sure there's a reason for the tri-bool currently.

Rapptz (approved):

PR #6497, opened 2021-03-01
Add Permissions attributes

TheOneMusic:
## Summary

This pull request adds these new things:
- `Permissions.default()` which corresponds to the default Discord permissions for roles
- Possibility to add and subtract [`Permissions`](https://github.com/Rapptz/discord.py/blob/master/discord/permissions.py#L47) directly. For example, we can now use `Permissions.text() + Permissions.voice()` to make a [`Permissions`](https://github.com/Rapptz/discord.py/blob/master/discord/permissions.py#L47) with all text and voice permissions, or `Permissions.all() - Permissions.general()` to make a [`Permissions`](https://github.com/Rapptz/discord.py/blob/master/discord/permissions.py#L47) with every permissions, except the **General** permissions. This is helpful to avoid dealing with [`Permissions.update()`](https://github.com/Rapptz/discord.py/blob/master/discord/permissions.py#L217) when adding/subtracting [`Permissions`](https://github.com/Rapptz/discord.py/blob/master/discord/permissions.py#L47) instances directly.
- Boolean method which returns `True` if and only if there are permissions (permission value different from 0), and `is_empty` method that returns the opposite - similar to [`PermissionOverwrite.is_empty()`](https://github.com/Rapptz/discord.py/blob/master/discord/permissions.py#L561)

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/permissions.py
+++ b/discord/permissions.py
@@ -131,6 +131,26 @@ class Permissions(BaseFlags):
     __lt__ = is_strict_subset
     __gt__ = is_strict_superset
 
+    def __add__(self, other):
+        if not isinstance(other, Permissions):
+            raise TypeError("cannot compare {} with {}".format(self.__class__.__name__, other.__class__.__name__))
+        return Permissions(self.value | other.value)
+
+    def __sub__(self, other):
+        if not isinstance(other, Permissions):
+            raise TypeError("cannot compare {} with {}".format(self.__class__.__name__, other.__class__.__name__))
+        return Permissions(self.value & (~other.value))
+
+    def __bool__(self):
+        return bool(self.value)
+
+    def is_empty(self):
+        """Returns ``True`` if there are no permissions, ``False`` otherwise.
+        
+        .. versionadded:: 1.7
+        """
+        return self.value == 0
+
     @classmethod
     def none(cls):
         """A factory method that creates a :class:`Permissions` with all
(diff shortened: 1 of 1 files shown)


discord/permissions.py
```diff
@@ -131,11 +131,40 @@ def is_strict_superset(self, other):
...
+        if not isinstance(other, Permissions):
+            raise TypeError("cannot compare {} with {}".format(self.__class__.__name__, other.__class__.__name__))
+        return Permissions(self.value & (~other.value))
+
+    def __bool__(self):
+        return bool(self.value)
+
+    def is_empty(self):
+        """Returns ``True`` if there are no permissions, ``False`` otherwise.
+
```
NCPlayz, line 149:
```suggestion

```

discord/permissions.py
```diff
@@ -131,11 +131,40 @@ def is_strict_superset(self, other):
...
     def none(cls):
         """A factory method that creates a :class:`Permissions` with all
         permissions set to ``False``."""
         return cls(0)
+    
+    @classmethod
+    def default(cls):
+        """A factory method that creates a :class:`Permissions` with only
+        the default permissions set to ``True``.
+
```
NCPlayz, line 164:
```suggestion

```

NCPlayz (changes requested):
Trailing whitespace.

Rapptz:
I'm not really a fan of these.

1. `default()` is more work to maintain with the UI changes.
2. I don't really think adding or removing permissions via subtraction and addition is a good interface. 
3. `is_empty` doesn't conceptually make sense on `Permissions`, whereas an overwrite can be empty and disengaged.
4. I'm neutral on `__bool__`, I think it could be handy but I'm not really sure.

PR #6505, opened 2021-03-03
Add Command Error Handling Example

ImVaskel:
## Summary

Adds a command error handler example.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/error_handling.py
+++ b/examples/error_handling.py
@@ -1,6 +1,5 @@
 import discord
 from discord.ext import commands
-from discord.ext.commands import BucketType
 import traceback
 
 description = '''
@@ -26,12 +25,12 @@ async def add(ctx, arg1: int, arg2: int):
 @commands.is_owner()
 @bot.command()
 async def owner_check(ctx):
-    await ctx.send("{} you do own this bot!".format(ctx.author.name))
+    await ctx.send("{}, you do own this bot!".format(ctx.author))
 
 # Here is a command with a cooldown.
 # The syntax for cooldown is rate / per / type.
 # Here we are doing 1 per 5 seconds per member.
-@commands.cooldown(1, 5, BucketType.member)
+@commands.cooldown(1, 5, commands.BucketType.member)
 @bot.command()
 async def cooldown_example(ctx):
     await ctx.send("{}, you are not on cooldown!".format(ctx.author))
@@ -46,9 +45,8 @@ async def on_command_error(ctx, error):
         return
 
     # Check for ctx.cog and return if it has a local handler.
-    elif ctx.cog:
-        if ctx.cog.has_error_handler():
-            return
+    elif ctx.cog and ctx.cog.has_error_handler():
+        return
 
     # The errors below are in a format that is readable, so we can just call `str` on them.
     readable_errors = (
--- a/examples/error_handling.py
+++ b/examples/error_handling.py
@@ -6,11 +6,8 @@ description = '''
 A basic bot featuring error handling for command errors
 '''
 
-# Here we are using regular intents, since they don't really matter for this example.
-intents = discord.Intents.default()
-
-# Constructing the bot with a prefix, our above intents, and the description
-bot = commands.Bot(command_prefix='!', intents=intents, description=description)
+# Constructing the bot with a prefix and the description
+bot = commands.Bot(command_prefix='!', description=description)
 
 # Example command to showcase a potential error raised.
 # The important ones to see are `MissingRequiredArgument` and `BadArgument`.
@@ -41,7 +38,7 @@ async def cooldown_example(ctx):
 @bot.event
 async def on_command_error(ctx, error):
     # Check if the command has a local handler.
-    if ctx.command.has_error_handler():
+    if ctx.command and ctx.command.has_error_handler():
         return
 
     # Check for ctx.cog and return if it has a local handler.
--- a/examples/error_handling.py
+++ b/examples/error_handling.py
@@ -6,11 +6,11 @@ description = '''
 A basic bot featuring error handling for command errors
 '''
 
-# Here we are using regular intents, since they don't really matter for this example
+# Here we are using regular intents, since they don't really matter for this example.
 intents = discord.Intents.default()
 
 # Constructing the bot with a prefix, our above intents, and the description
-bot = commands.Bot(command_prefix="!", intents=intents, description=description)
+bot = commands.Bot(command_prefix='!', intents=intents, description=description)
 
 # Example command to showcase a potential error raised.
 # The important ones to see are `MissingRequiredArgument` and `BadArgument`.
@@ -19,13 +19,13 @@ bot = commands.Bot(command_prefix="!", intents=intents, description=description)
 @bot.command()
 async def add(ctx, arg1: int, arg2: int):
     added = arg1 + arg2
-    await ctx.send("{} + {} = {}".format(arg1, arg2, added))
+    await ctx.send('{} + {} = {}'.format(arg1, arg2, added))
 
 # This command showcases the `is_owner` check, which will raise `commands.NotOwner` if the user doesn't own the bot.
 @commands.is_owner()
 @bot.command()
 async def owner_check(ctx):
-    await ctx.send("{}, you do own this bot!".format(ctx.author))
+    await ctx.send('{}, you do own this bot!'.format(ctx.author))
 
 # Here is a command with a cooldown.
 # The syntax for cooldown is rate / per / type.
@@ -33,7 +33,7 @@ async def owner_check(ctx):
 @commands.cooldown(1, 5, commands.BucketType.member)
 @bot.command()
 async def cooldown_example(ctx):
-    await ctx.send("{}, you are not on cooldown!".format(ctx.author))
+    await ctx.send('{}, you are not on cooldown!'.format(ctx.author))
 
 # Here we are **overriding** on_command_error.
 # You can also use `bot.listen()` if you prefer not to override it, though there isn't a difference here.
@@ -72,7 +72,7 @@ async def on_command_error(ctx, error):
     # Here we send a custom error message if the user is on cooldown.
     # We also format the cooldown seconds to 2 digits.
     elif isinstance(error, commands.CommandOnCooldown):
-        await ctx.send("You are on cooldown! Please wait `{}` seconds".format(round(error.retry_after, 2)))
+        await ctx.send('You are on cooldown! Please wait `{}` seconds'.format(round(error.retry_after, 2)))
 
     # If the error isn't picked up by any of our other checks, then we should just print it.
     # **NOTE**: This error is *not* being raised. It is being printed.
@@ -80,4 +80,4 @@ async def on_command_error(ctx, error):
         traceback.print_exc()
 
 # Run the bot. You should really read the token from a configuration file.
-bot.run("TOKEN HERE")
+bot.run('TOKEN HERE')
--- a/examples/error_handling.py
+++ b/examples/error_handling.py
@@ -1,0 +1,85 @@
+import discord
+from discord.ext import commands
+from discord.ext.commands import BucketType
+import traceback
+
+description = '''
+A basic bot featuring error handling for command errors
+'''
+
+# Here we are using regular intents, since they don't really matter for this example
+intents = discord.Intents.default()
+
+# Constructing the bot with a prefix, our above intents, and the description
+bot = commands.Bot(command_prefix="!", intents=intents, description=description)
+
+# Example command to showcase a potential error raised.
+# The important ones to see are `MissingRequiredArgument` and `BadArgument`.
+# `MissingRequiredArgument` will be raised when a required argument is not passed.
+# Bad Argument will be raised in this case when the arg cannot be converted to an `int`.
+@bot.command()
+async def add(ctx, arg1: int, arg2: int):
+    added = arg1 + arg2
+    await ctx.send("{} + {} = {}".format(arg1, arg2, added))
+
+# This command showcases the `is_owner` check, which will raise `commands.NotOwner` if the user doesn't own the bot.
+@commands.is_owner()
+@bot.command()
+async def owner_check(ctx):
+    await ctx.send("{} you do own this bot!".format(ctx.author.name))
+
+# Here is a command with a cooldown.
+# The syntax for cooldown is rate / per / type.
+# Here we are doing 1 per 5 seconds per member.
+@commands.cooldown(1, 5, BucketType.member)
+@bot.command()
+async def cooldown_example(ctx):
+    await ctx.send("{}, you are not on cooldown!".format(ctx.author))
+
+# Here we are **overriding** on_command_error.
+# You can also use `bot.listen()` if you prefer not to override it, though there isn't a difference here.
+# The event takes 2 args, `ctx` and `error`.
+@bot.event
+async def on_command_error(ctx, error):
+    # Check if the command has a local handler.
+    if ctx.command.has_error_handler():
+        return
+
+    # Check for ctx.cog and return if it has a local handler.
+    elif ctx.cog:
+        if ctx.cog.has_error_handler():
+            return
+
+    # The errors below are in a format that is readable, so we can just call `str` on them.
+    readable_errors = (
+        commands.BadArgument,
+        commands.MissingRequiredArgument,
+        commands.NotOwner
+    )
+
+    # The error may sometimes be wrapped in `CommandInvokeError`, which would cause our `isinstance` to fail,
+    # to fix this, we unwrap the error.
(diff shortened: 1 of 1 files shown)


examples/error_handling.py
```diff
@@ -0,0 +1,83 @@
+import discord
+from discord.ext import commands
+import traceback
+
+description = '''
+A basic bot featuring error handling for command errors
+'''
+
+# Here we are using regular intents, since they don't really matter for this example
```
mesub7, line 9:
```suggestion
# Here we are using regular intents, since they don't really matter for this example.
```

examples/error_handling.py
```diff
@@ -0,0 +1,83 @@
...
+intents = discord.Intents.default()
+
+# Constructing the bot with a prefix, our above intents, and the description
+bot = commands.Bot(command_prefix="!", intents=intents, description=description)
```
mesub7, line 13:
```suggestion
bot = commands.Bot(command_prefix='!', intents=intents, description=description)
```

examples/error_handling.py
```diff
@@ -0,0 +1,83 @@
...
+
+# Example command to showcase a potential error raised.
+# The important ones to see are `MissingRequiredArgument` and `BadArgument`.
+# `MissingRequiredArgument` will be raised when a required argument is not passed.
+# Bad Argument will be raised in this case when the arg cannot be converted to an `int`.
+@bot.command()
+async def add(ctx, arg1: int, arg2: int):
+    added = arg1 + arg2
+    await ctx.send("{} + {} = {}".format(arg1, arg2, added))
```
mesub7, line 22:
```suggestion
    await ctx.send('{} + {} = {}'.format(arg1, arg2, added))
```

examples/error_handling.py
```diff
@@ -0,0 +1,83 @@
...
+
+# This command showcases the `is_owner` check, which will raise `commands.NotOwner` if the user doesn't own the bot.
+@commands.is_owner()
+@bot.command()
+async def owner_check(ctx):
+    await ctx.send("{}, you do own this bot!".format(ctx.author))
```
mesub7, line 28:
```suggestion
    await ctx.send('{}, you do own this bot!'.format(ctx.author))
```

examples/error_handling.py
```diff
@@ -0,0 +1,83 @@
...
+
+# Here is a command with a cooldown.
+# The syntax for cooldown is rate / per / type.
+# Here we are doing 1 per 5 seconds per member.
+@commands.cooldown(1, 5, commands.BucketType.member)
+@bot.command()
+async def cooldown_example(ctx):
+    await ctx.send("{}, you are not on cooldown!".format(ctx.author))
```
mesub7, line 36:
```suggestion
    await ctx.send('{}, you are not on cooldown!'.format(ctx.author))
```

examples/error_handling.py
```diff
@@ -0,0 +1,83 @@
...
+        return
+    
+    # Checking for our readable errors and then calling `str` on them.
+    elif isinstance(error, readable_errors):
+        await ctx.send(str(error))
+
+    # Here we send a custom error message if the user is on cooldown.
+    # We also format the cooldown seconds to 2 digits.
+    elif isinstance(error, commands.CommandOnCooldown):
+        await ctx.send("You are on cooldown! Please wait `{}` seconds".format(round(error.retry_after, 2)))
```
mesub7, line 75:
```suggestion
        await ctx.send('You are on cooldown! Please wait `{}` seconds'.format(round(error.retry_after, 2)))
```

examples/error_handling.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # If the error isn't picked up by any of our other checks, then we should just print it.
+    # **NOTE**: This error is *not* being raised. It is being printed.
+    else:
+        traceback.print_exc()
+
+# Run the bot. You should really read the token from a configuration file.
+bot.run("TOKEN HERE")
```
mesub7, line 83:
```suggestion
bot.run('TOKEN HERE')
```

mesub7 (changes requested):
Looks good to me. Just a few nits (to keep it consistent with the other examples).

examples/error_handling.py
```diff
@@ -0,0 +1,83 @@
...
+async def cooldown_example(ctx):
+    await ctx.send('{}, you are not on cooldown!'.format(ctx.author))
+
+# Here we are **overriding** on_command_error.
+# You can also use `bot.listen()` if you prefer not to override it, though there isn't a difference here.
+# The event takes 2 args, `ctx` and `error`.
+@bot.event
+async def on_command_error(ctx, error):
+    # Check if the command has a local handler.
+    if ctx.command.has_error_handler():
```
Zomatree, line 44:
```suggestion
    if ctx.command and ctx.command.has_error_handler():
```
ctx.command can be none

examples/error_handling.py
```diff
@@ -0,0 +1,83 @@
...
+
+description = '''
+A basic bot featuring error handling for command errors
+'''
+
+# Here we are using regular intents, since they don't really matter for this example.
+intents = discord.Intents.default()
+
+# Constructing the bot with a prefix, our above intents, and the description
+bot = commands.Bot(command_prefix='!', intents=intents, description=description)
```
Zomatree, line 13:
```suggestion
# Constructing the bot with a prefix and the description
bot = commands.Bot(command_prefix='!', description=description)
```
no need to manually set intents to default

Zomatree (changes requested):

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
+import discord
+from discord.ext import commands
+import traceback
+
+description = '''
+A basic bot featuring error handling for command errors
+'''
+
+# Constructing the bot with a prefix and the description
+bot = commands.Bot(command_prefix='!', description=description)
```
Daggy1234, line 10: Maybe mention that if the intents aren't passed, they resort to default

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
...
+async def on_command_error(ctx, error):
+    # Check if the command has a local handler.
+    if ctx.command and ctx.command.has_error_handler():
+        return
+
+    # Check for ctx.cog and return if it has a local handler.
+    elif ctx.cog and ctx.cog.has_error_handler():
+        return
+
+    # The errors below are in a format that is readable, so we can just call `str` on them.
```
Daggy1234, line 48: It would be cool to have a link for the error hierarchy in the docs

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # Here we send a custom error message if the user is on cooldown.
+    # We also format the cooldown seconds to 2 digits.
+    elif isinstance(error, commands.CommandOnCooldown):
+        await ctx.send('You are on cooldown! Please wait `{}` seconds'.format(round(error.retry_after, 2)))
+
+    # If the error isn't picked up by any of our other checks, then we should just print it.
+    # **NOTE**: This error is *not* being raised. It is being printed.
+    else:
+        traceback.print_exc()
```
Daggy1234, line 77: Explain what this is?

Daggy1234 (changes requested):

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
...
+@bot.command()
+async def add(ctx, arg1: int, arg2: int):
+    added = arg1 + arg2
+    await ctx.send('{} + {} = {}'.format(arg1, arg2, added))
+
+# This command showcases the `is_owner` check, which will raise `commands.NotOwner` if the user doesn't own the bot.
+@commands.is_owner()
+@bot.command()
+async def owner_check(ctx):
+    await ctx.send('{}, you do own this bot!'.format(ctx.author))
```
AbstractUmbra, line 25:
```suggestion
    await ctx.send('{}, you own this bot!'.format(ctx.author))
```
Weird additional word?

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
...
+async def cooldown_example(ctx):
+    await ctx.send('{}, you are not on cooldown!'.format(ctx.author))
+
+# Here we are **overriding** on_command_error.
+# You can also use `bot.listen()` if you prefer not to override it, though there isn't a difference here.
+# The event takes 2 args, `ctx` and `error`.
+@bot.event
+async def on_command_error(ctx, error):
+    # Check if the command has a local handler.
+    if ctx.command and ctx.command.has_error_handler():
```
AbstractUmbra, line 41:
```suggestion
    if ctx.command.has_error_handler():
```
`ctx.command` would only be None in the event of `commands.CommandNotFound` and would also not have an error handler, so this is unnecessary.

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
...
+        return
+
+    # The errors below are in a format that is readable, so we can just call `str` on them.
+    readable_errors = (
+        commands.BadArgument,
+        commands.MissingRequiredArgument,
+        commands.NotOwner
+    )
+
+    # The error may sometimes be wrapped in `CommandInvokeError`, which would cause our `isinstance` to fail,
```
AbstractUmbra, line 55:
```suggestion
    # Exceptions that do not derive from `commands.CommandError` are wrapped in a `commands.CommandInvokeError` exception. Unwrap this using the following: 
```

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
...
+    # to fix this, we unwrap the error.
+    if isinstance(error, commands.CommandInvokeError):
+        error = error.original
+
+    # We use isinstance to check the type of error. 
+    if isinstance(error, commands.CommandNotFound):
+        # We don't really care if the command is not found, so we can just return.
```
AbstractUmbra, line 62:
```suggestion
        # We can suppress the `CommandNotFound` exception to suppress extra noise.
```
Wording is too informal.

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
...
+        return
+    
+    # Checking for our readable errors and then calling `str` on them.
```
AbstractUmbra, line 65:
```suggestion
    # Checking for our readable error types and then sending their string format to the invoking channel.
```

AbstractUmbra (changes requested):
Couple of wording nits here and there.

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
...
+bot = commands.Bot(command_prefix='!', description=description)
+
+# Example command to showcase a potential error raised.
+# The important ones to see are `MissingRequiredArgument` and `BadArgument`.
+# `MissingRequiredArgument` will be raised when a required argument is not passed.
+# Bad Argument will be raised in this case when the arg cannot be converted to an `int`.
+@bot.command()
+async def add(ctx, arg1: int, arg2: int):
+    added = arg1 + arg2
+    await ctx.send('{} + {} = {}'.format(arg1, arg2, added))
```
SebbyLaw, line 19:
```suggestion
async def divide(ctx, dividend: float, divisor: float):
    divided = dividend / divisor
    await ctx.send('{} / {} = {}'.format(dividend, divisor, divided))
```
This is perhaps a good way of showcasing the unwrapping of `CommandInvokeError` - to catch `ZeroDivisonError`.

examples/error_handling.py
Zomatree: This will raise an attribute error as None doesn't have a `has_error_handler` attribute, so this is necessary.

ImVaskel: This raises an error if `ctx.command` is `None`.

examples/error_handling.py
ImVaskel: I don't think clarification about intents is important in this situation, as none are used.

Daggy1234: > 
> 
> I don't think clarification about intents is important in this situation, as none are used.

disagree, examples may be used individually. All I said was clarify, that as no intents are passed, intents.default are used

pikaninja:
Would it be worth moving the event to a subclass? Many other examples such as reaction_roles.py, edits.py and deleted.py are in subclasses.

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
+import discord
+from discord.ext import commands
+import traceback
+
+description = '''
+A basic bot featuring error handling for command errors
+'''
+
+# Constructing the bot with a prefix and the description
```
Rapptz, line 9:
```suggestion
```

This comment is pointless.

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
...
+# Here is a command with a cooldown.
+# The syntax for cooldown is rate / per / type.
+# Here we are doing 1 per 5 seconds per member.
+@commands.cooldown(1, 5, commands.BucketType.member)
+@bot.command()
+async def cooldown_example(ctx):
+    await ctx.send('{}, you are not on cooldown!'.format(ctx.author))
+
+# Here we are **overriding** on_command_error.
+# You can also use `bot.listen()` if you prefer not to override it, though there isn't a difference here.
```
Rapptz, line 36:
```suggestion
# You can also use `bot.listen()` if you prefer not to override it
```

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
...
+# The event takes 2 args, `ctx` and `error`.
+@bot.event
+async def on_command_error(ctx, error):
+    # Check if the command has a local handler.
```
Rapptz, line 40:
```suggestion
    # If the command has a local error handler then don't handle it.
```

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
...
+    if ctx.command and ctx.command.has_error_handler():
+        return
+
+    # Check for ctx.cog and return if it has a local handler.
```
Rapptz, line 44:
```suggestion
    # If the command's cog has an error handler then don't handle the error
```

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
...
+    elif ctx.cog and ctx.cog.has_error_handler():
+        return
+
+    # The errors below are in a format that is readable, so we can just call `str` on them.
```
Rapptz, line 48: All errors are "readable" when casted into `str`. Having this code in the library's example gives off the impression that this isn't true, which I don't like doing.

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
...
+    readable_errors = (
+        commands.BadArgument,
+        commands.MissingRequiredArgument,
+        commands.NotOwner
+    )
+
+    # Exceptions that do not derive from `commands.CommandError` are wrapped in a `commands.CommandInvokeError` exception. Unwrap this using the following: 
+    # to fix this, we unwrap the error.
+    if isinstance(error, commands.CommandInvokeError):
```
Rapptz, line 57:
```suggestion
    # Exceptions that do not derive from `commands.CommandError` are wrapped in a
    # `commands.CommandInvokeError` exception. To get the original error back we
    # can use the `original` attribute of the error.
    if isinstance(error, commands.CommandInvokeError):
```

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
...
+        error = error.original
+
+    # We use isinstance to check the type of error. 
+    if isinstance(error, commands.CommandNotFound):
+        # We can suppress the `commands.CommandNotFound` exception to suppress extra noise.
+        return
+    
+    # Checking for our readable error types and then sending their string format to the invoking channel.
+    elif isinstance(error, readable_errors):
+        await ctx.send(str(error))
```
Rapptz, line 67: I've already mentioned my opinion on this, so I'm mentioning it applies to this code snippet as well.

examples/error_handling.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # Here we send a custom error message if the user is on cooldown.
+    # We also format the cooldown seconds to 2 digits.
+    elif isinstance(error, commands.CommandOnCooldown):
+        await ctx.send('You are on cooldown! Please wait `{}` seconds'.format(round(error.retry_after, 2)))
```
Rapptz, line 72:
```suggestion
        await ctx.send('You are on cooldown! Please wait `{0:.2f}` seconds'.format(error.retry_after))
```

Rapptz (changes requested):
I'm not really sure how I feel about the implementation of this example.

On one hand, it does show case the global error handler and how to override it but it fails to actually locally handle any errors even though the global error handler makes note of it.

As this example is designed, it pushes people to use the global error handler to handle all sorts of errors, even those that are local only to a few examples. Likewise, it makes opinions on what's worth sending and what isn't based on criteria that I don't personally agree with -- all errors are designed to be readable when casted to string.

I have a minor nit with the styling of the example as well since there are excessive newlines with the if-elif-else chain it makes it easy to miss whether they're multiple if statements or elif statements but this is a bit nitpicky.

PR #6509, opened 2021-03-05
Rework the reaction roles example

NCPlayz:
## Summary

Minor changes for the example to work as intended. Also allows Unicode emojis and custom emojis, although maybe the union-type for the `emoji_to_role` key might not be favourable.

Reviews are welcome.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/examples/reaction_roles.py
+++ b/examples/reaction_roles.py
@@ -8,25 +8,28 @@ class RoleReactClient(discord.Client):
 
         self.role_message_id = 0  # ID of message that can be reacted to to add role
         self.emoji_to_role = {
-            'partial_emoji_1': 0,  # ID of role associated with partial emoji object 'partial_emoji_1'
-            'partial_emoji_2': 0  # ID of role associated with partial emoji object 'partial_emoji_2'
+            '🔴': 0,  # ID of the role associated with unicode emoji '🔴'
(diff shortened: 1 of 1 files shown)


examples/reaction_roles.py
```diff
@@ -8,27 +8,30 @@ def __init__(self, *args, **kwargs):
 
         self.role_message_id = 0  # ID of message that can be reacted to to add role
         self.emoji_to_role = {
-            partial_emoji_1: 0,  # ID of role associated with partial emoji object 'partial_emoji_1'
-            partial_emoji_2: 0  # ID of role associated with partial emoji object 'partial_emoji_2'
+            '🔴': 0,  # ID of the role associated with unicode emoji '🔴'
```
mikeshardmind, line 11: I'm curious if there should be a mention of variation selectors and normalizing for them, most users probably don't care that there's a difference between A down arrow, and the same arrow with a variation selector applied to it. Examples here don't have this issue, but it might still be prudent to include

Rapptz (commented):
Feels like a better implementation would have just been to use `PartialEmoji` instances since that was within the spirit of the original example.

Rapptz (commented):
Some of the renames and changes here are superfluous and unnecessary but the underlying example is fine I guess.

PR #6522, opened 2021-03-11
Make Embed.copy() a deep copy

diceroll123:
## Summary

I'm requesting that Embed.copy become a less shallow copy. The only difference here is that the fields will be copied by value instead of the whole list reference, because that's the only thing that persists between copies.

small repro:

- make embed
- loop
	- clear fields
	- add some fields
	- yield embed.copy()

all yielded embeds would have the final embed's fields.

---

I removed `self.__slots__` and replaced it with a hard-coded `'_field'`-less tuple of __slots__. There are multiple ways to do what I'm trying to solve here and this may not have been the cleanest, so I'm all ears for suggestions.

*marked as breaking change as it does actually change how the end results work*

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -557,9 +557,16 @@ class Embed:
         # add in the raw data into the dict
         result = {
             key[1:]: getattr(self, key)
-            for key in self.__slots__
+            for key in ('title', 'url', 'type', '_timestamp', '_colour', '_footer',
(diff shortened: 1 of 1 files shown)


Rapptz:
`Embed.copy` is shallow by design (I have no plans to change it). If you want a deepcopy the stdlib provides `copy.deepcopy`. This is in line with every `copy()` method in the stdlib.

PR #6528, opened 2021-03-13
Ignore has_permissions and has_permissions in Dms

Luc1412:
## Summary

This PR fixes an error for commands used in DMs with has_permissions or has_permissions check.
This allows creating commands which can be used in DMs and on a server. If the command is used on a server, permissions get checked, otherwise, it gets ignored.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1779,6 +1779,8 @@ def has_permissions(**perms):
         raise TypeError('Invalid permission(s): %s' % (', '.join(invalid)))
 
     def predicate(ctx):
+        if ctx.guild is None:
+            return True
         ch = ctx.channel
         permissions = ch.permissions_for(ctx.author)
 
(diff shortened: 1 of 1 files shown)


PikalaxALT:
This could result in undesirable behavior.

Luc1412:
I think people who like to ignore DMs could use the guild_only() check. It currently throws for me a missing permissions error even though someone uses the command in DMs and my guild_only check won't get reached.

Phxntxm:
> I think people who like to ignore DMs could use the guild_only() check. It currently throws for me a missing permissions error even though someone uses the command in DMs and my guild_only check won't get reached.

Sure but that's *your* use case, not everyone's use case is different. The idea of "I want to check for permissions, if they don't have permissions then don't run this command" makes complete sense since you don't have those permissions in a DM. Since permissions are often used to lock down what you can do in a guild... using these commands in DMs doesn't make much sense. This may fit to your specific use case, but not everyones' is the same and this isn't really a change I think that should be in the library, I think this should be done in your own code if you want this behavior.

Rapptz:
These checks were designed to work within a DM context as well, so removing that is a breaking change and not desirable.

PR #6530, opened 2021-03-14
Include group args in command signature for the help page

### LINKED ISSUE #6529: Subcommand signature in default help formatter doesn't include group's args
### Summary

When a parent group has arguments, they aren't shown in subcommand's help

### Reproduction Steps

1. Use default help formatter (so just don't specify one at all)
2. Put
```py
class X(commands.Cog):
  @commands.group(invoke_without_command=True)
  async def cogname(self, ctx):
    pass

  @cogname.group()
  async def edit(self, ctx, channel: discord.TextChannel):
    pass

  @edit.group(name='name')
  async def edit_name(self, ctx):
    pass

  @edit_name.group(name='format')
  async def edit_name_format(self, ctx, option: str):
    pass
```
3. Start the bot
4. Try running `[p]help cogname edit name format` (`[p]` being bot's prefix)

### Expected Results

I expected the output to be:
```
[p]cogname edit <channel> name format <option>
```

### Actual Results

The output was:
```
[p]cogname edit name format <option>
```

### Intents

discord.Intents.all()

### System Information

```
- Python v3.8.8-final
- discord.py v1.6.0-final
- aiohttp v3.7.3
```

### Checklist

Note: When `invoke_without_command` is set to `True`, the signature should continue to not include the arguments of the parent as then, those arguments are not required to run subcommands.

jack1142:
## Summary

Fixes #6529 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/help.py
+++ b/discord/ext/commands/help.py
@@ -437,15 +437,24 @@ class HelpCommand:
             The signature for the command.
         """
 
-        parent = command.full_parent_name
+        parent = command.parent
+        entries = []
+        while parent is not None:
+            if not parent.signature or parent.invoke_without_command:
+                entries.append(parent.name)
+            else:
+                entries.append(parent.name + ' ' + parent.signature)
+            parent = parent.parent
+        parent_sig = ' '.join(reversed(entries))
+
         if len(command.aliases) > 0:
             aliases = '|'.join(command.aliases)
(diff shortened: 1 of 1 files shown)


Rapptz (approved):
There might be an optimisation opportunity here with reordering the way the boolean comparisons are done but this PR is fine as-is. Thanks.

PR #6539, opened 2021-03-17
Handle priority_speaker permission differently

Ultimator14:
## Summary

When a member is the server owner or has the administrator permission, discord.py automatically sets all the `guild_permission` flags for this member. However in discord, the `priority_speaker` flag is handled independently from the administrator permission (see screenshot from discord integrations).  
This pull request prevents the `priority_speaker` flag from being automatically set if the member has the administrator permission or is the server owner.

permissions

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/member.py
+++ b/discord/member.py
@@ -481,15 +481,15 @@ class Member(discord.abc.Messageable, _BaseUser):
         administrator implication.
         """
 
-        if self.guild.owner_id == self.id:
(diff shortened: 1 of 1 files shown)


Rapptz:
This is a client bug. Testing the actual behaviour in the app shows that administrator or being owner does actually give you priority speaker.

PR #6562, opened 2021-03-22
Add an optional seed parameter in Colour.random

Maselkov:
## Summary

Adds an optional seed parameter in Color.Random
Example use case: using a user's id to generate a consistent, random embed color without the need to store it.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/colour.py
+++ b/discord/colour.py
@@ -127,12 +127,14 @@ class Colour:
             The random algorithm works by choosing a colour with a random hue but
             with maxed out saturation and value.
 
+        .. versionadded:: 1.6
+
         Parameters
         ------------
         seed: Optional[Union[:class:`int`, :class:`str`, :class:`float`, :class:`bytes`, :class:`bytearray`]]
             The seed to initialize the colour with.
 
-        .. versionadded:: 1.6
+            .. versionadded:: 1.7
         """
         rand = random
         if seed:
--- a/discord/colour.py
+++ b/discord/colour.py
@@ -119,7 +119,7 @@ class Colour:
         return cls(0)
 
     @classmethod
-    def random(cls):
+    def random(cls, seed=None):
         """A factory method that returns a :class:`Colour` with a random hue.
 
         .. note::
@@ -127,9 +127,17 @@ class Colour:
             The random algorithm works by choosing a colour with a random hue but
             with maxed out saturation and value.
 
+        Parameters
+        ------------
+        seed: Optional[Union[:class:`int`, :class:`str`, :class:`float`, :class:`bytes`, :class:`bytearray`]]
+            The seed to initialize the colour with.
+
         .. versionadded:: 1.6
         """
-        return cls.from_hsv(random.random(), 1, 1)
+        rand = random
+        if seed:
+            rand = random.Random(seed)
+        return cls.from_hsv(rand.random(), 1, 1)
 
     @classmethod
     def teal(cls):
--- a/discord/colour.py
+++ b/discord/colour.py
@@ -119,7 +119,7 @@ class Colour:
         return cls(0)
 
     @classmethod
-    def random(cls, seed=None):
+    def random(cls, *, seed=None):
         """A factory method that returns a :class:`Colour` with a random hue.
 
         .. note::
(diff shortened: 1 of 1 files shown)


discord/colour.py
```diff
@@ -119,17 +119,25 @@ def default(cls):
...
         .. note::
 
             The random algorithm works by choosing a colour with a random hue but
             with maxed out saturation and value.
 
+        Parameters
+        ------------
+        seed: Optional[Union[:class:`int`, :class:`str`, :class:`float`, :class:`bytes`, :class:`bytearray`]]
+            The seed to initialize the colour with.
+
```
SebbyLaw, line 134: Parameters should be below the 1.6 versionadded marker

Should also have a 1.7 versionadded under the seed parameter

SebbyLaw (changes requested):

discord/colour.py
Bluenix2: No? Should it not be a function-wide `versionchanged`?

ShineyDev: The [documentation](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionchanged) implicitly states that `.. versionchanged::` should be used here, but I've seen both used in this context throughout the library - though I do also recall a consistency PR changing this.

Rapptz:
I'm not sure I really get the point of this. The use case seems really weak.

Maselkov:
I mostly added it because I generate groups of related roles based on a parameter and having a single, randomized color for the whole group is convenient. The use case presented is just another one I quickly came up with. You can always generate the color by yourself, but I think this is a nice shortcut.

discord/colour.py
```diff
@@ -119,7 +119,7 @@ def default(cls):
         return cls(0)
 
     @classmethod
-    def random(cls):
+    def random(cls, seed=None):
```
Rapptz, line 122:
```suggestion
    def random(cls, *, seed=None):
```

discord/colour.py
```diff
@@ -128,8 +128,18 @@ def random(cls):
             with maxed out saturation and value.
 
         .. versionadded:: 1.6
+
+        Parameters
+        ------------
+        seed: Optional[Union[:class:`int`, :class:`str`, :class:`float`, :class:`bytes`, :class:`bytearray`]]
+            The seed to initialize the colour with.
```
Rapptz, line 135:
```suggestion
            The seed to initialize the RNG with. If ``None`` is passed the default RNG is used. 
```

discord/colour.py
```diff
@@ -128,8 +128,18 @@ def random(cls):
...
+
+            .. versionadded:: 1.7
         """
-        return cls.from_hsv(random.random(), 1, 1)
+        rand = random
+        if seed:
+            rand = random.Random(seed)
```
Rapptz, line 141:
```suggestion
        rand = random if seed is None else random.Random(seed)
```

Rapptz (changes requested):

Midorina:
I don't really get the point of this. Exact color int or a seed, you store something either way. Could you provide an example use case where this addition will be useful?

pikaninja:
They left one in the pr's description?
> Example use case: using a user's id to generate a consistent, random embed color without the need to store it.
You can't give such a large number as the color int

Midorina:
@pikaninja oh, makes sense, thanks

Vexs:
I brought this up in the discord server but it's worth actually publishing I guess- you don't need to use seeding to get a random color here, you can just use `id % 255` to get a random-enough color.

Maselkov:
> you can just use `id % 255` to get a random-enough color.
That would lead to a very limited color range :p. And to be fair, my own use case involves strings being the seed rather than ids. The use case I presented above is something I thought might be somewhat more common

Anyway, I won't argue whether this change is worth adding or not. My viewpoint might be skewed and maybe it is indeed too niche to be worth merging. Still, I think that it could be useful from time to time.

Vexs:
color.random() uses HSV- the hue is the only thing randomly selected. I thought HSV was 0-255 per, but it's 0-1. `id%255^3` would give you the whole range of rgb, but as HSV with SV maxed is better, you'd want something like:
`color.from_hsv((id%1000)/1000, 1, 1)` (this isn't quite perfect, but w/e). 
As for handling strings, you could just `sum()` the `ord()`, or bitshift them around, etc- and then take that number and mod it.

Rapptz (approved):
The code is fine now, thanks.

I'm still not really sure if I understand the use cases but I guess there might be something I'm missing. It's fine though, thanks for the PR.

PR #6563, opened 2021-03-22
[commands] allow arbitrary callables in cooldown

sgtlaggy:
## Summary

This allows for easier use of non-default bucket keys.
This is especially useful in cases of global- or cog-check cooldowns that should apply to each command individually.

Previously:

```py
class CustomMapping(CooldownMapping):
    def _bucket_key(self, ctx):
      return (ctx.command.name, ctx.author.id)

mapping = CustomMapping.from_cooldown(1, 30, BucketType.user)  # anything but .default can be used here
```
or
```py
mappings = {}
default_mapping = CooldownMapping.from_cooldown(1, 30, BucketType.user)

def bot_check(ctx):
    mapping = mappings.setdefault(ctx.command.name, default_mapping.copy())
```

With this change:
```py
mapping = CooldownMapping.from_cooldown(1, 30, lambda ctx: (ctx.command.name, ctx.author.id))
```

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1959,7 +1959,7 @@ def cooldown(rate, per, type=BucketType.default):
         The number of times a command can be used before triggering a cooldown.
     per: :class:`float`
         The amount of seconds to wait for a cooldown when it's been triggered.
-    type: Union[:class:`.BucketType`, Callable[:class:`.Message`, Any]]
+    type: Union[:class:`.BucketType`, Callable[[:class:`.Message`], Any]]
         The type of cooldown to have. If callable, should return a key for the mapping.
     """
 
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -1959,8 +1959,8 @@ def cooldown(rate, per, type=BucketType.default):
         The number of times a command can be used before triggering a cooldown.
     per: :class:`float`
         The amount of seconds to wait for a cooldown when it's been triggered.
-    type: :class:`.BucketType`
-        The type of cooldown to have.
+    type: Union[:class:`.BucketType`, Callable[:class:`.Message`, Any]]
+        The type of cooldown to have. If callable, should return a key for the mapping.
     """
 
     def decorator(func):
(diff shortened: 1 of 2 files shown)


discord/ext/commands/core.py
```diff
@@ -1959,8 +1959,8 @@ def cooldown(rate, per, type=BucketType.default):
         The number of times a command can be used before triggering a cooldown.
     per: :class:`float`
         The amount of seconds to wait for a cooldown when it's been triggered.
-    type: :class:`.BucketType`
-        The type of cooldown to have.
+    type: Union[:class:`.BucketType`, Callable[:class:`.Message`, Any]]
```
zephyrkul, line 1962:
```suggestion
    type: Union[:class:`.BucketType`, Callable[[:class:`.Context`], Any]]
```

sgtlaggy: Though `Context` does fit the interface here, a `Message` is passed to `get_bucket` by the lib and subsequently `get_key`, see 
[core.py#L769](https://github.com/Rapptz/discord.py/blob/master/discord/ext/commands/core.py#L769).
The example in the issue was for a non-standard case where you would be making the mapping yourself, so you could pass whatever value you wanted.

discord/ext/commands/core.py
```diff
@@ -1959,8 +1959,8 @@ def cooldown(rate, per, type=BucketType.default):
         The number of times a command can be used before triggering a cooldown.
     per: :class:`float`
         The amount of seconds to wait for a cooldown when it's been triggered.
-    type: :class:`.BucketType`
-        The type of cooldown to have.
+    type: Union[:class:`.BucketType`, Callable[[:class:`.Message`], Any]]
+        The type of cooldown to have. If callable, should return a key for the mapping.
```
Rapptz, line 1963:
```suggestion
        The type of cooldown to have. If callable, should return a key for the mapping.
        
        .. versionchanged:: 1.7
            Callables are now supported for custom bucket types.
```

Rapptz (changes requested):

Rapptz (approved):

PR #6565, opened 2021-03-23
Clarify docs to state that on_reaction_remove requires Intents.members

### LINKED ISSUE #6010: `on_reaction_remove()` not called, while `on_reaction_add()` is called on the same message.
## Summary

If I have both a `on_reaction_add` and a `on_reaction_remove()`. See https://stackoverflow.com/questions/64685581/python-discord-py-on-reaction-removed-not-called

## Reproduction Steps

This is a minimal example I have writen to reproduce the problem:

```python
import discord
import json

client = discord.Client()


@client.event
async def on_reaction_add(reaction, user):
    print("on_reaction_add called")


@client.event
async def on_reaction_remove(reaction, user):
    print("on_reaction_remove called")


# same problem if I remove this part
@client.event
async def on_raw_reaction_remove(payload):
     print("on_raw_reaction_remove called")


with open("secret.json") as f:
    config = json.load(f)
token = config["token"]
client.run(token)
```

Now write a new message in the server (to force it being included in the cache) and remove it again.

## Expected Results

I would expect the following in the console:

```
on_reaction_add called
on_reaction_remove called
on_raw_reaction_remove called
```

## Actual Results

However, `on_reaction_remove` is never executed.

The actual output is:

```
on_reaction_add called
on_raw_reaction_remove called
```

Note that I get the same problem if I don't include the `on_raw_reaction_remove()` callback, in that case only `on_reaction_add` is printed.

## Intents

I am using default intents. This includes reactions, and the bot is able to add reactions, and receive `on_reaction_add()` properly. I have tried many possible intents combinations but none seem to make a difference.

## Checklist

- [x ] I have searched the open issues for duplicates.
- [x ] I have shown the entire traceback, if possible.
- [x ] I have removed my token from display, if visible.
- [x ] I have provided the intents that my bot is using.

## System Information

I am using version 1.5.1. I verified and got the same issue in version 1.5.0 and 1.4.2.

```
- Python v3.7.9-final
- discord.py v1.5.1-final
- aiohttp v3.6.3
- system info: Windows 10 10.0.17763
```

TK421bsod:
## Summary
Intents.members are required for on_reaction_remove to fire, but this isn't stated in the docs currently.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -512,6 +512,11 @@ to handle it, which defaults to print a traceback and ignoring the exception.
         To get the message being reacted, access it via :attr:`Reaction.message`.
 
     This requires both :attr:`Intents.reactions` and :attr:`Intents.members` to be enabled.
+    
+    .. note::
+
+        Consider using :func:`on_raw_reaction_remove` if you need this and do not want
+        to enable the members intent.
 
     :param reaction: The current state of the reaction.
     :type reaction: :class:`Reaction`
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -511,7 +511,7 @@ to handle it, which defaults to print a traceback and ignoring the exception.
 
         To get the message being reacted, access it via :attr:`Reaction.message`.
 
-    This requires :attr:`Intents.reactions` to be enabled.
+    This requires both :attr:`Intents.reactions` and :attr:`Intents.members` to be enabled.
 
     :param reaction: The current state of the reaction.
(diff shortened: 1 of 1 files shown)


mikeshardmind (commented):
This should probably refer users to the raw event to be consistent with [choices made around the reaction add event](https://github.com/Rapptz/discord.py/pull/6429#issuecomment-780417744)

PR #6568, opened 2021-03-23
Add scopes to utils.oauth_url

nihaals:
## Summary

This makes the `oauth_url` helper more useful, for example, you can give it `applications.commands`.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -126,7 +126,7 @@ def deprecated(instead=None):
         return decorated
     return actual_decorator
 
-def oauth_url(client_id, permissions=None, guild=None, redirect_uri=None):
+def oauth_url(client_id, permissions=None, guild=None, redirect_uri=None, scopes=None):
     """A helper function that returns the OAuth2 URL for inviting the bot
     into guilds.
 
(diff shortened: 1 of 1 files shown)


nihaals:
I'm unsure if you should be forced to pass in `bot` when using it, it might be better to always add that (but maybe someone wants to use it without `bot`?)

discord/utils.py
```diff
@@ -141,13 +141,16 @@ def oauth_url(client_id, permissions=None, guild=None, redirect_uri=None):
         The guild to pre-select in the authorization screen, if available.
     redirect_uri: :class:`str`
         An optional valid redirect URI.
+    scopes: List[:class:`str`]
```
nihaals, line 144: This could be changed to `Iterator` but I'm not sure if that fits with the rest of the documentation. `set` could also be used

Gobot1234: It would be Iterable not Iterator

Rapptz: Yeah `Iterable` would work here.

discord/utils.py
```diff
@@ -141,13 +141,16 @@ def oauth_url(client_id, permissions=None, guild=None, redirect_uri=None):
...
+        An optional valid list of scopes. Defaults to ``['bot']``.
```
Rapptz, line 145: This is missing `versionadded`.

Rapptz (changes requested):

Rapptz (approved):
LGTM. Thanks.

PR #6573, opened 2021-03-24
Add a delete_markdown helper function

Edwinexd:
## Summary

Add a helper function to delete Discord's markdown. In my case I need to extract text cleaned of markdown from some user-provided input. Thought I'd make a PR as others might have a similar use case.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -676,11 +676,14 @@ class clean_content(Converter):
         Whether to use nicknames when transforming mentions.
     escape_markdown: :class:`bool`
         Whether to also escape special markdown characters.
+    remove_markdown: :class:`bool`
+        Whether to also escape special markdown characters. This option is not supported with ``escape_markdown``
     """
-    def __init__(self, *, fix_channel_mentions=False, use_nicknames=True, escape_markdown=False):
+    def __init__(self, *, fix_channel_mentions=False, use_nicknames=True, escape_markdown=False, remove_markdown=False):
         self.fix_channel_mentions = fix_channel_mentions
         self.use_nicknames = use_nicknames
         self.escape_markdown = escape_markdown
+        self.remove_markdown = remove_markdown
 
     async def convert(self, ctx, argument):
         message = ctx.message
@@ -731,6 +734,8 @@ class clean_content(Converter):
 
         if self.escape_markdown:
             result = discord.utils.escape_markdown(result)
+        elif self.remove_markdown:
+            result = discord.utils.remove_markdown(result)
 
         # Completely ensure no mentions escape:
         return discord.utils.escape_mentions(result)
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -678,6 +678,8 @@ class clean_content(Converter):
         Whether to also escape special markdown characters.
     remove_markdown: :class:`bool`
         Whether to also escape special markdown characters. This option is not supported with ``escape_markdown``
+        
+        .. versionadded:: 1.7
     """
     def __init__(self, *, fix_channel_mentions=False, use_nicknames=True, escape_markdown=False, remove_markdown=False):
         self.fix_channel_mentions = fix_channel_mentions
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -677,7 +677,7 @@ class clean_content(Converter):
     escape_markdown: :class:`bool`
         Whether to also escape special markdown characters.
     remove_markdown: :class:`bool`
-        Whether to also escape special markdown characters. This option is not supported with ``escape_markdown``
+        Whether to also remove special markdown characters. This option is not supported with ``escape_markdown``
         
         .. versionadded:: 1.7
     """
--- a/discord/message.py
+++ b/discord/message.py
@@ -784,9 +784,9 @@ class Message(Hashable):
 
         .. note::
 
-            This *does not* escape markdown. If you want to escape
-            markdown then use :func:`utils.escape_markdown` along
-            with this function.
+            This *does not* affect markdown. If you want to escape
+            or remove markdown then use :func:`utils.escape_markdown` or :func:`utils.remove_markdown` along
+            with this function respectively.
         """
 
         transformations = {
--- a/discord/message.py
+++ b/discord/message.py
@@ -785,8 +785,8 @@ class Message(Hashable):
         .. note::
 
             This *does not* affect markdown. If you want to escape
-            or remove markdown then use :func:`utils.escape_markdown` or :func:`utils.remove_markdown` along
-            with this function respectively.
+            or remove markdown then use :func:`utils.escape_markdown` or :func:`utils.remove_markdown` 
+            respectively, along with this function.
         """
 
         transformations = {
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -479,13 +479,15 @@ _MARKDOWN_ESCAPE_COMMON = r'^>(?:>>)?\s|\[.+\]\(.+\)'
 
 _MARKDOWN_ESCAPE_REGEX = re.compile(r'(?P<markdown>%s|%s)' % (_MARKDOWN_ESCAPE_SUBREGEX, _MARKDOWN_ESCAPE_COMMON), re.MULTILINE)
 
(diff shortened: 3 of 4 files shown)


lgaan:
[This already exists](https://discordpy.readthedocs.io/en/latest/api.html#discord.utils.escape_markdown)

Edwinexd:
> [This already exists](https://discordpy.readthedocs.io/en/latest/api.html#discord.utils.escape_markdown).

Hey! Apricate the input but this does not already exist. Here is the difference between deleting and escaping markdown: input string with markdown: `**markdowned**`, escaping markdown: `\*\*markdowned\*\*`, deleting markdown: `markdowned`. Please lmk if I misunderstood you 👍

Vexs:
If this was to be added it should probably be be worked into both [commands.clean_content](https://discordpy.readthedocs.io/en/latest/ext/commands/api.html#discord.ext.commands.clean_content) and [message.clean_content](https://discordpy.readthedocs.io/en/latest/api.html?highlight=clean_content#discord.Message.clean_content). Also, you should probably refactor existing code from escape_markdown to be used in both instead of copying. 

Remove_markdown is a better name I think.

Edwinexd:
> If this was to be added it should probably be be worked into both [commands.clean_content](https://discordpy.readthedocs.io/en/latest/ext/commands/api.html#discord.ext.commands.clean_content) and [message.clean_content](https://discordpy.readthedocs.io/en/latest/api.html?highlight=clean_content#discord.Message.clean_content). Also, you should probably refactor existing code from escape_markdown to be used in both instead of copying. 
> 
> 
> 
> Remove_markdown is a better name I think. 

Hey! Appreciate the input, I'll get to work on those changes over the next few days👍

discord/utils.py
```diff
@@ -479,6 +479,36 @@ def resolve_template(code):
...
+        if a URL in the text contains characters such as ``_`` then it will
+        be left alone.
+        Defaults to ``True``.
+    Returns
+    --------
+    :class:`str`
+        The text with the markdown special characters deleted.
+    """
+
+    url_regex = r'(?P<url><[^: >]+:\/[^ >]+>|(?:https?|steam):\/\/[^\s<]+[^<.,:;\"\'\]\s])'
```
Rapptz, line 499: If this is being reused it's probably better to refactor it as a global constant.

discord/utils.py
```diff
@@ -479,6 +479,36 @@ def resolve_template(code):
...
+        The text to delete markdown from.
+    ignore_links: :class:`bool`
+        Whether to leave links alone when removing markdown. For example,
+        if a URL in the text contains characters such as ``_`` then it will
+        be left alone.
+        Defaults to ``True``.
+    Returns
+    --------
+    :class:`str`
+        The text with the markdown special characters deleted.
```
Rapptz, line 496:
```suggestion
    r"""A helper function that remove markdown characters.

    Parameters
    -----------
    text: :class:`str`
        The text to delete markdown from.
    ignore_links: :class:`bool`
        Whether to leave links alone when removing markdown. For example,
        if a URL in the text contains characters such as ``_`` then it will
        be left alone. Defaults to ``True``.

    Returns
    --------
    :class:`str`
        The text with the markdown special characters deleted.
```

Rapptz (changes requested):
I'm not sure I like this implementation, or even the concept of this function.

Right now, this is essentially just removing special characters indiscriminately which is hardly useful for a library helper. For example if the text says `10 * 5` then the `*` will be removed, rendering the original meaning of the sentence obsolete. However making this a bit more aware seems like it'd be difficult to do and have potential gotchas.

Edwinexd:
> Right now, this is essentially just removing special characters indiscriminately which is hardly useful for a library helper. For example if the text says `10 * 5` then the `*` will be removed, rendering the original meaning of the sentence obsolete. However making this a bit more aware seems like it'd be difficult to do and have potential gotchas.

Hey! That is true, however I can't figure out a simple way to implement support for that without opening up for potential syntax abuse. Anyway made some changes here and there lmk what you think 👍

Rapptz (changes requested):
Thanks for the PR and the changes.

This is missing the `versionadded` directives in the added functions and parameters. The next version is going to be `1.7`.

If the behaviour of the function can't change to be markdown aware then that limitation should be documented using the `.. note` directive because it does remove some of the usefulness of it from my perspective.

Other than that this looks fine to me.

discord/utils.py
```diff
@@ -479,6 +479,46 @@ def resolve_template(code):
...
+    Returns
+    --------
+    :class:`str`
+        The text with the markdown special characters removed.
+    
+    .. versionadded:: 1.7
+    
+    .. note::
+            This function is not markdown aware and may remove meaning from the original text. For example,
+            if the input contains ``10 * 5`` then it will be converted into ``10  5``.
```
Rapptz, line 507: This stuff goes before `Parameters`.

Rapptz (changes requested):

discord/ext/commands/converter.py
```diff
@@ -676,11 +676,16 @@ class clean_content(Converter):
         Whether to use nicknames when transforming mentions.
     escape_markdown: :class:`bool`
         Whether to also escape special markdown characters.
+    remove_markdown: :class:`bool`
+        Whether to also escape special markdown characters. This option is not supported with ``escape_markdown``
```
NCPlayz, line 680:
```suggestion
        Whether to also remove special markdown characters. This option is not supported with ``escape_markdown``
```

discord/message.py
```diff
@@ -784,9 +784,9 @@ def clean_content(self):
 
         .. note::
 
-            This *does not* escape markdown. If you want to escape
-            markdown then use :func:`utils.escape_markdown` along
-            with this function.
+            This *does not* affect markdown. If you want to escape
+            or remove markdown then use :func:`utils.escape_markdown` or :func:`utils.remove_markdown` along
+            with this function respectively.
```
NCPlayz, line 789:
```suggestion
            or remove markdown then use :func:`utils.escape_markdown` or :func:`utils.remove_markdown` 
            respectively, along with this function.
```

discord/utils.py
```diff
@@ -479,6 +479,46 @@ def resolve_template(code):
 
 _MARKDOWN_ESCAPE_REGEX = re.compile(r'(?P<markdown>%s|%s)' % (_MARKDOWN_ESCAPE_SUBREGEX, _MARKDOWN_ESCAPE_COMMON), re.MULTILINE)
 
+_URL_REGEX = r'(?P<url><[^: >]+:\/[^ >]+>|(?:https?|steam):\/\/[^\s<]+[^<.,:;\"\'\]\s])'
+
+_MARKDOWN_STOCK_REGEX = r'(?P<markdown>[_\\~|\*`]|%s)' % _MARKDOWN_ESCAPE_COMMON
+
+def remove_markdown(text, *, ignore_links=True):
+    r"""A helper function that remove markdown characters.
```
NCPlayz, line 487:
```suggestion
    r"""A helper function that removes Discord's markdown.
```

NCPlayz (changes requested):

discord/message.py
Edwinexd: Good catches, appreciate it! 👍

discord/utils.py
NCPlayz, line 487:
```suggestion
    r"""A helper function that removes markdown characters.
```

discord/utils.py
```diff
@@ -479,6 +479,46 @@ def resolve_template(code):
...
+    :class:`str`
+        The text with the markdown special characters removed.
+    """
+
+    def replacement(match):
+        groupdict = match.groupdict()
+        is_url = groupdict.get('url')
+        if is_url:
+            return is_url
+        return ''
```
ToxicKidz, line 515: I don't see why not just do `return groupdict.get('url', '')`

ToxicKidz (changes requested):

discord/utils.py
Rapptz: This one's a good point.

Rapptz (approved):

Rapptz:
Thanks for the PR. Sorry for the seemingly excessive back and forth.

PR #6574, opened 2021-03-24
Convert title/description/url to str

pikaninja:
## Summary

As mentioned in bikeshedding in the discord.py server, there was an oversight where these 3 values were not cast to str though it was specified in the documentation. This PR fixes this.

## Checklist

    - [ x] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -111,10 +111,19 @@ class Embed:
             colour = kwargs.get('color', EmptyEmbed)
 
         self.colour = colour
-        self.title = str(kwargs.get('title', EmptyEmbed))
+        self.title = kwargs.get('title', EmptyEmbed)
         self.type = kwargs.get('type', 'rich')
-        self.url = str(kwargs.get('url', EmptyEmbed))
-        self.description = str(kwargs.get('description', EmptyEmbed))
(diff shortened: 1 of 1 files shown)


discord/embeds.py
```diff
@@ -111,10 +111,10 @@ def __init__(self, **kwargs):
             colour = kwargs.get('color', EmptyEmbed)
 
         self.colour = colour
-        self.title = kwargs.get('title', EmptyEmbed)
+        self.title = str(kwargs.get('title', EmptyEmbed))
```
ToxicKidz, line 114: Casting `str` to `EmptyEmbed` will make the `embed's` title for example look like this:
image

ToxicKidz (changes requested):

discord/embeds.py
pikaninja: how would you suggest I resolve this? I could remove the str here, and put an isinstance check for EmptyEmbed later? Or is there a better way you can think of?

Rapptz (dismissed):
Should probably use `try`/`except` instead of the current implementation.

PR #6576, opened 2021-03-24
Moved region attribute from guild to voice channel

TheOneMusic:
## Summary

Discord recently wrote [this article](https://support.discord.com/hc/en-us/articles/360060570993-Voice-Regions-Update) to state that they will change the way region works on guilds. The guilds will no longer have a region attribute: instead, it will be made per voice channel.

This pull request doesn't completely fix the problem, but it's a good start. I though we could set this to version 1.8.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -29,7 +29,7 @@ import asyncio
 
 import discord.abc
 from .permissions import Permissions
-from .enums import ChannelType, try_enum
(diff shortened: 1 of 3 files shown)


Rapptz:
Breaking changes aren't meant for 1.x, it's not semver compatible.

PR #6577, opened 2021-03-24
Moved region attribute from guild to voice channel (2.0)

TheOneMusic:
## Summary

Following [this article](https://support.discord.com/hc/en-us/articles/360060570993-Voice-Regions-Update), Discord will now remove the `region` attribute from guilds to instead implement it per voice channel.

**This pull request doesn't fix everything, but it's a good start.** Because it is a breaking change it is set for **version 2.0**.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -29,7 +29,7 @@ import asyncio
 
 import discord.abc
 from .permissions import Permissions
-from .enums import ChannelType, try_enum
+from .enums import VoiceRegion, ChannelType, try_enum
 from .mixins import Hashable
 from . import utils
 from .asset import Asset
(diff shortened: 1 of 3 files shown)


Rapptz (changes requested):
I don't think this PR really belongs in 2.0, since we want to support 1.7 for a bit longer.

The old attribute can just stay and have pending removal in another API version but there's no need to remove it now, especially since it technically still works.

Likewise, this would need support in `VoiceChannel.edit` and I'm pretty sure the key for voice regions in voice channels is `rtc_region` not `region` with `null` being "automatic", not a value of 'automatic'.

Please don't close and remake this PR like last time, it causes clutter. You can edit a PR by pushing commits to the branch the current PR is using.

Thanks.

TheOneMusic:
This has successfully been implemented now :)

TheOneMusic:
This has successfully been implemented now :)

PR #6583, opened 2021-03-28
Create decompressed_raw_receive event

Keith-Cancel:
A rather simple event to avoid the needing to having to handle decompressing any received data unlike on_socket_raw_receive(msg).

## Summary

A rather simple addition. It simply creates an event called on_decompressed_raw_receive(msg). on_socket_raw_receive(msg) one is required to handle decompression to get any easily readable output when logging or debugging. So this event makes things a little simpler.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -427,6 +427,9 @@ class DiscordWebSocket:
             msg = self._zlib.decompress(self._buffer)
             msg = msg.decode('utf-8')
             self._buffer = bytearray()
+
+        self._dispatch('decompressed_raw_receive', msg)
+
         msg = json.loads(msg)
 
         log.debug('For Shard ID %s: WebSocket Event: %s', self.shard_id, msg)
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -200,7 +200,7 @@ to handle it, which defaults to print a traceback and ignoring the exception.
 
 .. function:: on_disconnect()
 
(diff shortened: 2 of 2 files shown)


Rapptz:
From a performance perspective I don't really want to do this. Dispatches are expensive when in a hot loop, which the gateway is, and at the moment (I plan to change this later) everyone pays the price for it.

Keith-Cancel:
Then is there a reason on_socket_raw_receive(msg) is always dispatched as well?

Rapptz:
Backwards compatibility.

Bluenix2:
What if there was an accompanying parameter passed to the bot? As in, you'd pass something to bot that enables the use of this? Meaning that for "production use" you can disable the whole dispatch call

Rapptz:
> What if there was an accompanying parameter passed to the bot? As in, you'd pass something to bot that enables the use of this? Meaning that for "production use" you can disable the whole dispatch call

That design doesn't really scale well. Like I said, turning off dispatches so you don't pay the price for it is something I plan on doing later and is out of the scope of this PR. However I'm not really sure these events are useful anymore so their fate is on thin ice from my perspective.

Keith-Cancel:
Fair enough if this is gonna get changed anyways some time in the future, I guess probably would be better to add the equivalent then.

Keith-Cancel:
Should I close this is this pull request then?

Rapptz:
Yeah I think this won't be explored right now. Sorry.

PR #6584, opened 2021-03-28
Fix mention prefixes

Miolus:
## Summary

It fixes mention-prefixes not being triggered when users add 1 space after the mention by themselves.

## Checklist

- [ x] This PR fixes an issue.
- [ x] This PR adds something new (e.g. new method or parameters).
- [ x] This PR is **not** a code change (e.g. documentation, README, ...)

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -46,7 +46,7 @@ def when_mentioned(bot, msg):
 
     These are meant to be passed into the :attr:`.Bot.command_prefix` attribute.
     """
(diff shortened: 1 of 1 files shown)


Rapptz:
I think this design space might be better suited for a boolean that tells the view to strip after the prefix is found but before parsing commands rather than doing it this way.

Rapptz:
This has been implemented in https://github.com/Rapptz/discord.py/commit/862d509d2e39f9d550ec3dff1e7d08fc6b3b5b8d using the aforementioned more robust method.

PR #6587, opened 2021-03-29
[Docs] Adding an Advanced quickstart option.

SuperstalkerX:
Note: This is just the first rough draft. Making this PR for input if I'm headed in the right or wrong direction.

## Summary

Adding an Advanced quickstart option for experienced discord.py users or programmers.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -74,3 +74,372 @@ On other systems:
     $ python3 example_bot.py
 
 Now you can try playing around with your basic bot.
+
+An Advanced Bot
+---------------
+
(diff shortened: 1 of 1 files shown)


Rapptz:
Heya.

I appreciate the underlying concept but this might be better suited written as a tutorial by someone with more experience with the library, its terminology, and their best practices. Topics like this are a bit too complicated to do without excessive input through some other platform (e.g. Discord) since there's a lot to cover. There are too many mistakes (some minor, some major) on here to really tackle individually and review.

So yeah, while the intention behind this is fine I think it needs to be done better and with a completely separate approach to pedagogy than the one currently presented.

Thanks anyway.

PR #6589, opened 2021-03-30
[docs] suggest presence in constructor instead of change_presence

sgtlaggy:
## Summary

Change the `How do I set the “Playing” status?` FAQ entry to point out the `activity` keyword arg in the `Client` constructor.
Also replaces the examples to show setting the activity by the constructor instead of `change_presence`.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -85,7 +85,8 @@ in the repository.
 How do I set the "Playing" status?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-There is a method for this under :class:`Client` called :meth:`Client.change_presence`.
+You may specify the ``activity`` keyword argument in the :class:`Client` constructor.
+There is also a method for this called :meth:`Client.change_presence`.
 The relevant aspect of this is its ``activity`` keyword argument which takes in an :class:`Activity` object.
 
 The status type (playing, listening, streaming, watching) can be set using the :class:`ActivityType` enum.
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -85,8 +85,15 @@ in the repository.
...
-The relevant aspect of this is its ``activity`` keyword argument which takes in an :class:`Activity` object.
+The ``activity`` keyword argument may be passed in the :class:`Client` constructor, given an :class:`Activity` object.
+
+There is also a method for this called :meth:`Client.change_presence`.
+
+.. warning::
+
+    It is highly discouraged to use :meth:`Client.change_presence` or API calls in :func:`on_ready` as this event may be called many times while running, not just once.
+
+    Discord has a high chance to completely disconnect you during the ``READY`` or ``GUILD_CREATE`` events (``1006`` close code) and there is nothing you can do to prevent it.
```
Rapptz, line 96:
```suggestion
    There is a high chance of disconnecting if presences are changed right after connecting.
```

I want to keep specifics and theory guessing out of it. I don't actually know why it disconnects you (I've tried figuring it out to no avail).

Rapptz (changes requested):
I think this is okay but removing the dynamic presence changing is also a bit odd. Some people want to be able to change their presences at runtime as well and this doesn't really point them to that direction.

sgtlaggy:
> Some people want to be able to change their presences at runtime as well and this doesn't really point them to that direction.

I feel like this is something users should be expected to extrapolate from the information given. `change_presence` is mentioned in the answer still, and its usage is no different from passing activity via the constructor.
Maybe the second line could be rewritten to "`change_presence` may be used to change the activity after the bot has started."

My main idea with the PR is to steer users away from using `change_presence` in cases where they don't need it. If they only see an example using the `Client` class, but they're using the `Bot` class, most of the users this PR targets will likely ignore that and go straight for the `change_presence` example.

Rapptz:
As it stands this entry recommends `activity` without actually detailing the difference between it and `change_presence`. If the wording could be modified slightly then I'd be fine with it.

Rapptz (approved):
Thanks. This is good enough.

PR #6594, opened 2021-03-31
Update background_task example to use ext.tasks

SebbyLaw:
## Summary

Updates `background_task.py` to use the tasks extension.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
Previous example still needs to exist somewhere since some people might be better off not using the task extension. Probably a new file named `background_task_asyncio.py`

PR #6603, opened 2021-04-01
Add StoreChannelConverter

DoggieLicc:
Add a channel converter to convert an argument to a StoreChannelConverter

## Summary

This is to add a converter to convert an argument to a StoreChannel.

When i was making a command to get a channel's info... i found out that StoreChannel doesn't have its own converter
(or even proper docs) so I want to add one because it'd be useful even if only a few guilds have store channels

Its in the same style as all the other channel converters

## Checklist

- [ ] This PR fixes an issue.


Rapptz (approved):
Thanks. There's some missing `versionadded` but I can add those myself.

Issue #6508, opened 2021-03-05
[commands]commands.Choice


sevenc-nanashi:
### Summary

I want a new type for commands.

### What is the feature request for?

discord.ext.commands

### The Problem

This feature will help to make choices.
Like setting custom system message, buy/sell choice etc.

#### commands.Choice
Like typing.Union, commands.Greedy.
Can use like these:

`Choice["join", "leave"]`
`Choice["everyone", Union[discord.Member, discord.Role]]`

#### commands.errors.UnexpectedChoice
Raise when unexpected choice passed.
##### Attribute
UnexpectedChoice.expected_choices: List
`Choice["join", "leave"]` -> `["join", "leave"]`
`Choice["everyone", Union[discord.Member, discord.Role]]` -> `["everyone", Union[discord.Member, discord.Role]]`
```py
@bot.command()
async def set_sysmsg(ctx, event_name: commands.Choice["join", "leave"], *, message):
  # Set db and do something
  await ctx.send("Successfully updated system message.")
```

#### How does it works

User#0001: /set_sysmsg join Hello, {name}!
Bot#9999: Successfully updated system message.

User#0001: /set_sysmsg invalid This is invalid!
Bot#9999: commands.errors.UnexpectedChoice: Unexpected choice passed.

### The Current Solution

```py
@bot.command()
async def set_sysmsg(ctx, event_name, *, message):
  if event_name not in ("join", "leave"):
    return await ctx.send("Event name isn't valid.")
  # Set db and do something
  await ctx.send("Successfully updated system message.")
```

#### How does it works

User#0001: /set_sysmsg join Hello, {name}!
Bot#9999: Successfully updated system message.

User#0001: /set_sysmsg invalid This is invalid!
Bot#9999: Event name isn't valid.

Gobot1234:
This should hopefully be possible in V2 with a `typing(_extensions).Literal` converter.

sevenc-nanashi:
what is "V2"?

mesub7:
V2 is the next major version of discord.py (after 1.7).

Rapptz:
Fixed by #6619 and using `typing.Literal`.

PR #6619, opened 2021-04-04
[commands]Add typing.Literal converter

SigmathBits:
## Summary

Add functionality for a Literal converter using Python 3.8+'s `typing.Literal`. Written to not break for earlier compatible versions of Python and tested on Python 3.8 and Python 3.7.

This pull request also adds a new exception, `BadLiteralArgument`, for failed attempted Literal conversion. The way a Literal parameter is show in a command signature by default is as a literal list of its values. Additional changes made so that `typing.Literal`, `typing.Union`, `typing.Optional`, and `Greedy` should all work together properly and display properly in a command signature.

This also largely addresses issue #6508. 

## Checklist

- [x] This PR fixes an issue.


NCPlayz:
Since the next major version is targeted at 3.8, this can now be simplified.

Rapptz:
This missed the 1.7 review window, if you want you can retarget this to v2.0 which uses 3.8 as a minimum Python version and might simplify the code.

SigmathBits:
Alright, should be retargeted for v2.0 instead now!

discord/ext/commands/core.py
```diff
@@ -476,31 +476,42 @@ async def _actual_conversion(self, ctx, converter, argument, param):
...
+        if origin is typing.Literal:
+            errors = []
+            for literal in self._flattened_typing_literal_args(converter):
+                try:
+                    value = await self._actual_conversion(ctx, type(literal), argument, param)
+                except CommandError as exc:
+                    errors.append(exc)
+                else:
+                    if value == literal:
                         return value
```
bijij, line 511: Only convert to each unique type once.

```suggestion
            conversions = {}
            literal_args = self._flattened_typing_literal_args(converter)
            for literal in literal_args:
                literal_type = type(literal)
                try:
                    value = conversions[literal_type]
                except KeyError:
                    try:
                        value = await self._actual_conversion(ctx, literal_type, argument, param)
                    except CommandError as exc:
                        errors.append(exc)
                        conversions[literal_type] = object()
                        continue
                    else:
                        conversions[literal_type] = value
                    
                if conversions[literal_type] == literal:
                    return value
```

discord/ext/commands/core.py
```diff
@@ -476,31 +476,42 @@ async def _actual_conversion(self, ctx, converter, argument, param):
...
 
-                # if we're  here, then we failed all the converters
-                raise BadUnionArgument(param, converter.__args__, errors)
+            # if we're here, then we failed to match all the literals
+            raise BadLiteralArgument(param, tuple(self._flattened_typing_literal_args(converter)), errors)
```
bijij, line 514: Use a shared variable from the above for loop for the flattened args.

bijij (changes requested):
Could optimise by only converting to each unique type once, useful if a custom type has a high cost to convert to.

discord/ext/commands/core.py
SigmathBits: `self._flattened_typing_literal_args` is currently a generator, so it can't be reused in this way. Unless you think this should be changed?

bijij: I suggest storing the output of the generator in a tuple before iterating at the start.
And iterating over that tuple as well as passing it into this exception's constructor, saves the function call.

SigmathBits: Alright, cool. Would there possibly be some performance penalty to pre-generating the generator for the for-loop since it can exit early? When the exception isn't raised.

bijij: likely yes, but i wouldn't be able to tell you if running the generator twice is better or not.

could keep a list of args and append to it as you iterate, as you'd only ever need the full list for the exception if it had already exhausted all args.

SigmathBits: Appends might become slow, at least if there's a lot of literals. The penalty is probably negligible either way. Pre-generation seems the simplest and looks the best, I think, so I'll just go ahead with that solution.

SigmathBits:
Also, ideally this would somehow have an option for case insensitivity when comparing against string literals, but I don't know how this could be implemented. Maybe it could be tied to `ctx.bot.case_insensitive` parameter, but that might not be ideal.

discord/ext/commands/core.py
```diff
@@ -984,31 +1005,44 @@ def short_doc(self):
...
             greedy = isinstance(param.annotation, converters._Greedy)
 
+            # for typing.Literal[...], typing.Optional[typing.Literal[...]], and Greedy[typing.Literal[...]], the
+            # parameter signature is a literal list of its values
+            if self._is_typing_literal(param.annotation):
+                name = '|'.join(value_str(v) for v in self._flattened_typing_literal_args(param.annotation))
+            elif self._is_typing_optional(param.annotation) and self._is_typing_literal(typing.get_args(param.annotation)[0]):
+                name = '|'.join(value_str(v) for v in self._flattened_typing_literal_args(typing.get_args(param.annotation)[0]))
+            elif greedy and self._is_typing_literal(param.annotation.converter):
+                name = '|'.join(value_str(v) for v in self._flattened_typing_literal_args(param.annotation.converter))
```
Rapptz, line 1044: There's a lot of code duplication here and a lot of double checking. The performance of this could probably be better as a result.

Rapptz (changes requested):
Thanks for the PR, sorry for the small delay in review. It's been a bit hectic.

I have a few changes here and some minor concerns -- mainly performance (though it's minor). I think repeated calls to `get_origin` and `get_args` leads to unexpected performance consequences. The old code that checked the attribute directly was pretty much free -- however the new `typing` methods have a lot of checks going in them which makes them less than free. I think as a result it's better to save calls and cache those calls as much as possible so they're not repeatedly called.

There were plans to refactor the code in core to do that, though maybe this can be saved for after this PR.

Thanks again, this was something I was planning on doing and it saved me some time.

discord/ext/commands/core.py
```diff
@@ -984,31 +1005,44 @@ def short_doc(self):
...
+    def _is_typing_literal(self, annotation):
+        return typing.get_origin(annotation) is typing.Literal
 
-        return annotation.__args__[-1] is type(None)
+    def _flattened_typing_literal_args(self, annotation):
+        for literal in typing.get_args(annotation):
+            if self._is_typing_literal(literal):
+                yield from self._flattened_typing_literal_args(literal)
+            else:
+                yield literal
```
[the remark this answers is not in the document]
jack1142: Not on Python 3.8 it seems:
image

discord/ext/commands/errors.py
```diff
@@ -646,7 +648,10 @@ def _get_name(x):
             try:
                 return x.__name__
             except AttributeError:
-                return x.__class__.__name__
+                pass
+            if typing.get_origin(x) is not None:
+                return repr(x)
+            return x.__class__.__name__
```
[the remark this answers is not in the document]
SigmathBits: This was added purely here so that `typing.Literal` arguments (and typing hints in general) show as their `repr` when `BadUnionArgument` is raised instead of as `_SpecialForm` or something like `_LiteralGenericAlias`, which isn't as useful to the user. How might this be moved back? (And to what `except` clause?) Or should this just be removed?

discord/ext/commands/core.py
```diff
@@ -476,31 +476,52 @@ async def _actual_conversion(self, ctx, converter, argument, param):
...
+
+                try:
+                    value = await self.do_conversion(ctx, conv, argument, param)
+                except CommandError as exc:
+                    errors.append(exc)
+                else:
+                    return value
+
+            # if we're here, then we failed all the converters
+            raise BadUnionArgument(param, typing.get_args(converter), errors)
```
SigmathBits, line 500: @Rapptz, should I change this as well to cache `get_args`, or would that be feature creep?

discord/ext/commands/core.py
SigmathBits: Should hopefully be more optimised now.

discord/ext/commands/core.py
Rapptz: This can be done in another PR. It's fine

discord/ext/commands/errors.py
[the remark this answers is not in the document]
Rapptz: What I meant was rather than having a `pass` within the `except` case, this code should be inside of it instead. As far as I know `typing` types do not have `__name__` so it'll fall under the `AttributeError` case.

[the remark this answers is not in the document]
SigmathBits: Oh, right, I understand now; my bad!

Rapptz:
This currently has merge conflicts by the way.

discord/ext/commands/core.py
```diff
@@ -1037,9 +1072,9 @@ def signature(self):
                 else:
                     result.append(f'[{name}...]')
             elif greedy:
-                result.append(f'[{name}]...')
-            elif self._is_typing_optional(param.annotation):
-                result.append(f'[{name}]')
+                result.append('[%s]...' % name)
+            elif optional:
+                result.append('[%s]' % name)
```
Rapptz, line 1077:
```suggestion
                result.append(f'[{name}]...')
            elif optional:
                result.append(f'[{name}]')
```

Rapptz (changes requested):
Just one more thing, it probably got caught in your merge commit.

Rapptz (approved):
Thanks a lot. Looks good to me now.

PR #6621, opened 2021-04-04
Use `typing.Protocol`s

Gobot1234:
## Summary

Using `typing.Protocol`s allows for some abc and Converter logic simplification.

## Checklist

- [ ] This PR fixes an issue.


Rapptz (commented):
Looks mostly fine. I have a few questions though:

1. Why keep `abc.abstractmethod` decorators?
2. Why not turn the other ABCs into `Protocol` too? They don't have to be runtime checkable.
3. Maybe the documentation should be updated at some point to mention the usage of protocols rather than ABCs.

PR #6626, opened 2021-04-04
Make `commands.Greedy` a `typing.Generic`

Gobot1234:
## Summary

Makes `commands.Greedy` a `typing.Generic`.

## Checklist

- [ ] This PR fixes an issue.


discord/ext/commands/converter.py
```diff
@@ -823,28 +828,55 @@ def repl(obj):
...
-    def __getitem__(self, params):
+    def __class_getitem__(cls, params: Union[tuple[T], T]) -> Greedy[T]:
         if not isinstance(params, tuple):
             params = (params,)
         if len(params) != 1:
             raise TypeError('Greedy[...] only takes a single argument')
         converter = params[0]
 
-        if not (callable(converter) or isinstance(converter, Converter) or hasattr(converter, '__origin__')):
+        origin = getattr(converter, "__origin__", None)
```
Rapptz, line 871:
```suggestion
        origin = getattr(converter, '__origin__', None)
```

discord/ext/commands/converter.py
```diff
@@ -54,6 +57,8 @@
     'Greedy',
 )
 
+T = TypeVar("T")
```
Rapptz, line 60:
```suggestion
T = TypeVar('T')
```

discord/ext/commands/converter.py
```diff
@@ -823,28 +828,55 @@ def repl(obj):
         # Completely ensure no mentions escape:
         return discord.utils.escape_mentions(result)
 
-class _Greedy:
-    __slots__ = ('converter',)
 
-    def __init__(self, *, converter=None):
+GreedyBase = list if sys.version_info >= (3, 9) else List
```
Rapptz, line 832: I don't really know how I feel about guards like these.

Right now we're targeting 3.8 and in 3.9 we don't have to use `List`. In the future, i.e. like 4 years from now `List` will probably break. By that time (assuming Discord and this library survives that long) we'd have moved on from 3.8 and the minimum Python version would be greater than 3.9. So I'm not sure if littering the code with these guards accomplishes much.

discord/ext/commands/converter.py
```diff
@@ -823,28 +828,55 @@ def repl(obj):
...
+
+        if not (callable(converter) or isinstance(converter, Converter) or origin is not None):
             raise TypeError('Greedy[...] expects a type or a Converter instance.')
 
-        if converter is str or converter is type(None) or converter is _Greedy:
+        if converter in (str, type(None)) or origin is Greedy:
             raise TypeError(f'Greedy[{converter.__name__}] is invalid.')
 
-        if getattr(converter, '__origin__', None) is typing.Union and type(None) in converter.__args__:
+        if origin is Union and type(None) in get_args(converter):
```
Rapptz, line 879: I don't think `get_args` should be globally in scope with that name. Could re-export it as `get_typing_args` instead. This isn't a strong opinion of mine but at first glance this could be confusing.

Rapptz (changes requested):
Seems mostly okay.

discord/ext/commands/converter.py
```diff
@@ -784,41 +828,36 @@ async def convert(self, ctx: Context, argument: str) -> str:
...
-        transformations.update(
-            (f'<@{member_id}>', resolve_member(member_id))
-            for member_id in message.raw_mentions
-        )
-
-        transformations.update(
-            (f'<@!{member_id}>', resolve_member(member_id))
-            for member_id in message.raw_mentions
-        )
+        transformations.update((f'<@!{member_id}>', resolve_member(member_id)) for member_id in message.raw_mentions)
```
Rapptz, line 852: Please disable auto formatting on these lines, they're written that way for readability.

discord/ext/commands/converter.py
```diff
@@ -784,41 +828,36 @@ async def convert(self, ctx: Context, argument: str) -> str:
...
+
             def resolve_role(_id, *, _find=ctx.guild.get_role):
                 r = _find(_id)
                 return '@' + r.name if r else '@deleted-role'
 
-            transformations.update(
-                (f'<@&{role_id}>', resolve_role(role_id))
-                for role_id in message.raw_role_mentions
-            )
+            transformations.update((f'<@&{role_id}>', resolve_role(role_id)) for role_id in message.raw_role_mentions)
```
Rapptz, line 860: Please disable auto formatting on these lines, they're written that way for readability.

discord/ext/commands/converter.py
```diff
@@ -834,28 +873,51 @@ def repl(obj):
...
+    """
+
     __slots__ = ('converter',)
 
-    def __init__(self, *, converter=None):
+    def __init__(self, *, converter: T):
         self.converter = converter
 
-    def __getitem__(self, params):
+    def __class_getitem__(cls, params: Union[tuple[T], T]) -> Greedy[T]:
```
Rapptz, line 904:
```suggestion
    def __class_getitem__(cls, params: Union[Tuple[T], T]) -> Greedy[T]:
```

Rapptz (changes requested):

discord/ext/commands/converter.py
Rapptz: This is still being formatted.

Rapptz (approved):

PR #6632, opened 2021-04-05
Fixed `HTTPClient.delete_messages()` method

EQUENOS:
## Summary

In some reason, URL `"/channels/{channel_id}/messages/bulk_delete"` doesn't work anymore, but `"/channels/{channel_id}/messages/bulk-delete"` does (there's dash instead of underscore).
It's also specified that way in official documentation.

**UPD:** turns out, this applies only to API `v8`

## Checklist

- [x] Tested the code with/without the fix
- [x] This PR fixes an issue.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -391,7 +391,7 @@ class HTTPClient:
         return self.request(r, reason=reason)
 
     def delete_messages(self, channel_id, message_ids, *, reason=None):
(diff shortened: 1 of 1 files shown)


Rapptz:
I'm not sure why this breaks. We're currently using v7 where this route is available. 

This should be part of a larger change where v8 is used instead of v6, which will be done in the future. Thanks for the PR though.

PR #6633, opened 2021-04-05
use `typing.AsyncIterator` for iterators

NCPlayz:
## Summary

Make `iterators._AsyncIterator` use [`typing.AsyncIterator`](https://docs.python.org/3/library/typing.html#typing.AsyncIterator). 

Also types relevant methods bound to `T`. 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/iterators.py
+++ b/discord/iterators.py
@@ -26,8 +26,7 @@ from __future__ import annotations
 
 import asyncio
 import datetime
-import sys
-from typing import TYPE_CHECKING, TypeVar, Optional, Any, Callable, Union
+from typing import TYPE_CHECKING, TypeVar, Optional, Any, Callable, Union, List, AsyncIterator, Coroutine
 
 from .errors import NoMoreItems
 from .utils import time_snowflake, maybe_coroutine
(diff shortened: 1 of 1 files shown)


discord/iterators.py
```diff
@@ -22,20 +22,39 @@
...
+    from .message import Message
+    from .audit_logs import AuditLogEntry
+    from .guild import Guild
+
+if sys.version_info >= (3, 9):
+    from collections.abc import AsyncIterator, Coroutine
+else:
+    from typing import AsyncIterator, Coroutine
+
+T = TypeVar("T")
```
Rapptz, line 49:
```suggestion
T = TypeVar('T')
```

discord/iterators.py
```diff
@@ -22,20 +22,39 @@
...
+_Predicate = Callable[[T], Union[T, Coroutine[Any, Any, T]]]
```
Rapptz, line 50: This seems wrong to me.

A predicate is `(T) -> bool` not `(T) -> T`. Granted, it should probably be "convertible to bool" not really strictly `bool`.

discord/iterators.py
```diff
@@ -60,40 +79,35 @@ async def find(self, predicate):
...
-        return _MappedAsyncIterator(self, func)
+    def map(self, predicate: _Predicate[T]) -> _MappedAsyncIterator[T]:
+        return _MappedAsyncIterator(self, predicate)
 
-    def filter(self, predicate):
+    def filter(self, predicate: _Predicate[T]) -> _FilteredAsyncIterator[T]:
         return _FilteredAsyncIterator(self, predicate)
 
-    async def flatten(self):
+    async def flatten(self) -> list[T]:
```
Rapptz, line 93: This requires 3.9.

```suggestion
    async def flatten(self) -> List[T]:
```

Rapptz (changes requested):

Rapptz (approved):
LGTM.

PR #6634, opened 2021-04-05
Use `contextlib.asynccontextmanager` for `abc.Messageable.typing`

NCPlayz:
## Summary

Simplifies the code for `abc.Messageable.typing`, using the [`@asynccontextmanager` decorator](https://docs.python.org/3/library/contextlib.html#contextlib.asynccontextmanager).

Also removes regular (sync) context manager behaviour.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
I'm not really a fan of these considering how slow they end up being.

PR #6637, opened 2021-04-05
Use covariant TypeVar for protocol

mikeshardmind:
## Summary

Fixes a typing issue with covariance.

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -70,7 +70,7 @@ def _get_from_guilds(bot, getter, argument):
 
 
 _utils_get = discord.utils.get
-T = TypeVar("T", covariant=True)
+T = TypeVar('T', covariant=True)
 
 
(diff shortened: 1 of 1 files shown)


Gobot1234:
Shouldn't this be a separate one just for Greedy as Converter is now generic and uses T?

discord/ext/commands/converter.py
```diff
@@ -70,7 +70,7 @@ def _get_from_guilds(bot, getter, argument):
 
 
 _utils_get = discord.utils.get
-T = TypeVar("T")
+T = TypeVar("T", covariant=True)
```
Rapptz, line 73:
```suggestion
T = TypeVar('T', covariant=True)
```

Rapptz (approved):
Looks fine. Just need to change the quotes on this.

PR #6641, opened 2021-04-06
Add more type information to public API of converters

mikeshardmind:
## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -171,7 +171,7 @@ class MemberConverter(IDConverter[discord.Member]):
             return None
         return members[0]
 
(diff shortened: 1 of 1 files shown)


mikeshardmind:
This revealed a different issue with just the message and partial message converters:

Return type "Coroutine[Any, Any, Message]" of "convert" incompatible with return type "Coroutine[Any, Any, PartialMessage]" in supertype "Converter"

Moving it out of the class, or making it a staticmethod and using as such should both work here (it's only subclasses for cod reuse, and it doesn't need to be an instance method), but I didn't want to have scope creep with the PR

PR #6643, opened 2021-04-06
Fix attribute error when converter doesn't have a convert attribute

ToxicKidz:
## Summary

This pull request fixes a `AttributeError` when the converter passed to [`_actual_conversion`](https://github.com/Rapptz/discord.py/blob/master/discord/ext/commands/core.py#L433-L472) doesn't have the `convert` attribute.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
I'll merge this as-is but I think a cleaner solution is needed. Thanks.

PR #6651, opened 2021-04-06
Fix AttributeError on HTTPClient.send_file to be send_files

diceroll123:
## Summary

Fixes sending files with a Message, the old method was removed

## Checklist

- [ ] This PR fixes an issue.


Rapptz (changes requested):
As mentioned on Discord, rename the HTTPClient function rather than the call site.

Rapptz (approved):

PR #6655, opened 2021-04-06
Fix regression in Activity relying on `try_enum`

NCPlayz:
## Summary

Fixes a regression caused by 008432c creating an unknown type when given an enum value.

```py
>>> from discord import Activity, ActivityType
>>> Activity(type=ActivityType.watching, name='everyone')
<Activity type=<ActivityType.unknown_ActivityType.watching: <ActivityType.watching: 3>> name='everyone' url=None details=None application_id=None session_id=None emoji=None>
>>> _.to_dict() # (called internally)
  File "discord.py/discord/activity.py", line 220, in to_dict
    ret['type'] = int(self.type)
TypeError: __int__ returned non-int (type _EnumValue_ActivityType)
```

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/activity.py
+++ b/discord/activity.py
@@ -185,7 +185,10 @@ class Activity(BaseActivity):
         self.flags = kwargs.pop('flags', 0)
         self.sync_id = kwargs.pop('sync_id', None)
         self.session_id = kwargs.pop('session_id', None)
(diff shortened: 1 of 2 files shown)


Rapptz:
This probably needs some performance testing since this is a sensitive part of the code. I'm unsure what happened here but I think the bug might be elsewhere.

PR #6656, opened 2021-04-06
Add typing for flags

NCPlayz:
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/flags.py
+++ b/discord/flags.py
@@ -22,6 +22,10 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 DEALINGS IN THE SOFTWARE.
 """
 
+from __future__ import annotations
+
+from typing import Any, Callable, ClassVar, Dict, Iterator, List, Optional, Tuple, Type, TypeVar, overload
+
 from .enums import UserFlags
 
 __all__ = (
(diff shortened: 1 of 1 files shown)


discord/flags.py
```diff
@@ -410,14 +434,14 @@ def all(cls):
         return self
 
     @classmethod
-    def none(cls):
+    def none(cls: Type[I]) -> I:
```
Rapptz, line 437: I don't think this pattern actually needs to be generic.

```py
class C:
    @classmethod
    def foo(cls) -> 'C':
        return cls()
```

The following works and checks without any extra work. Admittedly, if for whatever reason these types would be inherited then the whole thing kind of falls apart. I'm not sure why the typing module doesn't have a `typing.Self` and ` typing.SelfType` honestly considering how common these two are but that's an aside.

It can stay as-is, I was just trying to understand why the generic was there.

However, note that rather than specifying the 3 different bound `TypeVar` you can just use the `BaseFlag` bound `TypeVar` and the type checker will accept it.

```py
T = TypeVar('T', bound='A')

class A:
    pass

class B(A):
    @classmethod
    def foo(cls: Type[T]) -> T:
        return cls()

b = B.foo() # resolved type: B
```

This is because all flags are required to have `BaseFlags` as its bottom type.

discord/flags.py
```diff
@@ -32,17 +36,28 @@
...
+    def __get__(self, instance: BF, owner: Type[BF]) -> bool:
+        ...
+
+    def __get__(self, instance: Optional[BF], owner: Type[BF]) -> Any:
         if instance is None:
             return self
         return instance._has_flag(self.flag)
 
-    def __set__(self, instance, value):
+    def __set__(self, instance: Any, value: bool) -> None:
```
Rapptz, line 60:
```suggestion
    def __set__(self, instance: BF, value: bool) -> None:
```

discord/flags.py
```diff
@@ -85,30 +105,30 @@ def _from_value(cls, value):
...
 
-    def __ne__(self, other):
+    def __ne__(self, other: Any) -> bool:
         return not self.__eq__(other)
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         return hash(self.value)
 
     def __repr__(self):
```
Rapptz, line 117: This could use `-> str` for consistency's sake.

Rapptz (changes requested):

Rapptz (approved):

PR #6667, opened 2021-04-07
Add typings for channels and `PartialUser`

NCPlayz:
## Checklist

- [ ] This PR fixes an issue.


discord/types/permissions.py
```diff
@@ -0,0 +1,33 @@
...
+
+from .snowflake import Snowflake
+from typing import Literal, TypedDict
+
+
+class Overwrite(TypedDict):
+    id: Snowflake
+    type: Literal[0, 1]
+    allow: str
+    deny: str
```
Rapptz, line 33: These are v8, we're currently in v6. It's fine to keep it I guess, just making a note of it.

I'm not sure this is worth having in its own file. Might be better off merging it with `channel.py` since permission overwrites are only a thing there. Should probably also call it `PermissionOverwrite` since that's what the object is called.

discord/types/user.py
```diff
@@ -0,0 +1,33 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+"""
+
+from discord.types.snowflake import Snowflake
```
Rapptz, line 25:
```suggestion
from .snowflake import Snowflake
```

Rapptz (changes requested):

Rapptz (approved):

PR #6668, opened 2021-04-07
Add typings for `Message`, `Emoji`, and `Member`

NCPlayz:
Reviews welcome.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/types/emoji.py
+++ b/discord/types/emoji.py
@@ -22,27 +22,20 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 DEALINGS IN THE SOFTWARE.
 """
 
-from .snowflake import Snowflake
-from typing import Literal, Optional, TypedDict
+from typing import Optional, TypedDict
+from .snowflake import Snowflake, SnowflakeList
+from .user import User
 
 
-class PartialUser(TypedDict):
-    id: Snowflake
-    username: str
-    discriminator: str
-    avatar: Optional[str]
+class PartialEmoji(TypedDict):
+    id: Optional[Snowflake]
+    name: Optional[str]
 
 
-PremiumType = Literal[0, 1, 2]
-
-
-class User(PartialUser, total=False):
-    bot: bool
-    system: bool
-    mfa_enabled: bool
-    local: str
-    verified: bool
-    email: Optional[str]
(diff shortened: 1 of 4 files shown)


discord/types/emoji.py
```diff
@@ -0,0 +1,41 @@
...
+"""
+
+from typing import Optional, TypedDict
+from .snowflake import Snowflake, SnowflakeList
+from .user import User
+
+
+class PartialEmoji(TypedDict):
+    id: Optional[Snowflake]
+    name: Optional[str]
```
Rapptz, line 32: According to the docs, `name` can only be `Optional` for the reaction version.

Maybe split it into two -- `ReactionEmoji` with `name: Optional[str]` and `Emoji` with `name: str`.

Alternatively, maybe `Emoji` can override the `name: str` locally, but I can't test how `TypedDict` works with that.

discord/types/message.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class Attachment(TypedDict):
+    id: Snowflake
+    filename: str
+    size: int
+    url: str
+    proxy_url: str
+    height: Optional[int]
+    width: Optional[int]
```
Rapptz, line 56: The docs are wrong here -- these keys are not always available.

Here's an actual attachment payload:
```json
    {
      "content_type": "text/plain; charset=Windows-1252",
      "filename": "...",
      "id": "829481043115245608",
      "proxy_url": "...",
      "size": 7214,
      "url": "..."
    }
```

Note that `content_type` is also missing. It seems to be always be available regardless of file.

discord/types/message.py
```diff
@@ -0,0 +1,134 @@
...
+    format_type: StickerFormatType
+
+
+class _MessageOptional(TypedDict, total=False):
+    guild_id: Snowflake
+    member: Member
+    mention_channels: List[ChannelMention]
+    reactions: List[Reaction]
+    nonce: Union[int, str]
+    webhook_id: int
```
Rapptz, line 108:
```suggestion
    webhook_id: Snowflake
```

Rapptz (changes requested):

Rapptz (approved):

Rapptz:
I decided I could probably fix these locally since I need these typings for something else right now.

PR #6688, opened 2021-04-09
Add typings for invites, templates, and bans

NCPlayz:
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/template.py
+++ b/discord/template.py
@@ -22,6 +22,7 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 DEALINGS IN THE SOFTWARE.
 """
 
+from typing import Any, Optional, TYPE_CHECKING, overload
 from .utils import parse_time, _get_as_snowflake, _bytes_to_base64_data
 from .enums import VoiceRegion
 from .guild import Guild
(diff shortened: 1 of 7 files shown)


discord/template.py
```diff
@@ -197,7 +201,25 @@ async def sync(self):
...
+        ...
+
+    @overload
+    async def edit(
+        self,
+        *,
+        name: Optional[str] = ...,
+        description: Optional[str] = ...,
+    ) -> None:
+        ...
```
Rapptz, line 220: These two are the same.

I know you're doing this because the `overload` decorator requires two overloads for it to work (a bizarre limitation) and this is because kwargs typing is basically impossible.

Unfortunately due to limitations of kwargs typing this means that using kwargs doesn't play well. I wonder if it'd be better to spread the kwargs instead. For this in particular (two parameters) it's not too bad, though for other things it's going to be a disaster.

Alternatively rather than spelling out *every* kwargs twice in both the overload sets, one can just be an empty function: 

```py
@overload
async def edit(self) -> None:
    ...
```

I haven't tested this out but it'd hope it's valid -- even if the construction makes no sense.

discord/template.py
```diff
@@ -197,7 +201,20 @@ async def sync(self):
...
+
+    @overload
+    async def edit(self, **kwargs) -> None:
```
Rapptz, line 214:
```suggestion
    async def edit(self) -> None:
```

Rapptz (approved):

PR #6691, opened 2021-04-10
[commands] Provide a dynamic cooldown system

Phxntxm:
## Summary

This provides a way to allow users a more dynamic system to control cooldowns, by allowing a callable to be used that returns a Cooldown object. This callable can even return None if no cooldown is meant to be applied in this situation, allowing for essentially a way to "whitelist" certain people/guilds/scenarios from the cooldown.

Couple notes:
- I moved the type to the cooldown mapping object. I'm unsure if this is really desirable, but since the checking of the bucket type only ever happened in CooldownMapping I figured this was fine. This was what I figured would be required to ensure it was the same BucketType no matter what Cooldown object the dynamic method provided returns. 
- To not have to add another if check on every cooldown check (inside of `CooldownMapping.create_bucket` to see if the cooldown is a callable/just a Cooldown object) I instead created a new class for the dynamic usage

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/cooldowns.py
+++ b/discord/ext/commands/cooldowns.py
@@ -34,6 +34,7 @@ __all__ = (
     'BucketType',
     'Cooldown',
     'CooldownMapping',
+    'DynamicCooldownMapping',
     'MaxConcurrency',
 )
 
@@ -69,18 +70,14 @@ class BucketType(Enum):
 
 
 class Cooldown:
-    __slots__ = ('rate', 'per', 'type', '_window', '_tokens', '_last')
+    __slots__ = ('rate', 'per', '_window', '_tokens', '_last')
 
-    def __init__(self, rate, per, type):
+    def __init__(self, rate, per):
         self.rate = int(rate)
         self.per = float(per)
-        self.type = type
         self._window = 0.0
         self._tokens = self.rate
         self._last = 0.0
-
-        if not callable(self.type):
-            raise TypeError('Cooldown type must be a BucketType or callable')
 
     def get_tokens(self, current=None):
         if not current:
@@ -128,15 +125,19 @@ class Cooldown:
         self._last = 0.0
 
     def copy(self):
-        return Cooldown(self.rate, self.per, self.type)
+        return Cooldown(self.rate, self.per)
 
     def __repr__(self):
         return f'<Cooldown rate: {self.rate} per: {self.per} window: {self._window} tokens: {self._tokens}>'
 
 class CooldownMapping:
-    def __init__(self, original):
+    def __init__(self, original, type):
+        if not callable(type):
+            raise TypeError('Cooldown type must be a BucketType or callable')
+
         self._cache = {}
         self._cooldown = original
+        self._type = type
 
     def copy(self):
         ret = CooldownMapping(self._cooldown)
@@ -152,7 +153,7 @@ class CooldownMapping:
         return cls(Cooldown(rate, per, type))
 
     def _bucket_key(self, msg):
-        return self._cooldown.type(msg)
+        return self._type(msg)
 
     def _verify_cache_integrity(self, current=None):
         # we want to delete all cache objects that haven't been used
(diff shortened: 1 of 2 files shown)


discord/ext/commands/cooldowns.py
```diff
@@ -128,15 +124,19 @@ def reset(self):
...
-        return Cooldown(self.rate, self.per, self.type)
+        return Cooldown(self.rate, self.per)
 
     def __repr__(self):
         return f'<Cooldown rate: {self.rate} per: {self.per} window: {self._window} tokens: {self._tokens}>'
 
 class CooldownMapping:
-    def __init__(self, original):
+    def __init__(self, original, type):
+        if not isinstance(type, BucketType):
```
Rapptz:
```suggestion
        if not callable(type):
```

This is a regression.

discord/ext/commands/cooldowns.py
```diff
@@ -128,15 +124,19 @@ def reset(self):
...
+            raise TypeError('Cooldown type must be a BucketType')
```
Rapptz:
```suggestion
            raise TypeError('Cooldown type must be a BucketType or callable')
```

discord/ext/commands/core.py
```diff
@@ -1979,9 +1986,48 @@ def cooldown(rate, per, type=BucketType.default):
...
+    """
+    if not callable(cooldown):
+        raise TypeError("A callable must be provided")
+
+    def decorator(func):
+        if isinstance(func, Command):
+            func._buckets = DynamicCooldownMapping(cooldown, type)
         else:
-            func.__commands_cooldown__ = Cooldown(rate, per, type)
+            func.__commands_cooldown__ = cooldown, type
```
Rapptz, line 2030:
```suggestion
            func.__commands_cooldown__ = DynamicCooldownMapping(cooldown, type)
```

discord/ext/commands/core.py
```diff
@@ -256,7 +257,12 @@ def __init__(self, func, **kwargs):
         except AttributeError:
             cooldown = kwargs.get('cooldown')
         finally:
-            self._buckets = CooldownMapping(cooldown)
+            if cooldown is None:
+                self._buckets = CooldownMapping(cooldown, BucketType.default)
+            elif callable(cooldown[0]):
+                self._buckets = DynamicCooldownMapping(*cooldown)
```
Rapptz, line 263:
```suggestion
            elif isinstance(cooldown, CooldownMapping):
                self._buckets = cooldown
```

discord/ext/commands/core.py
```diff
@@ -256,7 +257,12 @@ def __init__(self, func, **kwargs):
...
+            else:
+                self._buckets = CooldownMapping(*cooldown)
```
Rapptz, line 265: This doesn't work if `cooldown` isn't a tuple. However, there are no cases where this is true now, so probably remove it.

discord/ext/commands/core.py
```diff
@@ -1979,9 +1986,48 @@ def cooldown(rate, per, type=BucketType.default):
 
     def decorator(func):
         if isinstance(func, Command):
-            func._buckets = CooldownMapping(Cooldown(rate, per, type))
+            func._buckets = CooldownMapping(Cooldown(rate, per), type)
+        else:
+            func.__commands_cooldown__ = Cooldown(rate, per), type
```
Rapptz, line 1991:
```suggestion
            func.__commands_cooldown__ = CooldownMapping(Cooldown(rate, per), type)
```

Rapptz (changes requested):

discord/ext/commands/cooldowns.py
```diff
@@ -188,8 +188,20 @@ def update_rate_limit(self, message, current=None):
 
 class DynamicCooldownMapping(CooldownMapping):
 
+    def __init__(self, factory, type):
+        if not callable(type):
+            raise TypeError('Cooldown type must be a BucketType or callable')
+
+        self._cache = {}
+        self._factory = factory
+        self._type = type
```
Rapptz, line 197:
```suggestion
    super().__init__(None, type)
    self._factory = factory
```

Rapptz (changes requested):

Rapptz (approved):
LGTM.

PR #6704, opened 2021-04-11
change command.params back to an OrderedDict

Zomatree:
## Summary

fixes command.clean_params as it relies on `popitem` having a `last` parameter

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -37,6 +37,7 @@ import inspect
 import datetime
 import types
 import sys
+import collections
 
 import discord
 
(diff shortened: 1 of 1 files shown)


discord/ext/commands/core.py
```diff
@@ -143,10 +144,10 @@ def resolve_annotation(annotation: Any, globalns: Dict[str, Any], cache: Dict[st
         annotation = ForwardRef(annotation)
     return _evaluate_annotation(annotation, globalns, cache)
 
-def get_signature_parameters(function: types.FunctionType) -> Dict[str, inspect.Parameter]:
+def get_signature_parameters(function: types.FunctionType) -> collections.OrderedDict[str, inspect.Parameter]:
```
jack1142, line 147: The return type hint used here is Python 3.9 syntax. Python 3.8 only has the generic for `OrderedDict` in `typing.OrderedDict`

Rapptz:
I think this is better off being done in `clean_params`. Unfortunately Python already did this breaking change, I'm just doing it too. Since dicts are ordered now there's a guarantee here so might as well take advantage of it.

PR #6728, opened 2021-04-14
Add `clean_prefix` attribute to commands.Context

MrKomodoDragon:
## Summary

Adds the `clean_prefix` attribute to commands.Context.

Reviews welcome

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/context.py
+++ b/discord/ext/commands/context.py
@@ -212,7 +212,11 @@ class Context(discord.abc.Messageable):
 
     @property
     def clean_prefix(self):
-        """:class:`str`: The cleaned up invoke prefix. i.e. mentions are ``@name`` instead of ``<@id>``."""
+        """clean_prefix: :class:`str`
+        The cleaned up invoke prefix. i.e. mentions are ``@name`` instead of ``<@id>``.
+        
+        .. versionadded:: 2.0
+        """
         user = self.context.guild.me if self.context.guild else self.context.bot.user
         # this breaks if the prefix mention is not the bot itself but I
         # consider this to be an *incredibly* strange use case. I'd rather go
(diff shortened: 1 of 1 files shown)


discord/ext/commands/context.py
```diff
@@ -54,6 +54,10 @@ class Context(discord.abc.Messageable):
         :func:`on_command_error` event then this dict could be incomplete.
     prefix: :class:`str`
         The prefix that was used to invoke the command.
+    clean_prefix: :class:`str`
```
Gobot1234, line 57: This shouldn't be in this section and should be merged into the property docstring

discord/ext/commands/context.py
```diff
@@ -206,6 +210,21 @@ def valid(self):
     async def _get_channel(self):
         return self.channel
 
+    @property
+    def clean_prefix(self):
+        """clean_prefix: :class:`str`
```
SebbyLaw, line 215: This isn't how property docstrings are formatted. See other properties for details

discord/ext/commands/context.py
```diff
@@ -206,6 +206,17 @@ def valid(self):
     async def _get_channel(self):
         return self.channel
 
+    @property
+    def clean_prefix(self):
+        """:class:`str`The cleaned up invoke prefix. i.e. mentions are ``@name`` instead of ``<@id>``."""
```
Rapptz, line 211:
```suggestion
        """:class:`str`: The cleaned up invoke prefix. i.e. mentions are ``@name`` instead of ``<@id>``.

        .. versionadded:: 2.0
        """
```

Rapptz (changes requested):
If this is going to live in context then it should be removed from `HelpCommand` too.

Rapptz (approved):

PR #6730, opened 2021-04-14
Fix guild.chunk() creating a hanging waiter

diceroll123:
## Summary

If you're trying to chunk a guild that the bot is not in, it'll just hang on the chunk coro forever. It's weird that one would even encounter this, I know.

I'm not 100% sure if this fix should be returning a coro here or not. As-is, seems to work.

In the meantime, if you're affected by this, the smallest fix for this would be putting `guild.chunk()` under an `if bot.get_guild(guild.id):`

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -2218,7 +2218,8 @@ class Guild(Hashable):
         if not self._state._intents.members:
             raise ClientException('Intents.members must be enabled to use this.')
 
-        return await self._state.chunk_guild(self, cache=cache)
+        if not self._state.is_guild_evicted(self):
+            return await self._state.chunk_guild(self, cache=cache)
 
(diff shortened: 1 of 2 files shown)


diceroll123:
TBH I kind of expect this to raise some kind of exception by default, so I'm all ears for changes.

mikeshardmind:
It might be better to just have chunk requests error out if they don't receive *anything* back within a certain time frame, we don't know what all the failure conditions could be here. (If the websocket doesn't send back a single chunk related to the request within a minute of the request, stop waiting on it, raise an error to all waiters)

This approach would require some slight changes to ChunkRequest, and should probably also include some internal queuing of chunk requests to not be making too many of them at once.

PR #6754, opened 2021-04-19
Embed inline default false

vivisrc:
## Summary

Sets the default `inline` value to `False` on embed fields. This would make the behaviour consistent with Discord's API when the value is omitted.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
Thanks, but I prefer the default being as it is and I don't strictly adhere to the defaults the API sets if I feel like there can be a better default.

PR #6762, opened 2021-04-20
[commands] Add support for aliasing to FlagConverter

bijij:
## Summary

Adds support for flag aliasing to FlagConverter

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/flags.py
+++ b/discord/ext/commands/flags.py
@@ -101,6 +101,7 @@ class Flag:
     """
 
     name: str = MISSING
+    aliases: List[str] = MISSING
     attribute: str = MISSING
     annotation: Any = MISSING
     default: Any = MISSING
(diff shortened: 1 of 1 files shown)


Rapptz (commented):
I think the aliases could be stored in the flags mapping, not really sure why they're separate at the moment.

bijij:
> I think the aliases could be stored in the flags mapping, not really sure why they're separate at the moment.

I figured this wouldn't be desired as it would allow accessing flag values by alias at runtime.

so 
```py
class Foo(FlagConverter):
    a: int = flag(aliases=['b'])
    
@command()
async def my_command(ctx, *, flags: Foo):
    foo.b  # this would be legal.
```
unless you would only setattr for non alias invocations?

Rapptz:
That wouldn't be legal, the attribute name is stored separately from the alias or flag name.

Rapptz (approved):

PR #6763, opened 2021-04-21
raise InvalidArgument when using overwrites

ebbit1q:
## Summary
when creating a channel with the overwrites set but using the wrong type
for the keys a not so descriptive AttributeError was raised as the
assumed to be a Member target does not have an id attribute.

instead the target is checked and a clear InvalidArgument is raised when
the overwrites dict has the wrong class as a key.

## Checklist

    - [ ] I have updated the documentation to reflect the changes.(change reflects documentation)
- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -295,6 +295,13 @@ class GuildChannel(Protocol):
         if overwrites is not None:
             perms = []
             for target, perm in overwrites.items():
+                if isinstance(target, Role):
+                    payload_type = _Overwrites.ROLE
+                elif isinstance(target, Member):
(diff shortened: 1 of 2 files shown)


NCPlayz:
This disallows usage of `discord.Object` as a substitute for `discord.Member` which may not be favourable.

ebbit1q:
I considered this and using a try statement instead, yet the documentation states the keys are either Role or Member.

It only needs that id though, it could be possible to just have passing the id be a third option and document it properly.

Rapptz:
Overwrite stuff is planning on changing in the future so this patch will be superceded by then.

PR #6767, opened 2021-04-21
[commands] Set constructible FlagConverter flags to not be required by default.

bijij:
## Summary

This PR sets Flags of type FlagConverter which are constructible to be optional.

## Checklist

- [x] If code changes were made then they have been tested. (By NC)
- [ ] This PR fixes an issue.


Rapptz:
This nested flag behaviour is definitely weird and not something I want to explicitly give the OK on. I figure there's going to be weird parsing ambiguities there. Nevertheless, this patch is fine but questionable at best.

PR #6793, opened 2021-04-26
Adjust quickstart to not show commands example

HyperGH:
## Summary

I thought this could be changed because new users might get the wrong impression that this is the intended
way to create commands, instead of using ext.commands. I also added a reference to ext.commands in the steps below.
P.S.: I am about 99% certain the way I did that reference is wrong, so if someone more experienced could take a look, it would
be greatly appreciated! Thanks!

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -32,7 +32,7 @@ It looks something like this:
         if message.author == client.user:
             return
 
-        if message.content == 'Hello World!':
+        if message.content.startswith('hello'):
             await message.channel.send('Hello!')
 
     client.run('your token here')
@@ -54,8 +54,9 @@ There's a lot going on here, so let's walk you through it step by step.
 4. Since the :func:`on_message` event triggers for *every* message received, we have to make
    sure that we ignore messages from ourselves. We do this by checking if the :attr:`Message.author`
    is the same as the :attr:`Client.user`.
-5. Afterwards, we check if the :class:`Message.content` is ``'Hello World!'``. If it is,
-   then we send a message in the channel it was used in with ``'Hello!'``. 
+5. Afterwards, we check if the :class:`Message.content` starts with ``'hello'``. If it does,
+   then we send a message in the channel it was used in with ``'Hello!'``. This is a basic way of 
+   handling commands, which can be later automated with the :ref:`ext.commands` framework.
 6. Finally, we run the bot with our login token. If you need help getting your token or creating a bot,
    look in the :ref:`discord-intro` section.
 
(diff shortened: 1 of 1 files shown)


docs/quickstart.rst
```diff
@@ -32,7 +32,7 @@ It looks something like this:
         if message.author == client.user:
             return
 
-        if message.content.startswith('$hello'):
+        if message.content == ('Hello World!'):
```
Rapptz, line 35:
```suggestion
        if message.content == 'Hello World!':
```

docs/quickstart.rst
```diff
@@ -54,8 +54,9 @@ There's a lot going on here, so let's walk you through it step by step.
 4. Since the :func:`on_message` event triggers for *every* message received, we have to make
    sure that we ignore messages from ourselves. We do this by checking if the :attr:`Message.author`
    is the same as the :attr:`Client.user`.
-5. Afterwards, we check if the :class:`Message.content` starts with ``'$hello'``. If it is,
-   then we send a message in the channel it was used in with ``'Hello!'``.
+5. Afterwards, we check if the :class:`Message.content` is ``'Hello World!'``. If it is,
+   then we send a message in the channel it was used in with ``'Hello!'``. This is a basic way of 
+   handling commands, which can be later automated with the :ref:`ext.commands` framework.
```
Rapptz, line 59:
```suggestion
5. Afterwards, we check if the :class:`Message.content` is ``'Hello World!'``. If it is,
   then we send a message in the channel it was used in with ``'Hello!'``. 
```

Rapptz (changes requested):

HyperGH:
Updated the PR based on feedback on Discord

docs/quickstart.rst
HyperGH: I strongly think that a reference to ext.commands should be included somewhere on this page, if not here then somewhere else.

docs/quickstart.rst
```diff
@@ -32,7 +32,7 @@ It looks something like this:
         if message.author == client.user:
             return
 
-        if message.content.startswith('$hello'):
+        if message.content.startswith('hello'):
```
Rapptz, line 35:
```suggestion
        if message.content.startswith('$hello'):
```

docs/quickstart.rst
```diff
@@ -54,8 +54,9 @@ There's a lot going on here, so let's walk you through it step by step.
 4. Since the :func:`on_message` event triggers for *every* message received, we have to make
    sure that we ignore messages from ourselves. We do this by checking if the :attr:`Message.author`
    is the same as the :attr:`Client.user`.
-5. Afterwards, we check if the :class:`Message.content` starts with ``'$hello'``. If it is,
-   then we send a message in the channel it was used in with ``'Hello!'``.
+5. Afterwards, we check if the :class:`Message.content` starts with ``'hello'``. If it does,
```
Rapptz, line 57:
```suggestion
5. Afterwards, we check if the :class:`Message.content` starts with ``'$hello'``. If it does,
```

Rapptz (changes requested):

Rapptz (approved):

PR #6804, opened 2021-04-27
[Interactions] Create User only when in DMs

AXVin:
## Summary
Makes it so that User object is created only when self.guild_id is not present

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/interactions.py
+++ b/discord/interactions.py
@@ -111,11 +111,7 @@ class Interaction:
         except KeyError:
             self.message = None
 
-        try:
-            self.user = User(state=self._state, data=data['user'])
-        except KeyError:
-            self.user = None
-
+        self.user = None
         # TODO: there's a potential data loss here
         if self.guild_id:
             guild = self.guild or Object(id=self.guild_id)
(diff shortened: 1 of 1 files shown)


discord/interactions.py
```diff
@@ -111,18 +111,20 @@ def _from_data(self, data: InteractionPayload):
...
-            self.user = None
-
+        self.user = None
         # TODO: there's a potential data loss here
         if self.guild_id:
             guild = self.guild or Object(id=self.guild_id)
             try:
                 self.user = Member(state=self._state, guild=guild, data=data['member'])
             except KeyError:
                 pass
```
apple502j, line 121: What happens here then?

AXVin: Moved it outside the conditional to avoid putting it twice in `except KeyError`
Which was previously being set to None as well

discord/interactions.py
```diff
@@ -111,18 +111,20 @@ def _from_data(self, data: InteractionPayload):
         except KeyError:
             self.message = None
 
-        try:
-            self.user = User(state=self._state, data=data['user'])
-        except KeyError:
-            self.user = None
-
+        self.user = None
```
Rapptz, line 114:
```suggestion
        self.user: Optional[Union[User, Member]] = None
```

PR #6805, opened 2021-04-27
Add utils.MISSING

NCPlayz:
## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -80,6 +80,9 @@ DISCORD_EPOCH = 1420070400000
 
 
 class _MissingSentinel:
+    def __eq__(self, other):
+        return False
+
     def __repr__(self):
         return 'MISSING'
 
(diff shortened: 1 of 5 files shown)


Rapptz (commented):
Some of these sentinels are used differently from one another so I don't think this is a simple replacement like implied here. Some sentinels have different `__repr__` because of how I want them to show up in the documentation. I could forego this if the consistency is desirable, though honestly making sentinels is pretty cheap.

When I brought up the `MISSING` thing in the d.py server I was thinking of the sentinels literally called `MISSING`, not refactoring every sentinel in the library.

discord/utils.py
```diff
@@ -79,6 +79,17 @@
 DISCORD_EPOCH = 1420070400000
 
 
+class _MissingSentinel:
+    def __eq__(self, other):
+        return False
+
+    def __repr__(self):
+        return 'MISSING'
```
Rapptz, line 87:
```suggestion
        return '...'
```

Might as well use this since it's less noisy.

PR #6806, opened 2021-04-27
Auto-detect embed instance in message content

electromeow:
This is a little feature inspired by discord.js. With this new feature, `await discord.abc.Messageable.send(embed)` usage will be available just like In discord.js. It will check if the content is an instance of discord.Embed and embed param is None, then it will be used as embed. `await discord.abc.Messageable.send(embed=embed)` will be also available. But this feature will make it easier and prevent responses like `<discord.embeds.Embed object at 0x00000000>` happening because of inattention.
Sorry for my bad English if I couldn't explain properly. But I hope that you understood.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -1070,7 +1070,7 @@ class Messageable(Protocol):
         else:
             if isinstance(content, DiscordEmbed):
                 embed = content.to_dict()
-                content = None
+            content = None
 
         if allowed_mentions is not None:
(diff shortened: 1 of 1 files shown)


Rapptz:
Heya, this code is already mostly complicated and I prefer being explicit over implicit for the most part so I like and prefer the kwarg approach better here. 

Thanks anyway.

PR #6822, opened 2021-04-30
Remove extra http.send_typing call in async context manager

diceroll123:
## Summary

Removes an unneeded second immediate call to http.send_typing. 

## Checklist

- [ ] This PR fixes an issue.


diceroll123:
Actually, realized neither of the lines mattered, so removed both.

Rapptz:
These lines of code are there for a reason. I think it's more likely that `__enter__` will be getting removed and the context manager reworked.

PR #6829, opened 2021-05-01
Stop _edit_webhook_helper from passing an array with None to request

jmvlangen:
## Summary

Fixes an error that occurs when a method that uses _edit_webhook_helper does not include a file

## Checklist

- [x] This PR fixes an issue.


Rapptz:
Thanks. That code will actually no longer used and will be deleted in a separate branch.

PR #6834, opened 2021-05-02
Change doc of Webhook.send

sevenc-nanashi:
## Summary
I changed docs of Webhook.send.
I put period to some exceptions, and I made returns easier to understand.
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/webhook/async_.py
+++ b/discord/webhook/async_.py
@@ -1162,16 +1162,16 @@ class Webhook(BaseWebhook):
         Forbidden
             The authorization token for the webhook is incorrect.
         TypeError
-            You specified both ``embed`` and ``embeds`` or ``file`` and ``files``
+            You specified both ``embed`` and ``embeds`` or ``file`` and ``files``.
         ValueError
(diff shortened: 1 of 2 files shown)


discord/webhook/async_.py
```diff
@@ -1162,16 +1162,16 @@ async def send(
...
-            The length of ``embeds`` was invalid
+            The length of ``embeds`` was invalid.
         InvalidArgument
             There was no token associated with this webhook.
 
         Returns
         ---------
         Optional[:class:`WebhookMessage`]
-            The message that was sent.
+            The message that was sent, if ``wait`` is ``True``.
```
Rapptz, line 1174: This is already mentioned above. Not sure why it'd be repeated, but if so it'd have to be like this;
```suggestion
            If ``wait`` is ``True`` then the message that was sent, otherwise ``None``.
```

Rapptz (changes requested):
The SyncWebhook docs should be updated too.

Rapptz (approved):

PR #6841, opened 2021-05-03
Feature/threads

Slexyy:
## Summary

This pull request would fix the message.content blank print when printing the contents of the message. Seems like message.content is blank on bigger servers when printing the message.content.

## Checklist

- [x ] This PR is a breaking change (e.g. methods or parameters removed/renamed)

### DIFF
--- a/discord/__init__.py
+++ b/discord/__init__.py
@@ -56,6 +56,7 @@ from .raw_models import *
 from .team import *
 from .sticker import *
 from .interactions import *
+from .threads import *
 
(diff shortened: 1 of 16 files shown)


Ak4zh:
@Slexyy Message.content is blank in user-bots and is intentional.
Discord and this library owners does not want us to use user-bots so this will not be fixed.

I have found an alternative library which allows user-bots:
https://github.com/Merubokkusu/Discord-S.C.U.M

Slexyy:
Thank you for your answer @Ak4zh

PR #6846, opened 2021-05-04
Clarify where abc.Messageable.fetch_message finds its messages from

Jackywathy:
Clarify the "destination" that messages are fetched from. Fetch_message only finds a message with message_id from the channel that Messageable wraps, and not from other channels in (for example) the guild.

## Summary

Clarify what where fetch_message for abc.Messageable finds messages from.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -1155,7 +1155,8 @@ class Messageable(Protocol):
     async def fetch_message(self, id):
         """|coro|
 
-        Retrieves a single :class:`~discord.Message` from the destination.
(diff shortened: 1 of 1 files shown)


Rapptz:
Where was this wording confusing, contextually?

Jackywathy:
fetch_message only finds messages in the current channel, otherwise a 404 is returned. At least for me it was a little ambiguous for user / context objects as to what exactly the 'destination' that the function retrieved from.

PR #6896, opened 2021-05-11
add template url

SebbyLaw:
## Summary

Implements `Template.url` which return's the template's url

## Checklist

- [ ] This PR fixes an issue.


ImVaskel:
> Should this also implement `__str__` in line with Invite?

I think it should, this would match `Invite`'s `__str__`.

PR #6813, opened 2021-04-28
Type up **kwargs of various methods

NCPlayz:
## Summary

Add rich type hints to various keyword-only methods in discord.py. Will reformat with black once done. Reviews welcome.

## Checklist

- [ ] If code changes were made then they have been tested.
    - [ ] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.
- [x] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

PR #6897, opened 2021-05-11
Fix issues with imports causing NameErrors

bijij:
## Summary

Fixes NameErrors caused by lack of `from __future__ import annotations` directives in  #6813

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -26,6 +26,7 @@ from __future__ import annotations
 
 import copy
 import asyncio
+from datetime import datetime
 from typing import Any, Dict, List, Mapping, Optional, TYPE_CHECKING, Protocol, Type, TypeVar, Union, overload, runtime_checkable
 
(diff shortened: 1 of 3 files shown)


discord/abc.py
```diff
@@ -52,8 +53,6 @@
 T = TypeVar('T', bound=VoiceProtocol)
 
 if TYPE_CHECKING:
-    from datetime import datetime
```
Rapptz, line 55: For this one I think `SnowflakeTime` can be moved inside this block instead. I'm not sure it's used anywhere outside of a type checking context anyway.

Rapptz (changes requested):

discord/abc.py
bijij: Moved definition to inside the block.

Rapptz (approved):

PR #6899, opened 2021-05-11
use premium_subscriber_role over premium_since for boosters

NCPlayz:
## Summary

Since [`premium_since` is documented as both optional and nullable](https://discord.com/developers/docs/resources/guild#guild-member-object), it seems more appropriate to check if they have the premium subscriber role.

## Checklist

- [x] This PR fixes an issue.


Rapptz:
Yeah it's optional. However, if it's present then it means they boosted, which is what this is checking. What bug is this fixing?

PR #6902, opened 2021-05-13
[commands] Update command.parent and command.parents docs

NextChai:
## Summary

Updated Command docs -> `parent`, `parents` and `root_parent` return a Group, not a Command. Although **Group** inherits from **Command**, a **Group** is not a **Command** and can be confusing for some.

Edited: added *root_parent*

## Checklist

- [ ] This PR fixes an issue.


proguy914629bot (approved):

proguy914629bot (approved):

Rapptz:
This change is technically wrong if someone implements their own subclass from `Command` that doesn't inherit from `Group` but I guess this is rare enough where this doesn't matter.

PR #6926, opened 2021-05-19
make create webhook consistent to create channel

pikaninja:
## Summary

in the create webhook method, name was a kwarg, but in create_x_channel, name is a positional, this pr makes them consistent

## Checklist

- [] This PR fixes an issue.


Rapptz:
This is probably not the approach I would have gone. I suspect requiring the `name` kwarg in the channel creation would have been more consistent.

PR #6932, opened 2021-05-20
add a webhook example

Euphyzr:
## Summary

Shows how to use `discord.Webhook` to send async and sync webhook messages and also using a discord client by fetching guild webhooks.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/basic_webhook.py
+++ b/examples/basic_webhook.py
@@ -1,0 +1,103 @@
+import asyncio
+import discord
+
+async def async_webhook():
+    import aiohttp
+
+    async with aiohttp.ClientSession() as session:
+        # webhook url anatomy:
+        # https://discord.com/api/webhooks/<id>/<token>
+
+        webhook = discord.Webhook.from_url(
+            # To create and get the webhook url:
+            # Server Settings > Integration > Webhooks > New Webhook > Copy Webhook Url
+            url = "your webhook url here",
+            # adapters are what define how your request should be handled
+            # here discord.AsyncWebhookAdapter is used since we're using aiohttp
+            adapter=discord.AsyncWebhookAdapter(session),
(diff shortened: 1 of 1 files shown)


apple502j:
1. We should have separate files for the three
2. and that's not how you use webhook.

Euphyzr:
I wished to make the webhook example self contained, thus one file. But I guess I can separate the client one. Regarding the usage, I didn't get how it's wrong. Can you please let me know? You wouldn't use webhooks to reply to a message true, but I wanted to highlight the mechanism.

numbermaniac:
Webhooks are often used for fetching something from an external website and posting it into a Discord channel; e.g. GitHub webhooks, Twitter webhooks, and so on. It's unusual for bots within Discord to be interacting with them in this way.

apple502j:
See https://discordpy.readthedocs.io/en/master/api.html#webhook-support for how to use webhook on discord.py 2.0.

Euphyzr:
Changed the example to delete and log when message contains "banned words" in a certain guild using webhook. Also, use webhook to log unhandled event exceptions. Hopefully, this is a more practical example. Updated the code to discord.py 2.0 as well.

Rapptz (changes requested):
As it stands I'm not really a big fan of this example. I think it tries to do too much rather than doing a simple thing.

Also please note that the library uses single quotes, not double quotes.

Tari-dev (changes requested):
Loop kwarg of aiohttp.ClientSession is deprecated since version 2.0.

JDJGInc:
You should probaly redesign it, so it's properly passing the session to the bot, like don't use a self.loop
Those don't exist in later aiohttp versions

PR #6937, opened 2021-05-21
added embeds.py example

Daishiky:
added example for embeds in the **examples** folder.

### DIFF
--- a/examples/embeds.py
+++ b/examples/embeds.py
@@ -1,0 +1,2 @@
+import discord
+from discord.ext import commands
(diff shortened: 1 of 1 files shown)


kaylynn234:
As it stands, this is superfluously commented and the comments themselves are quite poor. Additionally the duplication here is unheloful and the formatting is not in line with the rest of the library. Likewise I'm not sure if this is really a "practical" example of an embed - the command stands on its own without much context or reason for implementation. This example doesn't really convey why you should use an embed, and considering parts of it are invalid (i.e an argument to `url` that is badly formed) I don't think it's really helpful to others. As far as I know rapptz generally prefers discussion before examples PRs are opened, so in the future you may want to discuss things like this within [discord.py's support server](https://discord.gg/dpy) before opening a PR.

Rapptz:
This example is exceedingly low quality, sorry.

PR #6956, opened 2021-05-26
Add Member.get_role

mikeshardmind:
## Summary

  - Adds an efficient way to check if a member has a role by ID.

     This is done in a way consistent with the existing user API of the
     library.

     The more debated Member.has_role_id/has_role is intentionally not
     included for review at this time given the heavy bikeshedding of it.
 
If this is not something to be included yet or needs further discussion first, I have no issues with this being closed or delayed, but I figured getting the part that wasn't as controversial and was beneficial to those currently directly using the internals started was worth it.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/member.py
+++ b/discord/member.py
@@ -830,3 +830,20 @@ class Member(discord.abc.Messageable, _BaseUser):
             user_id = self.id
             for role in roles:
                 await req(guild_id, user_id, role.id, reason=reason)
+
+    def get_role(self, role_id: int) -> Optional[discord.Role]:
+        """Returns a role with the given ID from roles which the member has.
+
+        .. versionadded:: 2.0
+
+        Parameters
+        -----------
+        role_id: :class:`int`
+            The ID to search for.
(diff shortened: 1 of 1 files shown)


discord/member.py
```diff
@@ -830,3 +830,20 @@ async def remove_roles(self, *roles: Snowflake, reason: Optional[str] = None, at
             user_id = self.id
             for role in roles:
                 await req(guild_id, user_id, role.id, reason=reason)
+
+    def get_role(self, role_id: int) -> Optional[discord.Role]:
+        """Returns a role with the given ID from roles which the member is in.
```
numbermaniac, line 835:
```suggestion
        """Returns a role with the given ID from roles which the member has.
```
I think it's a bit more common to say that a member *has* certain roles, rather than saying that the member is *in* those roles.

Rapptz: Ditto

PR #6974, opened 2021-05-28
Add Role.is_assignable

tylergibbs2:
## Summary

Adds the ability for bots to quickly and efficiently check whether or not a `Role` is assignable to members by the logged in user.

Pros:

* Solves an issue that is commonly brought up in the help channels (Role hierarchy management/checking)
* Precedence with this type of method in the library (`Emoji.is_usable`)
* In case of future updates by Discord, this new method allows for the library to update the check rather than having library users update their own implementation of the check.

Bikeshed discussion: https://discord.com/channels/336642139381301249/603069307286454290/847640778485268520

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/role.py
+++ b/discord/role.py
@@ -258,6 +258,13 @@ class Role(Hashable):
         """
         return self.tags is not None and self.tags.is_integration()
 
+    def is_assignable(self):
+        """:class:`bool`: Whether the role is able to be assigned by the logged in user.
+
+        .. versionadded:: 2.0
+        """
+        return not self.is_default() and not self.managed and self.guild.me.top_role > self
+
     @property
     def permissions(self):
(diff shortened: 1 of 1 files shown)


discord/role.py
```diff
@@ -258,6 +258,13 @@ def is_integration(self):
         """
         return self.tags is not None and self.tags.is_integration()
 
+    def is_assignable(self):
+        """:class:`bool`: Whether the role is able to be assigned by the logged in user.
```
Rapptz, line 262:
```suggestion
        """:class:`bool`: Whether the role is able to be assigned or removed by the bot.
```

Rapptz (commented):
There can be some optimisation opportunities here but it's outside the scope of the PR. There's a minor documentation nit.

PR #6977, opened 2021-05-28
Fix bug in ui.View __slot__

sizumita:
## Summary

There was an error because _stopped and _timeout_handler were not in __slots__.

## Checklist

- [x] This PR fixes an issue.


Rapptz:
Slots are being removed locally since it doesn't make sense for a class that can be subclassed.

PR #6978, opened 2021-05-28
Fix `role.is_assignable()`

TheOneMusic:
## Summary

Since a discord client can be the owner of a maximum of 10 guilds, `role.is_assignable()` should take care of this case. This PR fixes this issue.

## Checklist

- [x] This PR fixes an issue.


Rapptz (commented):
Performance on this could be improved still but thanks.

PR #6985, opened 2021-05-29
Add documentation to user.mutual_guilds

lukenamop:
## Summary

`user.mutual_guilds` requires `Intents.members` (privileged intent) to be enabled because it uses `guild.get_member()`, this is important for anyone who may want to use it to know.

## Checklist

- [ ] This PR fixes an issue.


JDJGInc:
Yeah this would should help some people :D

Rapptz:
This note is actually not true. It requires cache to be enabled, which is similar but different. I also think it's kind of implied/obvious.

wblazej (approved):
That's right, it's quite helpful

PR #6986, opened 2021-05-29
Add Command.extras

mikeshardmind:
## Summary

Adds Command.extras as per discussion in discord.

I'm not sure if this should be marked as breaking or not, I went with yes since this is a class designed to be subclassable and this is adding a kwarg.

I went with a copy at _ensure_assignment_on_copy but there may be a reason to use a deep copy instead:

This may also benefit from examples of use at some point (I can add them to this PR if needed/preferable)

I have done a small bit of testing of this, but given the original issue that led to command objects being copying, I'm leaving this as a draft until I get a moment to set up a minimal test that it is suitable in that case as well.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -215,6 +215,10 @@ class Command(_BaseCommand):
         If ``True``\, cooldown processing is done after argument parsing,
         which calls converters. If ``False`` then cooldown processing is done
         first and then the converters are called second. Defaults to ``False``.
+    extras: :class:`dict`
+        A dict of user provided extras to attach to the Command.
+
+        .. versionadded:: 2.0
     """
 
     def __new__(cls, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


discord/ext/commands/core.py
```diff
@@ -393,6 +398,8 @@ def _ensure_assignment_on_copy(self, other):
             other._buckets = self._buckets.copy()
         if self._max_concurrency != other._max_concurrency:
             other._max_concurrency = self._max_concurrency.copy()
+        if self.extras != other.extras:
+            other.extras = self.extras.copy()
```
Rapptz, line 402: This is already handled since it's passed as a kwarg so it has no need being here.

mikeshardmind: I was concerned about the interaction with issue #426 is this is treated as mutable data by the user and this is done only be assignment

PR #7010, opened 2021-06-03
New example

FrostiiWeeb:
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/advanced/advanced_bot.py
+++ b/examples/advanced/advanced_bot.py
@@ -1,0 +1,17 @@
+import discord
+from discord.ext import commands
+from datetime import datetime
+
+class MyAdvancedBot(commands.Bot):
+    def __init__(self, *args, **kwargs):
+        super().__init__(command_prefix="example ", *args, **kwargs)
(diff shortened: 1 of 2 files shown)


mesub7 (changes requested):
I wouldn't say this is _really_ advanced and there a quite a few things that need to be changed. 

• Sections of code need to be appropriately commented.
• Single quotes should be used where possible. 
• Have you thought about file persistence?

ppotatoo:
There isn't much use case for an economy without persistent storage.

itzsimpleboii8282 (commented):
I do not understand the intention of creating it, we already have an example like this which shows about creating a basic bot. The code is also missing comments to explain what a specific thing does.

Rapptz:
Sorry this example is too low quality.

FrostiiWeeb:
Yes, now I realize that it's too "low quality" because of these reasons:
Not using an actual database
it isn't too advanced, it's just a simple cog with economy that anybody can do.

PR #7043, opened 2021-06-10
[commands] Add Case-Insensitive to BotBase.get_cog()

IIHERO4:
## Summary

Add Case-Insensitive to cog lookup func ( .get_cog() )
I did it because the '[p]help Cog' is case-sensitive
also please take note that i changed .add_cog() to use .get_cog() instead of self.__cogs.get(cog_name), I dont know supposed to be case-senstive

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -518,7 +518,7 @@ class BotBase(GroupMixin):
             raise TypeError('cogs must derive from Cog')
 
         cog_name = cog.__cog_name__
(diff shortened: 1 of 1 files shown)


Rapptz:
This changes the implementation from O(1) to O(N*M) if this is enabled. This type of stuff is purposefully not done because it would cause conflicts in the help command, imagine e.g. a cog named `Help` and a command named `help`.

PR #7057, opened 2021-06-13
Show decorator usage instead of signature in docs

Gorialis:
## Summary

This change makes decorators show their usage instead of their signature, which helps avoid confusing situations where the intended usage of a decorator is not entirely clear.

Before:
image
image
image
image

After:
image
image
image
image

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -40,7 +40,10 @@ Client
 
 .. autoclass:: Client
     :members:
-    :exclude-members: fetch_guilds
+    :exclude-members: fetch_guilds, event
+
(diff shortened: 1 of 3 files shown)


ToxicKidz (changes requested):
I think [Cog.listener](https://discordpy.readthedocs.io/en/master/ext/commands/api.html#discord.ext.commands.Cog.listener) was missed here.

Rapptz (approved):
Not sure how I feel about this but I'll give it a go.

PR #7080, opened 2021-06-18
Add parameter "embeds" to Messageable.send, http.send_files, http.send_message

AomiVel:
## Summary

Add a new parameter to messageable.send and something related to this.
This change were made for [new API change](https://discord.com/developers/docs/change-log#support-for-multiple-embeds-in-message-routes).

## Checklist

- [ ] This PR fixes an issue.


sevenc-nanashi (changes requested):
`@overload` should be updated too.

discord/http.py
```diff
@@ -428,6 +429,9 @@ def send_message(
 
         if embed:
             payload['embed'] = embed
```
sevenc-nanashi, line 431:
```suggestion
        if embed:
            payload['embeds'] = [embed]
```
Changelog says
> The existing embed field is considered deprecated and will be removed in the next API version.
so embed should use embeds field.

sevenc-nanashi (changes requested):

discord/http.py
AomiVel: The  embed field isn't delete yet so I haven't made change like you.
When the embed field were deleted from API (=API version were updated), me or someone will make change like you.
(But maybe someone other than me will make a pull request like your change and it will be merged.)

JelNiSlaw: There is no point not to change it now, so you should do it

AomiVel: That's for sure. I fixed it.

discord/abc.py
```diff
@@ -1197,6 +1201,8 @@ async def send(self, content=None, *, tts=False, embed=None, file=None,
             Indicates if the message should be sent using text-to-speech.
         embed: :class:`~discord.Embed`
             The rich embed for the content.
```
Cryptex-github, line 1203: Should embed be marked as deprecated?

discord/abc.py
```diff
@@ -1197,6 +1201,8 @@ async def send(self, content=None, *, tts=False, embed=None, file=None,
...
+        embeds: List[:class:`~discord.Embed`]
+            A list of embeds to upload. Must be a maximum of 10.
```
Cryptex-github, line 1205: I think is better to state that is added in version 2.0, like view.

sevenc-nanashi:
```suggestion
        embeds: List[:class:`~discord.Embed`]
            A list of embeds to upload. Must be a maximum of 10.
            
            .. versionadded:: 2.0
```

Cryptex-github (changes requested):
You might want to update edit_message as well.

discord/abc.py
```diff
@@ -1185,9 +1185,13 @@ async def send(self, content=None, *, tts=False, embed=None, file=None,
...
-
-        If the ``embed`` parameter is provided, it must be of type :class:`~discord.Embed` and
-        it must be a rich embed type.
+        
+        To upload a single embed, the ``embed`` parameter should be used with a
+        single :class:`~discord.Embed` object. To upload multiple embeds, the ``embeds``
+        parameter should be used with a :class:`list` of :class:`~discord.Embed` objects.
+        **Specifying both parameters will lead to an exception**.
+        If the ``embed`` or ``embeds`` parameter is provided, it must be of 
+        type :class:`~discord.Embed` and it must be a rich embed type.
```
Rapptz, line 1194:
```suggestion
```

Rapptz (commented):
I'm not really a fan of the way `edit` was written but I'll just fix it locally. Thanks.

PR #7081, opened 2021-06-18
Add: add utils.to_timestamp_format

sevenc-nanashi:
## Summary

I added a function that makes time mention (Like: `<t:1624048212>`) from datetime.
Note: It only works in Canary, but I think it will work in stable when 2.0 is released.
image

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -509,6 +509,7 @@ class NSFWLevel(Enum):
     age_restricted = 3
 
 class TimeMentionType(Enum):
+    default         = "f"
     short_date_time = "f"
     long_date_time  = "F"
     short_date      = "d"
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -508,6 +508,15 @@ class NSFWLevel(Enum):
     safe = 2
     age_restricted = 3
 
+class TimeMentionType(Enum):
+    short_date_time = "f"
+    long_date_time  = "F"
+    short_date      = "d"
+    long_date       = "D"
+    short_time      = "t"
+    long_time       = "T"
+    relative        = "R"
+
 T = TypeVar('T')
 
 def create_unknown_value(cls: Type[T], val: Any) -> T:
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -508,7 +508,7 @@ class NSFWLevel(Enum):
     safe = 2
     age_restricted = 3
 
-class TimeMentionType(Enum):
+class TimestampFormat(Enum):
     default         = "f"
     short_date_time = "f"
     long_date_time  = "F"
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -569,7 +569,7 @@ def to_time_mention(time: datetime.datetime) -> str:
     :class:`str`
         A datetime mention for given datetime.
     """
-    return "<t:{}>".format(int(time.timestamp()))
+    return f"<t:{int(time.timestamp())}>"
 
 
 def valid_icon_size(size: int) -> bool:
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -561,8 +561,13 @@ def to_time_mention(time: datetime.datetime) -> str:
 
     Parameters
     -----------
-    when: :class:`datetime.datetime`
+    time: :class:`datetime.datetime`
         The ``datetime`` to make mention.
+
+    Returns
+    --------
+    :class:`str`
+        A datetime mention for given datetime.
     """
     return "<t:{}>".format(int(time.timestamp()))
 
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -572,7 +572,7 @@ def to_time_mention(time: datetime.datetime, mention_type: Optional[TimeMentionT
     :class:`str`
         A datetime mention for given datetime.
     """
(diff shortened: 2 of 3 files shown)


discord/utils.py
```diff
@@ -554,6 +554,19 @@ def utcnow() -> datetime.datetime:
...
+    """A helper function to make a time mention from ``datetime.datetime``.
+
+    .. versionadded:: 2.0
+
+    Parameters
+    -----------
+    when: :class:`datetime.datetime`
+        The ``datetime`` to make mention.
+    """
+    return "<t:{}>".format(int(time.timestamp()))
```
pikaninja, line 572: Given that this targets master, this can be done with a f string

lgaan:
You’ll need to implement the formatting flags along with this. See image for examples.

sevenc-nanashi:
I am not native of English, so typo change request is welcome.

discord/utils.py
sevenc-nanashi: It uses fstring now. thanks

bijij:
Might be nice to implement a `class` which accepts a datetime and via `__format__` and `TimeMentionType` values as the `format_spec` yield the same result.

sevenc-nanashi:
it's good, but i think it should support other timestamp (Like: embed), so it should be other pr.

discord/enums.py
```diff
@@ -508,6 +508,16 @@ class NSFWLevel(Enum):
...
+class TimeMentionType(Enum):
+    default         = "f"
+    short_date_time = "f"
+    long_date_time  = "F"
+    short_date      = "d"
+    long_date       = "D"
+    short_time      = "t"
+    long_time       = "T"
+    relative        = "R"
+
```
bijij, line 520: I'm not sure this needs to be an Enum, just type hinting the argument with `typing.Literal` probably suffices just file.

Either way unsure if `Mention` if the correct name for this, it's not neccisarrily a mention. just a markdown represenation of a unix timestamp.

Api Docs call this a style so perhaps

`UnixTimestampStyle = Literal["f", "F", "d", "D", "t", "T", "R"]` 

in utils.py would be better suited. can just link to the API docs for what each style does anyway,

discord/utils.py
```diff
@@ -554,6 +555,29 @@ def utcnow() -> datetime.datetime:
...
+    --------
+    :class:`str`
+        A datetime mention for given datetime.
+    """
+    if mention_type is None or mention_type is TimeMentionType.default:
+        return f"<t:{int(time.timestamp())}>"
+    else:
+        return f"<t:{int(time.timestamp())}:{mention_type.value}>"
+
+
```
bijij, line 580: Same as above, unsure about the naming, `to_timestamp_format` might be a better suited name?

bijij (commented):
I agree d.py needs a helper method to create these, but I'm not certain about some of the choices made.

discord/enums.py
nihaals: I disagree, an enum seems to be more fitting for this, and just using `Literal` wouldn't really fit in with the rest of the lib

sevenc-nanashi:
I just made a typo(I was editing on browser and I pushed Enter accidentally), so i force pushed

discord/utils.py
sevenc-nanashi: I changed to `to_timestamp_format`, thanks

discord/utils.py
```diff
@@ -554,6 +555,29 @@ def utcnow() -> datetime.datetime:
...
+    Returns
+    --------
+    :class:`str`
+        A datetime mention for given datetime.
+    """
+    if mention_type is None or mention_type is TimeMentionType.default:
+        return f"<t:{int(time.timestamp())}>"
+    else:
+        return f"<t:{int(time.timestamp())}:{mention_type.value}>"
```
bijij, line 578:
```suggestion
    if format_type is None or format_type is TimestampFormat.default:
        return f"<t:{int(time.timestamp())}>"
    else:
        return f"<t:{int(time.timestamp())}:{format_type.value}>"
```

bijij (changes requested):
These variables need to be renamed.

discord/utils.py
sevenc-nanashi: Thanks!

sevenc-nanashi: Thanks!

Rapptz:
Thanks but I don't think this is the approach I want to take for this feature. I think there are two types of syntaxes that can be used:

```py
f'{discord.utils.timestamp(dt):R}'
```

or
```py
f'{discord.utils.format_dt(dt, "R")}'
```

Compared to this PR:
```py
f'{discord.utils.to_timestamp_format(dt, discord.TimestampFormat.relative)}'
```

You'll note the first two of these have a few quirks:

1. They don't use an enum
2. They are inside an f-string

The reason for this is because I imagine a vast majority of use cases for this formatting tool is going to be, well, for formatting. Since formatting is a string based operation these will generally be placed within an f-string for the majority of uses. Having an enum or typing out the opposite string literal will have low ergonomics. I'm not sure which approach I'm going to take for this feature, but it's not the one in this PR, sorry.

PR #7087, opened 2021-06-20
Add a __str__ to EmptyEmbed so that __repr__ is not used

Bluenix2:
## Summary

Doing `str()` on EmptyEmbed would give `Empty.Embed` instead of an empty string as could be expected.


Rapptz:
I'm not sure I expect an empty string here. When you do `str(None)` you get `None`.

PR #7088, opened 2021-06-20
Localize Embed.timestamp during assignment

thegamecracks:
## Summary

This provides consistency between setting the timestamp through the constructor and through assignment, since originally it would interpret naive datetimes in constructor as local time but interpret the same datetime during `.timestamp =` assignment as UTC.

## Checklist

- [ ] This PR fixes an issue.


quiprr (approved):
haven't tested but lgtm

PR #7100, opened 2021-06-23
Set Message.guild from guild_id if unavailable through Message.channel 

LostLuma:
## Summary

Fixes `Message.guild` being `None` when `Message.channel` is `None` or is only an `Object`.

I ran into this a bunch due to the current issues with unarchiving threads, but it should also be useful in case there are new text channel types.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -657,7 +657,11 @@ class Message(Hashable):
         self.nonce = data.get('nonce')
         self.stickers = [Sticker(data=d, state=state) for d in data.get('stickers', [])]
         self.components = [_component_factory(d) for d in data.get('components', [])]
-        self.guild = state._get_guild(utils._get_as_snowflake(data, 'guild_id'))
+
+        try:
+            self.guild = channel.guild
+        except AttributeError:
+            self.guild = state._get_guild(utils._get_as_snowflake(data, 'guild_id'))
 
         try:
             ref = data['message_reference']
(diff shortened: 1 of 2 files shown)


NCPlayz (approved):

discord/message.py
```diff
@@ -656,6 +658,11 @@ def __init__(
         self.stickers = [Sticker(data=d, state=state) for d in data.get('stickers', [])]
         self.components = [_component_factory(d) for d in data.get('components', [])]
 
+        try:
+            self.guild = channel.guild
+        except AttributeError:
+            self.guild = state._get_guild(utils._get_as_snowflake(data, 'guild_id'))
```
NCPlayz, line 664:
```suggestion
        try:
            guild = channel.guild
        except AttributeError:
            guild: Optional[Guild] = state._get_guild(utils._get_as_snowflake(data, 'guild_id'))

        self.guild = guild
```
this should ideally work, assuming a solution for microsoft/pyright#2036 is made.

NCPlayz (approved):
Seems alright, just some annotation suggestions.

Rapptz:
I'm not sure I quite understand where this became an issue. Thread channels should have a `guild` attribute.

LostLuma:
The issue I ran into is that unarchiving a thread currently does not add it to the bots cache, as we discussed on Discord once before.

In that case `Message.channel` is only an `Object`, and `Message.guild` is None even though thread messages only exist in guilds.

Rapptz:
Well the changeset seems fine so I'll merge as-is.

Issue #7113, opened 2021-06-27
Button label field is typing.Optional, and defaults to None, but is a Discord required field.


mirandazellnik:
### Summary

The definition of `discord.ui.Button` shows the `label` field is an `Optional[str]`, which defaults to `None`. In fact, this field is required to successfully send a button.

### Reproduction Steps

- Create an instance of discord.ui.Button without setting the label
- Assign it to a view, and try to send that view.

### Minimal Reproducible Code

```python
@bot.command()
async def send_button(ctx: commands.Context):
    view = discord.ui.View()
    button = discord.ui.Button(style=discord.ButtonStyle.primary)
    view.add_item(button)
    await ctx.send('Message', view=view)
```

### Expected Results

A message with a button would be sent.

[The docs show](https://discordpy.readthedocs.io/en/master/api.html?highlight=view#id11) that the `label` field is marked as `Optional[str]`. This is in contrast to other fields which need to be specified at some point, such as the `description` field of an embed. The description field defaults to a special "Empty" value, and, importantly, has a type of `str` (no Optional) to note that it is required.

### Actual Results

You'll get this error:
```
File "/.../discord/ext/commands/core.py", line 125, in wrapped
    ret = await coro(*args, **kwargs)
  File "main.py", line 120, in send_button
    await ctx.send('Message', view=view)
  File "/.../discord/abc.py", line 1316, in send
    data = await state.http.send_message(channel.id, content, tts=tts, embed=embed,
  File "/.../discord/http.py", line 333, in request
    raise HTTPException(r, data)
discord.errors.HTTPException: 400 Bad Request (error code: 50035): Invalid Form Body
In components.0.components.0.label: This field is required
```

### Intents

Default intents

### System Information

Linux
discord.py v2.0.0a3267+gd30fea5b

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.

Zomatree:
`Optional` means you can pass either the type inside it or `None`. `label` is optional because you can send a button with an emoji without needing to pass a label

PR #7114, opened 2021-06-27
Change button label field to a required field

mirandazellnik:
## Summary

As per Rapptz/discord.py#7113, adjust `label` field to be similar to the `description` field of an embed. This might not be the ideal way of doing it, I just tried to mimic the embed's way of doing it as closely as possible.

## Checklist

    - [X] I have updated the documentation to reflect the changes. (**I believe the documentation should automatically update with a type change? If not, how do I change it?**)
- [X] This PR fixes an issue.

### DIFF
--- a/discord/ui/button.py
+++ b/discord/ui/button.py
@@ -39,9 +39,22 @@ __all__ = (
     'button',
 )
 
+class _EmptyButton:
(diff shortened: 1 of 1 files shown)


Rapptz:
This change is incorrect as stated in the original issue.

PR #7120, opened 2021-06-28
Rename `slowmode_delay` to `slowmode`

jay3332:
## Summary

This is a breaking change, and since development of 2.0 is going on right now, I feel like this would be the time to make this PR.

This has been discussed [before](https://discord.com/channels/336642139381301249/603069307286454290/834256491362910239) but not added.

I feel like `slowmode_delay` is a bit too verbose for me. For newcomers to discord.py, it is not obvious what the slowmode attribute is and would usually require a documentation lookup.

`slowmode_delay` was not to be consistent with the API anyways, as it is officially `rate_limit_per_user`  
If consistency with the API is preferred over this, renaming to `rate_limit_per_user` would be fine with me, too.  

I chose `slowmode` over `rate_limit_per_user` as I felt "slowmode" is what everyone calls it and would take less time to get used to

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -296,7 +296,7 @@ class GuildChannel:
             parent_id = parent and parent.id
 
         try:
-            options['rate_limit_per_user'] = options.pop('slowmode_delay')
+            options['rate_limit_per_user'] = options.pop('slowmode')
         except KeyError:
             pass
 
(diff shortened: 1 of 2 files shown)


Rapptz:
I'm against this change. Thanks though. The current name makes it obvious that it's an attribute that signals a numeric value, Yours does not.

PR #7123, opened 2021-06-28
Replace API-deprecated stickers with sticker_items

apple502j:
## Summary
Replaces API-deprecated `Message.stickers` with `Message.sticker_items`. See also discord/discord-api-docs#3171

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/asset.py
+++ b/discord/asset.py
@@ -215,15 +215,6 @@ class Asset(AssetMixin):
             animated=animated,
         )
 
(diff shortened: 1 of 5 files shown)


Rapptz (changes requested):
I don't think removing the `Sticker` type is the way to go. I suspect keeping `Message.sticker` will be fine but only as a `List[StickerItem]`.

PR #7132, opened 2021-06-29
add persistent view in on_ready to avoid loop issues

NCPlayz:
## Checklist

- [ ] This PR fixes an issue.


mikeshardmind:
I'm not a fan of this change in an example. Doing things like this in on_ready is generally a bad idea. The example being one where it is fine due to how it is handled will still lead to people doing other things that aren't in on_ready

PR #7136, opened 2021-06-30
Add missing payload types to http.py

bijij:
## Summary

Adds a few missing payload types to http.py

- edit template
- edit widget
- create / edit stage instance
- upsert global commands

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -1146,7 +1146,7 @@ class StageChannel(VocalGuildChannel):
 
             payload['privacy_level'] = privacy_level.value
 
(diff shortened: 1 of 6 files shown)


Rapptz:
I don't think I want to use typed dicts for payload types

bijij:
A fair few already do, so figured at that point all should. 🤷

Rapptz:
The ones that use it also function as models, they're not exclusive to the edit payload.

Rapptz:
Discussed in the Discord.

PR #7141, opened 2021-06-30
Fix documentation for `Select.callback`

jay3332:
## Summary

The documentation for the corresponding decorator specifies using 3 parameters (`self, select, interaction`), while the documentation only specifies two. This fixes that.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
This is wrong.

PR #7156, opened 2021-07-02
Added the ability to use `Client.fetch_channel` on Threads.

### LINKED ISSUE #7155: Fetch for threads
### Summary

Simular to channels a way to fetch threads

The core library

### The Problem

getting threads is fine now, as they are just very little, but when its out of beta there will be way more. This might result in them not beeing in cache. Trying to use fetch_channel results in `discord.errors.InvalidData: Unknown channel type 12 for channel ID 00000000000.`   (the 00000 would be replaced by the actual id of course) 

### The Ideal Solution

Adding either support for threads to fetch_channel (my favourite solution) or supply a fetch_thread command (this might just be an alias to fetch_channel if the first solution is viable)

### The Current Solution

None

### Additional Context

We talked about that on the discord server https://discord.com/channels/336642139381301249/669155775700271126/860627360252428329

https://discord.com/developers/docs/resources/channel#channel-object-channel-types describes the types for the different threads

Umbra stated that the issue is in https://github.com/Rapptz/discord.py/blob/master/discord/client.py#L1403-L1405

Because there is no entry in https://github.com/Rapptz/discord.py/blob/d7ed88459341527a69e8d8a7a77ad1d9c5e2832c/discord/channel.py#L1914-L1921 to accomodate thread types yet

AbstractUmbra:
## Summary

Fixes #7155 

Added the ability to use `Client.fetch_channel` using Thread ids as per the normal workflow.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -1907,6 +1907,8 @@ def _guild_channel_factory(channel_type: Union[ChannelType, int]):
         return StoreChannel, value
     elif value is ChannelType.stage_voice:
         return StageChannel, value
+    elif value in (ChannelType.news_thread, ChannelType.public_thread, ChannelType.private_thread):
+        return Thread, value
     else:
         return None, value
 
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -686,7 +686,7 @@ class TextChannel(discord.abc.Messageable, discord.abc.GuildChannel, Hashable):
                 type=ChannelType.public_thread.value,
             )
 
-        return Thread(guild=self.guild, data=data)
+        return Thread(guild=self.guild, state=self._state, data=data)
 
     def archived_threads(
         self,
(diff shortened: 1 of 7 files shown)


veni-vidi-code:
Doesnt this also add a new method?

veni-vidi-code:
But thanks for the fast work

AbstractUmbra:
> Doesnt this also add a new method?

No, this uses the existing `Client.fetch_channel` method, I simply edited the channel type coercion used during this method.

Rapptz (changes requested):
I don't think this code works. For starters there are multiple places where threads are constructed, not just there. Secondly, I'm not sure adding this into the guild_channel_factory is a good idea since that's meant for things that reside in `Guild.channels` for typing reasons. By adding this there the typing of that method has broken.

Rapptz (approved):

PR #7157, opened 2021-07-02
Add `.is_nsfw()` to Thread.

AbstractUmbra:
## Summary

This PR adds `Thread.is_nsfw()` for end user convenience. 
As Threads cannot be inherently NSFW themselves, this is read from the parent channel instead.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/threads.py
+++ b/discord/threads.py
@@ -275,6 +275,8 @@ class Thread(Messageable, Hashable):
         i.e. :meth:`.TextChannel.is_nsfw` is ``True``.
         """
         parent = self.parent
+        if parent is None:
+            raise ClientException('Parent channel not found')
         return parent.is_nsfw()
 
     def permissions_for(self, obj: Union[Member, Role], /) -> Permissions:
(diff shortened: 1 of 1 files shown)


Rapptz (changes requested):
Needs a `None` check for parent channel similar to `category_id`.

discord/threads.py
```diff
@@ -268,6 +268,17 @@ def is_news(self) -> bool:
...
+    def is_nsfw(self) -> bool:
+        """:class:`bool`: Whether the thread is NSFW or not.
+
+        An NSFW thread is a thread that has a parent that is an NSFW channel,
+        i.e. :meth:`.TextChannel.is_nsfw` is ``True``.
+        """
+        parent = self.parent
+        if parent is None:
+            raise ClientException('Parent channel not found')
+        return parent.is_nsfw()
```
Rapptz, line 280:
```suggestion
        parent = self.parent
        return parent is not None and parent.is_nsfw()
```

I figure this is better than throwing in a boolean method.

Rapptz (changes requested):

PR #7174, opened 2021-07-05
[commands] update clean_content to work when standalone

z03h:
## Summary

Changes clean_content converter to not use `ctx.message.raw_mentions`/`role_mentions` so it gives more expected results when used standalone outside of command conversion. Also fixes bug where specifying `use_nicknames=False` could resolve users outside of the current guild.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
Was it necessary to rewrite the entire function for this supposed bug fix? `clean_content` used those attributes by design because the `mentions` field requires the users to actually be in the mention regardless of `allowed_mentions`. The point of the function is to clean *all* mentions, even potential ones that are ignored from `allowed_mentions`. Likewise, `role_mentions` field requires role cache and has the same gotchas as the one I mentioned for users. I see no real reason to change the attributes.

z03h:
The rewrite was to make it searches arg for mentions instead of only replacing mentions that are in ctx.message.content.
Intention was to add ability to clean mentions if you use it outside of type hints like `await clean_content().convert(ctx, random_string)`. Unless that's an unintended use for converters?

discord/ext/commands/converter.py
```diff
@@ -818,67 +818,58 @@ class clean_content(Converter[str]):
...
+                      '@&': resolve_role,
+                      }
+
+        def repl(match):
+            type = match[1]
+            id = int(match[2])
+            transformed = transforms[type](id)
+            return transformed or match[0]
+
+        pattern = re.compile(r'<(@[!&]?|#)(\d{15,20})>')
```
Rapptz, line 871:
```suggestion
        pattern = re.compile(r'<(@[!&]?|#)([0-9]{15,20})>')
```

If this compilation isn't stored then it's pointless, the `re` module already caches recently used regex.

discord/ext/commands/converter.py
```diff
@@ -818,67 +818,58 @@ class clean_content(Converter[str]):
...
+                return f'#{c.name}' if c else '#deleted-channel'
+        else:
+            def resolve_channel(id):
+                return
+
+        transforms = {'@': resolve_member,
+                      '@!': resolve_member,
+                      '#': resolve_channel,
+                      '@&': resolve_role,
+                      }
```
Rapptz, line 863:
```suggestion
        transforms = {
            '@': resolve_member,
            '@!': resolve_member,
            '#': resolve_channel,
            '@&': resolve_role,
        }
```

discord/ext/commands/converter.py
```diff
@@ -818,67 +818,58 @@ class clean_content(Converter[str]):
         .. versionadded:: 1.7
     """
 
-    def __init__(self, *, fix_channel_mentions: bool = False, use_nicknames: bool = True, escape_markdown: bool = False, remove_markdown: bool = False) -> None:
+    def __init__(self,
+                 *,
+                 fix_channel_mentions: bool = False,
+                 use_nicknames: bool = True,
+                 escape_markdown: bool = False,
+                 remove_markdown: bool = False) -> None:
```
Rapptz, line 826:
```suggestion
    def __init__(
        self,
        *,
        fix_channel_mentions: bool = False,
        use_nicknames: bool = True,
        escape_markdown: bool = False,
        remove_markdown: bool = False,
    ) -> None:
```

discord/ext/commands/converter.py
```diff
@@ -818,67 +818,58 @@ class clean_content(Converter[str]):
...
-            (f'<@{member_id}>', resolve_member(member_id))
-            for member_id in message.raw_mentions
-        )
-
-        transformations.update(
-            (f'<@!{member_id}>', resolve_member(member_id))
-            for member_id in message.raw_mentions
-        )
-        # fmt: on
+    async def convert(self, ctx, arg):
```
Rapptz, line 832:
```suggestion
    async def convert(self, ctx: Context, argument: str) -> str:
```

Changing this signature made no sense.

Rapptz (changes requested):
Overall this is missing type hints. This also doesn't match the style of the rest of the library.

z03h:
🙃 I'm dented

discord/ext/commands/converter.py
```diff
@@ -818,67 +818,60 @@ class clean_content(Converter[str]):
...
-            (f'<@{member_id}>', resolve_member(member_id))
-            for member_id in message.raw_mentions
-        )
-
-        transformations.update(
-            (f'<@!{member_id}>', resolve_member(member_id))
-            for member_id in message.raw_mentions
-        )
-        # fmt: on
+    async def convert(self, ctx: Context, arg: str) -> str:
```
Rapptz, line 834:
```suggestion
    async def convert(self, ctx: Context, argument: str) -> str:
```

Parameter names are part of the API contract, you can't change them.

Rapptz (changes requested):
Type hints are still missing but it's fine it can be dealt with afterwards if you're not familiar with it.

z03h:
ah I missed that, and type hinting all the `resolve_x`?

Rapptz:
If you're not familiar with it it's fine.

z03h:
I think that's it, not sure if I missed something

Rapptz (approved):

PR #7176, opened 2021-07-05
[types] Type-hint backoff.py

bijij:
## Summary

This PR uses a a constrained typevar on the `integral` argument, to determine the return type of `delay`.

image

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/backoff.py
+++ b/discord/backoff.py
@@ -29,7 +29,7 @@ import time
 import random
 from typing import Callable, Generic, Literal, TypeVar, Type, overload, Union
 
-T = TypeVar('T', bound=Union[int, float])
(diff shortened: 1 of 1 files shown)


discord/backoff.py
```diff
@@ -22,14 +22,20 @@
...
 """
 
+from __future__ import annotations
+
+
 import time
 import random
+from typing import Callable, Generic, Literal, TypeVar, Type, overload, Union
+
+T = TypeVar('T', bound=Union[int, float])
```
Gobot1234, line 32: Shouldn't these be constrains rather than bounds?

bijij: You're correct yes. fixed.

PR #7182, opened 2021-07-06
Add created_at attribute for threads.

Chrovo:
Added "created_at" attribute for threads. It returns the time at which the thread was created in UTC.

## Summary
Other classes like discord.TextChannel have a "created_at" attribute. However, threads don't have one at the moment. The changes I have made was to add them to Thread class.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/threads.py
+++ b/discord/threads.py
@@ -26,6 +26,7 @@ from __future__ import annotations
 from typing import Callable, Dict, Iterable, List, Optional, Union, TYPE_CHECKING
 import time
 import asyncio
+import datetime
 
 from .mixins import Hashable
 from .abc import Messageable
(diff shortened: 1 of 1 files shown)


Rapptz:
Hm. Considering thread IDs can be the message ID that started this message I'm not sure this is correct. I also thought I had this already but I guess I didn't.

Chrovo:
I tried checking which one it was via the following code:
```python
@bot.event
async def on_thread_join(thread):
    print(datetime.datetime.utcnow())
 
 @bot.event
 async def on_message(message):
     print(f"{datetime.datetime.utcnow()} - {message.id}") #Just to differentiate the message event with the thread event. Also for differentiating the starter message from the message with the same id as the thread channel.
  
@bot.command()
async def test(ctx):
  if isinstance(ctx.channel, discord.Thread):
    print(f"{discord.utils.snowflake_time(ctx.channel.id)}")
```
On the "on_thread_join" event, it printed out "2021-07-06 23:59:04.240560" and on the "on_message" event, it printed out "2021-07-06 23:59:04.216911". These are indeed different as you aforementioned. What's more is when I use the test command, it printed out "2021-07-06 23:59:04.173000+00:00", a different time than both the other two.

PR #7189, opened 2021-07-07
Add a new view example for link buttons

Ay-355:
## Summary

This PR adds a new view example showcasing how to create a link button.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/views/link.py
+++ b/examples/views/link.py
@@ -1,0 +1,40 @@
+from discord.ext import commands
+
+import discord
+
+class GoogleBot(commands.Bot):
+    def __init__(self):
+        super().__init__(command_prefix=commands.when_mentioned_or('$'))
+
+    async def on_ready(self):
+        print(f'Logged in as {self.user} (ID: {self.user.id})')
+        print('------')
+
+
+# Define a simple View that gives us a google link button.
+# We take in `query` as the query that the command author requests for
+class Google(discord.ui.View):
+    def __init__(self, query: str):
+        super().__init__()
+        # we need to replace spaces with plus signs to make a valid url. Discord will raise an error if it isn't valid.
+        self.query = query.replace(' ', '+')
+
+        # Link buttons cannot be made with the decorator
+        # Therefore we have to manually create one.
+        # We make a valid google search url with the query and add it to the url parameter.
+        # Then we add it to the view
+        self.add_item(
+            discord.ui.Button(style=discord.ButtonStyle.link, label='Click Here', url='https://www.google.com/search?q=' + self.query)
+        )
+
+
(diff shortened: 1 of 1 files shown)


Rapptz:
I don't really think this is a good example of a link button. Also, this constructor got simplified so someone doesn't have to do it this way.

Ay-355:
Are there any other examples of a link button that would be a better fit for the repo, or is the example not needed?

examples/views/link.py
```diff
@@ -0,0 +1,40 @@
...
+        print('------')
+
+
+# Define a simple View that gives us a google link button.
+# We take in `query` as the query that the command author requests for
+class Google(discord.ui.View):
+    def __init__(self, query: str):
+        super().__init__()
+        # we need to replace spaces with plus signs to make a valid url. Discord will raise an error if it isn't valid.
+        self.query = query.replace(' ', '+')
```
Rapptz, line 20: No, you want to use [`urllib.parse.quote_plus`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote_plus). Be sure to import it at the top as `from urllib.parse import quote_plus`.

```suggestion
        # we need to quote the query string to make a valid url. Discord will raise an error if it isn't valid.
        query = quote_plus(query)
        url = f'https://www.google.com/search?q={query}'
```

examples/views/link.py
```diff
@@ -0,0 +1,40 @@
...
+
+        # Link buttons cannot be made with the decorator
+        # Therefore we have to manually create one.
+        # We make a valid google search url with the query and add it to the url parameter.
+        # Then we add it to the view
+        self.add_item(
+            discord.ui.Button(style=discord.ButtonStyle.link, label='Click Here', url='https://www.google.com/search?q=' + self.query)
+        )
```
Rapptz, line 28: The construction could be severely simplified.

```suggestion
        self.add_item(discord.ui.Button(label='Click Here', url=url))
```

Rapptz (changes requested):

Rapptz (approved):
Thanks.

examples/views/link.py
```diff
@@ -0,0 +1,39 @@
...
+        self.add_item(discord.ui.Button(label='Click Here', url=url))
+
+
+bot = GoogleBot()
+
+
+@bot.command()
+async def google(ctx: commands.Context, *, query: str):
+    """Returns a google link for a query"""
+    await ctx.send(f"Google Result for: `{query}`", view=Google(query))
```
Rapptz, line 36:
```suggestion
    await ctx.send(f'Google Result for: `{query}`', view=Google(query))
```

PR #7190, opened 2021-07-08
Re-try requests on 504 error and raise correct 5xx error

LostLuma:
## Summary

Re-tries requests that timed out instead of simply raising `HTTPException` and ensures all 5xx responses raise `DiscordServerError` instead of `HTTPException`.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -321,8 +321,8 @@ class HTTPClient:
 
                             continue
 
-                        # we've received a 500 or 502, unconditional retry
-                        if response.status in {500, 502}:
+                        # we've received a 500, 502, or 504, unconditional retry
+                        if response.status in {500, 502, 504}:
                             await asyncio.sleep(1 + tries * 2)
                             continue
 
@@ -331,7 +331,7 @@ class HTTPClient:
                             raise Forbidden(response, data)
                         elif response.status == 404:
                             raise NotFound(response, data)
-                        elif response.status == 503:
+                        elif response.status >= 500:
                             raise DiscordServerError(response, data)
                         else:
                             raise HTTPException(response, data)


discord/http.py
```diff
@@ -331,7 +331,7 @@ async def request(
                             raise Forbidden(response, data)
                         elif response.status == 404:
                             raise NotFound(response, data)
-                        elif response.status == 503:
```
Taarek, line 334: I think it was made this why because Discord's load balancer for channel operations (Send, fetch messages, etc) sometimes returns a 503 response, and it was a bit confusing, thus why it got its own ``DiscordServerError`` error class.

Meanwhile, you can get a huge range of 500 class errors, most of them coming from CloudFlare having partial or temporary outages. Maybe is slightly confusing if they would suddenly raise ``DiscordServerError`` but CloudFlare is having an outage.

LostLuma: The documentation states that this error is for any 5xx error returned by the API:

https://github.com/Rapptz/discord.py/blob/8df35c83a9b5a1aa8e671a58263455d22d5c8097/discord/errors.py#L170-L171

And there is already other code after the retry loop and in both webhook implementations doing the same:

https://github.com/Rapptz/discord.py/blob/8df35c83a9b5a1aa8e671a58263455d22d5c8097/discord/http.py#L349-L350

https://github.com/Rapptz/discord.py/blob/8df35c83a9b5a1aa8e671a58263455d22d5c8097/discord/webhook/async_.py#L195-L196

Taarek: Oh, apologies in that case! 

I was just thinking about this https://github.com/Rapptz/discord.py/blob/d30fea5b0dcba9cd130026b56ec01e78bd788aff/docs/whats_new.rst#bug-fixes-4

But there's also this https://github.com/Rapptz/discord.py/blob/d30fea5b0dcba9cd130026b56ec01e78bd788aff/docs/whats_new.rst#new-features-2
> Add more concrete exceptions for 500 class errors under :class:`DiscordServerError` (:issue:`5797`)

So I guess it was an oversight, sorry once again! 💜

PR #7234, opened 2021-07-14
Added pagination example with buttons

Prince2347X:
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/views/pagination.py
+++ b/examples/views/pagination.py
@@ -1,113 +1,72 @@
 import discord
-
 from discord.ext import commands
 
+
 class PaginationView(discord.ui.View):
-
-    current = 0
-
-    def __init__(self, embed_list: list):
-
+    current = 0 # It's not in the constructor because it'll be reset everytime the view is updated.
+    def __init__(self, embed_list):
         super().__init__()
-
         self.embed_list = embed_list
+        if len(embed_list) == 1:
+            self.next.disabled = True
+            self.next.style = discord.ButtonStyle.red
+        self.count.label = f"Page {self.current + 1}/{len(self.embed_list)}"
 
-    async def on_timeout(self) -> None:
 
-        self.clear_items()
-        self.message.edit(view=self)
+    # To make sure only ctx.author can interact.   
+    async def interaction_check(self, interaction: discord.Interaction) -> bool:
+        if self.user == interaction.user:
+            return True
+        await interaction.response.send_message(f'Only `{self.user.name}` can interact. Run the command if you want to.', ephemeral=True)
+        return False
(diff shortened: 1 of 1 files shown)


examples/views/pagination.py
```diff
@@ -0,0 +1,72 @@
+import discord
+from discord.ext import commands
+
+
+class PaginationView(discord.ui.View):
+    current = 0 # It's not in the constructor because it'll be reset everytime the view is updated.
```
Phxntxm, line 6: No it wouldn't? If it would, so would the `embed_list`

examples/views/pagination.py
```diff
@@ -0,0 +1,72 @@
...
+
+    # To make sure only ctx.author can interact.   
+    async def interaction_check(self, interaction: discord.Interaction) -> bool:
+        if self.user == interaction.user:
+            return True
+        await interaction.response.send_message(f'Only `{self.user.name}` can interact. Run the command if you want to.', ephemeral=True)
+        return False
+        
+    @discord.ui.button(label="Previous", style=discord.ButtonStyle.red, disabled=True) # you can add emojis too..
+    async def previous(self, button, interaction):
```
Phxntxm, line 24: Why are some things type hinted, but some aren't

examples/views/pagination.py
```diff
@@ -0,0 +1,72 @@
...
+        self.current -= 1
+        self.next.disabled = False
```
Phxntxm, line 26: Why is this here? Is this implying that `embed_list` can be modified and next should be updated to reflect that? If so, then there is no handling of to disabling the button later

examples/views/pagination.py
```diff
@@ -0,0 +1,72 @@
...
+        self.current += 1
+        self.previous.disabled = False
+        self.previous.style = discord.ButtonStyle.green
+        self.count.label = f"Page {self.current + 1}/{len(self.embed_list)}"
+        if self.current == len(self.embed_list) -1 :
+            button.disabled = True
+            button.style = discord.ButtonStyle.red
+            await interaction.response.edit_message(embed=self.embed_list[self.current], view=self)
+        else:
+            await interaction.response.edit_message(embed=self.embed_list[self.current], view=self)
```
Phxntxm, line 53: In general, this is nearly the same exact thing as `previous`. This should really be moved into a function that handles updating information

Phxntxm (changes requested):

Prince2347X:
Noted.

Rapptz:
Sorry, this example doesn't match the style and quality of the pre-existing examples.

PR #7256, opened 2021-07-20
[commands] Change: Allow nickname that prefixed with @

sevenc-nanashi:
## Summary

I made MemberConverter to be able to convert nickname that prefixed with `@`.
On Desktop client, mention doesn't be copied correctly(It'll be `@Nickname` as string).

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
I'm not a fan of this behaviour. I think I've rejected it before as well.

PR #7260, opened 2021-07-21
Add reason kwarg to more methods.

NCPlayz:
## Summary

Add reason kwarg to more methods.
Reference: discord/discord-api-docs@ab9e385

Also made bans send a JSON payload rather than use query parameters.
Also does not format `roles` for `include_roles` in prune/estimate_pruned_members as a comma separated string, which seemed to work for me.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
> Also made bans send a JSON payload rather than use query parameters.

Why?

>  Also does not format roles for include_roles in prune/estimate_pruned_members as a comma separated string, which seemed to work for me.

Why?

Issue #2518, opened 2020-01-21
Documentation and examples promote hard coded tokens


iomintz:
The documentation [[1]] [[2]] [[3]] and examples [[4]] use hard coded tokens. Users should not be encouraged to write their token directly in their bot code. `python -m discord` is the only one that makes a separate config.py file. I suggest that the docs use some other method of obtaining the token. My suggestion is simply:

```py
with open('config.txt') as f:
    token = f.read()
```

above any `bot.run` or `client.run` lines, and modify the prose in the quickstart guide [[2]] to match. Another method (one that is simple and built in, of course) might be better.

[1]: https://discordpy.readthedocs.io/en/latest/intro.html
[2]: https://discordpy.readthedocs.io/en/latest/quickstart.html
[3]: https://github.com/Rapptz/discord.py/blob/master/README.rst#quick-example
[4]: https://github.com/Rapptz/discord.py/tree/a0b3e61b51a238c6a67fb2b1a6a94e38a67ccd7c/examples

timothypidashev:
#7272 will fix this

PR #7272, opened 2021-07-23
[docs]Remove any references encouraging users to write their token directly in their bot code. 

timmypidashev:
## Summary

PR for issue: #2518
Removes any references encouraging users to write their token directly in their bot code. 

```python
with open('token.txt') as f:
      token = f.read()

client.run(token) #or bot when in context
```

## Checklist

- [x] This PR fixes an issue.


lgaan:
I don’t think this is really needed.

The examples are there to provide a basic, one-file approach to basic concepts in the library; nowhere in the examples folder is a reference to loading tokens from separate files nor does this file exist leading to potentially more confusion.

Of course my say isn’t final, but this is just what I think.

JelNiSlaw:
I think reading an environment variable would be a better solution

timothypidashev:
I dont exactly think this is needed as well, but im just doing my part to clean up the issues page

Phxntxm:
> I think reading an environment variable would be a better solution

I do not, I've seen far too many times where people come from these service that use `.env` file as a standard to import env variables. They then think that's normal, they have to do nothing and the file is magically loaded which is NOT the case in normal environments. I would prefer something that is an example to do something like this, reading from a file, where it's more direct what is happening.

Using environment variables then means you have to teach the user how to actually *create* environment variables, expanding the scope of this. Putting some text in a file is something you do not have to teach

apple502j:
I'd say we should use `.env` and mention `python-dotenv`.

Phxntxm:
> I'd say we should use `.env` and mention `python-dotenv`.

I VERY STRONGLY disagree with this, due to the first part I mentioned in my first reply. I have seen way too many people be bit by this from heroku/replit to support this

lgaan:
> I'd say we should use `.env` and mention `python-dotenv`.

Back to my previous point, examples are meant to not rely on external packages.

timothypidashev:
> Using environment variables then means you have to teach the user how to actually _create_ environment variables, expanding the scope of this. Putting some text in a file is something you do not have to teach

I agree with @Phxntxm. Enlarging the scope of this is the worst path to take in this situation. We simply don't want to promote hard-coding tokens into files that will most likely make it onto a repository later

examples/views/link.py
```diff
@@ -36,4 +38,4 @@ async def google(ctx: commands.Context, *, query: str):
     await ctx.send(f'Google Result for: `{query}`', view=Google(query))
 
 
-bot.run('token')
+bot.run(token)
```
ioistired, line 41: mising newline at EOF

examples/views/persistent.py
```diff
@@ -60,4 +62,4 @@ async def prepare(ctx: commands.Context):
     await ctx.send("What's your favourite colour?", view=PersistentView())
 
 
-bot.run('token')
+bot.run(token)
```
ioistired, line 65: missing a newline at EOF here too

ioistired (changes requested):

examples/views/link.py
timothypidashev: will push in a sec

examples/views/persistent.py
timothypidashev: done

ImportTaste:
If this isn't approved, at the very least, there should be a comment in the example files warning users never to share any file that has their token inside of it right above the line where they put their token in.

timothypidashev:
> If this isn't approved, at the very least, there should be a comment in the example files warning users never to share any file that has their token inside of it right above the line where they put their token in.

agree

Hyperclaw79:
Discord itself warns us while creating the token....I personally don't think there's a need to again tell people to not hard-code the token. It's something that should be pretty obvious....

oliver-ni:
> Discord itself warns us while creating the token....I personally don't think there's a need to again tell people to not hard-code the token. It's something that should be pretty obvious....

They don't, actually.

Hyperclaw79:
> They don't, actually.

oh....you're right. For some reason, I remember seeing a warning when we click on Copy/Reveal. Guess I confused it with some other app.
Either way, that warning should be provided by Discord itself, rather than a library.

timothypidashev:
> Either way, that warning should be provided by Discord itself, rather than a library.
I agree Discord itself should provide a warning, but discord.py is not out of the picture either, and implying hard-coding tokens all throughout the docs and examples is not going to benefit anyone, in my humble opinion.

timothypidashev:
The only inconsistency I can think of atm for this pr is the use of both `bot` and `client` when setting up the bot class. Should I make changes(switch all to client or bot), or keep as is?

ioistired:
It's consistent with the fact that all instances of `commands.Bot` are called `bot` and all instances of `discord.Client` are called `client`, so keep it as is.

timothypidashev:
> It's consistent with the fact that all instances of `commands.Bot` are called `bot` and all instances of `discord.Client` are called `client`, so keep it as is.

That is a good point, I will keep as is, thank you for pointing this out!

timmypidashev:
Closing this pull request as it is stale and not relevant to current development.

Issue #7231, opened 2021-07-14
Components get moved randomly in a full Action Row on message/view update


Luc1412:
### Summary

The 3 Link buttons get moved into the Select ActionRow which causes an error

### Reproduction Steps

1. Use the following code below and start the view
2. select the 2nd select item multiple times
3. An error will randomly? occur

### Minimal Reproducible Code

```python
class TestView(discord.ui.View):

    def __init__(self):
        super().__init__()
        help_select = discord.ui.Select(options=[
            discord.SelectOption(label='Home', value='home', emoji='🏠', default=True),
            discord.SelectOption(label='Name', value='stats', emoji='📊'),
        ])
        help_select.callback = self.select_page
        self.add_item(help_select)

        self.add_item(discord.ui.Button(style=discord.ButtonStyle.url, label='Example 1', url=f'https://youtube.com'))
        self.add_item(discord.ui.Button(style=discord.ButtonStyle.url, label='Example 2', url='https://discord.com'))
        self.add_item(discord.ui.Button(style=discord.ButtonStyle.url, label='Example 3', url='https://google.com'))

    async def select_page(self, interaction: discord.Interaction):
        select: discord.ui.Select = discord.utils.get(self.children, custom_id=interaction.data['custom_id'])
        if not select.values:
            return
        await interaction.message.edit(view=self)
```

### Expected Results

The components just normally update.

`to_components()` looks like:
[{'type': 1, 'components': [{'type': 3, 'custom_id': 'f0a5a63f62e6b58e179eadba3014652b', 'min_values': 1, 'max_values': 1, 'options': [{'label': 'Home', 'value': 'home', 'default': True, 'emoji': {'name': '🏠'}}, {'label': 'Name', 'value': 'stats', 'default': False, 'emoji': {'name': '📊'}}]}]}, {'type': 1, 'components': [{'type': 2, 'style': 5, 'label': 'Example 1', 'disabled': False, 'url': 'https://youtube.com'}, {'type': 2, 'style': 5, 'label': 'Example 2', 'disabled': False, 'url': 'https://discord.com'}, {'type': 2, 'style': 5, 'label': 'Example 3', 'disabled': False, 'url': 'https://google.com'}]}]

### Actual Results

An error occurrs:
```py
[...]

PR #7285, opened 2021-07-27
Add support for discord.ui.Select.disabled

Luc1412:
Fix typo in the select decorator TypeError

Took 16 minutes

## Summary

- Add support for discord.ui.Select.disabled
- Fixed typo in discord.ui.select decorator TypeError
- ~~Solved #7231~~
- Add missing `super().__init__()` for `discord.ui.Select` 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ui/select.py
+++ b/discord/ui/select.py
@@ -93,7 +93,7 @@ class Select(Item[V]):
         'min_values',
         'max_values',
         'options',
-        'disabled'
+        'disabled',
     )
 
     def __init__(
--- a/discord/ui/select.py
+++ b/discord/ui/select.py
@@ -275,14 +275,14 @@ class Select(Item[V]):
         self._selected_values = data.get('values', [])
 
     @classmethod
-    def from_component(cls: Type[S], select: SelectMenu) -> S:
+    def from_component(cls: Type[S], component: SelectMenu) -> S:
         return cls(
-            custom_id=select.custom_id,
-            placeholder=select.placeholder,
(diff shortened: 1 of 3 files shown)


discord/ui/select.py
```diff
@@ -91,6 +93,7 @@ class Select(Item[V]):
         'min_values',
         'max_values',
         'options',
+        'disabled'
```
NCPlayz, line 96:
```suggestion
        'disabled',
```

Luc1412: solved

discord/ui/select.py
```diff
@@ -260,13 +275,14 @@ def refresh_state(self, interaction: Interaction) -> None:
         self._selected_values = data.get('values', [])
 
     @classmethod
-    def from_component(cls: Type[S], component: SelectMenu) -> S:
+    def from_component(cls: Type[S], select: SelectMenu) -> S:
```
Rapptz, line 278: No.

A parameter name is part of the contract. This doesn't really matter right now and this is an internal function but you shouldn't just rename parameters.

Rapptz (changes requested):

discord/ui/select.py
Luc1412: Reverted it. My intention was to make it more consistent. The `discord.ui.Button.from_component` function also uses `button` (not `component `) as parameter name.

Luc1412:
After further investigation it seems like #7231 still occurr.

Issue #7284, opened 2021-07-26
[discordpy v2] In discord.Select, options that are defined as "default" are not present in Select.values If the user does not change the selected options


AiroPi:
### Summary

When you create a UI.Select, with default values, and try to retrieve the selected values with Select.values, the list is empty.

### Reproduction Steps

When creating a Select list with a validation button : 
```py
class CustomView(ui.View):
    def __init__(self, bot: MyBot) -> None:
        super().__init__()
        self.bot = bot

    @ui.select(options=[discord.SelectOption(label='Français', value='fr', emoji='🇫🇷', default=True),
                        discord.SelectOption(label='English', value='en', emoji='🇺🇸', default=True)],
               max_values=2)
    async def callback(self, *_) -> None: ...

    @ui.button(label="Validate", row=1)
    async def validate(self, button, inter: discord.Interaction) -> None:
        print(self.children[0].values)  # Show [], but should show ['fr', 'en'] because they are selected by default
```

### Expected Results

`self.children[0].values` should show the default selected values.

### Actual Results

`self.children[0].values` show an empty list.

### Intents

all

### System Information

- Python v3.9.6-final
- discord.py v2.0.0-alpha
    - discord.py pkg_resources: v2.0.0a3391+gf14e5843
- aiohttp v3.7.4.post0
- system info: Windows 10 10.0.19041

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.

AkshuAgarwal:
Something similar for me too.
The selected values is an empty list during initialization and if I'm not wrong it don't get updated if we don't edit the Select and just try to get those values.
So maybe, we can just manually append default values into the list.

![image](https://user-images.githubusercontent.com/83153460/127195179-3b683974-08fe-4c61-b638-58aac0ec6b50.png)

I also made a PR (#7286) which fixed the issue for me.

adshivsat:
In Discord, Select Eject Parameter's and Switch o Off load drivers in order to initiate the R187 Protocol

PR #7286, opened 2021-07-27
Fix default options not present in Select.values

AkshuAgarwal:
## Summary

This fixes the issue #7284 
Default values are not present in `_selected_values` during initialisation, and only gets updated after state gets refreshed.
So we can put those (default) values manually.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ui/select.py
+++ b/discord/ui/select.py
@@ -103,10 +103,10 @@ class Select(Item[V]):
         options: List[SelectOption] = MISSING,
         row: Optional[int] = None,
     ) -> None:
-        self._selected_values: List[str] = []
         self._provided_custom_id = custom_id is not MISSING
         custom_id = os.urandom(16).hex() if custom_id is MISSING else custom_id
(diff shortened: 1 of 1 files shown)


Rapptz (changes requested):
I don't know how I feel about this.

I think this needs elaboration from the developers before I commit to a specific fix.

Also there are merge conflicts.

AkshuAgarwal:
I think like this is a Discord Limitation since API does not send initial values set during sending Message (as per the [Docs](https://discord.com/developers/docs/interactions/message-components#select-menus) say) -
image

And the library consider those values as None by default. But it gets fixed when user interacts with Dropdown at least once.

I don't know about the Discord's side but in library, instead of marking them blank by default, adding the `default` values into `_selected_values` during initialization returns correct values to the user. If the user make changes to Dropdown, everything gets handled as they're right now.

Also resolved the merge conflicts. 😄

PR #7289, opened 2021-07-28
Do not try to check length of None embeds

Galarzaa90:
## Summary

When passing `None` in the `embeds` parameter of `Webhook.send`, a `TypeError` is raised due to attempting to check the length of `None`.

Unlike other parameters, like `content` and `embed` and `view`, `embeds` doesn't consider `None` as an empty value.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/webhook/async_.py
+++ b/discord/webhook/async_.py
@@ -438,9 +438,12 @@ def handle_message_parameters(
 
     payload = {}
     if embeds is not MISSING:
-        if len(embeds) > 10:
+        if embed is None:
+            payload['embeds'] = []
+        elif len(embeds) > 10:
             raise InvalidArgument('embeds has a maximum of 10 elements.')
-        payload['embeds'] = [e.to_dict() for e in embeds]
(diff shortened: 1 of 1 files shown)


Rapptz:
`None` isn't a valid type for `embeds`.

Galarzaa90:
Why is the signature (and validations) of `Webhook.send` so different to `Messageable.send`? It is counterintuitive that one accepts None values fine, while the other doesn't. Or is `Message.send` yet to be updated?

Rapptz:
I haven't gotten around to changing `Messageable.send` yet.

Galarzaa90:
And is there a reason why the other arguments (view, content, embed) handle `None` values even if they have the same signature (`... = MISSING`)

Rapptz:
A coincidence from handling message updates (which allows `None`) and message creation in the same code. Ultimately the typed and documented signature is the source of truth for what is accepted from a consumer's perspective.

PR #7297, opened 2021-07-29
Add persistent view in a task instead of on_ready

SYCKGit:
## Summary
There's no need to use on_ready for adding a persistent view, a startup task can be created instead.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
This example's state is currently temporary since it's going to be written when async init is required.

PR #7300, opened 2021-07-30
Switch to API v9 for thread support.

Artemis21:
## Summary

Switch to API v9.

According to [the docs](https://discord.com/developers/docs/topics/threads#backwards-compatibility)

> Threads are only available in API v9. Bots that do not update to API v9 will not receive most gateway events for threads, or things that happen in threads (such as Message Create).

This makes thread support seem useless without updating to API v9.

## Checklist

- [x] This PR fixes an issue.


Rapptz:
The gateway is already at v9

PR #7312, opened 2021-07-31
Add FormattedDatetimeConverter

Daggy1234:
## Summary

Add datetime converter. This is a draft, hoping for fixes. Uses regex to match discord style timestamps and returns a datetime as well as a flag if present. Draft because at the moment discord has no way of letting regular users send formatted timestamps.

## Checklist

- [ ] This PR fixes an issue.


Daggy1234:
Extremely sorry for the lack of a description. Clicked on the create button by mistake while adding the description.

JelNiSlaw:
I think `re.fullmatch` should be used instead of `re.match`. That would make `$` unnecessary.
Also the rest of the library's code uses `[0-9]` instead of `\d` (I'm not sure why).

discord/ext/commands/converter.py
```diff
@@ -822,6 +824,27 @@ async def convert(self, ctx: Context, argument: str) -> discord.PartialEmoji:
...
+        Raise :exc:`.FormattedDatetimeConversionFailure` instead of generic :exc:`.BadArgument`
+     """
+
+    async def convert(self, ctx: Context, argument: str) -> Tuple[datetime.datetime, Optional[str]]:
+        match = re.match(r'<t:(-?\d+)(:[tTdDfFR])?>$', argument)
+
+        if match:
+            try:
+                return datetime.datetime.fromtimestamp(int(match.group(1))), match.group(2)
+            except:
```
JelNiSlaw, line 844: You shouldn't be using a bare except

apple502j:
> Also the rest of the library's code uses `[0-9]` instead of `\d` (I'm not sure why).

Because:
```py
>>> match(r"^\d+$","٠1１")
<re.Match object; span=(0, 3), match='٠1１'>
```

NCPlayz (changes requested):
I feel like this would be better if it was a utility function rather than a converter. Something along the lines of `resolve_formatted_timestamp` maybe. It doesn't seem right for the programmer to need to access the converter directly to use as the annotation, and it seems unlikely that a user would be sending a formatted timestamp (over natural language) to specify a date/time argument for a command anyway. 

Just my two cents!

Daggy1234:
So I added the util, but I still kept the converter. The converter also uses the util, and my basic reasoning for retaining the converter was that  eventually discord would add a way for users to send these. Also moved the regex to use [0-9] and added a time zone to the datetime.

discord/utils.py
```diff
@@ -1012,3 +1012,29 @@ def format_dt(dt: datetime.datetime, /, style: Optional[TimestampStyle] = None)
...
+    Returns
+    --------
+    Optional[Tuple[:class:`datetime.datetime`, Optional[:class:`str`]]
+        A tuple containing the datetime and the flag if present. Otherwise None.
+    """
+    match = re.match(r'<t:(-?[0-9]+)(:[tTdDfFR])?>$', formatted_dt)
+    if match:
+        try:
+            return datetime.datetime.fromtimestamp(int(match.group(1)), tz=datetime.timezone.utc), match.group(2)
+        except:
```
Gobot1234, line 1038: You're using a bare except again

discord/utils.py
```diff
@@ -1012,3 +1012,29 @@ def format_dt(dt: datetime.datetime, /, style: Optional[TimestampStyle] = None)
...
+            return
+    return
```
Gobot1234, line 1040: This is unnecessary

Daggy1234:
In terms of exceptions, for the `int` cast a ValueError, datetime.datetime.fromtimestamp can either raise an OverflowError or a OSError. Both are handled as well. https://docs.python.org/3/library/datetime.html#datetime.date.fromtimestamp

JelNiSlaw:
Is it possible for `int` to raise `ValueError` here?

Daggy1234:
> Is it possible for `int` to raise `ValueError` here?

if the regex works probably not. The `int` is for pyright that was complaining a lot.

Rapptz:
I'm personally not sure I get the utility of this one. Users aren't necessarily aware of this format at the moment and having to type it manually seems like a chore.

Daggy1234:
Yeah! hence, the utility. Developers can use the utility to parse the timestamps from content bots may send, like editing a timestamp field in an embed. The converter is auxiliary, I can totally remove it, but if discord were to add the ability to send timestamps it exists.

PR #7323, opened 2021-08-01
Fix webhook attribute error on send file=None

Brettanda:
## Summary

When sending an async webhook message, setting the file attribute to None an attribute error would occur. This fixes that.

## Checklist

- [ ] This PR fixes an issue.


Brettanda:
Never mind, I misunderstood the missing utility and ellipsis literal

PR #7324, opened 2021-08-02
Fix CustomView(View)().from_message()

ahmadelafify:
## Summary

When creating a CustomView object, CustomView.from_message() returns raw View rather than a CustomView

## Checklist

- [x] This PR fixes an issue.


bijij:
from_message returning a `View` class is intentional.

ahmadelafify:
Yeah never mind, read the documentation. Says "This always returns a :class:`View` and not one of its subclasses", I thought it was a bug.

PR #7331, opened 2021-08-03
Adds Shorthand methods for Locking & Archiving threads

Hyperclaw79:
## Summary

Adds `thread.lock()` and `thread.archive()` which serve as quick aliases for `thread.edit` with archive/locked=True.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/threads.py
+++ b/discord/threads.py
@@ -560,6 +560,38 @@ class Thread(Messageable, Hashable):
 
         await self._state.http.edit_channel(self.id, **payload)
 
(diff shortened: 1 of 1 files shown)


Rapptz:
I'm against this because it requires 4 verbs for what is essentially 2 bools. `archive`, `unarchive`, `lock`, `unlock`. Better to just re-route everything to `edit`.

Hyperclaw79:
Ah fair point!

PR #7335, opened 2021-08-04
Added colors

Middledot:
## Summary

Add 2 colors to colours.py because I thought it might be useful

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/colour.py
+++ b/discord/colour.py
@@ -312,22 +312,16 @@ class Colour:
     @classmethod
     def embed(cls: Type[CT]) -> CT:
         """A factory method that returns a :class:`Colour` with a value of ``0x2F3136``.
+        This color will match the embed, making it appear with round corners.
 
         .. versionadded:: 2.0
         """
         return cls(0x2F3136)
     
     @classmethod
-    def dark_channel(cls: Type[CT]) -> CT:
-        """A factory method that returns a :class:`Colour` with a value of ``0x36393E``.
-
-        .. versionadded:: 2.0
-        """
-        return cls(0x36393E)
-    
-    @classmethod
-    def light_channel(cls: Type[CT]) -> CT:
+    def light_theme(cls: Type[CT]) -> CT:
         """A factory method that returns a :class:`Colour` with a value of ``0xFFFFFF``.
+        This will appear transparent on Discord's light theme.
 
         .. versionadded:: 2.0
         """
--- a/discord/colour.py
+++ b/discord/colour.py
@@ -310,6 +310,15 @@ class Colour:
         return cls(0xFEE75C)
     
     @classmethod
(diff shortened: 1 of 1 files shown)


itzsimpleboii8282:
What are these colors? Are they somehow related to those colors in discord or in standard colour section.

Middledot:
> What are these colors? Are they somehow related to those colors in discord or in standard colour section.

They're both discord colors, one to match the embed color, and the other for light mode as there's one for dark mode

JelNiSlaw:
discord.py only has colors from Discord's color picker and branding. You can easily use your own colors if you need them.

Middledot:
> discord.py only has colors from Discord's color picker and branding. You can easily use your own colors if you need them.

These are colors from discord's themes.
There's the white mode color and the color of the embed

JelNiSlaw:
A better name would be just `white` instead of `light_theme`. What's the use case of an embed color? Because "invisible embed borders" are not an official feature, and if you really want them you can make use the color value yourself.

Middledot:
> A better name would be just `white` instead of `light_theme`. What's the use case of an embed color? Because "invisible embed borders" are not an official feature, and if you really want them you can make the color value yourself.

The embed color could be an alternative to putting the hex value for a full one color embed.

ahts97:
You checked the checkbox stating that you have updated the documentation to reflect the changes - you did not do that. Please pay attention to what are you doing. We can now doubt your code has been tested if you did not even read what have you checked.

Either way, there is absolute no reason to set the theme colours as presets, some of the arguments being:
- These colours are not transparent - some people use the light theme, some use the dark theme - this change is simply misleading;
- They are not official branding colours, meaning it's not supported by Discord and is a subject to change without prior announcements;
- The colours might vary on different platforms - small colour shifts make an unpleasant visual experience.

You said it's often used - myself, being on Discord from around 2017, I have never seen a single bot use it. You aren't everyone.

You can store these colours for yourself in an easily accessible place, but not in the actual library code.

Middledot:
> There is absolute no reason to set the theme colours as presets, some of the arguments being:
> 
> * These colours are not transparent / embed matching - some people use the light theme, some use the dark theme - this change is simply misleading;
> * They are not official branding colours, meaning it's not supported by Discord and is a subject to change without prior announcements;
> * The colours might vary on different platforms - small colour shifts make an unpleasant visual experience.
> 
> You said it's often used - myself, being on Discord for around 4 years, I have never seen a single bot use it. You aren't everyone.
> 
> You can store these colours for yourself in an easily accessible place, but not in the actual library code.

Since the docs are autogenerated, I assumed the changing the documentation meant adding or editing the noted under the function, so my bad.

Even if users have different modes, there is already a dark theme color, so I thought there should be the opposite themed colors?

PR #7348, opened 2021-08-08
added a shortcut to `reset_cooldown` and some type hinting

Nirlep5252:
## Summary

Added a shortcut for `Command.reset_cooldown` in `Context` so instead of using `ctx.command.reset_cooldown(ctx)` I can just use `ctx.reset_cooldown()` 👍 
And added some type hinting in the `Context` class so users can get better autocomplete in IDEs 👀 

## Checklist

- [ ] This PR fixes an issue.


JelNiSlaw:
I'm not sure what's the advantage of this shortcut. You can subclass `Context` and add this method yourself without modifying the library.

JDJGInc:
I would actually use this method, @JelNiSlaw.

scrazzz:
`ctx.reset_cooldown` sounds vague in my opinion.

Rapptz:
I'm not really a fan of this shortcut. Sorry.

PR #7373, opened 2021-08-12
Use channel's default_auto_archive_duration for new threads

jgayfer:
## Summary

Currently creating a new thread uses a default auto archive duration of
1440 minutes, or 1 day.

Rather than prescribing our own default, we can use the default auto
archive duration that is set on the channel. This ensures that newly
created threads will respect the default auto archive duration as
prescribed by the user.

My initial approach was to simply omit the `auto_archive_duration` from
the request, but when doing so, the thread is always created with an
auto archive duration of 1440 minutes, regardless of the channel's
default setting.

See the following PR for more discussion on the matter:

  https://github.com/discord/discord-api-docs/pull/3567

It's unclear if this is intended behaviour or not, but we might as well
work with it for the time being

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -662,7 +662,7 @@ class TextChannel(discord.abc.Messageable, discord.abc.GuildChannel, Hashable):
         *,
         name: str,
         message: Optional[Snowflake] = None,
-        auto_archive_duration: ThreadArchiveDuration = 1440,
+        auto_archive_duration: ThreadArchiveDuration = MISSING,
         type: Optional[ChannelType] = None,
         reason: Optional[str] = None,
     ) -> Thread:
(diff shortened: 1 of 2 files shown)


Rapptz (commented):
The library generally uses `MISSING` instead of `Optional[X]` for type checking purposes and because `None` is overloaded in the API to mean "not set". I think for here it looks superficially okay, except this is no longer symmetrical with `edit` (i.e. `Thread.edit(auto_archive_duration=None)` doesn't work. Note that I'm not saying to make it work, I'm just giving extra rationale why `MISSING` is the better option from a library design perspective.

jgayfer:
@Rapptz Thanks for the feedback! I wasn't aware of `MISSING`, but it makes sense to differentiate between values that aren't provided versus those that are intentionally blank.

I changed the default values from `None` to `MISSING`, and updated the type hints accordingly.

Rapptz (approved):
LGTM.

PR #7374, opened 2021-08-12
Adding `on_check_failure` to `View` 

Kile:
## Summary

This pull request adds the Coro `on_check_failure` to the `View` class which is called when `interaction_check` returns `False` and defers the interaction, if not previously responded to, per default.

Why this is needed:
`interaction_check`'s purpose is to check if the interaction should go through, not really to handle the case of it not going through. A separate method handling what to do when the check fails seems more clean and useful to deal with that.
So thing code which would previously look like
```py
async def interaction_check(self, interaction: discord.Interaction) -> bool:
    if not some_check():
        await interaction.response.defer()
        return False
    return True
```
looks now like
```py
async def on_check_failure(self, interaction: discord.Interaction) -> None:
    await interaction.response.defer()

async def interaction_check(self, interaction: discord.Interaction) -> bool:
    return some_check():
```

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ui/view.py
+++ b/discord/ui/view.py
@@ -301,7 +301,7 @@ class View:
         This is useful to override if, for example, you want to ensure that the
         interaction author is a given user.
 
-        When this returns ``False`` :meth:`InteractionResponse.defer` is automatically called
+        When this returns ``False`` :meth:`InteractionResponse.defer` is automatically called if the interaction hasn't been responded to
 
         The default implementation of this returns ``True``.
 
@@ -356,7 +356,8 @@ class View:
 
             allow = await self.interaction_check(interaction)
             if not allow:
-                await interaction.response.defer()
+                if not interaction.response.is_done():
+                    await interaction.response.defer()
                 return
 
             await item.callback(interaction)
--- a/discord/ui/view.py
+++ b/discord/ui/view.py
@@ -354,6 +354,7 @@ class View:
 
             allow = await self.interaction_check(interaction)
             if not allow:
+                await interaction.response.defer()
                 return
 
             await item.callback(interaction)
--- a/discord/ui/view.py
+++ b/discord/ui/view.py
@@ -301,6 +301,8 @@ class View:
         This is useful to override if, for example, you want to ensure that the
         interaction author is a given user.
 
+        When this returns ``False`` :meth:`InteractionResponse.defer` is automatically called
+
         The default implementation of this returns ``True``.
 
         .. note::
--- a/discord/ui/view.py
+++ b/discord/ui/view.py
@@ -343,7 +343,7 @@ class View:
         interaction: :class:`~discord.Interaction`
             The interaction that occurred.
         """
-        if not interaction.response._responded:
+        if not interaction.response.is_done():
             await interaction.response.defer()
 
     async def on_error(self, error: Exception, item: Item, interaction: Interaction) -> None:
--- a/discord/ui/view.py
+++ b/discord/ui/view.py
@@ -301,14 +301,14 @@ class View:
         This is useful to override if, for example, you want to ensure that the
         interaction author is a given user.
 
-        When this returns ``False`` :meth:`InteractionResponse.defer` is automatically called if the interaction hasn't been responded to
-
         The default implementation of this returns ``True``.
 
         .. note::
 
             If an exception occurs within the body then the check
             is considered a failure and :meth:`on_error` is called.
+
+            If this returns ``False`` :meth:`on_check_failure` is called
 
         Parameters
         -----------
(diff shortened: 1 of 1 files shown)


jgayfer (commented):
What happens if we don't want to defer the interaction, such as send a message indicating that the user isn't permitted to perform the action from within the check?

Calling `defer` will [raise an exception](https://github.com/Rapptz/discord.py/blob/1e17b7fceaf7173b1666465e639d1bebaa126683/discord/interactions.py#L398-L399) if we've already responded to the interaction.

Kile:
Is there a property that allows me to check if there was already a response to the interaction?

Phxntxm:
I don't really like this idea, for the exact reason that you said:
> Since this is a free API call, it doesn't harm anything, just avoids the error message.

It's a free API call, there isn't a rate limit.... so the normal thing to do here is to actually make a response with something like `"You can't use this"` or whatever.

Kile:
If there is a way to check if there was already a response I am happy to check for that and only defer if there wasn't

nickofolas:
I feel that this is a functionality that, if desired, you should just implement yourself. Abstracting the control away within the library is unnecessary.

Kile:
Updated code, description and documentation reflecting that it only defers when the interaction hasn't been responded to before

Kile:
> I feel that this is a functionality that, if desired, you should just implement yourself. Abstracting the control away within the library is unnecessary.

as shown in my example, it is quite ugly to do myself and there is no drawback to implement this (with the change that this is only called when there was no previous response)

nickofolas:
> there is no drawback to implement this

You make the assumption here that every developer wishes to handle a check failure in the same way you do.

> it is quite ugly to do myself

I don't think aesthetics is the best reason to sacrifice control over how something like this is handled.

Kile:
> You make the assumption here that every developer wishes to handle a check failure in the same way you do.

this check failure can only be handled other than by defining it by catching it error in which case the interaction also wouldn't be defered. So no developer would really care about this deferering as it doesn't influence their code at all, it merely makes sure no error message is shown to the user if the interaction wasn't responded to

jgayfer:
I understand where you're coming from here.

The `interaction_check` method name doesn't do a great job communicating that the interaction must also be handled as part of the callback to avoid an error. Other checks in d.py need only return `True` or `False`.

I'm curious how this might look if we defined a `on_check_failure` method on the view. We _could_ have it default to deferring the view, or we could instead leave it empty, allowing a developer to subclass `View` and define their own defaults for how failed checks are handled.

Kile:
Yeah that sounds reasonable

Kile:
> I understand where you're coming from here.
> 
> The `interaction_check` method name doesn't do a great job communicating that the interaction must also be handled as part of the callback to avoid an error. Other checks in d.py need only return `True` or `False`.
> 
> I'm curious how this might look if we defined a `on_check_failure` method on the view. We _could_ have it default to deferring the view, or we could instead leave it empty, allowing a developer to subclass `View` and define their own defaults for how failed checks are handled.

I made an attempt on changing my pr to that

Kile:
I currently still `defer` it per default in `on_check_failure` but I wouldn't mind leaving it empty for this implementation

discord/ui/view.py
```diff
@@ -327,6 +329,23 @@ async def on_timeout(self) -> None:
...
+        This is good to overwrite when you want a custom error message
+
+        Per default this calls :meth:`InteractionResponse.defer` if the interaction has not been responded to before
+
+        Parameters
+        -----------
+        interaction: :class:`~discord.Interaction`
+            The interaction that occurred.
+        """
+        if not interaction.response._responded:
```
jgayfer, line 346: Probably better to use the public method.
```suggestion
        if not interaction.response.is_done():
```

Kile:
Just used in `on_error` like that so I used it like that too

jgayfer:
I don't mind this change.

Ultimately not up to me, but it's nice to have a sane default for those times where a failed check doesn't warrant any further action.

Daggy1234 (commented):
Just my two cents here, but I personally feel this goes against the very principle of using classes.

The point of OOP is you can subclass your view, and implement whatever in your `on_check_failure_func` , if you'd like you can defer. Forcing a non-sane default on the pretence that everyone would prefer it that way is not the best idea. 

Also, your argument falls flat, when you talk about ugliness. How is two lines of code in your subclassed on_check_failure is ugly? 

You also fail to address rate limits, as Sarc had mentioned.

My final thought would be that I like the error in the UI, the user gets a clear message that shows them that the interaction failed. Letting the user get a direct and clear signal from the UI itself is superior to having to send messages manually.

tl;dr just subclass and write your own, don't force defaults.

jgayfer:
@Daggy1234 I agree that forcing the default might not be great, but currently there isn't a notion of an `on_check_failure` - it could be handy to have that function in play so that we can subclass and override.

Currently the choices are to defer inside `interaction_check`, or to override a private method `_schedule_task`, which is never fun.

pikaninja:
This doesn't mean a default of defer is good, I personally do not want this, if you want something specific like this, you can do it yourself in a subclass.

Kile:
As I said, I'm fine with removing the default defering and instead leave that function empty if you'd prefer that

Rapptz:
The failure is by design. You return False, it means the interaction wasn't for you. You can send a response within the body if you want to set an ephemeral message. 

I don't get the point of having another function call that is devoid of context for the actual message sending aspect. The reason why this flow works with the command framework is because you can also raise an exception that gives you the context that is being lost here. Such mechanism wouldn't make sense here because it's much simpler, there's only a single call site to care about. Checks in the commands framework can come from many other places (global, cog, local functions) so the exception and the error handler are necessary. Again, not the case here where everything is entirely self-contained.

PR #7377, opened 2021-08-13
Typehinted raw_models.py

StockerMC:
## Summary
This PR typehints `discord/raw_models.py` and adds `discord/types/raw_models.py`

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/raw_models.py
+++ b/discord/raw_models.py
@@ -22,6 +22,25 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 DEALINGS IN THE SOFTWARE.
 """
 
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Optional, Set
+
+if TYPE_CHECKING:
+    from .types.raw_models import (
+        MessageDeleteEvent,
+        BulkMessageDeleteEvent,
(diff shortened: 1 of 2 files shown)


discord/raw_models.py
```diff
@@ -158,18 +177,18 @@ class RawReactionActionEvent(_RawReprMixin):
...
 
-    def __init__(self, data, emoji, event_type):
-        self.message_id = int(data['message_id'])
-        self.channel_id = int(data['channel_id'])
-        self.user_id = int(data['user_id'])
-        self.emoji = emoji
-        self.event_type = event_type
-        self.member = None
+    def __init__(self, data: ReactionActionEvent, emoji: PartialEmoji, event_type: str) -> None:
+        self.message_id: int = int(data['message_id'])
```
bryanforbes, line 181: The only properties that need a explicit types are `self.member` and `self.guild_id` (and the latter only needs a type hint on the first occurrence). The rest can be inferred.

bryanforbes (changes requested):
I made a comment about inferred types. The rest of the classes should be changed to use inferred types as well.

Rapptz (approved):
This is an okay first step. Ideally `types/raw_models.py` would really just be raw gateway event data but this can be done later.

PR #7378, opened 2021-08-13
Adding overwrite parameter in add_cog

Rishiraj0100:
## Checklist

- [ ] This PR fixes an issue.


Rapptz:
I see no point in this addition. This causes bad state management.

PR #7383, opened 2021-08-13
Added missing typehints to discord/template.py

StockerMC:
## Summary
Added missing typehints to discord/template.py

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/template.py
+++ b/discord/template.py
@@ -24,7 +24,7 @@ DEALINGS IN THE SOFTWARE.
 
 from __future__ import annotations
 
-from typing import Any, Optional, TYPE_CHECKING, cast, NoReturn, List
+from typing import Any, Optional, TYPE_CHECKING, cast, List
 from .utils import parse_time, _get_as_snowflake, _bytes_to_base64_data, MISSING
 from .enums import VoiceRegion
 from .guild import Guild
@@ -37,8 +37,6 @@ if TYPE_CHECKING:
     import datetime
     from .types.template import Template as TemplatePayload
     from .types.emoji import Emoji
-    from .state import ConnectionState
-    from .client import ClientUser
     from .flags import MemberCacheFlags
     from .user import User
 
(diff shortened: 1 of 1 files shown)


discord/template.py
```diff
@@ -150,18 +154,18 @@ def _store(self, data: TemplatePayload):
             source_serialised = data['serialized_source_guild']
             source_serialised['id'] = id
             state = _PartialTemplateState(state=self._state)
-            guild = Guild(data=source_serialised, state=state)  # type: ignore
+            guild = Guild(data=source_serialised, state=state)  # type: ignore - Guild expects a ConnectionState, we're passing a _PartialTemplateState
```
bryanforbes, line 157: Should `Guild` be updated to take a protocol that both `ConnectionState` and `_PartialTemplateState` meet?

discord/template.py
```diff
@@ -127,20 +131,20 @@ class Template:
         '_state',
     )
 
-    def __init__(self, *, state, data: TemplatePayload):
+    def __init__(self, *, state, data: TemplatePayload) -> None:
```
bryanforbes, line 134: Should `state` be typed?

StockerMC: Done.

discord/template.py
```diff
@@ -150,18 +153,18 @@ def _store(self, data: TemplatePayload):
             source_serialised = data['serialized_source_guild']
             source_serialised['id'] = id
             state = _PartialTemplateState(state=self._state)
-            guild = Guild(data=source_serialised, state=state)  # type: ignore
+            guild = Guild(data=source_serialised, state=state)  # type: ignore - Guild expects a ConnectionState, we're passing a _PartialTemplateState
 
-        self.source_guild = guild
-        self.is_dirty = data.get('is_dirty', None)
+        self.source_guild: Guild = guild # type: ignore - id is a required field on the discord API docs
```
Rapptz, line 158: I'm not sure I follow why an assignment needs to be `# type: ignore`'d.

discord/template.py
Rapptz: That's an open question worth considering down the line but a lot of models take in fake connection states and it's more trouble than it's worth compared to just lying like is done here.

discord/template.py
StockerMC: `self._state._get_guild(id)` returns `Unknown | None`. Instead of a `# type: ignore`, `and id` from `if guild is None and id` could also be removed because `guild` is set as a `Guild` instance in the if block.

PR #7385, opened 2021-08-13
Type-hint user.py

thetimtoy:
- [ ] This PR fixes an issue.


discord/user.py
```diff
@@ -51,42 +68,42 @@ class BaseUser(_UserTag):
         bot: bool
         system: bool
 
-    def __init__(self, *, state, data):
-        self._state = state
+    def __init__(self, *, state: ConnectionState, data: UserPayload) -> None:
+        self._state: ConnectionState = state
```
bryanforbes, line 72: I believe `_state` is inferred since `state` is typed, so typing `_state` is not needed

discord/user.py
```diff
@@ -51,42 +68,42 @@ class BaseUser(_UserTag):
...
-        self.id = int(data['id'])
-        self.discriminator = data['discriminator']
-        self._avatar = data['avatar']
-        self._banner = data.get('banner', None)
-        self._accent_colour = data.get('accent_color', None)
-        self._public_flags = data.get('public_flags', 0)
-        self.bot = data.get('bot', False)
-        self.system = data.get('system', False)
+    def _update(self, data: UserPayload) -> None:
+        self.name: str = data['username']
```
bryanforbes, line 94: Either the `if TYPE_CHECKING` block above should be removed, or the types covered in that block should be removed here.

discord/user.py
```diff
@@ -391,17 +413,17 @@ def __del__(self) -> None:
...
-        self = super()._copy(user)
+    def _copy(cls: Type[U], user: U) -> U:
+        self: U = super()._copy(user)
         self._stored = False
         return self
 
-    async def _get_channel(self):
-        ch = await self.create_dm()
+    async def _get_channel(self) -> DMChannel:
+        ch: DMChannel = await self.create_dm()
```
bryanforbes, line 422: This shouldn't need the explicit type

discord/user.py
```diff
@@ -434,10 +456,10 @@ async def create_dm(self):
...
         """
-        found = self.dm_channel
+        found: Optional[DMChannel] = self.dm_channel
         if found is not None:
             return found
 
-        state = self._state
-        data = await state.http.start_private_message(self.id)
-        return state.add_dm_channel(data)
+        state: ConnectionState = self._state
```
bryanforbes, line 463: This shouldn't need an explicit type

discord/user.py
```diff
@@ -434,10 +456,10 @@ async def create_dm(self):
         :class:`.DMChannel`
             The channel that was created.
         """
-        found = self.dm_channel
+        found: Optional[DMChannel] = self.dm_channel
```
bryanforbes, line 459: This shouldn't need an explicit type

discord/user.py
```diff
@@ -439,5 +470,5 @@ async def create_dm(self):
             return found
 
         state = self._state
-        data = await state.http.start_private_message(self.id)
-        return state.add_dm_channel(data)
+        data: DMChannelPayload = await state.http.start_private_message(self.id)
+        return state.add_dm_channel(data)
```
Rapptz, line 474: This needs a newline at the end.
```suggestion
        return state.add_dm_channel(data)

```

PR #7395, opened 2021-08-14
fix Message.is_system

SYCKGit:
## Summary

Message.is_system was checking if the type is MessageType.default but now there are other MessageTypes that are not system messages, this PR fixes this

## Checklist

- [x] This PR fixes an issue.


Rapptz (commented):
This also needs a documentation update explaining what it does. The previous definition of a system message was one that the Discord API specifically constructed. However I can do this one myself.

PR #7400, opened 2021-08-15
fix Guild.fetch_channel does not work for threads

la-png:
## Summary
Guild.fetch_channel did not work when the target was a thread.
I will solve this problem by adding threads to the target of _guild_channel_factory.

## Checklist

- [x] This PR fixes an issue.


Rapptz:
This is not a proper solution.

PR #7245, opened 2021-07-16
Add an example for the new dropdowns

Robin5605:
## Checklist

- [ ] If code changes were made then they have been tested.
    - [ ] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.
- [ ] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [x] This PR is **not** a code change (e.g. documentation, README, ...)

PR #7405, opened 2021-08-17
Add an example of view without subclassing the selectmenu

bartick:
It would be really great if you add an example of view without subclassing buttons or select menu.


Rapptz:
[This was deliberate](https://github.com/Rapptz/discord.py/pull/7245#pullrequestreview-711314528). I didn't want to push people into the clunky decorator syntax (even though it's there) and wanted to just recommend the cleaner syntax instead.

PR #7407, opened 2021-08-17
Fix incorrect type-hint

thetimtoy:
## Checklist

- [x If code changes were made then they have been tested.
- [x] This PR fixes an issue.


NCPlayz:
`timeout` being set to `None` indicates that the view is listened to forever. This is also used for persistent views, so removing the type hint would be incorrect.

PR #7411, opened 2021-08-18
[commands] reset view when Optional argument encounters parsing error

### LINKED ISSUE #7386: UnexpectedQuoteError for keyword only args when preceded by optional argument
### Summary

When a keyword only argument contains a single quote character such as " an error if thrown if that argument is preceded by an optional argument.

### Reproduction Steps

Reproduction code below. Here are the commands used to reproduce the bug in command response pairs

`-test1 abc"`
Throws UnexpectedQuoteError

`-test2 abc"`
`This works. abc"`

`-test3 1 abc"`
`This works. 1 abc"`

### Minimal Reproducible Code

```python
from typing import Optional

from discord.ext.commands import Bot

bot = Bot('-')


@bot.event
async def on_ready():
    print('ready')


@bot.command()
async def test1(ctx, optional: Optional[int], *, rest):
    await ctx.send(f'This does not work. {rest}')


@bot.command()
async def test2(ctx, *, rest):
    await ctx.send(f'This works. {rest}')


@bot.command()
async def test3(ctx, required: int, *, rest):
    await ctx.send(f'This works. {required} {rest}')


bot.run('token')
```

### Expected Results

I expected no error to be thrown when calling the test1 command.

### Actual Results

UnexpectedQuoteError was thrown when string contained a quote character.

### Intents

Default intents

### System Information

- Python v3.7.4-final
- discord.py v1.7.3-final
- aiohttp v3.7.4.post0
- system info: Windows 10 10.0.18362

### Checklist

### Additional Context

ExpectedClosingQuoteError is also thrown for test1 if the string starts with a quote while it is not thrown for the other two commands.

SebbyLaw:
## Summary

Resolves #7386 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -505,7 +505,14 @@ class Command(_BaseCommand):
         if consume_rest_is_special:
             argument = view.read_rest().strip()
         else:
-            argument = view.get_quoted_word()
+            try:
+                argument = view.get_quoted_word()
+            except ArgumentParsingError as exc:
+                if self._is_typing_optional(param.annotation):
+                    view.index = previous
+                    return None
+                else:
+                    raise exc
         view.previous = previous
 
         return await run_converters(ctx, converter, argument, param)


Rapptz (approved):
LGTM.

PR #7413, opened 2021-08-19
Make enums comparable

### LINKED ISSUE #7250: Comparing VerificationLevel gives unexpected results
### Summary

Comparing VerificationLevel does not work as expected

### Reproduction Steps

1. Tried to compare VerificationLevel 4 with 1
2. Received a different result than expected

### Minimal Reproducible Code

```python
>>> from discord import VerificationLevel
>>> VerificationLevel(4) > VerificationLevel(1) # "extreme" and "low"
False # !
```

### Expected Results

For the comparison to return `True`

### Actual Results

The comparison returned `False`

### Intents

-

### System Information

- Python v3.8.7-final
- discord.py v1.7.3-final
- aiohttp v3.7.3
- system info: Linux 5.10.49-1-MANJARO

### Checklist

### Additional Context

I was about to open a PR to use IntEnum instead but I see it was removed for compatibility with 3.5

SebbyLaw:
## Summary

Add comparison operations to Enum. Also added docs for supported operations on some enums where comparing values might be relevant. (NotificationLevel, NSFWLevel)

Not sure what to do about the supported operations containers on VerificationLevel and ContentFilter that were added in #688 and then broken in 0ddc6867e9e2c98d74a4ef85ac36a99fec5a1f03.

This PR closes #7250.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -58,12 +58,16 @@ __all__ = (
 )
 
 
-def _create_value_cls(name):
+def _create_value_cls(name, comparable):
     cls = namedtuple('_EnumValue_' + name, 'name value')
     cls.__repr__ = lambda self: f'<{name}.{self.name}: {self.value!r}>'
     cls.__str__ = lambda self: f'{name}.{self.name}'
+    if comparable:
+        cls.__le__ = lambda self, other: isinstance(other, self.__class__) and self.value <= other.value
+        cls.__ge__ = lambda self, other: isinstance(other, self.__class__) and self.value >= other.value
+        cls.__lt__ = lambda self, other: isinstance(other, self.__class__) and self.value < other.value
+        cls.__gt__ = lambda self, other: isinstance(other, self.__class__) and self.value > other.value
     return cls
-
 
 def _is_descriptor(obj):
     return hasattr(obj, '__get__') or hasattr(obj, '__set__') or hasattr(obj, '__delete__')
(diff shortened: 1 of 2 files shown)


Rapptz (changes requested):
I think it's better to make this opt-in rather than being unconditional as it is now. A lot of enums are strings and string comparisons are useless. This should really only be done for numerical, documented, enums.

SebbyLaw:
> Not sure what to do about the supported operations containers on VerificationLevel and ContentFilter that were added in #688 and then broken in 0ddc6867e9e2c98d74a4ef85ac36a99fec5a1f03.

Should these have a versionadded as well?

discord/enums.py
```diff
@@ -76,12 +76,19 @@ class EnumMeta(type):
         _enum_member_map_: ClassVar[Dict[str, Any]]
         _enum_value_map_: ClassVar[Dict[Any, Any]]
 
-    def __new__(cls, name, bases, attrs):
+    def __new__(cls, name, bases, attrs, **kwargs):
```
Rapptz, line 79:
```suggestion
    def __new__(cls, name, bases, attrs, *, comparable: bool = False):
```

discord/enums.py
```diff
@@ -76,12 +76,19 @@ class EnumMeta(type):
...
         member_mapping = {}
         member_names = []
 
         value_cls = _create_value_cls(name)
+
+        if kwargs.get('comparable'):
+            value_cls.__le__ = lambda self, other: isinstance(other, self.__class__) and self.value <= other.value
+            value_cls.__ge__ = lambda self, other: isinstance(other, self.__class__) and self.value >= other.value
+            value_cls.__lt__ = lambda self, other: isinstance(other, self.__class__) and self.value < other.value
+            value_cls.__gt__ = lambda self, other: isinstance(other, self.__class__) and self.value > other.value
```
Rapptz, line 90: Move this to `_create_value_cls`. This stuff was in that function to prevent the unnecessary closure.

Rapptz (changes requested):
The docs should be updated to reflect the changes.

Rapptz (approved):

PR #7414, opened 2021-08-19
Typehint state.py

StockerMC:
## Summary
This pull request typehints state.py

The other files that were edited were edited for typehinting state.py

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/state.py
+++ b/discord/state.py
@@ -102,7 +102,7 @@ class ChunkRequest:
                 if existing is None or existing.joined_at is None:
                     guild._add_member(member)
 
-    async def wait(self) -> None:
+    async def wait(self) -> List[Member]:
         future = self.loop.create_future()
         self.waiters.append(future)
         try:
--- a/discord/state.py
+++ b/discord/state.py
@@ -30,7 +30,7 @@ import copy
 import datetime
 import itertools
 import logging
-from typing import Dict, Optional, TYPE_CHECKING, Union, Callable, Any, List, TypeVar, Coroutine, Sequence, Tuple
+from typing import Dict, Optional, TYPE_CHECKING, Union, Callable, Any, List, TypeVar, Coroutine, Sequence, Tuple, Deque
 import inspect
 
 import os
@@ -241,9 +241,9 @@ class ConnectionState:
         # extra dict to look up private channels by user id
         self._private_channels_by_user: Dict[int, PrivateChannel] = {}
         if self.max_messages is not None:
-            self._messages: Optional[deque] = deque(maxlen=self.max_messages)
+            self._messages: Optional[Deque[Message]] = deque(maxlen=self.max_messages)
         else:
-            self._messages: Optional[deque] = None
+            self._messages: Optional[Deque[Message]] = None
 
     def process_chunk_requests(self, guild_id: int, nonce: Optional[str], members: List[Member], complete: bool) -> None:
         removed = []
@@ -1089,7 +1089,7 @@ class ConnectionState:
 
         # do a cleanup of the messages cache
         if self._messages is not None:
-            self._messages: Optional[deque] = deque((msg for msg in self._messages if msg.guild != guild), maxlen=self.max_messages)
+            self._messages: Optional[Deque[Message]] = deque((msg for msg in self._messages if msg.guild != guild), maxlen=self.max_messages)
 
         self._remove_guild(guild)
         self.dispatch('guild_remove', guild)
--- a/discord/state.py
+++ b/discord/state.py
@@ -288,8 +288,9 @@ class ConnectionState:
     def voice_clients(self) -> List[VoiceProtocol]:
         return list(self._voice_clients.values())
 
-    def _get_voice_client(self, guild_id: int) -> Optional[VoiceProtocol]:
-        return self._voice_clients.get(guild_id)
+    def _get_voice_client(self, guild_id: Optional[int]) -> Optional[VoiceProtocol]:
+        # the keys of self._voice_clients are ints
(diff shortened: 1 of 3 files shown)


discord/state.py
```diff
@@ -56,26 +54,43 @@
...
-
-    def add_members(self, members):
+    def __init__(self, guild_id: int, loop: asyncio.AbstractEventLoop, resolver: Callable[[int], Any], *, cache: bool = True) -> None:
+        self.guild_id: int = guild_id
+        self.resolver: Callable[[int], Any] = resolver
+        self.loop: asyncio.AbstractEventLoop = loop
+        self.cache: bool = cache
+        self.nonce: str = os.urandom(16).hex()
+        self.buffer: List[Member] = []
+        self.waiters: List[asyncio.Future] = []
```
Rapptz, line 91:
```suggestion
        self.waiters: List[asyncio.Future[List[Member]]] = []
```

discord/state.py
```diff
@@ -87,43 +102,48 @@ def add_members(self, members):
                 if existing is None or existing.joined_at is None:
                     guild._add_member(member)
 
-    async def wait(self):
+    async def wait(self) -> None:
```
Rapptz, line 105:
```suggestion
    async def wait(self) -> List[Member]:
```

discord/state.py
```diff
@@ -87,43 +102,48 @@ def add_members(self, members):
...
         future = self.loop.create_future()
         self.waiters.append(future)
         try:
             return await future
         finally:
             self.waiters.remove(future)
 
-    def get_future(self):
+    def get_future(self) -> asyncio.Future:
```
Rapptz, line 113:
```suggestion
    def get_future(self) -> asyncio.Future[List[Member]]:
```

Rapptz (changes requested):

discord/state.py
```diff
@@ -1311,32 +1353,32 @@ def get_channel(self, id):
...
-        self._ready_task = None
-        self.shard_ids = ()
-        self.shards_launched = asyncio.Event()
+        self.shard_ids: Union[List[int], range] = []
+        self.shards_launched: asyncio.Event = asyncio.Event()
 
-    def _update_message_references(self):
-        for msg in self._messages:
+    def _update_message_references(self) -> None:
+        for msg in self._messages or []:
```
Rapptz, line 1366: When this call happens `self._messages` is never `None` since it's in an `if` block.

discord/state.py
```diff
@@ -1216,32 +1256,34 @@ def parse_stage_instance_delete(self, data):
...
 
-            member, before, after = guild._update_voice_state(data, channel_id)
+            member, before, after = guild._update_voice_state(data, channel_id) # type: ignore
             if member is not None:
                 if flags.voice:
                     if channel_id is None and flags._voice_only and member.id != self_id:
-                        # Only remove from cache iff we only have the voice flag enabled
-                        guild._remove_member(member)
+                        # Only remove from cache if we only have the voice flag enabled
+                        # Member doesn't have
```
Rapptz, line 1277: This seems incomplete.

discord/state.py
```diff
@@ -561,16 +592,17 @@ def parse_message_update(self, data):
...
         emoji = PartialEmoji.with_state(self, id=emoji_id, animated=emoji.get('animated', False), name=emoji['name'])
         raw = RawReactionActionEvent(data, emoji, 'REACTION_ADD')
 
         member_data = data.get('member')
         if member_data:
-            guild = self._get_guild(raw.guild_id)
-            raw.member = Member(data=member_data, guild=guild, state=self)
+            # the guild won't be None here
+            guild = self._get_guild(raw.guild_id) # type: ignore
+            raw.member = Member(data=member_data, guild=guild, state=self) # type: ignore
```
Rapptz, line 605: This probably isn't true actually.

discord/state.py
```diff
@@ -661,41 +693,44 @@ def parse_presence_update(self, data):
...
             old_channel = copy.copy(channel)
-            channel._update_group(data)
+            # the channel is a GroupChannel
+            channel._update_group(data) # type: ignore
             self.dispatch('private_channel_update', old_channel, channel)
             return
 
         guild_id = utils._get_as_snowflake(data, 'guild_id')
-        guild = self._get_guild(guild_id)
+        guild = self._get_guild(guild_id) # type: ignore
```
Rapptz, line 733: I assume this was because `guild_id` is `Optional[int]`. The signature should be updated instead.

Rapptz (changes requested):

discord/state.py
StockerMC: `guild_id` can be None if there's a member object provided?

discord/state.py
StockerMC: Do you mean typehinting `guild_id` as `int` and `# type: ignore`'ing that line, or changing the signature of `_get_as_snowflake`? I don't see how `_get_as_snowflake`'s signature could be changed if that was the case.

discord/state.py
Rapptz: `guild` can be `None` (e.g. guild intent is off)

discord/state.py
Rapptz: I'm talking about `_get_guild`'s signature.

StockerMC: Alright. Should the signature for all `ConnectionState._get_x` methods that take an id be changed for the id to be `Optional[int]`?

Rapptz: Sure.

Rapptz (approved):
Thanks.

PR #7415, opened 2021-08-19
Re-define Member properties inferred from User to support type-checking

bijij:
## Summary

This pr re-defines all properties of `Member` derived from `User` as assignment leaves the type as `property`

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/member.py
+++ b/discord/member.py
@@ -272,28 +272,12 @@ class Member(discord.abc.Messageable, _UserTag):
         bot: bool
         system: bool
         created_at: datetime.datetime
-
-        @property
-        def default_avatar(self) -> Asset:
-            ...
-
-        @property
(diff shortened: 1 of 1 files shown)


discord/member.py
```diff
@@ -270,12 +272,28 @@ class Member(discord.abc.Messageable, _UserTag):
         bot: bool
         system: bool
         created_at: datetime.datetime
```
bryanforbes, line 274: `created_at` needs to have the same thing done to it

Rapptz (commented):
Maybe instead of this approach it should do e.g.

```py
default_avatar: Asset
```
instead?

I'm not sure if it being a `property` matters much to the type checker? I could go either way but this form is shorter and less obnoxious.

PR #7416, opened 2021-08-19
Add Thread to the return type of Client.get_channel and add type: ignore where needed

StockerMC:
## Summary
Lines 234, 235 and 303 were not `# type: ignore`'d because they will no longer error once #7414 is merged.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -206,6 +206,7 @@ class Client:
         loop: Optional[asyncio.AbstractEventLoop] = None,
         **options: Any,
     ):
+        # self.ws is set in the connect method
         self.ws: DiscordWebSocket = None  # type: ignore
         self.loop: asyncio.AbstractEventLoop = asyncio.get_event_loop() if loop is None else loop
         self._listeners: Dict[str, List[Tuple[asyncio.Future, Callable[..., bool]]]] = {}
--- a/discord/client.py
+++ b/discord/client.py
@@ -37,7 +37,7 @@ from .user import User, ClientUser
 from .invite import Invite
 from .template import Template
 from .widget import Widget
-from .guild import Guild, VocalGuildChannel
+from .guild import Guild
 from .emoji import Emoji
 from .channel import _threaded_channel_factory, PartialMessageable
 from .enums import ChannelType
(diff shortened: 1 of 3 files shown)


discord/client.py
```diff
@@ -682,7 +682,8 @@ def activity(self, value: Optional[ActivityTypes]) -> None:
         if value is None:
             self._connection._activity = None
         elif isinstance(value, BaseActivity):
-            self._connection._activity = value.to_dict()
+            # ConnectionState._activity is typehinted as ActivityPayload, we're passing Dict[str, Any]
+            self._connection._activity = value.to_dict() # type: ignore
```
bryanforbes, line 686: I'm not sure why this is needed since the return value of `BaseActivity.to_dict()` is typed as `ActivityPayload`

discord/client.py
```diff
@@ -1473,11 +1474,14 @@ async def fetch_channel(self, channel_id: int) -> Union[GuildChannel, PrivateCha
             raise InvalidData('Unknown channel type {type} for channel ID {id}.'.format_map(data))
 
         if ch_type in (ChannelType.group, ChannelType.private):
-            channel = factory(me=self.user, data=data, state=self._connection)
+            # the factory will be a DMChannel or GroupChannel here
+            channel = factory(me=self.user, data=data, state=self._connection) # type: ignore
```
bryanforbes, line 1478: I'm not sure what is being ignored. Isn't `factory` typed as `Any`?

bryanforbes (changes requested):

discord/client.py
StockerMC: The (inferred, since the return type for factory related methods in channel.py isn't typed yet) type of  `factory` is `"Type[Thread] | Type[DMChannel] | Type[GroupChannel] | Type[TextChannel] | Type[VoiceChannel] | Type[CategoryChannel] | Type[StoreChannel] | Type[StageChannel] |`

DMChannel and GroupChannel are the only types of those that would take the `me` parameter.

discord/client.py
bryanforbes: I didn't see that `value` is typed as `Optional[ActivityTypes]`, which is why `to_dict()` doesn't return `ActivityPayload`. This should stay as `type: ignore`

discord/client.py
bryanforbes: `mypy` doesn't infer return types, so I'm not used to `factory` being anything but `Any`. If `factory` is getting a type, then you definitely need to ignore when calling it.

Rapptz (approved):
LGTM.

PR #7418, opened 2021-08-19
Typehint shard.py and set force=True for disconnecting from voice clients in AutoShardedClient.close

StockerMC:
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/shard.py
+++ b/discord/shard.py
@@ -328,8 +328,9 @@ class AutoShardedClient(Client):
         self.__queue = asyncio.PriorityQueue()
 
     def _get_websocket(self, guild_id: Optional[int] = None, *, shard_id: Optional[int] = None) -> DiscordWebSocket:
-        if shard_id is None and guild_id is not None and self.shard_count is not None:
-            shard_id = (guild_id >> 22) % self.shard_count
+        if shard_id is None:
+            # guild_id won't be None if shard_id is None and shard_count won't be None here
+            shard_id = (guild_id >> 22) % self.shard_count # type: ignore
         return self.__shards[shard_id].ws
 
     def _get_state(self, **options: Any) -> AutoShardedConnectionState:
--- a/discord/shard.py
+++ b/discord/shard.py
@@ -55,7 +55,7 @@ __all__ = (
     'ShardInfo',
 )
 
-log = logging.getLogger(__name__)
+log: logging.Logger = logging.getLogger(__name__)
 
 class EventType:
     close = 0
--- a/discord/shard.py
+++ b/discord/shard.py
@@ -22,8 +22,9 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 DEALINGS IN THE SOFTWARE.
 """
 
+from __future__ import annotations
+
 import asyncio
-import itertools
 import logging
 
 import aiohttp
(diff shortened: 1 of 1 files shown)


discord/shard.py
```diff
@@ -34,22 +35,29 @@
...
 __all__ = (
     'AutoShardedClient',
     'ShardInfo',
 )
 
-log = logging.getLogger(__name__)
+log: logging.Logger = logging.getLogger(__name__)
 
 class EventType:
     close = 0
```
bryanforbes, line 63: These should all be annotated as `ClassVar[int]`

discord/shard.py
```diff
@@ -62,36 +70,36 @@ class EventType:
...
 
 class Shard:
-    def __init__(self, ws, client, queue_put):
-        self.ws = ws
-        self._client = client
-        self._dispatch = client.dispatch
-        self._queue_put = queue_put
-        self.loop = self._client.loop
-        self._disconnect = False
+    def __init__(self, ws: DiscordWebSocket, client: AutoShardedClient, queue_put: Callable[[Any], None]) -> None:
```
bryanforbes, line 92: It would be better if `queue_put` was typed as `Callable[[EventItem], None]`

discord/shard.py
```diff
@@ -62,36 +70,36 @@ class EventType:
...
+        self.ws: DiscordWebSocket = ws
+        self._client: Client = client
+        self._dispatch: Callable[..., None] = client.dispatch
+        self._queue_put: Callable[[Any], None] = queue_put
```
bryanforbes, line 96: The same goes for `_queue_put` here

discord/shard.py
```diff
@@ -101,25 +109,25 @@ def __init__(self, ws, client, queue_put):
         )
 
     @property
-    def id(self):
-        return self.ws.shard_id
+    def id(self) -> int:
+        return self.ws.shard_id # type: ignore
```
bryanforbes, line 113: Please add a comment to explain why this is ignored

bryanforbes (changes requested):

discord/shard.py
```diff
@@ -34,64 +35,71 @@
...
-    resume = 2
-    identify = 3
-    terminate = 4
-    clean_close = 5
+    close: ClassVar[int] = 0
+    reconnect: ClassVar[int] = 1
+    resume: ClassVar[int] = 2
+    identify: ClassVar[int] = 3
+    terminate: ClassVar[int] = 4
+    clean_close: ClassVar[int] = 5
```
Rapptz, line 68: Please don't.

discord/shard.py
```diff
@@ -315,18 +327,18 @@ def __init__(self, *args, loop=None, **kwargs):
         self._connection._get_client = lambda: self
         self.__queue = asyncio.PriorityQueue()
 
-    def _get_websocket(self, guild_id=None, *, shard_id=None):
-        if shard_id is None:
+    def _get_websocket(self, guild_id: Optional[int] = None, *, shard_id: Optional[int] = None) -> DiscordWebSocket:
+        if shard_id is None and guild_id is not None and self.shard_count is not None:
```
Rapptz, line 331: `_get_websocket` is always called with these being set.

discord/shard.py
```diff
@@ -355,11 +367,11 @@ def get_shard(self, shard_id):
             return ShardInfo(parent, self.shard_count)
 
     @property
-    def shards(self):
+    def shards(self) -> Mapping[int, ShardInfo]:
```
Rapptz, line 370:
```suggestion
    def shards(self) -> Dict[int, ShardInfo]:
```

Rapptz (changes requested):

discord/shard.py
StockerMC: Changed back to the original conditional and added a `# type: ignore` with the reason.

Rapptz (approved):

PR #7420, opened 2021-08-19
Add a few typehints to opus.py

StockerMC:
## Summary
Add a few typehints to opus.py to avoid pyright errors

## Checklist

- [ ] This PR fixes an issue.


discord/opus.py
```diff
@@ -76,15 +78,15 @@ class DecoderStruct(ctypes.Structure):
 CTL_SET_GAIN             = 4034
 CTL_LAST_PACKET_DURATION = 4039
 
-band_ctl = {
+band_ctl: Dict[str, int] = {
```
bryanforbes, line 81: Both of these `Dict`s should probably be done as `TypedDict`

Issue #7054, opened 2021-06-12
Add per-guild avatar support


JustAnyones:
### Summary

Add support for per-guild member avatars

### What is the feature request for?

The core library

### The Problem

See https://github.com/discord/discord-api-docs/pull/3081.
Feature implementation would allow bots to access per-guild member avatars.

### The Ideal Solution

Ideally, it would be implemented as a member object method like ``Member.avatar_url_as()``

### The Current Solution

Right now, I use a very hacky solution which includes _hackily_ getting the full member object and then checking whether the avatar field is available and returning the correct URL based on that.

```py
format = 'png'
size = 256

user_data = await client._get_state().http.get_member(guild.id, member.id)
avatar = user_data["avatar"]
if avatar is None:
    return None
# Basically, replicates Asset._from_avatar behaviour
format = 'gif' if avatar.startswith('a_') else format
return f"https://cdn.discordapp.com/guilds/{guild.id}/users/{member.id}/avatars/{avatar}.{format}?size={size}"
```

Rapptz:
This is being worked on already. Note that v1.x won't get any new features and asset was redesigned in v2 so your proposed API won't make it in.

PR #7423, opened 2021-08-19
Add per-guild member avatar support

JustAnyones:
## Summary

This pull request addresses the #7054 by adding support for per-guild member avatars. To put simply, this PR adds a ``Member.guild_avatar`` property to get member guild avatar. If the guild-specific avatar is unavailable, it just calls ``Member.avatar``.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/asset.py
+++ b/discord/asset.py
@@ -178,6 +178,17 @@ class Asset(AssetMixin):
         )
 
     @classmethod
+    def _from_guild_avatar(cls, state, guild_id: int, member_id: int, avatar: str) -> Asset:
+        animated = avatar.startswith('a_')
+        format = 'gif' if animated else 'png'
+        return cls(
+            state,
+            url=f"{cls.BASE}/guilds/{guild_id}/users/{member_id}/avatars/{avatar}.{format}?size=1024",
+            key=avatar,
+            animated=animated,
+        )
+
+    @classmethod
     def _from_icon(cls, state, object_id: int, icon_hash: str, path: str) -> Asset:
         return cls(
             state,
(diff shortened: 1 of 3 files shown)


Rapptz:
This feature is unstable (i.e. not documented to work according to Discord and can break at any time).

That being said, this has a few bikeshedding considerations that were bypassed in this PR which is also why it wasn't immediately implemented. We have a few choices here:

First, `Member.avatar` can return the guild avatar and fallback to the user avatar if not given. This also means making `Member.user` a public read-only attribute so you can access `Member.user.avatar` for the original avatar. Alternatively, `Member.user_avatar` as an attribute.

Second, this PR's implementation which is `Member.guild_avatar` while keeping `Member.avatar` the same.

Third, a `Member.display_avatar` which does the same thing `nick` does in this library. Basically returning guild avatar if available, otherwise fall back to user avatar. This property would have to be part of `User` as well. So.. `Member.avatar` is the user's avatar, `Member.guild_avatar` is *only* the guild avatar, and `Member.display_avatar` would do `guild_avatar` or fallback avatar.

There were too many choices and we didn't know which one to go with.

JustAnyones:
You bring up a good point. Personally, I feel like the third option would be the right choice as it would be symmetrical with the display name.

Moreover, in that implementation, you could tell whether the member has a guild avatar at all.

However, that's just my humble opinion and I would like if others could express their thoughts on it.

Rapptz:
I think the `display_name` proposal is the most popular given an informal vote if you want to go ahead with it.

JustAnyones:
I've observed the discord.py server, seing the third option be the popular one, I implemented the changes.

Rapptz (changes requested):
Mostly good, but this is missing the same attribute for `BaseUser`.

Rapptz (approved):
LGTM. Thanks.

PR #7438, opened 2021-08-21
Add views on every on_ready in persistent example

HyperGH:
## Summary

This is an attempt at clarifying the persistent view example and adding the views back on every `on_ready` as `READY` clears the state of said views, breaking them. (See #7437) This ensures that the views keep working after a disconnect for example. I have not tested this, but it shouldn't cause any issues.

## Checklist

- [x] This PR fixes an issue.


Rapptz:
This is the incorrect fix, thanks though.

HyperGH:
I assume then that the current behavior of persistent views is unintended then?

PR #7440, opened 2021-08-21
Add fetch_creation_message method to threads.

Chrovo:
Add a method to the discord.Thread class which allows you to fetch the message which created the thread from the parent channel.

## Summary
This pull request is for adding a method to the discord.Thread class, the method allows you to fetch the message which created the thread from the parent channel. 

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/threads.py
+++ b/discord/threads.py
@@ -684,6 +684,26 @@ class Thread(Messageable, Hashable):
             Deleting the thread failed.
         """
         await self._state.http.delete_channel(self.id)
+        
+    async def fetch_creation_message(self) -> Message:
+        """|coro|
+        Retrieves the message that was used to create this thread from the parent channel.
+        
+        Raises
+        --------
+        ~discord.NotFound
+            The message was not found.
+        ~discord.Forbidden
+            You do not have the permissions required to get the message.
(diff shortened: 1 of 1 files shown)


Rapptz:
There isn't actually a way to check if the thread ID is the message starter ID so this gives kind of a false premise. I'm not really sure I like this, sorry.

Chrovo:
I think you misunderstood, this method wasn't supposed to be fetching the starter message, but the message that was used to create this thread, the one from the parent channel. For example:
image

Rapptz:
That is called the starter message.

Chrovo:
Oh, my bad, I thought this was called the starter message: 
image
Additionally, isn't the thread ID the same as the starter message ID?

Rapptz:
Those two messages are the exact same (same ID and everything). They just show up twice for client purposes. Like I was trying to elude, sometimes the thread is started without a starter message so since there's no way to check whether the thread has one or not I don't feel comfortable with this feature.

Chrovo:
Oh, I thought it raising NotFound would be an indication to the user that the starter message was deleted or it didn't get started with a starter message.

PR #7446, opened 2021-08-22
Typehint opus.py

apple502j:
## Summary
Typehints opus.py. Checked on mypy.

- All native function calls are `#type: ignore`'d.
- `None` assignment inside destructors are `#type: ignore`'d.
- `Decoder.decode` now raises `InvalidArgument` instead of `OpusError` because `OpusError` cannot take a string as the error code. This is not technically breaking change, since it is unused and never documented.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/opus.py
+++ b/discord/opus.py
@@ -320,18 +320,18 @@ class Encoder(_OpusStruct):
 
     def __del__(self) -> None:
         if hasattr(self, '_state'):
-            _lib.opus_encoder_destroy(self._state) # type: ignore
+            _lib.opus_encoder_destroy(self._state)
             # This is a destructor, so it's okay to assign None
             self._state = None # type: ignore
 
     def _create_state(self) -> EncoderStruct:
         ret = ctypes.c_int()
-        return _lib.opus_encoder_create(self.SAMPLING_RATE, self.CHANNELS, self.application, ctypes.byref(ret)) # type: ignore
+        return _lib.opus_encoder_create(self.SAMPLING_RATE, self.CHANNELS, self.application, ctypes.byref(ret))
 
     def set_bitrate(self, kbps: int) -> int:
         kbps = min(512, max(16, int(kbps)))
 
-        _lib.opus_encoder_ctl(self._state, CTL_SET_BITRATE, kbps * 1024) # type: ignore
+        _lib.opus_encoder_ctl(self._state, CTL_SET_BITRATE, kbps * 1024)
         return kbps
 
     def set_bandwidth(self, req: BAND_CTL) -> None:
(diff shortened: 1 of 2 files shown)


discord/opus.py
```diff
@@ -273,8 +280,8 @@ class OpusError(DiscordException):
         The error code returned.
     """
 
-    def __init__(self, code):
-        self.code = code
+    def __init__(self, code: int) -> None:
```
Rapptz, line 283:
```suggestion
    def __init__(self, code: int):
```

discord/opus.py
```diff
@@ -300,92 +307,96 @@ def get_opus_version() -> str:
         return _lib.opus_get_version_string().decode('utf-8')
 
 class Encoder(_OpusStruct):
-    def __init__(self, application=APPLICATION_AUDIO):
+    def __init__(self, application: int=APPLICATION_AUDIO) -> None:
```
Rapptz, line 310:
```suggestion
    def __init__(self, application: int = APPLICATION_AUDIO):
```

discord/opus.py
```diff
@@ -172,7 +179,7 @@ def _err_ne(result, func, args):
         [ctypes.c_char_p, ctypes.c_int], ctypes.c_int, _err_lt),
 ]
 
-def libopus_loader(name):
+def libopus_loader(name: str) -> ctypes.CDLL:
```
Rapptz, line 182:
```suggestion
def libopus_loader(name: str) -> Any:
```

Instead of `# type: ignore`ing everything this is the better way to go about doing it since the type checking is unnecessary.

discord/opus.py
```diff
@@ -300,92 +307,96 @@ def get_opus_version() -> str:
...
-        ret = _lib.opus_encode(self._state, pcm, frame_size, data, max_data_bytes)
+        ret = _lib.opus_encode(self._state, pcm_ptr, frame_size, data, max_data_bytes) # type: ignore
 
-        return array.array('b', data[:ret]).tobytes()
+        # array can be initialized with bytes but mypy doesn't know
+        return array.array('b', data[:ret]).tobytes() # type: ignore
 
 class Decoder(_OpusStruct):
-    def __init__(self):
+    def __init__(self) -> None:
```
Rapptz, line 369:
```suggestion
    def __init__(self):
```

discord/opus.py
```diff
@@ -395,28 +406,36 @@ def _set_gain(self, adjustment):
...
+        ...
+    
+    @overload
+    def decode(self, data: Literal[None], *, fec: Literal[False]) -> bytes:
+        ...
+
+    def decode(self, data: Optional[bytes], *, fec: bool = False) -> bytes:
         if data is None and fec:
-            raise OpusError("Invalid arguments: FEC cannot be used with null data")
+            raise InvalidArgument("Invalid arguments: FEC cannot be used with null data")
```
Rapptz, line 438: Why was this changed?

Rapptz (changes requested):

discord/opus.py
apple502j: `OpusError` takes `code: int`, this is `str`

Rapptz (approved):
LGTM.

PR #7457, opened 2021-08-23
Fix async iterable not showing up attributetable

Gobot1234:
## Summary

Before it would show up as `def` now it correctly shows up as `async for`
image

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
This has two unrelated commits.

docs/extensions/attributetable.py
```diff
@@ -1,90 +1,127 @@
...
+from sphinx import addnodes
+from sphinx.application import Sphinx
+from sphinx.environment import BuildEnvironment
+from sphinx.locale import _
+from sphinx.util.docutils import SphinxDirective
+from sphinx.util.typing import OptionSpec
+
+
+def returns_async_iterable(object: Any) -> bool:
+    return_type = object.__annotations__.get("return")
```
Rapptz, line 16:
```suggestion
def returns_async_iterable(obj: Any) -> bool:
    return_type = obj.__annotations__.get('return')
```

docs/extensions/attributetable.py
```diff
@@ -1,90 +1,127 @@
...
+    if isinstance(return_type, str):
+        module = importlib.import_module(object.__module__)
+        try:
+            return_type = getattr(module, return_type)
+        except AttributeError:  # not a global, so probably not an async iterable
+            return False
```
Rapptz, line 22: This is definitely finicky and only works because incidentally a lot of the async iterables import their implementations. This might not end up being true in a future refactor.

Rapptz (changes requested):

docs/extensions/attributetable.py
Gobot1234: I understand, I'm just not entirely sure how you would avoid this unless you want to parse ast yourself.

Rapptz: It'd be better to parse the docstring than parse the types.

Gobot1234: I'm just using inspect.isasyncgenfunction now which I think is a much better solution than either of the two propsed

PR #7459, opened 2021-08-23
[types] Make VoiceProtocol/VoiceClient generic on client.

bijij:
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/voice_client.py
+++ b/discord/voice_client.py
@@ -44,7 +44,7 @@ import socket
 import logging
 import struct
 import threading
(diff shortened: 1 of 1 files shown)


Rapptz:
How are you expecting this to work? `connect` doesn't have the necessary information to propagate the generic so it'd be `VoiceClient[Unknown]`.

bijij:
~~It'll work for my duck type constructing before passing to connect meme~~
But yeah it won't work for the standard case.

bijij:
I guess `Client.voice_clients` in a subclass could be overwritten too to utilise the generic.

PR #7462, opened 2021-08-23
Add thread attribute to discord.Message class.

Chrovo:
## Summary
Add an attribute which returns the thread the message created, None if it did not create any threads or if the channel is not a TextChannel.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -988,6 +988,21 @@ class Message(Hashable):
             MessageType.application_command,
             MessageType.thread_starter_message,
         )
+    
+    @property
+    def thread(self) -> Optional[Thread]:
+        """Retrieves the thread the message has created ``None`` if no thread was created
+        with this message or if the channel is not a :class:`TextChannel`.
+        
+        Returns
+        --------
+        Optional[:class:`Thread`]
+            The thread this message has created.
+        """
+        try:
+            return self.channel.get_thread(self.id)
+        except AttributeError:
+            return None
 
     @utils.cached_slot_property('_cs_system_content')
     def system_content(self):


discord/message.py
```diff
@@ -988,6 +988,21 @@ def is_system(self) -> bool:
...
+        
+        Returns
+        --------
+        Optional[:class:`Thread`]
+            The thread this message has created.
+        """
+        try:
+            return self.channel.get_thread(self.id)
+        except AttributeError:
+            return None
```
jgayfer, line 1005: `get_thread` should return `None` if not found. The exception handler isn't needed.

Chrovo: It is, if the channel is DMChannel or something other than a text channel, it does not have the attribute get_thread. In which case, we will return None.

jgayfer: Right, gotcha! In that case, I'd personally check the channel type before hand and return early. Generally not a huge fan of leaning on exceptions for control flow.

Rapptz:
Thanks for the PR.

As it stands, this implementation is subpar and the attribute name is definitely ambiguous. When I first saw this PR my first reaction was commenting "The `channel` attribute already exists". I'm not entirely sure this is super useful to have. What's the use case here?

Chrovo:
Hm. I am just realizing the attribute name could lead to some confusion. Additionally, is there a better way to implement this, and what is bad about the current code? I noticed you called it "subpar" so I became a bit curious about it. Regarding your question, frankly, I also don't see a lot of use cases for this.

Rapptz:
A better implementation would be

```py
@property
def started_thread(self) -> Optional[Thread]:
    guild = self.guild
    if guild and self.type is MessageType.thread_starter_message:
        return guild.get_thread(self.id)
    return None
```

But since there's no use case for this then I don't really see why this should be added.

PR #7465, opened 2021-08-24
[commands] re-order union for checks

NCPlayz:
## Summary

Seemingly, pyright complains with checks involving lambdas with the current order of the Union, inferring that the first argument is a Cog instance:
```py
@check(lambda ctx: ctx.author.id == 0)
@bot.command()
async def test(ctx: Context[Bot]):
    ...
```

```
Argument of type "(ctx: Cog) -> Unknown" cannot be assigned to parameter "predicate" of type "Check" in function "check"
  Type "(ctx: Cog) -> Unknown" cannot be assigned to type "Check"
    Type "(ctx: Cog) -> Unknown" cannot be assigned to type "(_p0: Cog, _p1: Context[Any]) -> MaybeCoro[bool]"
      Function accepts too many positional parameters; expected 1 but received 2
    Type "(ctx: Cog) -> Unknown" cannot be assigned to type "(_p0: Context[Any]) -> MaybeCoro[bool]"
      Parameter 1: type "Context[Any]" cannot be assigned to type "Cog"
        "Context[Any]" is incompatible with "Cog" Pylance(reportGeneralTypeIssues)
```

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/_types.py
+++ b/discord/ext/commands/_types.py
@@ -37,7 +37,7 @@ Coro = Coroutine[Any, Any, T]
 MaybeCoro = Union[T, Coro[T]]
 CoroFunc = Callable[..., Coro[Any]]
 
-Check = Union[Callable[["Cog", "Context[Any]"], MaybeCoro[bool]], Callable[["Context[Any]"], MaybeCoro[bool]]]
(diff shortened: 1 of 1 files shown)


Rapptz:
I'm genuinely confused why this matters. This sounds like a pyright bug?

Rapptz:
This pyright bug will be fixed in the next pyright release so no need for this workaround.

PR #7493, opened 2022-03-06
Adds with_counts to dpy.

JDJGInc:
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -1307,7 +1307,7 @@ class Client:
         data = await self.http.get_template(code)
         return Template(data=data, state=self._connection)  # type: ignore
 
-    async def fetch_guild(self, guild_id: int, /, *, with_counts: Optional[bool] = True) -> Guild:
+    async def fetch_guild(self, guild_id: int, /, *, with_counts: bool = True) -> Guild:
         """|coro|
 
         Retrieves a :class:`.Guild` from an ID.
--- a/discord/client.py
+++ b/discord/client.py
@@ -1307,7 +1307,7 @@ class Client:
         data = await self.http.get_template(code)
         return Template(data=data, state=self._connection)  # type: ignore
 
-    async def fetch_guild(self, guild_id: int, /) -> Guild:
+    async def fetch_guild(self, guild_id: int, /, *, with_counts: Optional[bool] = True) -> Guild:
         """|coro|
 
         Retrieves a :class:`.Guild` from an ID.
@@ -1324,6 +1324,12 @@ class Client:
         .. versionchanged:: 2.0
 
             ``guild_id`` parameter is now positional-only.
+
+        with_counts: Optional[:class:`bool`]
+            Whether to include count information in the guild. This fills the
+            :attr:`.Guild.approximate_member_count` and :attr:`.Guild.approximate_presence_count`
+            attributes without needing any privileged intents. Defaults to ``True``.
+            .. versionadded:: 2.0
 
         Parameters
         -----------
@@ -1342,7 +1348,7 @@ class Client:
         :class:`.Guild`
             The guild from the ID.
         """
-        data = await self.http.get_guild(guild_id)
+        data = await self.http.get_guild(guild_id, with_counts=with_counts)
         return Guild(data=data, state=self._connection)
 
     async def create_guild(
--- a/discord/client.py
+++ b/discord/client.py
@@ -1329,6 +1329,7 @@ class Client:
             Whether to include count information in the guild. This fills the
             :attr:`.Guild.approximate_member_count` and :attr:`.Guild.approximate_presence_count`
             attributes without needing any privileged intents. Defaults to ``True``.
+
             .. versionadded:: 2.0
 
         Parameters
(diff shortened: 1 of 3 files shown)


discord/client.py
```diff
@@ -1325,6 +1325,12 @@ async def fetch_guild(self, guild_id: int, /) -> Guild:
 
             ``guild_id`` parameter is now positional-only.
 
+        with_counts: Optional[:class:`bool`]
+            Whether to include count information in the guild. This fills the
+            :attr:`.Guild.approximate_member_count` and :attr:`.Guild.approximate_presence_count`
+            attributes without needing any privileged intents. Defaults to ``True``.
+            .. versionadded:: 2.0
```
StockerMC, line 1332:
```suggestion

            .. versionadded:: 2.0
```

discord/client.py
```diff
@@ -1307,7 +1307,7 @@ async def fetch_template(self, code: Union[Template, str]) -> Template:
         data = await self.http.get_template(code)
         return Template(data=data, state=self._connection)  # type: ignore
 
-    async def fetch_guild(self, guild_id: int, /) -> Guild:
+    async def fetch_guild(self, guild_id: int, /, *, with_counts: Optional[bool] = True) -> Guild:
```
Gobot1234, line 1310: This type hint is wrong, it should only allow bool

JDJGInc: it's an Optional Kwarg.
I don't want to break it.

StockerMC: `Optional[T]` means `T | None`,  but it only accepts a `bool`

JDJGInc: do you mean with_counts: bool = True ?

Gobot1234: Yes

StockerMC: Yes, that's correct

JDJGInc: There you go

discord/guild.py
```diff
@@ -255,6 +255,19 @@ class Guild(Hashable):
...
+        using :meth:`Client.fetch_guild` with ``with_counts=True``.
+        .. versionadded:: 2.0
+
+    approximate_presence_count: Optional[:class:`int`]
+        The approximate number of members currently active in the guild.
+        This includes idle, dnd, online, and invisible members. Offline members are excluded.
+        This is ``None`` unless the guild is obtained using :meth:`Client.fetch_guild`
+        with ``with_counts=True``.
+    
+        .. versionchanged:: 2.0
```
Rapptz, line 270:
```suggestion
    approximate_member_count: Optional[:class:`int`]
        The approximate number of members in the guild. This is ``None`` unless the guild is obtained
        using :meth:`Client.fetch_guild` with ``with_counts=True``.

        .. versionadded:: 2.0
    approximate_presence_count: Optional[:class:`int`]
        The approximate number of members currently active in the guild.
        Offline members are excluded. This is ``None`` unless the guild is obtained using 
        :meth:`Client.fetch_guild` with ``with_counts=True``.
    
        .. versionchanged:: 2.0
```

discord/guild.py
```diff
@@ -463,6 +478,8 @@ def _from_data(self, guild: GuildPayload) -> None:
         self._public_updates_channel_id: Optional[int] = utils._get_as_snowflake(guild, 'public_updates_channel_id')
         self.nsfw_level: NSFWLevel = try_enum(NSFWLevel, guild.get('nsfw_level', 0))
         self.mfa_level: MFALevel = try_enum(MFALevel, guild.get('mfa_level', 0))
+        self.approximate_presence_count = guild.get("approximate_presence_count")
+        self.approximate_member_count = guild.get("approximate_member_count")
```
Rapptz, line 482:
```suggestion
        self.approximate_presence_count: Optional[int] = guild.get('approximate_presence_count')
        self.approximate_member_count: Optional[int] = guild.get('approximate_member_count')
```

discord/http.py
```diff
@@ -1075,8 +1075,9 @@ def get_guilds(
     def leave_guild(self, guild_id: Snowflake) -> Response[None]:
         return self.request(Route('DELETE', '/users/@me/guilds/{guild_id}', guild_id=guild_id))
 
-    def get_guild(self, guild_id: Snowflake) -> Response[guild.Guild]:
-        return self.request(Route('GET', '/guilds/{guild_id}', guild_id=guild_id))
+    def get_guild(self, guild_id: Snowflake, *, with_counts: bool = True) -> Response[guild.Guild]:
+        params = {"with_counts": int(with_counts)}
```
Rapptz, line 1079:
```suggestion
        params = {'with_counts': int(with_counts)}
```

Rapptz (changes requested):
This library uses single quotes not double quotes.

Did you check whether the appropriate API typings in `discord.types` need to be updated?

discord/guild.py
dolfies: IMO it would be better to just use the existing member_count property instead of a new attribute.

JDJGInc: If people want me to do that I can do that right now.

JDJGInc:
my current code for with_counts works with slash commands and normally with commands.

Rapptz (approved):

PR #7495, opened 2022-03-06
[DOCS] Reaction Guide

ImVaskel:
## Summary

Adds a reaction guide

## Checklist

- [ ] This PR fixes an issue.


docs/guide/topics/reactions.rst
```diff
@@ -0,0 +1,133 @@
+.. currentmodule:: discord
+
+.. _guide_topic_reactions
+
+Reactions
+==========
+
+Reactions are a feature within Discord where you can add emojis next to messages.
+They have many uses, but a common one for bots is for polls or giving feedback to a command without sending a new message.
+Anyone can react to a message.
```
slice, line 10:
```suggestion
Members with the :attr:`~Permissions.add_reactions` can add new reactions to messages. However, this permission is not necessary to add to an existing reaction.
```

The note on line 31 also gets this across, but I feel like it conflicts with this sentence as it is.

docs/guide/topics/reactions.rst
```diff
@@ -0,0 +1,133 @@
...
+    :scale: 50%
+
+
+.. note::
+
+    You must have the permission :attr:`~Permissions.read_message_history` to add reactions.
+    Additionally, if no one else has reacted to the message using this emoji, you will also need the
+    :attr:`~Permissions.add_reactions` permission.
+
+:meth:`~Message.add_reaction` can also take many types, not just :class:`str`. This includes :class:`~Emoji` or :class:`~PartialEmoji`.
```
slice, line 37:
```suggestion
In addition to :class:`str`, you can also pass :class:`~Emoji`, :class:`~PartialEmoji`, or :class:`~Reaction` to :meth:`~Message.add_reaction`.
```

Exhaustively listing everything that `add_reaction` can accept here might help.

docs/guide/topics/reactions.rst
```diff
@@ -0,0 +1,133 @@
...
+For example, the following shows an alternative way to add a custom emoji reaction:
+
+.. code-block:: python3
+
+    emoji = client.get_emoji(819046807370989599)
+    await message.add_reaction(emoji)
+
+You may notice that if you try to add a standard Discord emoji like ``:smile:``, you'll receive an error.
+This is because to add standard Discord emojis to a message, you must pass their unicode values.
+Because of this, to add 😄 / ``:smile:`` you would do either of the following:
```
slice, line 47:
```suggestion
When adding standard (i.e. not custom) emojis as a reaction, you cannot pass emoji shortcodes,
such as the string ``":smile:"`` in order to react with 😄. The Discord client recognizes these
shortcodes as a typing convenience, but the API is not aware of them. So, in your bot, you must
pass the actual Unicode codepoint as is (e.g., ``"😄"``). This can be done like so:
```

This seems a bit pedantic, but the nomenclature can be ambiguous. "Emoji" can mean actual Unicode codepoints, Discord's custom emoji, or non-standard emoji shortcodes. I tried to reword stuff to make it less confusing here.

docs/guide/topics/reactions.rst
```diff
@@ -0,0 +1,133 @@
...
+
+.. code-block:: python3
+
+    # pass the unicode emoji via the hexadecimal value
+    await message.add_reaction("\U0001f604")
```
slice, line 52:
```suggestion
    # Pass the emoji codepoint via 32-bit hexadecimal escape.
    # For more information, see: https://docs.python.org/3/howto/unicode.html#the-string-type
    await message.add_reaction("\U0001f604")
```

docs/guide/topics/reactions.rst
```diff
@@ -0,0 +1,133 @@
...
+    
+    # or
+
+    # get the unicode emoji via the name (obtained via the unicode database)
```
slice, line 56:
```suggestion
    # Alternatively, use the codepoint's name.
    # See: https://docs.python.org/3/library/unicodedata.html
```

docs/guide/topics/reactions.rst
```diff
@@ -0,0 +1,133 @@
...
+    users = await reaction.users().flatten()
+    # users is now a list of User...
+    winner = random.choice(users)
+    await channel.send(f'{winner} has won the raffle.')
+
+Next Steps
+~~~~~~~~~~~
+
+You can find further information about :class:`~Reaction` in its documentation.
+Additionally, :meth:`~Client.wait_for` can pair well with reactions.
```
slice, line 133: Might wanna link to the `wait_for` guide here? It doesn't exist yet though.

ImVaskel: i will link it when the guide exists, but currently it does not.

docs/guide/topics/reactions.rst
Vexs: `:name:` emojis are supported in other contexts as well- such as sending messages.

slice: Unless I'm missing something, shortcodes in messages don't seem to be postprocessed by the API in any way, so the remark "[...] the API is not aware of them" appears to hold:

image

The client parses them though, of course.

docs/guide/topics/reactions.rst
```diff
@@ -0,0 +1,133 @@
...
+.. image:: /images/guide/topics/reactions/remove_reaction_after.png
+    :scale: 50%
+
+Removing all reactions of a single emoji from a message is also possible: 
+
+.. code-block:: python3
+
+    await message.clear_reaction("<:amogus:819046807370989599>")
+
+It is also possible to remove all reactions from a message, not just a singular member or emoji:
```
Gorialis, line 94: This following sentence feels a touch staggered due to the previous sentence (e.g., saying "You can also x. You can also y. You can also z." with no bridges doesn't read very well).

I think it'd probably be best to fix it like:
```suggestion
Removing all reactions of a given emoji from a message is also possible: 

.. code-block:: python3

    await message.clear_reaction("<:amogus:819046807370989599>")

And we can also remove all reactions entirely:
```

docs/guide/topics/reactions.rst
ImVaskel: i feel like a note would be better than adding that, perhaps just removing the "anyone can react to a message"?

docs/guide/topics/reactions.rst
ImVaskel: i do not see a reason to introduce client semantics to the guide.

slice: I feel like reducing the amount of potential confusion is desirable here, even at the expense of the result being a bit too wordy or technical. I don't think it's particularly difficult or overtly technical to explain that the API does not recognize shortcodes once it's spelled out, and that emojis are semantically distinct from shortcodes in this way.

I do see that this can be over the top here, though. Oh well pensive

ImVaskel: i feel like maybe a "the client and api see emojis in different ways, so you should be doing "X" to make sure you appease the api (in other wording of course).

slice: Oh, that could help, yeah. That's basically what you had in the first place, though, I think. So it's fine without my suggestion, seems.

docs/guide/topics/reactions.rst
slice: > perhaps just removing the "anyone can react to a message"?

Yeah that seems good.

docs/guide/topics/reactions.rst
AbstractUmbra: The guide now exists and has a reference link, this can be added.

AbstractUmbra (commented):
Only one comment here.

docs/guide/topics/reactions.rst
```diff
@@ -0,0 +1,133 @@
...
+    :scale: 50%
+
+You can find the documentation for adding reactions to a message object at :meth:`~Message.add_reaction`.
+
+Removing a reaction
+~~~~~~~~~~~~~~~~~~~~~
+
+Removing reactions is similar to adding a reaction, but you must also specify a member to remove the reaction from.
+This can be anything that meets the abc :class:`~abc.Snowflake`, like :class:`~discord.Member` or :class:`~discord.Object`.
+This means that you cannot pass simply only an id, and you must instead pass something with an ``id`` attribute.
```
numbermaniac, line 71:
```suggestion
This means that you cannot simply pass only an id, you must instead pass an object with an ``id`` attribute.
```

IMO sounds a little bit more natural this way.

Soheab (changes requested):
This should cover super/burst reactions too.

PR #7496, opened 2022-03-06
Docs/threads docs

AbstractUmbra:
## Summary

This PR adds a section on the guide for threads.

## Checklist

- [ ] This PR fixes an issue.


docs/guide/topics/threads.rst
```diff
@@ -0,0 +1,206 @@
...
+.. currentmodule:: discord
+
+.. _guide_threads:
+
+Threads
+========
+
+If you've been on a public Discord for a while, you've probably seen a :ddocs:`topics/threads` or two in your time.
+
+They act as offshoots of regular text channels, allowing for a focused, but temporary, location to talk about a given topic.
```
Void-ux, line 12:
```suggestion
They act as offshoots of regular text channels, allowing for a focused, but temporary location to talk about a given topic.
```

tailoric: I am pretty sure the original punctuation is correct.
The 'but temporary' is an off clause, it is not essential to the meaning of the sentence, therefore has to be separated by two commas. You know this, because leaving the off clause out still keeps a valid sentence .
`They act as offshoots of regular text channels, allowing for a focused location to talk about a given topic.`
[See point 3 from this site.](https://owl.purdue.edu/owl/general_writing/punctuation/commas/extended_rules_for_commas.html)

IAmTomahawkx (approved):

Rapptz (changes requested):
This is mostly fine, but can we add a section about how to deal with threads when you receive a message from them.

AbstractUmbra:
> This is mostly fine, but can we add a section about how to deal with threads when you receive a message from them.

Started on this but needs more fleshing out. Having a bit of issues trying to think of gotchas relating to this to add in. Suggestions on this are very welcome!

docs/guide/topics/threads.rst
```diff
@@ -0,0 +1,245 @@
...
+
+
+Browsing threads in a channel
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can browse threads within :class:`~TextChannel` instances with the following methods:
+    - :attr:`~TextChannel.threads`
+    - :meth:`~TextChannel.archived_threads`
+
+The former is a property of the channel instance that returns a list of all valid :class:`~Thread` instances
```
tailoric, line 202: What does "valid" mean here?

Also should be important to mention in this section that `TextChannel.threads` will only contain active threads (excluding all archived threads which can only be fetched with the `archived_threads` iterator or `fetch_thread()`. Not even `channel.get_thread()` will get them since archived threads are not cached, they're evicted from the cache once they are archived even.)

docs/guide/topics/threads.rst
```diff
@@ -0,0 +1,245 @@
...
+
+We've chosen here to mention the author as soon as we make the thread.
+If we didn't do this, then the thread would be created, but we wouldn't necessarily be able to see it!
+
+Note that this mention is only guaranteed to work because we own the thread.
+Mentions from regular users wouldn't work in this thread, because it is not ``invitable``.
+
+You can also use the :meth:`~Thread.add_user` method to add someone, without sending a mention. This requires ``invitable`` to be ``False``.
+
+Unlike a public thread, private threads are not tied to messages, so they won't appear in the channel they're attached to.
```
tailoric, line 107: See other comment, public threads are not necessarily attached to a message

docs/guide/topics/threads.rst
```diff
@@ -0,0 +1,245 @@
...
+
+Simple, right? Let's discuss some of the finer details.
+
+Public threads
+~~~~~~~~~~~~~~~
+
+Public threads can be viewed by anyone who can see the channel they belong to.
+People can browse and join these threads as they please.
+
+All public threads are tied to a specific message. You can use a message that already exists, but you can also send a new message to attach the thread to.
```
tailoric, line 51: They are not necessarily tied to a specific message.
Public threads can be created without spawning from a message.
image

This is also possible in the UI
image
image

docs/guide/topics/threads.rst
```diff
@@ -0,0 +1,245 @@
...
+Adding members to threads
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We touched on it earlier a bit, but now we'll go into how to add people to threads.
+
+The go-to practice for doing this is pinging the member, like so:
+
+.. image:: /images/guide/threads/mention_member_to_thread.png
+
+The API has methods of adding members that do not require pinging them, which is implemented with :meth:`Thread.add_user`.
```
tailoric, line 128: Again kind of misleading wording since it still looks like a ping to the invited person. 


image
image

docs/guide/topics/threads.rst
```diff
@@ -0,0 +1,245 @@
...
+The current accepted values are:
+
+    - 1 hour
+    - 24 hours
+    - 3 days
+    - 1 week
+
+
+.. note::
+    The options for both "3 days" and "1 week" are locked behind server boost level 1 and 2, respectively.
```
tailoric, line 169: Possibly helpful to show how to write a check for if that feature is enabled (or rather if the guild fulfills the boost level)

docs/guide/topics/threads.rst
```diff
@@ -0,0 +1,245 @@
...
+    @bot.command()
+    async def secret_thread(ctx: commands.Context):
+        thread = await ctx.channel.create_thread(
+            name="Secret Birthday Planning!",
+            invitable=False
+        )
+        await thread.send(ctx.author.mention)
+
+We've chosen here to mention the author as soon as we make the thread.
+If we didn't do this, then the thread would be created, but we wouldn't necessarily be able to see it!
```
tailoric, line 100:
```suggestion
If we didn't do this, then the thread would be created, but the user, who invoked the command, wouldn't necessarily be able to see it!
```
"We" could mean the bot but the bot would be the only one to see the thread. 
"We" could also mean the developer writing the command however we want the person using the command to see the thread.

tailoric (commented):
- One inaccuracy about public threads in the guide.
- Otherwise some nitpicks and corrections to the wording.
- Two small wishes to expand on some examples. 

Overall I'd say however the guide is pretty far and well written and should give a complete overview over the capabilities of threads (unless I missed something)

docs/guide/topics/threads.rst
AbstractUmbra: The lines after your comment here do cover the rest of this comment.
I have fixed the "valid" mention locally, thank you

docs/guide/topics/threads.rst
```diff
@@ -0,0 +1,270 @@
...
+The current accepted values are:
+
+    - 1 hour
+    - 24 hours
+    - 3 days
+    - 1 week
+
+
+.. note::
+    The options for both "3 days" and "1 week" are locked behind server boost level 1 and 2, respectively.
```
Soheab, line 182: Don't know since when but this no longer seems to be the case? At least not in the UI ui.

docs/guide/topics/threads.rst
```diff
@@ -0,0 +1,270 @@
...
+            auto_archive_duration=duration,
+            message=ctx.message
+        )
+
+Deleting or archiving a thread
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can delete or force archive threads if you have the :attr:`~Permissions.manage_threads` permission.
+
+If you wish to force archive the thread you can use the :meth:`Thread.edit` method and set the ``archived=True`` keyword argument.
```
Soheab, line 216: The `locked` kwarg could also be mentioned I think. Most people don't knoW about it or what it does.

Soheab (commented):
Just two small things, looks good and has everything relevant otherwise.

docs/guide/topics/threads.rst
```diff
@@ -0,0 +1,270 @@
...
+
+.. note::
+    There is a limit of around 100 members per addition to a thread, so if your role contains more than 100 members they will not be added.
+
+
+Setting the archive duration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Threads automatically archive after a set period of time of inactivity. You can choose this period when creating the thread.
+The current accepted values are:
```
Soheab, line 173: Maybe relevant to the one below but the client only hides it from the channel list now as UI also mentions, not archive it.

docs/guide/topics/threads.rst
AbstractUmbra: You're right it's no longer gated.

tailoric (approved):
I don't see anything to fix or add right now so
LGTM

No767:
I will review this sometime this week for my input

docs/guide/topics/threads.rst
```diff
@@ -0,0 +1,256 @@
...
+            auto_archive_duration=duration,
+            message=ctx.message
+        )
+
+Deleting or archiving a thread
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can delete or force archive threads if you have the :attr:`~Permissions.manage_threads` permission.
+
+If you wish to force archive the thread you can use the :meth:`Thread.edit` method and set the ``archived=True`` keyword argument.
```
No767, line 199:
```suggestion
If you wish to force archive the thread you can use the :meth:`Thread.edit` method and set the `archived` keyword arugment to `True`
```
This above is more explicit now so people won't browse over it

AbstractUmbra: I disagree with this actually. The wording difference makes no change in reading to me.

docs/guide/topics/threads.rst
```diff
@@ -0,0 +1,256 @@
...
+~~~~~~~~~~~~~~~~
+
+Private threads exist as "invite only" threads.
+By default, they only contain the person who created them, but other people can be invited to participate.
+
+Private threads can optionally be 'invitable'.
+If a thread is invitable, it means anyone in that thread can invite others to it.
+Otherwise, only the owner and people with the :attr:`~Permissions.manage_threads` permission can add new people to the thread.
+
+Users with :attr:`~Permissions.manage_threads` ("Moderators") can also see a list of private threads they're not in, similar to public threads.
```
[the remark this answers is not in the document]
AbstractUmbra: Yes? The term for someone who can manage threads is a moderator, it doesn't matter what "staff" may be present.

docs/guide/topics/threads.rst
No767: Ok you do you (mark this as resolved pls)

docs/guide/topics/threads.rst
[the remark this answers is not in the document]
No767: Just to avoid confusion, it would be best to just remove it just to explicitly mention the permission that is needed instead. I could see this getting confusing for some

Soheab (approved):

PR #7499, opened 2022-03-06
Sharding guide

nihaals:
## To-do

- [ ] Link to [Discord docs](https://discord.com/developers/docs/topics/gateway#sharding)
- [ ] Include code examples
- [ ] Include more info about differences between `Client` and `AutoShardedClient`
- [ ] Insert newlines in long lines


docs/guide/sharding/index.rst
```diff
@@ -0,0 +1,31 @@
+.. currentmodule:: discord
+
+.. _guide_sharding
+
+Sharding
+==========
+
+For bots in a large number of guilds, sharding may be required. Sharding is where a subset of the bot's total guilds are processed in each gateway connection. This allows a bot to handle more events, by splitting them by connection and possibly process. Sharding is generally not recommended for bots in less than 1,000 guilds and is required by Discord when a bot is in over 2,500 guilds.
```
Gorialis, line 8: This is a good explanation, but for heading paragraphs I generally recommend trying to break the information into individually digestable, small units. For beginners, it makes the text a bit more approachable, and for novices, it helps quickly determine if the page covers the topic they're expecting.

Not putting all my chips in on this particular presentation, but:
```suggestion
When bots start to get considerably large, the amount of events they have to deal with can start to become problematic.

At high user and guild counts the incoming messages, typing events, and status updates can start to climb to being as frequent as hundreds per second.

To help deal with this large amount of traffic, Discord supports **sharding**, a feature where your bot can split its guilds amongst separate connections, reducing the amount of data each individual connection has to handle.

This not only helps Discord by reducing how much data they need to direct to one place, but also helps us, as we can split our bot's overall work across different connections, environments, or even entirely different machines. 

Discord recommends using sharding once you get over 1,000 guilds, and once you reach beyond 2,500 guilds, it becomes a requirement.

Let's discuss what options we have for setting up sharding within discord.py.
```

docs/guide/sharding/index.rst
```diff
@@ -0,0 +1,31 @@
...
+Specifying shard count and IDs
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You may want to specify a total shard count instead of relying on Discord's recommendation if you find you need more or less shards, or if you want to keep it consistent across restarts of your bot.
+
+.. warning::
+
+    When specifying a shard count, note that each shard must have less than 2,500 guilds.
+
+Specifying shard IDs is useful for bots running as multiple processes. For example, if a bot has 16 shards, you may have one process run shards 0-7 and another process run shards 8-15. These values can be specified with for example an environment variable to keep allow passing different values to each process. If you specify shard IDs, you must also specify a shard count.
```
Gorialis, line 31:
```suggestion
Specifying shard IDs is useful for bots running as multiple processes. For example, if a bot has 16 shards, you may have one process run shards 0-7 and another process run shards 8-15. These values can be specified with, for example, an environment variable to allow passing different values to each process. If you specify shard IDs, you must also specify a shard count.
```

Gorialis (changes requested):
Just copying over my old review comments since they still apply. Also, this document feels a bit short. Not sure what we can do about it at the moment, though.

docs/guide/sharding/index.rst
```diff
@@ -0,0 +1,31 @@
...
+
+Sharding
+==========
+
+For bots in a large number of guilds, sharding may be required. Sharding is where a subset of the bot's total guilds are processed in each gateway connection. This allows a bot to handle more events, by splitting them by connection and possibly process. Sharding is generally not recommended for bots in less than 1,000 guilds and is required by Discord when a bot is in over 2,500 guilds.
+
+Client vs AutoShardedClient
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Both :class:`~Client` and :class:`~ext.commands.AutoShardedClient` have auto sharded variants, :class:`~AutoShardedClient` and :class:`~ext.commands.AutoShardedBot` respectively.
```
HigherOrderLogic, line 13:
```suggestion
Both :class:`~Client` and :class:`~ext.commands.Bot` have auto sharded variants, :class:`~AutoShardedClient` and :class:`~ext.commands.AutoShardedBot` respectively.
```

docs/guide/sharding/index.rst
nihaals: > I agree, I think rewriting it in a form where you can both quickly skim *and* process/understand it in "chunks" makes it more useful for everyone

nihaals:
> I'm not sure how in-depth we want this page to be beyond explaining the concept of sharding. With code examples, do we need to provide anything more than an example with `shard_count` and `shard_ids`?
>
> There's also this suggestion but I'm not sure what we really need to include:
> > A thought about sharding: perhaps it could go over "[clustering]" as well, i.e. splitting shards between processes and having them communicate with each other.
>
> _Originally posted by @XuaTheGrate_

_Originally posted by @nihaals_

nihaals:
> "Clusters" are a term used for larger bots (20k+ guilds) that split up the processes into multiple sub-processes, each handling a handful of shards, for example:
> 
> * Cluster 1 (3k guilds)
>   * Shard 0 (1.5k guilds)
>   * Shard 1 (1.5k guilds)
> * Cluster 2 (3k guilds)
>   * Shard 2 (1.5k guilds)
>   * Shard 3 (1.5k guilds)
> * and so on...
> 
> 
> Some bots have the clusters "communicate" with each other (be it I.P.C. or sharing data through their database, or something rather).
> 
> This is used to reduce the CPU load overall as having all the shards clumped together can cause the process to hang, increasing latency and resources used to keep the bot running. Split processes use the resources more efficiently.

_Originally posted by @XuaTheGrate_

nihaals:
> I think I can extend my explanation of using multiple processes with this to explain why you might actually do that and a bit of how you might do it

_Originally posted by @nihaals_

docs/guide/sharding/index.rst
nihaals: Have we already used "us" and "let's" in other guides? It seems like an important thing to check now to make sure our tone is consistent across the guides. I stuck to avoiding it and using a more typical docs-style tone initially, although since this is a guide, I think either works as long as it's consistent.

PR #7506, opened 2022-03-06
Documenting that guild.chunk() returns list of members.

Scrumpyy:
## Checklist

- [ ] This PR fixes an issue.


Rapptz:
I'm curious if this return type needs to be updated to just `List[Member]`, but this might be outside the scope of this PR. Thanks.

PR #7507, opened 2022-03-06
[Guide] add guide pages for views/modals

bijij:
## Summary

Adds guide pages for views/modals

## Checklist

- [ ] This PR fixes an issue.


Gorialis (approved):
I can't see any problems that stand out to me.

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+components such as :class:`~discord.ui.TextInput`.
+
+This section will detail how to create and use these modals in your code.
+
+
+Defining a Modal
+------------------
+
+Modals share structural similarities with :ref:`Views <guide_interactions_views>`, both are used to design component based UIs 
+and share
```
ShineyDev, line 18:
```suggestion
and share 
```

and share what

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+------------------
+
+Modals share structural similarities with :ref:`Views <guide_interactions_views>`, both are used to design component based UIs 
+and share 
+
+Modals are created by subclassing the :class:`~discord.ui.Modal` class.
+
+For example, the following class definition:
+
+.. code-block:: python3
```
ShineyDev, line 24:
```suggestion
.. code-block:: python
```

deprecated i think

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+
+Defining a Modal
+------------------
+
+Modals share structural similarities with :ref:`Views <guide_interactions_views>`, both are used to design component based UIs 
+and share 
+
+Modals are created by subclassing the :class:`~discord.ui.Modal` class.
+
+For example, the following class definition:
```
ShineyDev, line 22:
```suggestion
For example:
```

continued below

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+        additional_information = discord.ui.TextInput(
+            label="Additional Information", style=discord.TextStyle.long, required=False
+        )
+
+        async def on_submit(self, interaction: discord.Interaction) -> None:
+            ... # do something with the data
+
+            await interaction.response.send_message(f'Thank you {self.name.value}, your submission was recorded.')
+
+produces a modal which appears on Discord as:
```
ShineyDev, line 38:
```suggestion
This produces a modal which appears on Discord as:
```

probably don't continue sentences around blocks of code

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+
+Let's break down the code:
+
+
+Title
+~~~~~~
+
+When defining the class a ``title`` keyword argument can be passed which sets the default title of the modal.
+Instances of the modal class can override this title by setting the ``title`` attribute or passing a ``title`` keyword argument to
+the constructor. Titles are required and are displayed at the top of the modal.
```
ShineyDev, line 50:
```suggestion
the constructor. The title is required and is displayed at the top of the modal.
```

one title

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+In modal classes, class-attributes which are instances of components are used to represent the fields of the modal.
+
+Discord requires a modal contain between 1 and 5 fields. 
+
+.. note:: 
+
+    Each modal fields is displayed in definition order. Our example from above defines ``name`` before it defines ``feedback``, 
+    that order is preserved when the FeedbackForm modal is displayed on Discord.
+
+Fields can also be added or removed from a modal instance using the :meth:`~discord.ui.View.add_item` and :meth:`~discord.ui.View.remove_item` methods.
```
ShineyDev, line 66:
```suggestion
Fields can also be added to or removed from a modal instance using the :meth:`~discord.ui.View.add_item` and :meth:`~discord.ui.View.remove_item` methods.
```

reads nicer

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+~~~~~~~
+
+In our example above we have three class-attributes: ``name``, ``feedback``, and ``additional_information``, all of which are instances of the :class:`~discord.ui.TextInput` class.
+In modal classes, class-attributes which are instances of components are used to represent the fields of the modal.
+
+Discord requires a modal contain between 1 and 5 fields. 
+
+.. note:: 
+
+    Each modal fields is displayed in definition order. Our example from above defines ``name`` before it defines ``feedback``,
```
ShineyDev, line 63:
```suggestion
    Each modal field is displayed in definition order. Our example from above defines ``name`` before it defines ``feedback``, 
```

one fields

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+.. note:: 
+
+    Each modal fields is displayed in definition order. Our example from above defines ``name`` before it defines ``feedback``, 
+    that order is preserved when the FeedbackForm modal is displayed on Discord.
+
+Fields can also be added or removed from a modal instance using the :meth:`~discord.ui.View.add_item` and :meth:`~discord.ui.View.remove_item` methods.
+
+.. warning:: 
+
+    Ensure that class-attribute defined fields do not conflict with attributes or methods defined in the class, as this
```
ShineyDev, line 70: probably tell me what attributes these are and where

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+    will cause unexpected behavior.
+
+
+You can customize individual modal instances with normal attribute access and assignment, For example, to change the label of the ``name`` field to ``"Your Name"``:
```
ShineyDev, line 74:
```suggestion
You can customize individual modal instances with normal attribute access and assignment, For example, to change the label of the ``name`` field to ``"Your Name"`` you can do the following:
```

i don't love "you can do the following" but it needs something

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+
+    form = FeedbackForm()
+    form.name.label = "Your Name"
+
+
+Handling User Responses
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Once the user has clicked `Submit` the :meth:`Modal.on_submit <discord.ui.Modal.on_submit>` callback is called. 
+The callback is passed a new :class:`~discord.Interaction` instance, which requires a response.
```
ShineyDev, line 86: what if i don't respond? what is the timeout?

ShineyDev (changes requested):
Review'd modals.rst.

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,328 @@
...
+
+Decorator-based Components
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In reality, our components are relatively simple, so we can use a helper function decorator instead.
+
+In this case, we're creating a button we can use the :func:`~discord.ui.button` decorator inside our ``View`` like so:
+
+.. code-block:: python3
+    :emphasize-lines: 12-15
```
ShineyDev, line 108:
```suggestion
    :emphasize-lines: 12-19
```

validate this

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,328 @@
...
+Our command might look like this:
+
+.. code-block:: python3
+
+    @discord.app_commands.command()
+    @discord.app_commands.describe(member="The member to ban.")
+    async def ban(interaction: discord.Interaction, member: discord.Member) -> None:
+        """Ban a member from the server."""
+        confirmation = Confirm(interaction.user)
+        await confirmation.send_message(f'Are you sure you want to ban {member.name}?', view=confirmation)
```
ShineyDev, line 156:
```suggestion
        await interaction.send_message(f'Are you sure you want to ban {member.name}?', view=confirmation)
```

unless i'm missing something

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,328 @@
...
+
+Let's make one such view.
+
+
+Designing a Persistent View
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We'll start by creating a class to represent the view, again like above we'll create a 
+subclass of :class:`~discord.ui.View`. We'll also create a :class:`~discord.ui.Select` component,
+which is a dropdown menu using the :func:`@select <discord.ui.select>` decorator. We'll set some
```
ShineyDev, line 187: i don't like the excessive use of "we'll"

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,328 @@
...
+        )
+        async def selector(
+            self,
+            component: discord.ui.Select['RoleSelector'],
+            interaction: discord.Interaction,
+        ) -> None:
+            raise NotImplementedError
+
+This is missing a few important things however, for example the select component 
+needs to know what roles to display in the dropdown, so we'll need to pass in the roles to the component,
```
ShineyDev, line 212: **we'll** work on this

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,328 @@
...
+            raise NotImplementedError
+
+This is missing a few important things however, for example the select component 
+needs to know what roles to display in the dropdown, so we'll need to pass in the roles to the component, 
+using the :meth:`~discord.ui.Select.add_option` method, we can iterate over the roles and add the
+role name as the ``label`` and the role ID as the ``value``.
+
+Additionally, since this view is persistent we'll need to specify a :attr:`~discord.ui.Select.custom_id` for our 
+:class:`~discord.ui.Select` component, which is used to identify the component when a user interacts with it.
+Since we could have multiple role selectors, we'll use the ID of the message the view is attached to as part of the ``custom_id``.
```
ShineyDev, line 218: wheel

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,328 @@
...
+            max_values=1,
+        )
+        async def selector(
+            self,
+            component: discord.ui.Select['RoleSelector'],
+            interaction: discord.Interaction,
+        ) -> None:
+            raise NotImplementedError
+
+We'll also need to add a body to our ``selector`` callback function which will assign the selected role to the user:
```
ShineyDev, line 247: image

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,328 @@
...
+
+Now our view class is complete we can manually attach an instance to a message to retrieve the ID.
+When we're doing this the ``custom_id`` can be set to anything since the view handling will be done 
+when we mark the view as persistent.
+
+.. code-block:: python3
+
+    ROLE_IDS = [490320652230852629, 714516281293799438, 859169678966784031]
+    roles = [guild.get_role(id) for id in ROLE_IDS]
+    await channel.send('Choose your house', view=RoleSelector(0, roles))
```
ShineyDev, line 295:
```suggestion
    await channel.send('Choose your house', view=RoleSelector(336642139381301249, roles))
```

just so it looks like an id

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,328 @@
...
+we can just access the first element directly. We can then use sets to determine what roles the user already has, and then with :meth:`Member.edit <discord.Member.edit>` 
+override the members roles.
+
+Our function body might look like this:
+
+.. code-block:: python3
+    :emphasize-lines: 21-24
+
+    class RoleSelector(discord.ui.View):
+        def __init__(self, guild: discord.Guild, roles: List[discord.Role]) -> None:
```
ShineyDev, line 259:
```suggestion
        def __init__(self, guild_id: int, roles: List[discord.Role]) -> None:
```

josh says snowflake is private

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,328 @@
...
+            super().__init__(timeout=None)
+
+            self.selector.custom_id = f'role_selector_{guild.id}'
```
ShineyDev, line 262:
```suggestion
            self.selector.custom_id = f'role_selector_{guild_id}'
```

continued from above

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,328 @@
...
+
+    client = discord.Client()
+    client.views_set = False
+
+    @client.event
+    async def on_ready():
+        if not client.views_set:
+            guild = client.get_guild(GUILD_ID)
+            roles = [guild.get_role(id) for id in ROLE_IDS]
+            client.add_view(RoleSelector(SELECTOR_MESSAGE_ID, roles))
```
ShineyDev, line 314:
```suggestion
            client.add_view(RoleSelector(GUILD_ID, roles), message_id=SELECTOR_MESSAGE_ID)
```

ope

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,328 @@
...
+with the :meth:`~discord.Client.add_view` method.
+
+.. code-block:: python3
+
+    GUILD_ID = 336642139381301249
+    SELECTOR_MESSAGE_ID = 881049369850306610
+    ROLE_IDS = [490320652230852629, 714516281293799438, 859169678966784031]
+
+    client = discord.Client()
+    client.views_set = False
```
ShineyDev, line 307: use setup_hook?

ShineyDev (changes requested):
Review'd views.rst.

ShineyDev (changes requested):
Can we also rename the image files from modals1 etc.? :)

ShineyDev (approved):

Soheab:
This is missing the recently added dynamic items. If it's still being worked.

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,332 @@
...
+        async def setup_role_selector(self) -> None:
+            await self.wait_until_ready()
+            guild = self.get_guild(GUILD_ID)
+            roles = [guild.get_role() for id in ROLE_IDS]
+            client.add_view(RoleSelector(SELECTOR_MESSAGE_ID, roles), message_id=SELECTOR_MESSAGE_ID)
+
+        async def setup_hook(self) -> None:
+            asyncio.create_task(self.setup_role_selector())
+
+    client = MyClient()
```
IAmTomahawkx, line 321:
```suggestion
    client = MyClient(intents=discord.Intents.default())
```

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,332 @@
...
+When we're doing this the ``custom_id`` can be set to anything since the view handling will be done 
+when we mark the view as persistent.
+
+.. code-block:: python
+
+    ROLE_IDS = [490320652230852629, 714516281293799438, 859169678966784031, 673195396834656257]
+    roles = [guild.get_role(id) for id in ROLE_IDS]
+    await channel.send('Choose your house', view=RoleSelector(0, roles)) # we don't know the message ID yet.
+
+Once we have a message ID we just need to create an instance of our view and attach it to the :class:`~discord.Client`
```
IAmTomahawkx, line 301: (RFC) While a bit out of the scope of this topic, I don't believe that anywhere in the guide we mentions *how* to get message IDs. Is that something worth adding here?

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,332 @@
...
+
+        @discord.ui.button(label="No", style=discord.ButtonStyle.primary)
+        async def no_button(
+            self,
+            interaction: discord.Interaction,
+            button: discord.ui.Button[Confirm],
+        ) -> None:
+            self.result = False
+            self.stop()
+            await interaction.response.defer()
```
IAmTomahawkx, line 130: Need to send a response here, or it'll show as failed on discord.

IAmTomahawkx (changes requested):

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+Modals are a form of interaction which prompt the user for additional information, they appear as a pop-up windows and accept input using
+components such as :class:`~discord.ui.TextInput`.
+
+This section will detail how to create and use these modals in your code.
+
+
+Defining a Modal
+------------------
+
+Modals share structural similarities with :ref:`Views <guide_interactions_views>`, both are used to design component based UIs.
```
aqwuah, line 17:
```suggestion
Modals share structural similarities with :ref:`Views <guide_interactions_views>`, both are used to design component-based UIs.
```

aqwuah (changes requested):
Added some grammar corrections and wording suggestions (some of which are nit-picky so feel free to disagree). Great job on the guides.

docs/guide/interactions/views.rst
bijij: I kind of hoped it was implied.
But i can be more specific.

Idea is the user sends the message first, then knows the ID then uses it.

bijij:
> This is missing the recently added dynamic items. If it's still being worked.

Added a new section on dynamic items

bijij:
Shift enter defaulted to close with comment

aqwuah (approved):
Looks good to me.

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,467 @@
...
+    GUILD_ID = 336642139381301249
+    SELECTOR_MESSAGE_ID = 881049369850306610
+    ROLE_IDS = [490320652230852629, 714516281293799438, 859169678966784031, 673195396834656257]
+
+    class MyClient(discord.Client):
+
+        async def setup_role_selector(self) -> None:
+            await self.wait_until_ready()
+            guild = self.get_guild(GUILD_ID)
+            roles = [guild.get_role() for id in ROLE_IDS]
```
Aryathel, line 316:
```suggestion
            roles = [guild.get_role(id) for id in ROLE_IDS]
```

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,467 @@
...
+when we mark the view as persistent.
+
+.. code-block:: python
+
+    ROLE_IDS = [490320652230852629, 714516281293799438, 859169678966784031, 673195396834656257]
+    roles = [guild.get_role(id) for id in ROLE_IDS]
+    await channel.send('Choose your house', view=RoleSelector(0, roles)) # we don't know the message ID yet.
+
+Once we have a message ID, we just need to create an instance of our view and attach it to the :class:`~discord.Client`
+with the :meth:`~discord.Client.add_view` method.
```
Aryathel, line 303:
```suggestion
Once we have a message ID, we just need to create an instance of our view and attach it to the :class:`~discord.Client`
with the :meth:`~discord.Client.add_view` method. By calling this method in our client setup and using a custom ID on all of the view components, the bot will add this view every time it is started. This is what makes a view persistent.
```
I'm not sure I am perfectly happy with my wording here either, but adding a little more detail describing what exactly this section is for and why this is important for a persistent view seems relevant.

Aryathel (changes requested):
Just a few edits/clarity pieces I think are worth looking at.

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+        additional_information = discord.ui.TextInput(
+            label='Additional Information', style=discord.TextStyle.long, required=False
+        )
+
+        async def on_submit(self, interaction: discord.Interaction) -> None:
+            ... # do something with the data
+
+            await interaction.response.send_message(f'Thank you {self.name.value}, your submission was recorded.')
+
+        async def on_error(self, error: Exception, interaction: discord.Interaction) -> None:
```
No767, line 116: The arguments for `on_error` are incorrect. The proper ones should be noted below:

`(self, interaction: discord.Interaction, error: Exception)`

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+            if not interaction.response.is_done():
```
No767, line 117: As the interaction will be sent once, there isn't a need to set a guard for it

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,467 @@
+.. currentmodule:: discord
+
+.. _guide_interactions_views:
+
+Views
+=======
+
+Views are the primary way to design message component based UIs.
```
No767, line 8: What is a view? What does a view represent? Someone who would read through this page may not understand the concept of a view, as it is not explained yet.

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+.. code-block:: python
+
+    @discord.app_commands.command()
+    async def feedback(interaction: discord.Interaction) -> None:
+        """Send important feedback."""
+        await interaction.response.send_modal(FeedbackForm())
+
+
+Once a modal is sent, a user has the option to submit values. These should be handled either in an implementation of the
+:meth:`~discord.ui.Modal.on_submit` method or via :meth:`Modal.wait() <discord.ui.Modal.wait>`,
```
No767, line 138: Kinda on the fence about including `Modal.wait` usage as I've never seen code that uses the coro

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+in a similar fashion to the :ref:`confirmation prompt View example <guide_interactions_views>`.
+
+.. note::
+
+    It is currently not possible to tell if a user has closed the modal prompt without submitting.
```
No767, line 143: It has never been to my knowledge

docs/guide/interactions/modals.rst
```diff
@@ -0,0 +1,143 @@
...
+            await interaction.response.send_message(f'Thank you {self.name.value}, your submission was recorded.')
+
+This produces a modal which appears on Discord as:
+
+.. image:: /images/guide/interactions/modal_example.png
+
+Let's break down the code:
+
+
+Title
```
No767, line 44: No brief example?

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,467 @@
...
+
+        @discord.ui.button(label='No', style=discord.ButtonStyle.primary)
+        async def no_button(
+            self,
+            interaction: discord.Interaction,
+            button: discord.ui.Button[Confirm],
+        ) -> None:
+            self.result = False
+            self.stop()
+            await interaction.response.defer()
```
No767, line 131: As noted, some sort of response needs to be sent to Discord. Most commonly, deleting the original response is what's implemented (or responding with a message saying that the user chose no)

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,467 @@
...
+
+In our case we're creating a confirmation prompt for our ban command, so we'll want to use the :meth:`InteractionResponse.send_message` method.
+
+Our command might look like this:
+
+.. code-block:: python
+
+    @discord.app_commands.command()
+    @discord.app_commands.describe(member='The member to ban.')
+    async def ban(interaction: discord.Interaction, member: discord.Member) -> None:
```
No767, line 157: I personally don't think a ban command is the best example of showing a confirmation view. You can find my concerns in `?tag nabc`. Instead, I recommend making a more simpler command, such as whether does the user want a cookie or not

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,467 @@
...
+
+To start out, our code might look like this:
+
+.. code-block:: python
+
+    class RoleSelector(discord.ui.View):
+        def __init__(self) -> None:
+            super().__init__(timeout=None)
+
+        @discord.ui.select(
```
No767, line 203: `discord.ui.select` has an kwarg called `cls`. In this, you pass in the type of select that you want. In this case, all we need to do is to pass in `discord.ui.RoleSelect`. The current implementation in this page relies on an old method on doing this and uses the base select type.

No767 (changes requested):
I left some feedback based off of a brief reading of the modal and view pages. Note that it is not a complete read and I may have missed some details along the way.

docs/guide/interactions/views.rst
bijij: `RoleSelect` doesn't provide the same functionality though (it lists all roles). If we'd rather use it then it would make more sense to just replace the entire example.

docs/guide/interactions/views.rst
bijij: Added some extra detail, however I'm not really sure there's much benefit.

docs/guide/interactions/views.rst
bijij: The idea here was to use somewhat practical examples which don't require much additional context. 
I'm not sure I 100% agree with your take on ban commands, (banning external users for example isn't particualrly easy with the UI, (that being said this command doesn't help with that either)).

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,467 @@
...
+
+This section details how to create and use views in various situations.
+
+
+Creating a View
+-----------------
+
+Let's start with a simple example, a confirmation prompt.
+
+By the end of this section, we'll have a generic confirmation prompt that we can use in a variety of situations.
```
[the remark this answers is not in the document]
bijij: Reworded.

docs/guide/interactions/views.rst
bijij: I added a note regarding this, could rework to cover more useful responses though.

bijij (commented):
Addressed some review comments.

docs/guide/interactions/modals.rst
No767: The guard will **always** be ran, so it's unnecessary

docs/guide/interactions/views.rst
No767: If you think about it, when most people don't understand what a view is, they want an core version of what views do. Like adding a button and getting a response back from it. You could rewrite the section to build a simple view that just counts how much time a button has been pressed, or an button that just responses with "Hello World!" back. It doesn't need to be as complex as a confirmation view tied with a ban command. I'd also recommend adding a note that notes to check out the examples for more detailed information.

docs/guide/interactions/views.rst
No767: Sounds good

docs/guide/interactions/modals.rst
No767: Well if that works, then it's fine to leave it there

docs/guide/interactions/modals.rst
No767: I might have misread, and was referring to what a modal does. You can mark this resolved

docs/guide/interactions/views.rst
No767: It would make more sense instead of roles, that they are something unique such as colors imo. this kinda depends if the example should be replaced or not

docs/guide/interactions/modals.rst
AbstractUmbra: Agreed on it being best practice, I would disregard this comment.

docs/guide/interactions/views.rst
AbstractUmbra: I fully agree with Josh on this one. The <cls>Select classes provide no filter (which is a bit of a silly move on Discord's side IMO>. As such I still view this as best practice as we can filter what is shown to the user, and not make an accidental footgun where a user could give themselves an admin role.

docs/guide/interactions/views.rst
bijij: Reworked this to be a /report command instead

docs/guide/interactions/views.rst
bijij: Added some prose, and linked to ddocs

Soheab (approved):
Just two small things that I'm not sure about whether they need to be mentioned, so it's up to you and others. It looks good otherwise.

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,483 @@
...
+    class CounterView(discord.ui.View):
+        def __init__(self) -> None:
+            super().__init__(timeout=None)
+            self.add_item(DynamicCounter())
+
+We'll also need some way to send a message with the view, so let's create a command to do that.
+
+.. code-block:: python
+    
+    @discord.app_commands.command()
```
No767, line 456:
```suggestion
    @app_commands.command()
```

Exact same thing with the other suggestions in regards to `app_commands` package utilization

No767 (changes requested):
I've left some of my suggestions. Overall, the guide is fine but make sure you address the suggestions that are pointed out in this review.

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,483 @@
...
+
+.. code-block:: python
+
+    GUILD_ID = 336642139381301249
+    SELECTOR_MESSAGE_ID = 881049369850306610
+    ROLE_IDS = [490320652230852629, 714516281293799438, 859169678966784031, 673195396834656257]
+
+    class MyClient(discord.Client):
+
+        async def setup_role_selector(self) -> None:
```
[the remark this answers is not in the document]
tailoric: It will not block **because** of the `asyncio.create_task()` in the `setup_hook()`
The warning in the docs is for the case of doing the `wait_until_ready()` in the `setup_hook()` directly.
`asyncio.create_task()` doesn't block the function from continuing like  `wait_for` and `wait_until_ready()` do, it schedules the task on the loop and then continues which prevents the deadlock that would be caused by a `wait_until_ready()`
Test the code yourself and you will see it doesn't deadlock the bot
or here a more minimal example showing that it won't deadlock
```py
#!/usr/bin/env python3

import discord
import config
import asyncio

class Bot(commands.Bot):
    def __init__(self, intents: discord.Intents, **kwargs):
        super().__init__(command_prefix=commands.when_mentioned_or('$'), intents=intents, **kwargs)

    async def ready_waiter(self):
        print("waiting for ready event....")
        await self.wait_until_ready()
        print("ready!!!!")

    async def setup_hook(self):
        asyncio.create_task(self.ready_waiter())

    async def on_ready(self):
        print(f'Logged on as {self.user} (ID: {self.user.id})')


intents.message_content = True
bot = Bot(intents=intents)

# write general commands here

bot.run(config.token)

```

Should show the following output:
```
(.venv) ➜  SetupBot python .\bot.py
2024-02-06 10:48:33 INFO     discord.client logging in using static token
waiting for ready event....
2024-02-06 10:48:34 INFO     discord.gateway Shard ID None has connected to Gateway (Session ID: e40b991f53a805e1a5f4a9148a1509a7).
ready!!!!    
```

The way the guide is doing is fine and actually helpful since it shows how to make a setup part that needs to wait for the cache to be ready, to add the view, which is a bit more complex than the linked example in your comment, which just uses static values for the sake of simplicity and showing what is needed for a persistent view (which means it doesn't need a ready cache) vs. what this example here is doing of showing what to do with more dynamic parts in a persistent view and objects needing to be in cache to actually add the view.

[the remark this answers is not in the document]
No767: @tailoric That is perfectly understandable, but *why* wait until the internal caches have been set in order to then add the view? Unless the role selector view depends on some internal cache to be set, I don't really see an purpose of using `wait_until_ready` in this context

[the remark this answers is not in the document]
tailoric: It does need the cache.
The view relies on a list of roles to get the name and id of the role to add to the select options.
See Lines [277-279 ](https://github.com/Rapptz/discord.py/blob/8a3e3e56e3e7eec10e680cd7f07628384b6217ed/docs/guide/interactions/views.rst?plain=1#L277-L279)this very file. That's why the `wait_until_ready` is needed in this example since this is supposed to be a permanent dropdown with self assignable roles.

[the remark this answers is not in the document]
No767: > wait_until_ready is needed in this example since this is supposed to be a permanent dropdown with self assignable roles.

@tailoric `wait_until_ready` is required because of the guild cache not being set within `setup_hook`. I have validated and tested this code on dpy 2.3.2 using Python 3.12.1 and yes it does require the cache to be set. The reason is not what you described, but rather because the guild is `None` as no data is populated within the cache at that time. Thus it errors out.

docs/guide/interactions/views.rst
```diff
@@ -0,0 +1,483 @@
...
+        async def selector(
+            self,
+            interaction: discord.Interaction,
+            component: discord.ui.Select['RoleSelector'],
+        ) -> None:
+            role_id = int(component.values[0])
+            role = discord.Object(id=role_id, type=discord.Role)
+            roles = set(interaction.user.roles) - set(self.roles) | {role}
+            await interaction.user.edit(roles=roles)
+            await interaction.response.defer()
```
No767, line 295: Instead of deferring the interaction and not acknowledging the response, you should send an message response instead. The contents of the message should tell the user that their role has been changed to the role that they have selected, and this should be ephemeral.

docs/guide/interactions/views.rst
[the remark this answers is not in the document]
tailoric: I don't think what I said was wrong here.
I said the cache needs to be ready to get the roles, which yes in turn you also need to do `get_guild` which also has to be cached so you can use `get_role` on the guild to get the role instances for the list that will be provided to the `__init__` of the view.
But I am also not interested into getting into unnecessary pedant-isms anyway.

The main takeaway and the reason for my reply to your initial review comment was to show and tell you that the `wait_until_ready` was necessary and your suggestion of moving the code directly into the `setup_hook` was not possible with how the view is currently constructed.
I also see the value in showing how to defer the creation of a persistent view, that needs some bot cache objects to be ready, before being re-added to the view store, after a restart, which this example shows off. 

This conversation however also showed that it may be beneficial to elaborate why this example is putting the initialization of the `RoleSelector` view into an `asyncio.create_task` to avoid the deadlock mentioned in the documentation.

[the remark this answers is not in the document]
No767: I agree with in this case. I'll leave it as is

PR #7513, opened 2022-03-06
Add the suppress_embeds parameter to send methods

StockerMC:
## Summary

This PR adds the ability to send messages with all embeds suppressed.

The following methods are modified:
- abc.Messageable.send
- Webhook.send
- SyncWebhook.send
- InteractionResponse.send_message

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -1183,6 +1183,7 @@ class Messageable:
         reference: Union[Message, MessageReference, PartialMessage] = ...,
         mention_author: bool = ...,
         view: View = ...,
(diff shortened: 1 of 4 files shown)


Rapptz:
I didn't end up doing this because I wasn't sure about the state of message flags in editing or sending. Right now since the flags are only for suppressing I think having a bool per flag is probably fine, though at some point when the number of flags exceeds like... 2 we might need to have a bare `flags` kwarg. Obviously this is outside of the scope of this PR though, just giving some contextual history for why this was missed.

Thanks.

PR #7518, opened 2022-03-06
Add example for application commands

VincentRPS:
## Summary

This PR adds a examples for the slash commands, groups and context menus

## Checklist

- [ ] This PR fixes an issue.


Robin5605:
I think it would also be helpful if you included nested groups, just so if people did want to use nested groups in their own code, it would be here. For those who just want regular slash commands, they can also refer to the example

Rapptz (changes requested):
I think this example should use a custom derived class like the other examples do instead of storing things in a global state. Please check that the styling is consistent with other examples as well, such as using the same `on_ready` printing style.

examples/application_commands/slash_command_trees.py
```diff
@@ -0,0 +1,44 @@
...
+
+client = discord.Client()
+
+# initiate our command tree, with the client
+apple_tree = app_commands.CommandTree(client)
+apples_picked = []
+
+@client.event
+async def on_ready():
+
```
Robin5605, line 12:
```
print(f'Logged in as {client.user} (ID: {client.user.id})')
print('------')
```
Just to maintain consistency with the other examples

examples/application_commands/slash_command_trees.py
```diff
@@ -0,0 +1,44 @@
...
+    # add the commands to our guild
```
Robin5605, line 13: Perhaps something more descriptive, it's easy to mistake using the `app_commands.commandTree.command()` decorator as actually adding the command to the guild, perhaps something like 
```
# Sync our local tree with Discord
``` 
could be more helpful

examples/application_commands/slash_command_trees.py
```diff
@@ -0,0 +1,44 @@
+import discord
+from discord import app_commands
+
+client = discord.Client()
```
EvieePy, line 4: As Danny has mentioned consider making this a class.
```py
class Client(discord.Client):
    
    def __init__(self):
        super().__init__()
        
        self.apples_picked = []
        
    async def on_ready(self):
        ...
    

client = Client()


# Do Slash commands here...


client.run('TOKEN')
```

examples/application_commands/slash_command_trees.py
```diff
@@ -0,0 +1,44 @@
...
+
+# initiate our command tree, with the client
+apple_tree = app_commands.CommandTree(client)
```
EvieePy, line 7: Consider renaming this and calling it just `tree`. The novelty is nice, but for an example it could be confusing to the first time user.

examples/application_commands/slash_command_trees.py
```diff
@@ -0,0 +1,44 @@
...
+
+@client.event
+async def on_ready():
+
+    # add the commands to our guild
+    await apple_tree.sync(guild=discord.Object(id=12345))
+
+# initiate the command on top of our coroutine
+@apple_tree.command(guild=discord.Object(id=12345))
+@app_commands.describe(apple='The Apple to Pick') # describe what the option is
```
EvieePy, line 18: Inline comment should be 2 spaces from any code...

examples/application_commands/slash_command_trees.py
```diff
@@ -0,0 +1,44 @@
...
+async def pick(interaction: discord.Interaction, apple: str):
+    """Pick an Apple from a tree""" # the docstring representing the commands description
```
EvieePy, line 20: Consider giving a more indepth run down in one of the docstrings explaining what it does? Just a thought.

examples/application_commands/slash_command_trees.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # make sure the apple isn't already picked
+    if apple in apples_picked:
+        await interaction.response.send_message('This Apple is already picked!')
```
EvieePy, line 24: You never return here if the apple was already picked, meaning it will continue and carry on to the `list.append()`.
You could also use an else statement if you prefer.

examples/application_commands/slash_command_trees.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # add the apple to the list
+    apples_picked.append(apple)
+    await interaction.response.send_message(f'You picked {apple}!')
+
+# initiate the command on top of our coroutine
+@apple_tree.command(guild=discord.Object(id=12345))
```
EvieePy, line 31: Instead of 12345 for guild id, maybe using a `GUILD_ID_HERE` placeholder to make it clearer?

examples/application_commands/slash_command_trees.py
```diff
@@ -0,0 +1,44 @@
...
+@app_commands.describe(apple='The Apple to Plant') # describe what the option is
+async def plant(interaction: discord.Interaction, apple: str):
+    """Plant a Apple""" # the docstring representing the commands description
+
+    # make sure the apple is picked
+    if apple not in apples_picked:
```
EvieePy, line 37: You never return here, meaning it will continue and carry on to the `list.remove()`.
This will lead to a `ValueError` if that "apple" has not been picked already.

You could also use an else statement if you prefer.

examples/application_commands/slash_command_trees.py
```diff
@@ -0,0 +1,44 @@
...
+        await interaction.response.send_message('This Apple hasn\'t been picked yet!')
+
+    # remove the apple from the list
+    apples_picked.remove(apple)
+    await interaction.response.send_message(f'You planted {apple}!')
+
+client.run('token')
```
EvieePy, line 44: 1 newline should be placed at end of file.

VincentRPS: not sure about github, but on my local thing atleast it shows there is a line placed in the end

EvieePy: That's not of the upmost importance. But it is PEP8 compliance.

examples/application_commands/context_menu_trees.py
```diff
@@ -0,0 +1,84 @@
...
+
+@client.tree.context_menu(name='Add friend', guild=guild)
+async def add_friend(interaction: discord.Interaction, member: discord.Member):
+    try:
+        # add the user as a friend
+        client.friends[member.id].append(interaction.user.id)
+    except KeyError:
+        # make the list if the user isn't there
+        client.friends[member.id] = [interaction.user.id]
+    except IndexError:
```
RGBCube, line 37: how would this work? there will just be multiple of the same id's.

examples/application_commands/context_menu_trees.py
```diff
@@ -0,0 +1,84 @@
...
+
+@client.tree.context_menu(name='Add bookmark', guild=guild)
+async def bookmark(interaction: discord.Interaction, message: discord.Message):
+    try:
+        # add the message to bookmarks
+        client.bookmarked_messages[interaction.user.id].append(message)
+    except KeyError:
+        # make the list if the user isn't there
+        client.bookmarked_messages[interaction.user.id] = [message]
+    except IndexError:
```
RGBCube, line 64: same here

RGBCube (changes requested):
`client.friends` is a dict, and wont raise `IndexError`

VincentRPS:
> `client.friends` is a dict, and wont raise `IndexError`

the dict can't but the list can, when doing .append etc.

Rapptz:
Hello.

First, thanks for the PR. However, I've decided not to go ahead with this PR due to both an increase in scope and insufficient quality compared to the rest of the examples. Sorry.

PR #7520, opened 2022-03-06
Create ClientStatus type to improve Member memory usage

LostLuma:
## Summary

Replaces the `_client_status` dict with a slotted class to reduce Member's memory footprint.
From some simple testing this reduces memory usage in my bot to ~960MiB from ~1125MiB.

I typed the attributes as `str` instead of `StatusType` so the `status.setter` in Member works typing wise, please let me know if I should reverse this and put an ignore there instead.

## Checklist

- [ ] This PR fixes an issue.


Rapptz (approved):
This seems more than fine, thanks!

PR #7521, opened 2022-03-06
Add an example for the new modals

Robin5605:
## Summary

It aims to show off the new modals, and the different options and features it has. It also serves to show users how it would work alongside slash commands, and a general bot structure.

## Checklist

- [ ] This PR fixes an issue.


examples/modal.py
```diff
@@ -0,0 +1,67 @@
+import discord
+from discord import app_commands, Object, ui
```
Rapptz, line 2:
```suggestion
from discord import app_commands
```

The examples generally don't do imports like these. Might as well be consistent.

examples/modal.py
```diff
@@ -0,0 +1,67 @@
...
+
+# Just default intents and a `discord.Client` instance
+# We don't need a `commands.Bot` instance because we are not
+# creating text-based commands.
+intents = discord.Intents().default()
```
Rapptz, line 7:
```suggestion
intents = discord.Intents.default()
```

examples/modal.py
```diff
@@ -0,0 +1,67 @@
...
+client = discord.Client(intents=intents)
+
+# We need an `discord.app_commands.CommandTree` instance to register slash commands
+tree = app_commands.CommandTree(client)
+
+# The guild in which this slash command will be registered.
+# As global commands can take up to an hour to propagate, it is ideal
+# to test it in a guild.
```
Rapptz, line 15: Generally, I use the term "application commands" instead of "slash commands" but I guess for examples this is fine.

examples/modal.py
```diff
@@ -0,0 +1,67 @@
...
+test_guild = Object(ID_HERE)
```
Rapptz, line 16:
```suggestion
TEST_GUILD = discord.Object(ID_HERE)
```

examples/modal.py
```diff
@@ -0,0 +1,67 @@
...
+
+@client.event
+async def on_ready():
+    print(f"Connected as {client.user}")
```
Rapptz, line 20: These print statements should be consistent with other examples.

examples/modal.py
```diff
@@ -0,0 +1,67 @@
...
+    # Sync the slash commands with Discord.
+    await tree.sync(guild=test_guild)
```
Rapptz, line 22:
```suggestion
    await tree.sync(guild=TEST_GUILD)
```

examples/modal.py
```diff
@@ -0,0 +1,67 @@
...
+class Feedback(ui.Modal, title='Feedback'):
+    # This will be a short input, where the user can enter their name
+    # It will also be required, denoted by the `required=True` kwarg
+    name = ui.TextInput(
+        label='Name', 
+        style=discord.TextStyle.short, 
+        placeholder='Your name here...',
+        required=True,
+        custom_id='modal_name'
+    )
```
Rapptz, line 33:
```suggestion
    name = ui.TextInput(label='Name', placeholder='Your name here...')
```

There's no need to re-specify defaults. Examples are meant to simple.

examples/modal.py
```diff
@@ -0,0 +1,67 @@
...
+    # This is a longer, paragraph style input, where user can submit feedback
+    # Unlike the name, it is not required. If filled out, however, it will
+    # only accept a maximum of 300 characters, denoted by the
+    # `max_length=300` kwarg.
+    feedback = ui.TextInput(
+        label='What do you think of this new feature?',
+        style=discord.TextStyle.long,
+        placeholder='Type your feedback here...',
+        required=False,
+        custom_id='modal_feedback',
```
Rapptz, line 44:
```suggestion
```

Custom IDs are automatically provided for one-off non-persistent UI components.

examples/modal.py
```diff
@@ -0,0 +1,67 @@
...
+        max_length=300
```
Rapptz, line 45:
```suggestion
        max_length=300,
```

Trailing comma is needed to appease black.

examples/modal.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+
+    async def on_submit(self, interaction: discord.Interaction):
+        # Self explanatory, this will be called when the modal is submitted
+        await interaction.response.send_message(f'Thanks for your feedback, {self.name.value}!', ephemeral=True)
+
+    async def on_error(self, error: Exception, interaction: discord.Interaction) -> None:
+        # Also self explanatory, this will be called when an error happens.
```
Rapptz, line 53: I'd like to avoid using terminology like "self-explanatory" since it can be misconstrued. It's better to explain if possible or just leave it out if it's actually self-explanatory.

examples/modal.py
```diff
@@ -0,0 +1,67 @@
...
+        await interaction.response.send_message('Oops! Something went wrong.', ephemeral=True)
```
Rapptz, line 54: I feel this implementation should do something with the error.

examples/modal.py
```diff
@@ -0,0 +1,67 @@
...
+        # Also self explanatory, this will be called when an error happens.
+        await interaction.response.send_message('Oops! Something went wrong.', ephemeral=True) 
+
+
+@tree.command(guild=test_guild, description="Submit feedback")
+async def feedback(interaction: discord.Interaction):
+    """Slash command that initiates the modal.
+    
+    Usage: /feedback
+    """
```
Rapptz, line 62:
```suggestion
```

examples/modal.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # Send the modal with an instance of our `Feedback` class
+    await interaction.response.send_modal(Feedback())
+
+client.run("TOKEN")
```
Rapptz, line 67:
```suggestion
client.run('token')

```

Consistency with other examples.

Rapptz (changes requested):
Hello. Thanks for the PR.

I think this is a good starting point but there are a few changes that should be done.

Robin5605:
Should we also move this modal example, a slash command example (I believe one is already PR'd), the context menu example (I can make one if no one else does within the next few days), and perhaps future additions that don't necessarily fit in the views folder or the general examples folder? We could move them into a `interactions` folder perhaps, since all 3 of the current new features have to do with interactions.

examples/modal.py
```diff
@@ -0,0 +1,69 @@
...
+        # Make sure we know what the error actually is
+        traceback.print_tb(error.__traceback__)
+
+
+@tree.command(guild=TEST_GUILD, description="Submit feedback")
+async def feedback(interaction: discord.Interaction):
+    # Send the modal with an instance of our `Feedback` class
+    await interaction.response.send_modal(Feedback())
+
+client.run('token')
```
Rapptz, line 69:
```suggestion
client.run('token')

```

Rapptz (approved):
I forgot one more minor consistency thing. Generally speaking we use subclasses instead of bare `Client`. I think for this example this is okay to deviate from though so it has no bearing on the PR. Thanks for the work on this one.

Rapptz:
As for directory structure, these can be shuffled around later I feel.

PR #7528, opened 2022-03-06
feat: ability to use asyncio.run

arHSM:
## Summary

This PR does the following things
makes use of `asyncio.run` for the `discord.Client.run` method
`discord.Client.start` if you don't want d.py to automatically call `asyncio.run`
blah blah blah

## Checklist
- [ ] This PR fixes an issue.


Rapptz (changes requested):
This PR removed the ability for people to set a custom `aiohttp.BaseConnector`.

I'm still not overly convinced on removing the `loop` caching and calling `get_running_loop` all the time. It might be better to cache the loop property at an appropriate time to do so, similar to how HTTPClient knows when its asynchronous initialisation is.

arHSM:
right the `BaseConnector`, I was actually thinking about it, till then this was the only safe way to assure that the [`loop`](https://github.com/aio-libs/aiohttp/blob/184274d9b28bbfa06ac60e48bf286a761c6a6cb0/aiohttp/connector.py#L226) attribute in the [base connector and the current loop are the same](https://github.com/aio-libs/aiohttp/blob/184274d9b28bbfa06ac60e48bf286a761c6a6cb0/aiohttp/helpers.py#L279-L294), do note that I was using aiohttp3.4.7post0 as per the requirements.txt


the newer version i.e 3.8 uses [`asyncio.get_running_loop`](https://github.com/aio-libs/aiohttp/blob/916b3eecda825cd42415b6f8821c035647baf890/aiohttp/connector.py#L207), which then again you can't pass it in the `__init__` method of the `Client` because there is no loop at that time.
But you can always pass it as a kwarg in `Client.run`


As for caching the loop, I guess we can do that 😩

arHSM:
Aigth @Rapptz I have re-added the loop attribute, and as for the connector...
Users will have to do something like this
```py
from discord import Client
from asyncio import run

client = Client(...)


async def main():
    await client.start("token", connector=...)

run(main)
```

Rapptz:
This unfortunately has merge conflicts now.

discord/app_commands/tree.py
```diff
@@ -23,6 +23,7 @@
 """
 
 from __future__ import annotations
+import asyncio
```
Rapptz, line 26: Unused?

discord/client.py
```diff
@@ -220,30 +178,23 @@ class Client:
...
-        The event loop that the client uses for asynchronous operations.
     """
 
     def __init__(
         self,
-        *,
-        loop: Optional[asyncio.AbstractEventLoop] = None,
         **options: Any,
     ):
+        self.loop = MISSING
```
Rapptz, line 187:
```suggestion
        self.loop: asyncio.AbstractEventLoop = MISSING
```

discord/client.py
```diff
@@ -485,7 +436,7 @@ async def before_identify_hook(self, shard_id: Optional[int], *, initial: bool =
 
     # login state management
 
-    async def login(self, token: str) -> None:
+    async def login(self, token: str, *, connector: Optional[aiohttp.BaseConnector] = None) -> None:
```
Rapptz, line 439: No, this should be passed at `__init__`.

Again, if this ends up erroring out then that's *fine*, the error is propagated to the user and anyone with sufficient knowledge of asyncio and aiohttp would know how to fix it.

discord/ext/commands/cooldowns.py
```diff
@@ -297,7 +297,7 @@ class _Semaphore:
 
     def __init__(self, number: int) -> None:
         self.value: int = number
-        self.loop: asyncio.AbstractEventLoop = asyncio.get_event_loop()
+        self.loop: asyncio.AbstractEventLoop = asyncio.get_running_loop()
```
Rapptz, line 300: This change will break decorators.

discord/ext/tasks/__init__.py
```diff
@@ -752,15 +743,6 @@ def loop(
...
-            func,
-            seconds=seconds,
-            minutes=minutes,
-            hours=hours,
-            count=count,
-            time=time,
-            reconnect=reconnect,
-            loop=loop,
-        )
+        return Loop[LF](func, seconds=seconds, minutes=minutes, hours=hours, count=count, time=time, reconnect=reconnect)
```
Rapptz, line 746: What's the point of this change?

discord/gateway.py
```diff
@@ -153,7 +153,7 @@ def run(self) -> None:
             if self._last_recv + self.heartbeat_timeout < time.perf_counter():
                 _log.warning("Shard ID %s has stopped responding to the gateway. Closing and restarting.", self.shard_id)
                 coro = self.ws.close(4000)
-                f = asyncio.run_coroutine_threadsafe(coro, loop=self.ws.loop)
+                f = asyncio.run_coroutine_threadsafe(coro, self.ws.loop)
```
Rapptz, line 156: What's the point of this change?

discord/gateway.py
```diff
@@ -166,7 +166,7 @@ def run(self) -> None:
             data = self.get_payload()
             _log.debug(self.msg, self.shard_id, data['d'])
             coro = self.ws.send_heartbeat(data)
-            f = asyncio.run_coroutine_threadsafe(coro, loop=self.ws.loop)
+            f = asyncio.run_coroutine_threadsafe(coro, self.ws.loop)
```
Rapptz, line 169: What's the point of this change?

discord/gateway.py
```diff
@@ -872,7 +872,7 @@ async def from_client(cls: Type[DVWS], client: VoiceClient, *, resume=False, hoo
         gateway = 'wss://' + client.endpoint + '/?v=4'
         http = client._state.http
         socket = await http.ws_connect(gateway, compress=15)
-        ws = cls(socket, loop=client.loop, hook=hook)
+        ws = cls(socket, loop=client.client.loop, hook=hook)
```
Rapptz, line 875: This feels incorrect.

Rapptz (changes requested):
I feel there are unfortunately a lot of questionable changesets here.

discord/ext/tasks/__init__.py
arHSM: The `loop` attribute was removed from this class

discord/gateway.py
arHSM: This is when I removed the `loop` attribute and then re-added it

discord/gateway.py
arHSM: This is also when I removed the `loop` attribute and then re-added it

discord/ext/commands/cooldowns.py
arHSM: even if I change this back to `get_event_loop`, in future Python versions when `get_event_loop` will act as an alias to `get_running_loop`, this will break, so I think we need a better solution

Rapptz: Actually, I double checked and this change is fine since these are only created within an async function.

Rapptz (approved):
Testing this seems to work fine, though the clean up noisy and not as clean as I would hope for it to be. I'm sure that can be improved in the future. For now, I'm willing to give this the OK.

Thanks for the PR.

PR #7530, opened 2022-03-06
Fixed audio delay when calling the after function

Serious-senpai:
## Summary
After the source has been exhausted or an error occurred, the `after` function is supposed to be called immediately. However, sometimes `source.cleanup()` may take several seconds to finish, which causes a noticeable delay when playing audio.

A possible solution is to call `self._call_after()` before `self.source.cleanup()` in `AudioPlayer.run()`.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/player.py
+++ b/discord/player.py
@@ -694,8 +694,8 @@ class AudioPlayer(threading.Thread):
             self._current_error = exc
             self.stop()
         finally:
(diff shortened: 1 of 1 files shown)


Rapptz (approved):
Thanks. LGTM.

PR #7532, opened 2022-03-06
Add Interaction.locale

LeoCx1000:
## Summary
Added `Interaction.locale`, being an optional Locale enum, to the Interaction class, as discord now [provides this data](https://discord.com/developers/docs/interactions/receiving-and-responding#:~:text=locale%3F***,the%20invoking%20user).
It's optional as it's [not available on ping interactions](https://discord.com/developers/docs/interactions/receiving-and-responding#:~:text=***%20This%20is%20available%20on%20all%20interaction%20types%20except%20PING).

## Note
*Not sure if the changes in `types/interactions.py` are in the right place.*

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/interactions.py
+++ b/discord/interactions.py
@@ -103,9 +103,8 @@ class Interaction:
         for 15 minutes.
     data: :class:`dict`
         The raw interaction data.
-    locale: Optional[:class:`Locale`]
-        The locale of the user invoking the interaction. This field
-        is not available on `InteractionType.ping` interactions.
+    locale: :class:`Locale`
+        The locale of the user invoking the interaction.
     guild_locale: Optional[:class:`Locale`]
         The preferred locale of the guild the interaction was sent from, if any.
     """
@@ -150,12 +149,6 @@ class Interaction:
         self.guild_id: Optional[int] = utils._get_as_snowflake(data, 'guild_id')
         self.application_id: int = int(data['application_id'])
 
-        self.locale: Optional[Locale]
-        try:
-            self.locale = try_enum(Locale, data['locale'])
-        except KeyError:
-            self.locale = None
-
         self.message: Optional[Message]
         try:
             # The channel and message payloads are mismatched yet handled properly at runtime
@@ -166,6 +159,7 @@ class Interaction:
         self.user: Union[User, Member] = MISSING
         self._permissions: int = 0
 
+        self.locale: Locale = try_enum(Locale, data.get('locale', 'en-US'))
         self.guild_locale: Optional[Locale]
 
         # TODO: there's a potential data loss here
--- a/discord/interactions.py
+++ b/discord/interactions.py
@@ -159,7 +159,7 @@ class Interaction:
         self.user: Union[User, Member] = MISSING
         self._permissions: int = 0
 
-        self.locale: Locale = try_enum(Locale, data.get('locale'))
+        self.locale: Locale = try_enum(Locale, data.get('locale', 'en_US'))
         self.guild_locale: Optional[Locale]
 
         # TODO: there's a potential data loss here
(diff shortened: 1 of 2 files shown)


StockerMC:
It would probably make sense for this PR to add `Interaction.guild_locale` as well. I would also suggest not making `locale` optional (i.e. using `data.get(...)`), and instead making `guild_locale` optional.

discord/interactions.py
```diff
@@ -103,6 +104,8 @@ class Interaction:
         for 15 minutes.
     data: :class:`dict`
         The raw interaction data.
+    locale: Optional[:class:`Locale`]
+        The locale of the interaction.
```
StockerMC, line 108: It would be good to specify that this is the user's locale.

LeoCx1000: Will do! I completely missed the `guild_locale` so I'll add that too.

Which locale should I default to if none is given (as in, in the ping type interaction) or what would be the best way to handle that.

StockerMC: I'm not sure defaulting to a specific language would be the best here. `try_enum` could return an unknown enum with a value of `None` if the locale wasn't provided, which should be okay considering it will be provided most of the time. You could have a note in the docs saying that it will be an unknown value if the interaction was a ping interaction, or something similar.

LeoCx1000: I see, I tried passing None but,
```python
  File "--------\discord.py\main.py", line 2, in <module>
    print(try_enum(Locale, None))
TypeError: __str__ returned non-string (type NoneType)
```
So I went back to defaulting 'en-US' which, I don't really like, but it's what `Guild.preferred_locale` does too.

discord/types/interactions.py
```diff
@@ -201,6 +201,7 @@ class ModalSubmitInteractionData(TypedDict):
 class _BaseInteractionOptional(TypedDict, total=False):
     guild_id: Snowflake
     channel_id: Snowflake
+    locale: str
```
StockerMC, line 204:
```suggestion
    locale: str
    guild_locale: str
```

discord/interactions.py
```diff
@@ -164,12 +173,16 @@ def _from_data(self, data: InteractionPayload):
                 # The fallback to Object for guild causes a type check error but is explicitly allowed here
                 self.user = Member(state=self._state, guild=guild, data=member)  # type: ignore
                 self._permissions = self.user._permissions or 0
+
+            self.guild_locale = try_enum(Locale, data.get('guild_locale', 'en-US'))
```
Rapptz, line 177: It might be better to move this up and use try except instead of storing it here.

Rapptz (changes requested):
There might be a need for a Locale.unknown value but this might be outside the scope of this PR.

discord/interactions.py
LeoCx1000: Fixed, I... think. Unless I misunderstood instructions >_<

discord/interactions.py
```diff
@@ -164,12 +177,16 @@ def _from_data(self, data: InteractionPayload):
...
```
Rapptz, line 181:
```suggestion
```

discord/interactions.py
```diff
@@ -164,12 +177,16 @@ def _from_data(self, data: InteractionPayload):
...
         else:
             try:
                 self.user = User(state=self._state, data=data['user'])  # type: ignore - The key is optional and handled
             except KeyError:
                 pass
 
+            self.guild_locale = None
+
```
Rapptz, line 189:
```suggestion
```

Rapptz (changes requested):

Rapptz (approved):

PR #7542, opened 2022-03-06
Add Attachment.ephemeral

StockerMC:
## Summary

This PR adds the `Attachment.ephemeral` attribute to determine whether an attachment was sent with an ephemeral message.

## Checklist

- [ ] This PR fixes an issue.


discord/message.py
```diff
@@ -170,10 +170,26 @@ class Attachment(Hashable):
     description: Optional[:class:`str`]
         The attachment's description. Only applicable to images.
 
+        .. versionadded:: 2.0
+    ephemeral: bool
```
Rapptz, line 174:
```suggestion
    ephemeral: :class:`bool`
```

PR #7545, opened 2022-03-06
Add client.setup_hook

GnomedDev:
## Summary

This adds `discord.Client.setup_hook`, a simple method to be overwritten by the user to provide an asynchronous startup point. This has been in [enhanced-discord.py](https://github.com/iDevision/enhanced-discord.py/blob/03c51210a54c529fd26011d0ca6dcab878fee2b7/discord/client.py#L635-L648) for a while and we have received quite a lot of praise for it. It solves the issue of many new users doing asynchronous setup on_ready, which may be called multiple times, or after other events are received and can even be overwritten without subclassing due to how `@client.event` blindly overwrites methods on the client instance. 
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -649,7 +649,7 @@ class Client:
         await self.login(token)
         await self.connect(reconnect=reconnect)
 
-    async def setup_hook(self) -> Any:
+    async def setup_hook(self) -> None:
         """|coro|
 
         A coroutine to be called to setup the bot, by default this is blank.
--- a/discord/client.py
+++ b/discord/client.py
@@ -488,7 +488,8 @@ class Client:
     async def login(self, token: str) -> None:
         """|coro|
 
-        Logs in the client with the specified credentials.
+        Logs in the client with the specified credentials and
+        calls the :meth:`setup_hook`.
 
 
         Parameters
@@ -511,6 +512,8 @@ class Client:
 
         data = await self.http.static_login(token.strip())
         self._connection.user = ClientUser(state=self._connection, data=data)
+
+        await self.setup_hook()
 
     async def connect(self, *, reconnect: bool = True) -> None:
         """|coro|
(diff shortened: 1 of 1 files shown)


Rapptz:
I think we were debating an event called on_ready_once instead of this but I don't inherently think this is inferior or bad. The event allows listeners to piggy back off of it if they were added before ready was fired. However with the pending change of making it async I'm not sure how that interaction I mentioned would hold.

I'll have to mull over this a bit. Note that docs building is failed, might be worth fixing this.

LostLuma:
I think both of these names could be a bit unclear, with there already being the `before_identify_hook` which has similar behavior it could be called `before_connect_hook` or something else with a `_hook` suffix to communicate its purpose better.

discord/client.py
```diff
@@ -644,8 +644,25 @@ async def start(self, token: str, *, reconnect: bool = True) -> None:
...
+        """|coro|
+
+        A coroutine to be called to setup the bot, by default this is blank.
+
+        To perform asynchronous setup after the bot is logged in but before
+        it has connected to the Websocket, overwrite this coroutine.
+
+        This is only called once, and will be called before any events are
+        dispatched, making it a better solution than doing such setup in the
+        :func:`on_ready` event.
```
Rapptz, line 660:
```suggestion
        :func:`~discord.on_ready` event.
```

This will fix the build error.

discord/client.py
```diff
@@ -644,8 +644,25 @@ async def start(self, token: str, *, reconnect: bool = True) -> None:
             An unexpected keyword argument was received.
         """
         await self.login(token)
+        await self.setup()
         await self.connect(reconnect=reconnect)
 
+    async def setup(self) -> Any:
```
Rapptz, line 650:
```suggestion
    async def setup_hook(self) -> None:
```

Rapptz (changes requested):
Right now this is only called from the `Client.start` method, which is purely meant to be called from a convenience point of view. I suggest, instead, that this is moved to `login` and documented as such. This would allow it to work in `AutoShardedClient` as well.

discord/client.py
GnomedDev: Why shouldn't this return Any, the return from `setup_hook` isn't used so shouldn't be enforced to be a specific type imo

Rapptz: Precisely because it's unused.

Rapptz (approved):
LGTM. Thanks.

PR #7560, opened 2022-03-06
Create basic_modal.py

dannynotsmart:
Simple example used for showing how modals are used.

## Summary

This example serves to show how you would use modals in your code. I chose to make it be a suggestion bot because majority of the users I feel will either go with mod applications, reports, or suggestions.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/basic_modal.py
+++ b/examples/basic_modal.py
@@ -37,7 +37,6 @@ class Suggestions(discord.ui.Modal, title='Server Suggestions'):
     )
 
     async def on_submit(self, interaction: discord.Interaction):
-
         suggestion = discord.Embed(
             title='New Suggestion',
             description=f'Suggestion from {interaction.user.mention}',
--- a/examples/basic_modal.py
+++ b/examples/basic_modal.py
@@ -2,17 +2,17 @@ from discord import app_commands
 
 import discord
 
-my_guild_id = 1234567 #id used for registering slash commands in a specific guild 
-suggestion_channel_id = 1234567 #where to send the suggestion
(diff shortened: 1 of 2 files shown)


examples/suggestions_modal.py
```diff
@@ -0,0 +1,81 @@
...
+            'Your suggestion has been sent successfully!', ephemeral=True
+        )
+
+
+client = SuggestionsBot()
+
+# We create a slash command that opens the modal
+@client.tree.command(
+    guild=discord.Object(id=my_guild_id),
+    name='suggest'
```
pikaninja, line 74: this isn't neccesary

dannynotsmart: ah my bad

dannynotsmart: i have updated the file

examples/basic_modal.py
```diff
@@ -0,0 +1,78 @@
...
+
+    reasoning = discord.ui.TextInput(
+        label='Provide a reasoning for this suggestion.',
+        min_length=10,
+        max_length=100,
+        style=discord.TextStyle.paragraph,
+    )
+
+    async def on_submit(self, interaction: discord.Interaction):
+
```
StockerMC, line 40:
```suggestion
```

dannynotsmart: Oops!

Rapptz:
Hello.

This example seems like a duplicate effort of #7521. Since that one was first, I think I'll close this one in favour of the other one. Sorry about this.

PR #7561, opened 2022-03-06
Fix `HelpCommand.invoked_with` raising an error if the context hasn't been set

StockerMC:
## Summary

This PR fixes `HelpCommand.invoked_with` by checking whether `ctx` is `MISSING` instead of `None`.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/help.py
+++ b/discord/ext/commands/help.py
@@ -45,6 +45,8 @@ __all__ = (
     'MinimalHelpCommand',
 )
 
+MISSING = discord.utils.MISSING
+
 # help -> shows info of bot on top/bottom and lists subcommands
 # help command -> shows detailed info of command
 # help command <subcommand chain> -> same as above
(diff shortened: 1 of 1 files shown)


discord/ext/commands/help.py
```diff
@@ -406,7 +406,7 @@ def invoked_with(self):
         """
         command_name = self._command_impl.name
         ctx = self.context
-        if ctx is None or ctx.command is None or ctx.command.qualified_name != command_name:
+        if ctx is discord.utils.MISSING or ctx.command is None or ctx.command.qualified_name != command_name:
```
Rapptz, line 409: Might be better to explicitly import this name since it's common.

Rapptz (changes requested):

Rapptz (approved):
Thanks.

PR #7574, opened 2022-03-07
Fix archived threads not getting evicted from cache.

### LINKED ISSUE #7541: Archived threads do not get evicted from Guild.threads cache
### Summary

Thread stays in Guild.threads when it's archived during bot's runtime.

### Reproduction Steps

Check the value of `guild.threads` after each step:
1. Create a public thread.
2. Archive a thread.
3. Check the value of `Guild.threads` and notice that the thread was not evicted from cache

### Minimal Reproducible Code

```python
Here's a few tests and their results:

print(guild.threads)

thread = await channel.create_thread(name="example-thread", type=discord.ChannelType.public_thread)
await asyncio.sleep(0)  # allow the gateway events to come
print(guild.threads)

await thread.edit(archived=True)
await asyncio.sleep(0)  # allow the gateway events to come
print(guild.threads)

await thread.edit(archived=False)
await asyncio.sleep(0)  # allow the gateway events to come
print(guild.threads)

await thread.edit(locked=True)
await asyncio.sleep(0)  # allow the gateway events to come
print(guild.threads)
```

Results:
```
[]
[<Thread id=950038334057644153 name='example-thread' parent=commands owner_id=536260979772882946 locked=False archived=False>]
[<Thread id=950038334057644153 name='example-thread' parent=commands owner_id=536260979772882946 locked=False archived=True>]
[<Thread id=950038334057644153 name='example-thread' parent=commands owner_id=536260979772882946 locked=False archived=False>]
[<Thread id=950038334057644153 name='example-thread' parent=commands owner_id=536260979772882946 locked=True archived=False>]
```

### Expected Results

I expected the thread to be evicted from the cache.

### Actual Results

The thread was not evicted from the cache.

### Intents

discord.Intents.all()

### System Information

λ python -m discord -v
- Python v3.8.10-final
- discord.py v2.0.0-alpha
    - discord.py pkg_resources: v2.0.0a3741+ga909c1ff
- aiohttp v3.7.4.post0
- system info: Windows 10 10.0.19043

### Checklist

### Additional Context

_No response_

nerdguyahmad:
## Summary

Fixes #7541

This PR fixes threads not getting removed from cache after they have been archived.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/state.py
+++ b/discord/state.py
@@ -857,10 +857,13 @@ class ConnectionState:
         if thread is not None:
             old = copy.copy(thread)
             thread._update(data)
+            if thread.archived:
+                guild._remove_thread(thread)
             self.dispatch('thread_update', old, thread)
         else:
             thread = Thread(guild=guild, state=guild._state, data=data)
-            guild._add_thread(thread)
+            if not thread.archived:
+                guild._add_thread(thread)
             self.dispatch('thread_join', thread)
 
     def parse_thread_delete(self, data: gw.ThreadDeleteEvent) -> None:


jgayfer:
Would we ever consider this as optional? I've come to rely on archived threads being cached.

In the scenario where a thread on a message has been archived, and I'd like to either revive it or create a new one, knowing if a thread already exists helps me decide if I need to unarchive, or create a new thread, saving a request.

Rapptz:
I think in order to facilitate that use case the eviction would need to dispatch a thread_remove event so user code can check if the removal reason is due to being archived. Not sure if that would be a proper solution for your use case but it'd allow you to add your own caching layer on top of the library.

The reason why eviction on archival is desirable is mainly because otherwise the memory of threads becomes unbounded since there wouldn't be another way to evict it.

jgayfer:
An event sounds like a reasonable solution.

I hear you, unbounded memory feels bad. Plus since archived threads aren't present in the initial guild create payload, we still end up with some unpredictable behaviour, as threads that have been archived since we connected to the gateway are present in the cache.

dolfies:
If you lose access to a channel the channel's threads also sit stale in cache. IDK if that's within the scope of this PR, though.

PR #7584, opened 2022-03-07
Fix TypeError in Range transformer from None value

diceroll123:
## Summary

As it says on the tin, it fixes TypeErrors from happening in the Range transformer when a None value is used.

## Checklist

- [ ] This PR fixes an issue.


Rapptz (approved):
Thanks.

PR #7586, opened 2022-03-07
Add Client.fetch_application

Soheab:
## Summary

This PR adds a `fetch_application()` method to `Client` and a matching `get_application()` in http.py. This method allows users to get the a `PartialAppInfo` object with an application id. Uses the following path: `/applications/:id/rpc`

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -1978,6 +1978,9 @@ class HTTPClient:
         return self.request(r, json=payload)
 
     # Misc
+    
+    def get_application(self, application_id: Snowflake, /) -> Response[appinfo.PartialAppInfo]:
+        return self.request(Route("GET", "/applications/{application_id}/rpc", application_id=application_id))
 
     def application_info(self) -> Response[appinfo.AppInfo]:
         return self.request(Route('GET', '/oauth2/applications/@me'))
--- a/discord/http.py
+++ b/discord/http.py
@@ -1980,7 +1980,7 @@ class HTTPClient:
     # Misc
 
     def get_application(self, application_id: Snowflake, /) -> Response[appinfo.PartialAppInfo]:
-        return self.request(Route("GET", "/applications/{application_id}/rpc", application_id=application_id))
+        return self.request(Route('GET', '/applications/{application_id}/rpc', application_id=application_id))
 
     def application_info(self) -> Response[appinfo.AppInfo]:
         return self.request(Route('GET', '/oauth2/applications/@me'))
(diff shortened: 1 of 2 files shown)


discord/http.py
```diff
@@ -1979,6 +1979,9 @@ def bulk_edit_guild_application_command_permissions(
 
     # Misc
 
+    def get_application(self, application_id: Snowflake, /) -> Response[appinfo.PartialAppInfo]:
+        return self.request(Route("GET", "/applications/{application_id}/rpc", application_id=application_id))
```
Rapptz, line 1983:
```suggestion
        return self.request(Route('GET', '/applications/{application_id}/rpc', application_id=application_id))
```

Why is this using the /rpc endpoint? Where is this endpoint documented?

Soheab: Discord does not want to document it for some reason. According to [advaith in the discord-developers server](https://discord.com/channels/613425648685547541/697489244649816084/950540148033081354).

Rapptz: I don't think I want to add an undocumented endpoint.

dolfies: FWIW this endpoint has existed for many years and doesn't even need authentication.

Rapptz: I'm sure it has, same with the activities endpoint and other undocumented ones. It's just these have no guarantee attached to them so if they break they stick with the library which has stronger versioning guarantees than Discord (e.g. look how long we had support for user accounts) and that's not good.

dolfies: I agree that there's no guarantee. However, something like this is extremely useful and its better to have it with the small possibility it breaks than to not have it at all (IMO).

Rapptz:
Closing since this is undocumented.

lol219:
But the nsfw application command was undocumented too

Soheab:
> But the nsfw application command was undocumented too and you added it.

That is an [upcoming feature](
https://gist.github.com/Stalruth/8258cd0cd7298ffaed6290bc7ecadb8e#apps-and-api-updates) for application commands. [This one won't ever be documented](https://github.com/discord/discord-api-docs/discussions/4620#discussioncomment-2385323) and is not something new.

lol219:
> > But the nsfw application command was undocumented too and you added it.
> 
> That is an [upcoming feature](
> https://gist.github.com/Stalruth/8258cd0cd7298ffaed6290bc7ecadb8e#ending-questions-to-mason) for application commands. [This one won't ever be documented](https://github.com/discord/discord-api-docs/discussions/4620#discussioncomment-2385323) and is not something new.

Ok

PR #7598, opened 2022-03-08
Fix for `_Bot` Context generic in converters.

Axelancerr:
## Summary
Changes `_Bot` in `converter.py` to use a TypeVar which allows for subclasses to properly typecheck when setting `Context`'s generic.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
I feel like this isn't the proper solution to this problem but I'm unsure why the TypeVar approach was reverted.

Axelancerr:
I've gone ahead and reverted back to the TypeVar after discussion in #lib-dev.

PR #7603, opened 2022-03-09
fix errors in ScheduledEvent and add url property 

z03h:
## Summary

Adds `ScheduledEvent.url` since they behave like invite urls where even a partial url creates a special embed.

add ability to remove cover_image with `image=None`.
add ability to remove end_time with `end_time=None` for voice/stage entity types.
fix `event.channel` attribute error by adding `event.guild` property
fix `data['entity_metadata']` key and attribute error for `location` if `entity_metadata` was missing.
fix `entity_id` using `data['id']` instead of `data['entity_id']`.
fix `invalid channel type for event` error when switching from entity type voice/stage to external by setting channel to `None`.
fix events having both location and channel attributes by setting location to `None` when changing `entity_type` from external to voice/stage.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/scheduled_event.py
+++ b/discord/scheduled_event.py
@@ -132,7 +132,7 @@ class ScheduledEvent(Hashable):
         self.name: str = data['name']
         self.description: str = data.get('description', '')
         self.entity_type = try_enum(EntityType, data['entity_type'])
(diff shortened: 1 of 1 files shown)


Rapptz (approved):
LGTM.

PR #7604, opened 2022-03-09
Update translations, add Crowdin CLI config

Gorialis:
## Summary

This adds a `crowdin.yml` for the `crowdin3` CLI tool, to make it easier to programmatically both upload translation sources and pull down new translations.

For some annoying implementation reasons in the CLI, there's a bit of a caveat in that we have to force use of server sources when downloading, else it tries to resolve downloaded translations into the wrong place, but uploading works as normal.

I will write a GitHub Actions workflow to somewhat automate this process at a later date, but it took me a long time debugging the CLI to even figure out why it was rejecting the translations in the first place, so I'll leave it until later.

At the very least, the Crowdin sources are now manually up to date, so the translation sources no longer reflect the version of the docs from 2 years ago, and the new stuff can start being translated.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
Good work. Could you fix the builds failing or would it cause Sphinx/Crowdin to complain?

Wolfhound905:
Hey @Gorialis, I was actually just looking today about adding Crowdin to my bot for localization. Do you know if there's any tutorials or something that can help me figure it out?

Issue #6652, opened 2021-04-06
AttributeError: _member_count in Guild


ioistired:
### Summary

Guild._member_count is sometimes unset

### Reproduction Steps

I'm summing my guilds' member count in on_shard_ready. My bot got disconnected from the internet, and after I got a new DHCP lease, I got this exception

### Minimal Reproducible Code

```python
# the actual code i'm using saves this to a database but this is easier to test
async def on_shard_ready(self, shard_id):
    member_count = sum(guild.member_count for guild in self.guilds if guild.shard_id == shard_id)
    print(f'Member count for shard {shard_id}: {member_count}')
```

### Expected Results

Member count is printed after a reconnect

### Actual Results

```
Ignoring exception in on_shard_ready
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/discord/client.py", line 343, in _run_event
    await coro(*args, **kwargs)
  File "bot.py", line 74, in on_shard_ready
    member_count = sum(guild.member_count for guild in guilds)
  File "bot.py", line 74, in <genexpr>
    member_count = sum(guild.member_count for guild in guilds)
  File "/usr/local/lib/python3.8/site-packages/discord/guild.py", line 741, in member_count
    return self._member_count
AttributeError: _member_count
```

### Intents

guilds, messages, guild_reactions, emojis

### System Information

- Python v3.8.9-final
- discord.py v1.6.0-final
- aiohttp v3.6.2
- system info: Linux 4.19.0-14-amd64 #1 SMP Debian 4.19.171-2 (2021-01-30)

### Checklist

I have searched the open issues for duplicates.
I have shown the entire traceback, if possible.
I have removed my token from display, if visible.

I'm not sure if the reconnect is related to this issue, but I haven't gotten it on a "cold' boot. Apparently this is supposed to be fixed by https://github.com/Rapptz/discord.py/commit/9f47d7f71c5f42d5f1957cdb53c76841c45327a6.

apple502j:
Are you sure the guild is not unavailable? Most of the attributes cannot be accessed for unavailable guilds.

Rapptz:
`on_shard_ready` is called when a specific shard is ready, however the remaining shard guilds are still loading. The other guilds are still in the `Bot.guilds` mapping so there's a chance they're not fully available when you access them and the library is waiting for the `GUILD_CREATE` stream.

ioistired:
[...]

PR #7605, opened 2022-03-09
Make `Guild.member_count` Optional

AbstractUmbra:
## Summary

According to multiple reports, this key is not always present in a guild payload. Currently the library only sets this attribute if it is not `None` resulting in a rare error during startup or when retrieving guilds.
Potentially a fix for #6652?

I have made this attribute and property `Optional` as a result, to avoid an `AttributeError`.

Sadly I cannot test this, due to potentially requiring an "unavailable" guild.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -972,6 +972,10 @@ class Guild(Hashable):
             Due to a Discord limitation, in order for this attribute to remain up-to-date and
             accurate, it requires :attr:`Intents.members` to be specified.
 
+        .. versionchanged:: 2.0
+
+            Now returns an ``Optional[int]``.
+
         """
         return self._member_count
 
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -512,7 +512,7 @@ class Guild(Hashable):
                 self._add_member(member)
 
         self._sync(guild)
-        self._large: Optional[bool] = None if member_count is None else self._member_count >= 250
+        self._large: Optional[bool] = None if self._member_count is None else self._member_count >= 250
 
         self.owner_id: Optional[int] = utils._get_as_snowflake(guild, 'owner_id')
         self.afk_channel: Optional[VocalGuildChannel] = self.get_channel(utils._get_as_snowflake(guild, 'afk_channel_id'))  # type: ignore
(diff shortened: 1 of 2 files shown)


Rapptz:
I don't know how I feel about the type of `member_count` changing to `Optional[int]`. Python ergonomics with optional types is subpar, especially when the data is always there. Not sure what the solution is but it is unfortunate.

dolfies:
Unless I'm misremembering, the only time `member_count` will not be present is when a guild is unavailable. In that case, can't we just default to 0? ~~You could argue it actually is 0 since no-one can access it.~~

EvieePy:
I don't think 0 makes a lot of sense here. The guild member count is still above 0, regardless of whether it is unavailable or not.
If you were going to return 0 you might as well just return None for some sanity. But I'm not sure... So goodlock with this.

dolfies:
None will make typing UX not amazing.

AbstractUmbra:
I am aware that marking this attribute as `Optional` is a far from ideal solution, however I believe that it's the lesser of two evils, considering the other option (at the moment) is a rare AttributeError that's been affecting people.

I'd rather a little extra work to appease a type checker than it just dying, personally.

discord/guild.py
```diff
@@ -966,14 +964,18 @@ def discovery_splash(self) -> Optional[Asset]:
...
 
         .. warning::
 
             Due to a Discord limitation, in order for this attribute to remain up-to-date and
             accurate, it requires :attr:`Intents.members` to be specified.
 
+        .. versionchanged:: 2.0
+
+            Now returns an ``Optional[int]``.
+
```
Rapptz, line 978: Extra newline.
```suggestion
```

Rapptz (changes requested):
Needs updating migrating.rst

dolfies:
> I am aware that marking this attribute as `Optional` is a far from ideal solution, however I believe that it's the lesser of two evils, considering the other option (at the moment) is a rare AttributeError that's been affecting people.
> 
> I'd rather a little extra work to appease a type checker than it just dying, personally.

That's why 0 is a good middle ground.

Rapptz (approved):

PR #7612, opened 2022-03-10
update ScheduledEventConverter docstring

mysistersbrother:
previous docstring didn't quite match the behaviour and could cause some confusion

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -877,9 +877,9 @@ class GuildStickerConverter(IDConverter[discord.GuildSticker]):
 
 class ScheduledEventConverter(IDConverter[discord.ScheduledEvent]):
     """Converts to a :class:`~discord.ScheduledEvent`.
-
-    All lookups are done for the local guild first, if available. If that lookup
-    fails, then it checks the client's global cache.
+    
+    Lookups are done for the local guild if available. Otherwise, it will
+    check the client's global cache.
 
     The lookup strategy is as follows (in order):
 
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -877,7 +877,6 @@ class GuildStickerConverter(IDConverter[discord.GuildSticker]):
 
 class ScheduledEventConverter(IDConverter[discord.ScheduledEvent]):
     """Converts to a :class:`~discord.ScheduledEvent`.
-    
     Lookups are done for the local guild if available. Otherwise, for a DM context,
     lookup is done by the global cache.
 
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -878,7 +878,7 @@ class GuildStickerConverter(IDConverter[discord.GuildSticker]):
 class ScheduledEventConverter(IDConverter[discord.ScheduledEvent]):
     """Converts to a :class:`~discord.ScheduledEvent`.
     
-    Lookups are done for the local guild if available. In a DM context, then the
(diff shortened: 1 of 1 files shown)


Rapptz:
I'm not sure how this is different?

mysistersbrother:
> I'm not sure how this is different?

The previous docstring said that if the local guild lookup failed it would search the client's global cache. This is supposed to make it more clear that it only searches the global cache in the case of there being no available local guild, i.e., a dm context.

discord/ext/commands/converter.py
```diff
@@ -877,9 +877,9 @@ async def convert(self, ctx: Context[_Bot], argument: str) -> discord.GuildStick
 
 class ScheduledEventConverter(IDConverter[discord.ScheduledEvent]):
     """Converts to a :class:`~discord.ScheduledEvent`.
-
-    All lookups are done for the local guild first, if available. If that lookup
-    fails, then it checks the client's global cache.
+
```
Rapptz, line 880:
```suggestion

```

discord/ext/commands/converter.py
```diff
@@ -877,9 +877,9 @@ async def convert(self, ctx: Context[_Bot], argument: str) -> discord.GuildStick
 
 class ScheduledEventConverter(IDConverter[discord.ScheduledEvent]):
     """Converts to a :class:`~discord.ScheduledEvent`.
-
-    All lookups are done for the local guild first, if available. If that lookup
-    fails, then it checks the client's global cache.
+    
+    Lookups are done for the local guild if available. Otherwise, it will
+    check the client's global cache.
```
Rapptz, line 882:
```suggestion
    Lookups are done for the local guild if available. In a DM context, then the
    lookup is done by the global cache.
```

Rapptz (commented):
In that case I think this wording might be more explicit.

discord/ext/commands/converter.py
```diff
@@ -877,6 +877,7 @@ async def convert(self, ctx: Context[_Bot], argument: str) -> discord.GuildStick
 
 class ScheduledEventConverter(IDConverter[discord.ScheduledEvent]):
     """Converts to a :class:`~discord.ScheduledEvent`.
+    
     Lookups are done for the local guild if available. In a DM context, then the
     lookup is done by the global cache.
```
asheeeshh, line 882:
```suggestion
    Lookups are done for the local guild if available. Otherwise, for a DM context,
    lookup is done by the global cache.
```

asheeeshh: @Rapptz ig this one sounds more clear and understandable.

Rapptz (approved):
Thanks.

Rapptz:
There seems to be trailing whitespace remaining.

discord/ext/commands/converter.py
Rapptz, line 880:
```suggestion

```

discord/ext/commands/converter.py
```diff
@@ -877,9 +877,8 @@ async def convert(self, ctx: Context[_Bot], argument: str) -> discord.GuildStick
 
 class ScheduledEventConverter(IDConverter[discord.ScheduledEvent]):
     """Converts to a :class:`~discord.ScheduledEvent`.
-
-    All lookups are done for the local guild first, if available. If that lookup
-    fails, then it checks the client's global cache.
+    Lookups are done for the local guild if available. Otherwise, for a DM context,
```
Rapptz, line 880:
```suggestion

    Lookups are done for the local guild if available. Otherwise, for a DM context,
```

PR #7617, opened 2022-03-10
Catch TypeError instead of AttributeError

Willy-C:
In the past, the attribute is not set if the guild is unavailable but now `Guild._member_count` is `Optional[int]` 

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/state.py
+++ b/discord/state.py
@@ -973,20 +973,16 @@ class ConnectionState:
         if self.member_cache_flags.joined:
             guild._add_member(member)
 
-        try:
+        if guild._member_count is not None:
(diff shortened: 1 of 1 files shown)


JDJGInc:
Can Confirm the error is TypeError.

ZenoABC:
Can confirm too.

Motzumoto:
Can also confirm

Rapptz (changes requested):
This PR needs to explicitly check if it's None instead of catching TypeError.

Rapptz (approved):
Thanks.

PR #7622, opened 2022-03-10
[Fix] Message.edit ignoring suppress kwarg

### LINKED ISSUE #7621: Message.edit ignores `suppress` kwarg
### Summary

When using the keyworded arg suppress in the Message.edit() method then 

### Reproduction Steps

Have a bot that has the `manage_message` permission.  
Have it try to suppress embeds on its own or other user's messages by editing it `message.edit(suppress=True)`  

### Minimal Reproducible Code

```python
msg = await ctx.send("https://gist.github.com/tailoric/94a58a29d150375584b885e563ed0b0f")
await msg.edit(suppress=True)
```

### Expected Results

The message's embeds getting suppressed (as if pressing the small x on the message in the client)

### Actual Results

Nothing happening. 
Message embeds still are displayed and checking via jishaku the `message.flags.suppress_embeds` is False

### Intents

Intents.all()

### System Information

- Python v3.9.4-final
- discord.py v2.0.0-alpha
    - discord.py pkg_resources: v2.0.0a3795+g89009759
- aiohttp v3.7.4.post0

### Checklist

### Additional Context

I think I already tracked down the problem because the `suppress` kwarg never get actually gets used for setting the message flag:

https://github.com/Rapptz/discord.py/blob/e15415413b24afb13fbb67e4a0c58a35a8ee2a84/discord/message.py#L1307-L1313

See older code version of message.py when it worked:

https://github.com/Rapptz/discord.py/blob/489e5f3288efb9b9e6649447209e1ac1746ccb32/discord/message.py#L1258-L1261

Testing on my own simply doing 

```py
        if suppress is not MISSING:
            flags = MessageFlags._from_value(self.flags.value)
            flags.suppress_embeds = suppress
```

Already solves the problem. 
I can make a pull request for this in a bit if this solution is good enough

tailoric:
## Summary

This PR is should fix #7621 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -1306,6 +1306,7 @@ class Message(Hashable):
         previous_allowed_mentions = self._state.allowed_mentions
         if suppress is not MISSING:
             flags = MessageFlags._from_value(self.flags.value)
+            flags.suppress_embeds = suppress
         else:
             flags = MISSING
 


Rapptz (approved):
Good catch, thanks.

PR #7629, opened 2022-03-10
Add `ConnectionState.message_cache_filter`

### LINKED ISSUE #7620: Add message cache policy/filtering
### Summary

Allow for filtering on which messages are stored in the internal message cache

The core library

### The Problem

The message cache has a fixed size. When you rely on this cache you might not care about about specific types of messages that will dequeue the messages you do care about.

For example you might not care about system/join messages, which can fill up the cache really fast.

### The Ideal Solution

Maybe something that's passed into the client similar to `MemberCacheFlags` but for messages, or a callable that takes a message to evaluate if it should be cached or not. 
Honestly wouldn't know the nicest solution for this.

I'll leave it up for discussion and if this is something that's nice to have, I could create a PR with the suggested solution if no one else feels like it.

### The Current Solution

Don't think there is a proper solution for this that doesn't involve touching the message cache yourself.

### Additional Context

One obvious use case for this is being able to track deleted/edited messages more reliably, which is obviously something discord doesn't want you to do with the new message intents, so maybe this isn't worth the effort unless there's another obvious use case.

GnomedDev:
## Summary

This closes #7620, with a basic predicate function. This is one option which I believe allows the most control, other than a `MessageCacheFlags` or similar which wouldn't be able to account for all of a user's intentions.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -218,8 +218,10 @@ class Client:
     message_cache_filter: Optional[Callable[[:class:`.Message`], :class:`bool`]]
         An optional function called on each message event to control whether to add it to cache.
 
-        This can be used for higher control over the message cache, in order to only cache messages with
-        certain attributes filled, for example. This does not bypass max_messages.
+        This gives greater control over the message cache. 
+        
+        .. note:: 
+            This does not bypass ``max_messages``.
 
         .. versionadded:: 2.0
 
--- a/discord/client.py
+++ b/discord/client.py
@@ -215,6 +215,13 @@ class Client:
         To enable these events, this must be set to ``True``. Defaults to ``False``.
 
         .. versionadded:: 2.0
+    message_cache_filter: Optional[Callable[[:class:`.Message`], :class:`bool`]]
+        Am optional function called on each message event to control whether to add it to cache or discard it.
+
+        This can be used for higher control over the message cache, in order to only cache messages with
+        certain attributes filled, for example. This does not bypass max_messages.
+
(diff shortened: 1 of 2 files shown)


discord/client.py
```diff
@@ -214,6 +214,13 @@ class Client:
         this is ``False`` then those events will not be dispatched (due to performance considerations).
         To enable these events, this must be set to ``True``. Defaults to ``False``.
 
+        .. versionadded:: 2.0
+    message_cache_filter: Optional[Callable[[:class:`.Message`], :class:`bool`]]
+        Am optional function called on each message event to control whether to add it to cache or discard it.
```
Gobot1234, line 219:
```suggestion
        An optional function called on each message event to control whether to add it to cache.
```

discord/client.py
```diff
@@ -214,6 +214,13 @@ class Client:
...
+
+        This can be used for higher control over the message cache, in order to only cache messages with
+        certain attributes filled, for example. This does not bypass max_messages.
```
Gobot1234, line 222:
```suggestion
        This gives greater control over the message cache. 
        
        .. note:: 
            This does not bypass ``max_messages``.
```

z03h:
If the filter func raises an error the client will crash.

GnomedDev:
Should error handling be added (complicating more, possible performance slowdown) or document it with a danger section?

Rapptz:
I don't think I want to move ahead with this, sorry.

PR #7635, opened 2022-03-11
Implement `Widget.presence_count` attribute and fix `Widget` types

m-etroo:
## Summary

 * Implements missing `Widget.presence_count` attribute
 * Fixes the return types of `Widget.invite_url` and `Widget.fetch_invite` to make them Optional
 * Removes `_WidgetOptional` since none of the widget object fields returned by the API are optional
 * Minor documentation changes

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/types/widget.py
+++ b/discord/types/widget.py
@@ -46,15 +46,12 @@ class WidgetMember(User, total=False):
     suppress: bool
 
 
-class _WidgetOptional(TypedDict, total=False):
-    channels: List[WidgetChannel]
-    members: List[WidgetMember]
-
-
-class Widget(_WidgetOptional):
+class Widget(TypedDict):
     id: Snowflake
     name: str
     instant_invite: Optional[str]
+    channels: List[WidgetChannel]
+    members: List[WidgetMember]
     presence_count: int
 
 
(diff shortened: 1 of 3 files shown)


discord/types/widget.py
```diff
@@ -49,13 +49,13 @@ class WidgetMember(User, total=False):
...
     channels: List[WidgetChannel]
     members: List[WidgetMember]
-    presence_count: int
 
 
 class Widget(_WidgetOptional):
     id: Snowflake
     name: str
     instant_invite: str
+    presence_count: int
```
Rapptz, line 58: The `instant_invite` type needs to be `Optional[str]`.

m-etroo: Thanks. Is there a reason why `channels` and `members` are left in the `_WidgetOptional` class? They are not listed as optional fields on the API docs.

Rapptz: Those can probably be moved too, assuming the documentation doesn't lie (dangerous assumption but oh well).

Rapptz (approved):
Thanks.

Rapptz (changes requested):
Sorry, I just realised this is a breaking change.

This needs to be documented in `migrating.rst`.

Rapptz (approved):
Thanks again.

PR #7646, opened 2022-03-11
Change wording of InvalidArgument removal in docs

quackbarc:
## Summary

This fixes the grammar of the `versionchanged` note in the docstrings and Japanese localization about replacing InvalidArgument with TypeError or ValueError .

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -752,8 +752,8 @@ class GuildChannel:
             await channel.set_permissions(member, overwrite=overwrite)
 
         .. versionchanged:: 2.0
-            Will raise :exc:`TypeError` instead of the now-removed
-           ``InvalidArgument``.
+            This function will now raise :exc:`TypeError` instead of
+            ``InvalidArgument``.
 
 
         Parameters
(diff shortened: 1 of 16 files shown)


Rapptz:
I think instead of that wording something like:

```
This function will now raise <exc> instead of ``InvalidArgument``.
``` is more readily understandable.

**Comment by quackbarc** [comment]
That sounds good too. I'll also get the localization thing reverted, thought the text over there had to be changed as well.

**Comment by quackbarc** [comment]
Actually, for the wording how about
```
This function will now raise <exc> instead of the now-removed ``InvalidArgument``.
```
Could the footnotes still be transparent about what `InvalidArgument` was upfront, or would that still hinder understandability?

Rapptz:
I just think it sounds awkward. Its removal is already mentioned in the migrating guide.

quackbarc:
Gotcha.

Rapptz (approved):
Thanks.

PR #7647, opened 2022-03-11
Update `@Client.event` to support aliases

thetimtoy:
## Summary

This addition should break very little existing code; here are a few examples:

```py
@client.event
async def on_ready():
    ...
   
@client.event('on_ready')
async def ready_listener():
    ...

@client.event()
async def on_ready():
    ...
```

NOTE: `@Client.event` is also made safe to call without arguments when used as a decorator.

The only case I can think of where this implementation breaks is if someone was calling the function like this: `client.event(coro=...)` (But this is unlikely).

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -44,6 +44,7 @@ from typing import (
     Tuple,
     TypeVar,
     Union,
+    overload,
 )
 
 import aiohttp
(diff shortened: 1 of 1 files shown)


thetimtoy:
```py
_log.debug('%s has successfully been registered as an event', coro.__name__)
```

I'm not sure whether to log the function's name or the event name on this line.

SebbyLaw:
What is the point of this?

joannx:
Not sure either, if you want a way to have multiple events, listeners are already a thing…

thetimtoy:
> What is the point of this?

To support aliases so that the name of the function doesn't have to be the event name.

kaylynn234:
This is already supported using [Bot.listen](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html?highlight=listener#discord.ext.commands.Bot.listen) and [Cog.listener](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html?highlight=listener#discord.ext.commands.Cog.listener). I don't think there's any need to duplicate the functionality they provide in this way.

thetimtoy:
> This is already supported using [Bot.listen](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html?highlight=listener#discord.ext.commands.Bot.listen) and [Cog.listener](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html?highlight=listener#discord.ext.commands.Cog.listener). I don't think there's any need to duplicate the functionality they provide in this way.

I was expecting this to be brought up. Except that the point of the `Bot.listen` and `Cog.listener` methods is to register listeners, which is not equivalent to `Client.event`. If anything, this should be more incentive to add this functionality; If the commands extension supports it for listeners (for no particular reason), why shouldn't the standard lib? Besides comparing an extension to the main library isn't really a fair comparision.

Rapptz:
The simplicity of the event coroutine is mostly by design right now. I've been kind of stepping away from using the decorator for events over the past few years and instead preferring subclassing instead, which does not have this concept of aliasing.

PR #7670, opened 2022-03-13
Fix Cog+Group subclasses adding subcommands to tree

z03h:
## Summary

Fix Cog subclasses that don't directly subclass app_commands.Group from adding app_commands.Command/Group in the class directly to the tree.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
Hm. I originally wanted to forbid the multi-inheritance case for groups. 

Mainly because `Group` operates only at one level of inheritance and doesn't walk MRO.

PR #7673, opened 2022-03-13
Make cog/extension load/unload methods async

bijij:
## Summary

See title

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/cog.py
+++ b/discord/ext/commands/cog.py
@@ -378,10 +378,10 @@ class Cog(metaclass=CogMeta):
         return not hasattr(self.cog_command_error.__func__, '__cog_special_method__')
 
     @_cog_special_method
-    def cog_load(self) -> None:
-        """A special method that is called when the cog gets loaded.
-
-        This function **can** be a coroutine.
+    async def cog_load(self) -> None:
+        """|maybecoro|
+        
+        A special method that is called when the cog gets loaded.
 
         Subclasses must replace this if they want special asynchronous loading behaviour.
         Note that the ``__init__`` special method does not allow asynchronous code to run
@@ -392,10 +392,10 @@ class Cog(metaclass=CogMeta):
         pass
 
     @_cog_special_method
-    def cog_unload(self) -> None:
-        """A special method that is called when the cog gets removed.
-
-        This function **can** be a coroutine.
+    async def cog_unload(self) -> None:
+        """|maybecoro|
+        
+        A special method that is called when the cog gets removed.
 
         Subclasses must replace this if they want special unloading behaviour.
 
@@ -487,6 +487,10 @@ class Cog(metaclass=CogMeta):
     async def _inject(self, bot: BotBase, override: bool, guild: Optional[Snowflake], guilds: List[Snowflake]) -> Self:
         cls = self.__class__
 
+        # we'll call this first so that errors can propagate without
+        # having to worry about undoing anything
+        await maybe_coroutine(self.cog_load)
+
         # realistically, the only thing that can cause loading errors
         # is essentially just the command loading, which raises if there are
         # duplicates. When this condition is met, we want to undo all what
@@ -522,11 +526,6 @@ class Cog(metaclass=CogMeta):
             for command in self.__cog_app_commands__:
                 # This is already atomic
                 bot.tree.add_command(command, override=override, guild=guild, guilds=guilds)
-
-        try:
-            await maybe_coroutine(self.cog_load)
-        except Exception:
-            pass
 
         return self
 
--- a/discord/ext/commands/cog.py
+++ b/discord/ext/commands/cog.py
@@ -380,7 +380,7 @@ class Cog(metaclass=CogMeta):
     @_cog_special_method
     async def cog_load(self) -> None:
         """|maybecoro|
-        
+
         A special method that is called when the cog gets loaded.
 
         Subclasses must replace this if they want special asynchronous loading behaviour.
@@ -394,7 +394,7 @@ class Cog(metaclass=CogMeta):
     @_cog_special_method
     async def cog_unload(self) -> None:
         """|maybecoro|
-        
+
         A special method that is called when the cog gets removed.
 
         Subclasses must replace this if they want special unloading behaviour.
--- a/discord/ext/commands/cog.py
+++ b/discord/ext/commands/cog.py
@@ -383,7 +383,9 @@ class Cog(metaclass=CogMeta):
 
         This function **can** be a coroutine.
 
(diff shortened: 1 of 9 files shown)


discord/ext/commands/cog.py
```diff
@@ -505,9 +521,14 @@ def _inject(self, bot: BotBase, override: bool, guild: Optional[Snowflake], guil
                 # This is already atomic
                 bot.tree.add_command(command, override=override, guild=guild, guilds=guilds)
 
+        try:
+            await maybe_coroutine(self.cog_load)
+        except Exception:
+            pass
```
bijij, line 527: Do we want errors on load to propagate? we don't do this for unload.

discord/ext/commands/cog.py
```diff
@@ -376,14 +377,29 @@ def has_error_handler(self) -> bool:
         """
         return not hasattr(self.cog_command_error.__func__, '__cog_special_method__')
 
+    @_cog_special_method
+    def cog_load(self) -> None:
+        """A special method that is called when the cog gets loaded.
+
+        This function **can** be a coroutine.
+
+        Subclasses must replace this if they want special loading behaviour.
```
bijij, line 386: Wording here is weird, init can do special loading behaviour anyway.
Unless we want to maybe use this as an opportunity to ban touching init?

doesn't need the super call in here too.

discord/ext/commands/cog.py
```diff
@@ -376,14 +377,29 @@ def has_error_handler(self) -> bool:
         """
         return not hasattr(self.cog_command_error.__func__, '__cog_special_method__')
 
+    @_cog_special_method
+    def cog_load(self) -> None:
```
Rapptz, line 381:
```suggestion
    async def cog_load(self) -> None:
```

discord/ext/commands/cog.py
```diff
@@ -376,14 +377,29 @@ def has_error_handler(self) -> bool:
...
+        """A special method that is called when the cog gets loaded.
+
+        This function **can** be a coroutine.
```
Rapptz, line 384:
```suggestion
        """|maybecoro|

        A special method that is called when the cog gets loaded.
```

discord/ext/commands/cog.py
Rapptz:
```suggestion
        Subclasses must replace this if they want special asynchronous loading behaviour.
        Note that the ``__init__`` special method does not allow asynchronous code to run
        inside it, thus this is helpful for setting up code that needs to be asynchronous.
```

discord/ext/commands/cog.py
```diff
@@ -376,14 +377,29 @@ def has_error_handler(self) -> bool:
...
+
+        Subclasses must replace this if they want special loading behaviour.
+
+        .. versionadded:: 2.0
+        """
+        pass
+
     @_cog_special_method
     def cog_unload(self) -> None:
```
Rapptz, line 393:
```suggestion
    async def cog_unload(self) -> None:
```

discord/ext/commands/cog.py
```diff
@@ -376,14 +377,29 @@ def has_error_handler(self) -> bool:
...
         """A special method that is called when the cog gets removed.
 
-        This function **cannot** be a coroutine. It must be a regular
-        function.
+        This function **can** be a coroutine.
```
Rapptz, line 396: Move this to use `|maybecoro|`.

discord/ext/commands/cog.py
Rapptz: If this is going to be propagated it needs to be in the beginning of the method rather than at the end as it is here. I think propagating is fine since `__init__` propagates. Unload doesn't need to because it's a finalizer, an error that occurs during finalisation is pretty much can't be cleanly handled anyway.

docs/migrating.rst
```diff
@@ -65,6 +65,64 @@ The following have been removed:
...
+            await client.start(TOKEN)
+
+    asyncio.run(main())
+
+A new :meth:`~Client.setup_hook` method has also been added to the :class:`Client` class.
+This method is called after login but before connecting to the discord gateway.
+
+It is intended to be used to setup various bot features in an asynchronous context.
+
+:meth:`~Client.setup_hook` can be defined by either subclassing the :class:`Client` class, or via the :meth:`Client.event` decorator.
```
Rapptz, line 100:
```suggestion
:meth:`~Client.setup_hook` can be defined by subclassing the :class:`Client` class.
```

docs/migrating.rst
```diff
@@ -65,6 +65,64 @@ The following have been removed:
...
+    client.run(TOKEN)
+
+    # using the @Client.event decorator
+    client = discord.Client()
+
+    @client.event
+    async def setup_hook(self):
+        print('This is asynchronous!')
+
+    client.run(TOKEN)
```
Rapptz, line 121: Don't recommend this.

```suggestion
```

docs/migrating.rst
```diff
@@ -1095,6 +1153,55 @@ The following changes have been made:
 Command Extension Changes
 ---------------------------
 
+.. _migrating_2_0_commands_extension_cog_async:
+
+Extension and Cog Loading / Unloading is Now Asynchronous
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+As an extension to :ref:`migrating_2_0_client_async_setup` changes the loading and unloading of extensions and cogs is now asynchronous.
```
Rapptz, line 1161:
```suggestion
As an extension to the :ref:`asyncio changes <migrating_2_0_client_async_setup>` the loading and unloading of extensions and cogs is now asynchronous.
```

docs/migrating.rst
```diff
@@ -1095,6 +1153,55 @@ The following changes have been made:
...
+
+.. code:: python
+
+    #before
+    bot.load_extension('my_extension')
+
+    #after using setup_hook
+    @bot.event
+    async def setup_hook():
+        await bot.load_extension('my_extension')
```
Rapptz, line 1194: We're mostly moving away from the free form decorators for a while now so it's best to not recommend this.

examples/basic_voice.py
```diff
@@ -132,10 +132,13 @@ async def ensure_voice(self, ctx):
                    description='Relatively simple music bot example',
                    intents=intents)
 
+@bot.event
+async def setup_hook():
+    await bot.add_cog(Music(bot))
+
```
Rapptz, line 138: Again

Rapptz (changes requested):

Rapptz (approved):
LGTM.

PR #7685, opened 2022-03-14
Rework help command to avoid a deepcopy on invoke

bijij:
## Summary

This PR is my first go-around at reworking help command to avoid a deepcopy.

I'm not 100% sure on the implementation here, but it's an option we can consider.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/context.py
+++ b/discord/ext/commands/context.py
@@ -354,6 +354,7 @@ class Context(discord.abc.Messageable, Generic[BotT]):
         """
         from .core import Group, Command, wrap_callback
         from .errors import CommandError
+        from .help import _context
 
         bot = self.bot
         cmd = bot.help_command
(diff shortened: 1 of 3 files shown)


discord/ext/commands/help.py
```diff
@@ -498,31 +398,54 @@ def replace(obj: re.Match, *, transforms: Dict[str, str] = self.MENTION_TRANSFOR
...
+        self._cog = None
 
-        Returns
-        --------
-        Optional[:class:`Cog`]
-            The cog that is currently set for the help command.
-        """
-        return self._command_impl.cog
+    @property
+    def cog(self) -> Optional[Cog]:
```
Gobot1234, line 438: Shouldn't this be CogT?

discord/ext/commands/help.py
```diff
@@ -219,92 +219,12 @@ def _not_overridden(f: FuncT) -> FuncT:
...
-        # revert back into their original methods
-        cog = self.cog
-        cog.get_commands = cog.get_commands.__wrapped__
-        cog.walk_commands = cog.walk_commands.__wrapped__
-        self.cog = None
+_context: ContextVar[Optional[Context]] = ContextVar('context', default=None)
 
 
-class HelpCommand:
+class HelpCommand(Command[Optional['Cog'], ... if TYPE_CHECKING else Any, Any], Generic[ContextT]):
```
Rapptz, line 225: Need to be in a separate line.

discord/ext/commands/help.py
```diff
@@ -336,88 +256,57 @@ class HelpCommand:
...
-        __original_args__: Tuple[Any, ...]
-
     def __new__(cls, *args: Any, **kwargs: Any) -> Self:
-        # To prevent race conditions of a single instance while also allowing
-        # for settings to be passed the original arguments passed must be assigned
-        # to allow for easier copies (which will be made when the help command is actually called)
-        # see issue 2123
+        # we need to wrap the command callback to ensure context is correctly set.
+        functools.update_wrapper(cls._set_context, cls.command_callback)
+        cls._set_context.__qualname__ = 'command_callback'
```
Rapptz, line 262: Not really a fan of this. I know this is attempting to lie to `utils.is_inside_class`. I have no idea what the wrapping is for.

discord/ext/commands/context.py
```diff
@@ -354,52 +354,47 @@ async def send_help(self, *args: Any) -> Any:
...
             return None
 
-        cmd = cmd.copy()
-        cmd.context = self  # type: ignore
+        _context.set(self)
+
         if len(args) == 0:
             await cmd.prepare_help_command(self, None)
-            mapping = cmd.get_bot_mapping()
+            entity = cmd.get_bot_mapping()
```
Rapptz, line 369: This change doesn't seem right.

Rapptz (changes requested):

discord/ext/commands/help.py
bijij: Wrapping is to have `self.context` set without requring subclasses to explicitly do so.

This was one of my own main gripes with the implimentation.

discord/ext/commands/context.py
bijij: `HelpCommand.send_bot_help` excepts a `Mapping`, so it's correct. 
In an earlier version of this PR i needed to make changes to how the call was done which I no longer do, so technically this refactor of `send_help` could be undone.

discord/ext/commands/help.py
bijij: I reverted back to overriding params in the constructor.

Rapptz (approved):

PR #7688, opened 2022-03-14
Fix message converter not inferring channel when missing

ChrisLovering:
## Summary

The docstring for both the MessageConverter and the PartialMessageConverter imply that they will infer the channel id from the context channel if one isn't given. However, as of [this commit](https://github.com/Rapptz/discord.py/commit/1a4e73d59932cdbe7bf2c281f25e32529fc7ae1f), it no longer does this.

Instead, passing just a messageID to either of the message converters will only return a Message object if the given message ID is in the bot's cached messages.

I am not familiar with contributing to this project, so please let me know if I need to document this fix somewhere.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -344,7 +344,7 @@ class PartialMessageConverter(Converter[discord.PartialMessage]):
         if not match:
             raise MessageNotFound(argument)
         data = match.groupdict()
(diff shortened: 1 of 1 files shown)


Rapptz (approved):
Good catch. Thanks for the PR.

> I am not familiar with contributing to this project, so please let me know if I need to document this fix somewhere.

Since this is a bug fix, there is no need to document it.

PR #7693, opened 2022-03-15
Replace Invariant container types with wider types where applicable.

bijij:
## Summary

This PR replaces Invariant type annotations (List/Dict) with wider ABCs where applicable.

## Checklist

- [ ] This PR fixes an issue.


discord/emoji.py
```diff
@@ -214,7 +214,7 @@ async def delete(self, *, reason: Optional[str] = None) -> None:
 
         await self._state.http.delete_custom_emoji(self.guild_id, self.id, reason=reason)
 
-    async def edit(self, *, name: str = MISSING, roles: List[Snowflake] = MISSING, reason: Optional[str] = None) -> Emoji:
+    async def edit(self, *, name: str = MISSING, roles: Collection[Snowflake] = MISSING, reason: Optional[str] = None) -> Emoji:
```
bijij, line 217: This could be `Iterable` if we want.

discord/guild.py
```diff
@@ -1063,7 +1065,7 @@ def _create_channel(
         self,
         name: str,
         channel_type: Literal[ChannelType.text],
-        overwrites: Dict[Union[Role, Member], PermissionOverwrite] = ...,
+        overwrites: Mapping[Union[Role, Member], PermissionOverwrite] = ...,
```
bijij, line 1068: I'm not sure How I feel about these `Mapping` replacements.
We used to do a `isinstance(x, dict)` check which had to be modified to support it.

Undo?

bijij (commented):
Most of these changes are just List>Sequence in the send/edit methods, I've commented on the cases I wasn't completely sure on.

discord/guild.py
Rapptz: I can't readily imagine someone using a non-dict `Mapping` being used here.

PR #7696, opened 2022-03-15
Crowdin Actions workflows

Gorialis:
## Summary

This adds a download and upload workflow for Crowdin translations.

Uploading is scheduled every commit on master, but is deployed to the 'Crowdin' environment, so the run is delayed by 30 minutes. If another commit hits master in that time, the old run will be cancelled in favor of the new run, so uploading actually occurs after 30 minutes of inactivity on master (but adjustable via Environment settings)

Downloading can be triggered by workflow dispatch, but otherwise, it gets triggered at 6pm UTC every Monday. This is because I figure weekends might be when a decent amount of translation work gets done, and 6pm UTC is usually after most Japanese people have gone to sleep, so this should result in a sync during relatively quiet Crowdin hours. The resulting download sync will be opened as a PR, updating the existing branch if it exists.

I also merged in the download sync I used for testing so we can have an up to date sync given it's already past the time this week.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/.github/workflows/crowdin_download.yml
+++ b/.github/workflows/crowdin_download.yml
@@ -1,0 +1,44 @@
+name: crowdin download
+
+on:
+  schedule:
+    - cron: '0 18 * * 1'
+  workflow_dispatch:
+
+jobs:
+  download:
+    runs-on: ubuntu-latest
+    environment: Crowdin
+    name: download
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          ref: master
+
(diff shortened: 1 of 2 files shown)


Gorialis:
PR checks failing due to a currently unresolvable bug with Sphinx sources. Will have to figure out a solution to down the line.

Gorialis:
For the sake of unblocking this, I've removed the latest translations and just left it as the Actions. I'll spend some time figuring out how to fix some of our ref-related issues with Sphinx later.

PR #7701, opened 2022-03-16
Forbid wait_until_ready inside setup_hook

apple502j:
## Summary
Raises an exception if `wait_until_ready` is called inside `setup_hook`.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -29,6 +29,7 @@ import datetime
 import logging
 import sys
 import traceback
+from contextvars import ContextVar
 from typing import (
     Any,
     AsyncIterator,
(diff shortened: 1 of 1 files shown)


Rapptz:
Realistically, `wait_for` suffers the same deadlock problem.

apple502j:
Hmm, it's unlikely someone would use `wait_for` there, but it's something we can fix.

Also, we might have to do inspect hack to fix tasks...

apple502j:
I think there are legitimate uses for wait_for, and it's unlikely people would misuse that.

PR #7705, opened 2022-03-17
Add `utils.convert_timestamp`

ChesterWOV-DPY:
#### New Feature
Add `utils.convert_timestamp(timestamp, tz=...)`.
Converts a discord timestamp (for example `<:9999999>`) to a datetime object using `datetime.fromtimestamp`. `tz` is the timezone to convert to.

Not sure if I should also make a Transformer/Converter.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -1150,7 +1150,7 @@ def format_dt(dt: datetime.datetime, /, style: Optional[TimestampStyle] = None) 
         return f'<t:{int(dt.timestamp())}>'
     return f'<t:{int(dt.timestamp())}:{style}>'
     
-def format_timestamp(timestamp: str, /, tz: datetime.timezone = MISSING) -> Optional[datetime.datetime]:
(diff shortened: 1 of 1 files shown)


Rapptz:
I'm not really convinced this function is necessary?

ChesterWOV-DPY:
I thought it might be useful if a bot needs to edit a timestamp, I mean, yeah you could hard code the new timestamp inside but well why don't you use `convert_timestamp(msg.content) + timedelta(days=1)`

Gorialis:
I think if this was implemented at all, it would be as a regexing cached slot property on Message similar to `Message.raw_mentions`, but I don't know if the demand is really there. There'd also be a question of how to name it in a non-confusing way (`Message.timestamps` sounds like it would end up being *very* confusing)

Rapptz:
I've denied this before (#7312) so not sure I'll accept this one.

PR #7709, opened 2022-03-18
add wait_for guide

pikaninja:
## Summary

This pr adds the wait_for guide in the documentation. This was done in conjunction with HayBaleAv. I believe there's a few odd wording in this, any feedback would be appreciated

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/topics/wait_for.rst
+++ b/docs/topics/wait_for.rst
@@ -1,0 +1,81 @@
+:orphan:
+.. currentmodule:: discord
+
+.. _guide_wait_for:
+
+Waiting For Events
+=============================
+When creating commands or handling events, you often find yourself wanting to wait for an action or user response. In discord.py, we use the :meth:`Client.wait_for` method. 
+
+What is ``wait_for``?
+~~~~~~~~~~~~~~~~~~~~~
+:meth:`Client.wait_for` for is similar to client.event/listen, however instead of calling a function, it instead holds execution of your code until the event has been dispatched, and returns the data that was in the event. 
+
+
+Here is a quick example
+
+.. code-block:: python3
+   :emphasize-lines: 2
+
+    await channel.send('say hi!')
+    message = await client.wait_for('message') # wait for a message, the wait_for will return the message it finds
+    await message.reply('hello!')
+
+The key line here is line 2. We are waiting for an event to happen, in this case a message event
+
+Wait_for is commonly used to control flow in a code, waiting for message replies or reactions to be added. You can wait for any event you want.
+
+Checks and Timeouts
+~~~~~~~~~~~~~~~~~~~~
+While the example above allows us to wait for an event, this will pass as soon as *any* event of that type is dispatched. Sometimes, we want to filter it. To do this, we use the check keyward argument.
+
+.. code-block:: python3
+   :emphasize-lines: 2, 3, 4
+
+    await message.channel.send('say hi!')
+    def check(m):
+        return m.author == message.author # check that the author is the one who sent the original message
+    msg = await client.wait_for('message', check=check)
+    await msg.reply('hello!')
+
+In this example, the wait_for will only terminate after the check returns ``True``, in this case after the message author is equal to the original author.
+The check takes the arguments the event would take, in this cause just ``m``, a message.
+
+Sometimes, we only want to wait for a specific amount of time, before timing out the wait_for and allowing the code to continue. 
+
+.. code-block:: python3
+   :emphasize-lines: 2, 3, 4, 5
+
+    await message.channel.send('say hi!')
+    try:
+        msg = await client.wait_for('message', timeout=5.0) # timeout in seconds.
+    except asyncio.TimeoutError: 
+        await message.channel.send('You did not respond :(')
+    else:
+        await msg.reply('hello!')
+
+We pass the timeout in seconds to the timeout kwarg of wait_for. If the event is not dispatched or the check is not passed in those seconds, the wait_for will terminate and raise :class:`asyncio.TimeoutError`.
+
+.. warning::
+
+    avoid using wait_for within a loop to catch events of a specific type. Due to the async nature of discord.py, events may be fired between loops, in return causing your wait_for to miss the event dispatch.
+
+Examples
+~~~~~~~~~~~~~~~~~~~~
+Wait for reaction
+
+.. code-block:: python3
+
+    def check(reaction, user):
+        return user == message.author and reaction.message == message # check that the reaction is on a specific message and by a specific user
+
+    reaction, user = await client.wait_for('reaction_add', check = check)
+    await message.channel.send(f'You reacted with {reaction.emoji}!')
+
+Inline check
+
+.. code-block:: python3 
+
+    await client.wait_for('message', check = lambda m: m.author == message.author)
+
+


docs/topics/wait_for.rst
```diff
@@ -0,0 +1,81 @@
...
+
+.. _guide_wait_for:
+
+Waiting For Events
+=============================
+When creating commands or handling events, you often find yourself wanting to wait for an action or user response. In discord.py, we use the :meth:`Client.wait_for` method. 
+
+What is ``wait_for``?
+~~~~~~~~~~~~~~~~~~~~~
+:meth:`Client.wait_for` for is similar to client.event/listen, however instead of calling a function, it instead holds execution of your code until the event has been dispatched, and returns the data that was in the event.
```
ChesterWOV-DPY, line 12: Remove the "for" after the method name?
```suggestion
:meth:`Client.wait_for` is similar to client.event/listen, however instead of calling a function, it instead holds execution of your code until the event has been dispatched, and returns the data that was in the event. 
```

ChesterWOV-DPY (changes requested):

mirnovov:
It might be a good idea to include basic instructions on using ```wait_for``` for multiple things simultaneously, as in the ```multiple wait_for``` tag on the discord server. Though of course, including the proviso that it's a more advanced topic when mentioning it.

pikaninja:
> It might be a good idea to include basic instructions on using `wait_for` for multiple things simultaneously, as in the `multiple wait_for` tag on the discord server. Though of course, including the proviso that it's a more advanced topic when mentioning it.

i did consider adding this, but i felt it was a bit too jank and confusing to add to the guide, idk though

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,81 @@
...
+Here is a quick example
+
+.. code-block:: python3
+   :emphasize-lines: 2
+
+    await channel.send('say hi!')
+    message = await client.wait_for('message') # wait for a message, the wait_for will return the message it finds
+    await message.reply('hello!')
+
+The key line here is line 2. We are waiting for an event to happen, in this case a message event
```
mysistersbrother, line 24:
```suggestion
The key line here is line 2. We are waiting for an event to happen, in this case, a message event.
```

consistency in punctuation

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,81 @@
...
+
+Wait_for is commonly used to control flow in a code, waiting for message replies or reactions to be added. You can wait for any event you want.
+
+Checks and Timeouts
+~~~~~~~~~~~~~~~~~~~~
+While the example above allows us to wait for an event, this will pass as soon as *any* event of that type is dispatched. Sometimes, we want to filter it. To do this, we use the check keyward argument.
```
mysistersbrother, line 30:
```suggestion
While the example above allows us to wait for an event, this will pass as soon as *any* event of that type is dispatched. Sometimes, we want to filter it. To do this, we use the check keyword argument.
```

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,81 @@
...
+.. code-block:: python3
+   :emphasize-lines: 2, 3, 4
+
+    await message.channel.send('say hi!')
+    def check(m):
+        return m.author == message.author # check that the author is the one who sent the original message
+    msg = await client.wait_for('message', check=check)
+    await msg.reply('hello!')
+
+In this example, the wait_for will only terminate after the check returns ``True``, in this case after the message author is equal to the original author.
```
mysistersbrother, line 41:
```suggestion
In this example, the wait_for will only terminate when the check returns ``True``, in this case when the message author is equal to the original author.
```

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,81 @@
...
+The check takes the arguments the event would take, in this cause just ``m``, a message.
```
mysistersbrother, line 42:
```suggestion
The check function takes the same arguments the event would take, in this case just ``m``, a message.
```

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,81 @@
...
+
+    await message.channel.send('say hi!')
+    try:
+        msg = await client.wait_for('message', timeout=5.0) # timeout in seconds.
+    except asyncio.TimeoutError: 
+        await message.channel.send('You did not respond :(')
+    else:
+        await msg.reply('hello!')
+
+We pass the timeout in seconds to the timeout kwarg of wait_for. If the event is not dispatched or the check is not passed in those seconds, the wait_for will terminate and raise :class:`asyncio.TimeoutError`.
```
mysistersbrother, line 57:
```suggestion
We pass the timeout in seconds to the timeout kwarg of wait_for. If the wait_for does not complete successfully within the given time it will be terminated and :class:`asyncio.TimeoutError` will be raised.
```

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,81 @@
...
+
+.. warning::
+
+    avoid using wait_for within a loop to catch events of a specific type. Due to the async nature of discord.py, events may be fired between loops, in return causing your wait_for to miss the event dispatch.
```
mysistersbrother, line 61:
```suggestion
    avoid using wait_for within a loop to catch events of a specific type. Due to the async nature of discord.py, events may be fired between loops causing your wait_for to miss the event dispatch.
```

mysistersbrother (changes requested):
some wording changes for extra clarity and typo fixes

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,80 @@
...
+
+.. _guide_wait_for:
+
+Waiting For Events
+=============================
+When creating commands or handling events, you often find yourself wanting to wait for an action or user response. In discord.py, we use the :meth:`Client.wait_for` method. 
+
+What is ``wait_for``?
+~~~~~~~~~~~~~~~~~~~~~
+:meth:`Client.wait_for` is similar to client.event/listen, however instead of calling a function, it instead holds execution of your code until the event has been dispatched, and returns the data that was in the event.
```
numbermaniac, line 12:
```suggestion
:meth:`Client.wait_for` is similar to client.event/listen, however instead of calling a function, it instead halts execution of your code until the event has been dispatched, and returns the data that was in the event. 
```

Perhaps "halts" is more appropriate here?

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,80 @@
...
+.. code-block:: python3
+   :emphasize-lines: 2
+
+    await channel.send('say hi!')
+    message = await client.wait_for('message') # wait for a message, the wait_for will return the message it finds
+    await message.reply('hello!')
+
+The key line here is line 2. We are waiting for an event to happen, in this case, a message event.
+
+Wait_for is commonly used to control flow in a code, waiting for message replies or reactions to be added. You can wait for any event you want.
```
numbermaniac, line 26:
```suggestion
``wait_for`` is commonly used to control code flow, waiting for message replies or reactions to be added. You can wait for any event you want.
```

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,80 @@
...
+
+Checks and Timeouts
+~~~~~~~~~~~~~~~~~~~~
+While the example above allows us to wait for an event, this will pass as soon as *any* event of that type is dispatched. Sometimes, we want to filter it. To do this, we use the check keyword argument.
```
numbermaniac, line 30:
```suggestion
While the example above allows us to wait for an event, this will pass as soon as *any* event of that type is dispatched. Sometimes, we want to filter it. To do this, we use the ``check`` keyword argument.
```

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,80 @@
...
+.. code-block:: python3
+   :emphasize-lines: 2, 3, 4
+
+    await message.channel.send('say hi!')
+    def check(m):
+        return m.author == message.author # check that the author is the one who sent the original message
+    msg = await client.wait_for('message', check=check)
+    await msg.reply('hello!')
+
+In this example, the wait_for will only terminate when the check returns ``True``, in this case when the message author is equal to the original author.
```
numbermaniac, line 41:
```suggestion
In this example, ``wait_for`` will only terminate when the check returns ``True``, in this case when the message author is equal to the original author.
```
There's a lot of other `wait_for`s below this, but I'm not sure if we should put all of them into inline codeblocks.

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,80 @@
...
+
+    await message.channel.send('say hi!')
+    try:
+        msg = await client.wait_for('message', timeout=5.0) # timeout in seconds.
+    except asyncio.TimeoutError: 
+        await message.channel.send('You did not respond :(')
+    else:
+        await msg.reply('hello!')
+
+We pass the timeout in seconds to the timeout kwarg of wait_for. If the wait_for does not complete successfully within the given time it will be terminated and :class:`asyncio.TimeoutError` will be raised.
```
numbermaniac, line 57:
```suggestion
We pass the timeout in seconds to the ``timeout`` kwarg of wait_for. If the wait_for does not complete successfully within the given time it will be terminated and :class:`asyncio.TimeoutError` will be raised.
```

numbermaniac (commented):
A couple of small suggested changes.

docs/topics/wait_for.rst
AbstractUmbra, line 41:
```suggestion
In this example, the wait_for will only accept the incoming event and move on when the check returns ``True``, in this case when the message author is equal to the original author.
```

Feels like "terminate" is not the correct word here. Sounds like it would end the process, not a success criteria.

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,80 @@
...
+   :emphasize-lines: 2, 3, 4
+
+    await message.channel.send('say hi!')
+    def check(m):
+        return m.author == message.author # check that the author is the one who sent the original message
+    msg = await client.wait_for('message', check=check)
+    await msg.reply('hello!')
+
+In this example, the wait_for will only terminate when the check returns ``True``, in this case when the message author is equal to the original author.
+The check function takes the same arguments the event would take, in this case just ``m``, a message.
```
AbstractUmbra, line 42:
```suggestion
The check function takes the same arguments the event would take, in this case a paramter named ``m``, which represents a ``discord.Message`` instance. This is similar to `:meth:`Client.on_message`.
```

Feels like it was not explained all too cleanly.

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,80 @@
...
+        await msg.reply('hello!')
+
+We pass the timeout in seconds to the timeout kwarg of wait_for. If the wait_for does not complete successfully within the given time it will be terminated and :class:`asyncio.TimeoutError` will be raised.
+.. warning::
+
+    avoid using wait_for within a loop to catch events of a specific type. Due to the async nature of discord.py, events may be fired between loops, causing your wait_for to miss the event dispatch.
+
+Examples
+~~~~~~~~~~~~~~~~~~~~
+Wait for reaction
```
AbstractUmbra, line 64: I believe this should be fleshed out a little more. Meaning that we covered the waiting for a message in great detail yet this other event has a throwaway remark.

Perhaps elaborate a bit on the fact that it takes the same parameters as the matching event name and that you can explicitly see it has 2 parameters in the check, and that you actually receive a tuple from the method and not a single object like previously.

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,80 @@
...
+    reaction, user = await client.wait_for('reaction_add', check = check)
+    await message.channel.send(f'You reacted with {reaction.emoji}!')
+
+Inline check
+
+.. code-block:: python3 
+
+    await client.wait_for('message', check = lambda m: m.author == message.author)
+
+
```
AbstractUmbra, line 80: Something @Gorialis has been highlighting is that these guide documents should end with some prose. I believe here you should mention the event api docs and their relations here.

AbstractUmbra (commented):
I have added a handful of comments below.

Since we are now a strongly typed library I believe our examples should be typed too.

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,92 @@
...
+    def check(reaction: discord.Reaction, user: discord.User):
+        return user == message.author and reaction.message == message # check that the reaction is on a specific message and by a specific user
+
+    reaction: discord.Reaction, user: discord.User = await client.wait_for('reaction_add', check = check)
+    await message.channel.send(f'You reacted with {reaction.emoji}!')
+
+Notive the ``reaction`` event, unlike the ``message`` event, takes 2 arguments. Thus, the check function takes the same arguments as that, ``reaction, user``.
+Additionally, the wait_for will now return a tuple of the arguments the event recieved, which we are unwraping when recieving. 
+This is how it works for *any* event that takes more than one argument, a few examples being ``typing`` or ``message_edit``.
```
numbermaniac, line 77:
```suggestion
Notice the ``reaction_add`` event, unlike the ``message`` event, takes 2 arguments. Thus, the check function takes the same arguments as that, ``reaction, user``.
Additionally, the wait_for will now return a tuple of the arguments the event received, which we are unwrapping when receiving. 
This is how it works for *any* event that takes more than one argument, a few examples being ``typing`` or ``message_edit``.
```

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Inline check
+++++++++++++
+
+.. code-block:: python3 
+
+    await client.wait_for('message', check = lambda m: m.author == message.author)
+
+As the check kwarg simply takes a function, we can make it inline by making use of a lambda function
```
numbermaniac, line 86:
```suggestion
As the ``check`` kwarg simply takes a function, we can make it inline by making use of a lambda function.
```

numbermaniac (commented):
Just some quick fixes to the new text added to the guide.

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,92 @@
...
+
+.. code-block:: python3 
+
+    await client.wait_for('message', check = lambda m: m.author == message.author)
+
+As the ``check`` kwarg simply takes a function, we can make it inline by making use of a lambda function.
+
+Closing Remarks
+~~~~~~~~~~~~~~~~~~~~
+``wait_for`` is a powerful tool used often to wait for responses in code. The examples above only shows 2 types of ``wait_for``, reactions and messages, but you can wait for any event! A full list of events can be seen here: :ref:`event reference <discord-api-events>`.
```
numbermaniac, line 90:
```suggestion
``wait_for`` is a powerful tool used often to wait for responses in code. The examples above show only 2 types of ``wait_for``, reactions and messages, but you can wait for any event! A full list of events can be seen here: :ref:`event reference <discord-api-events>`. 
```

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,92 @@
...
+
+.. code-block:: python3
+
+    def check(reaction: discord.Reaction, user: discord.User):
+        return user == message.author and reaction.message == message # check that the reaction is on a specific message and by a specific user
+
+    reaction: discord.Reaction, user: discord.User = await client.wait_for('reaction_add', check = check)
+    await message.channel.send(f'You reacted with {reaction.emoji}!')
+
+Notice the ``reaction_add`` event, unlike the ``message`` event, takes 2 arguments. Thus, the check function takes the same arguments as that, ``reaction, user``.
```
dartmern, line 75: You should emphasize that it takes the same arguments as the event.

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,92 @@
...
+~~~~~~~~~~~~~~~~~~~~
+Wait for reaction
++++++++++++++++++
+
+.. code-block:: python3
+
+    def check(reaction: discord.Reaction, user: discord.User):
+        return user == message.author and reaction.message == message # check that the reaction is on a specific message and by a specific user
+
+    reaction: discord.Reaction, user: discord.User = await client.wait_for('reaction_add', check = check)
```
Soheab, line 72: This isn't valid syntax.

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,94 @@
...
+
+.. _guide_wait_for:
+
+Waiting For Events
+=============================
+When creating commands or handling events, you often find yourself wanting to wait for an action or user response. In discord.py, we use the :meth:`Client.wait_for` method. 
+
+What is ``wait_for``?
+~~~~~~~~~~~~~~~~~~~~~
+:meth:`Client.wait_for` is similar to client.event/listen, however instead of calling a function, it instead halts execution of your code until the event has been dispatched, and returns the data that was in the event.
```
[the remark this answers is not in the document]
pikaninja: oops i didnt realize that either

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,86 @@
...
+    await msg.reply('hello!')
+
+In this example, ``wait_for`` will only accept the incoming message when the predicate returns ``True``.
+The arguments passed to the predicate match the signature of the corresponding event.
+For example, when defining a message event handler we do `async def on_message(message)` - a predicate would similarly be defined with `def check(message)`.
+
+Likewise, a predicate for `reaction_add` would take `def check(reaction, user)`.
+
+It's common to also include a timeout in addition to a check so our code doesn't potentially end up waiting indefinitely, or to add a limited time window for the author to send a message.
+.. code-block:: python3
```
AbstractUmbra, line 50:
```suggestion
It's common to also include a timeout in addition to a check so our code doesn't potentially end up waiting indefinitely, or to add a limited time window for the author to send a message.

.. code-block:: python3
```

A newline is necessary here, the page builds incorrectly otherwise.

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,86 @@
+:orphan:
+.. currentmodule:: discord
+
+.. _guide_wait_for:
+
+Waiting for Events
+=============================
+A common step when creating commands or handling events involves having to wait for an action or user-input. To aid with this, the library offers a special method :meth:`Client.wait_for` to wait for an event in-line instead of having to define a separate event listener.
+What is ``wait_for``?
```
AbstractUmbra, line 9:
```suggestion
A common step when creating commands or handling events involves having to wait for an action or user-input. To aid with this, the library offers a special method :meth:`Client.wait_for` to wait for an event in-line instead of having to define a separate event listener.

What is ``wait_for``?
```
A new line is necessary here for the header to be declared correctly.

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,86 @@
+:orphan:
+.. currentmodule:: discord
+
+.. _guide_wait_for:
+
+Waiting for Events
+=============================
```
AbstractUmbra, line 7:
```suggestion
Waiting for Events
==================
```

Header declaration only needs to match the line length of the header.

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,86 @@
...
+A common step when creating commands or handling events involves having to wait for an action or user-input. To aid with this, the library offers a special method :meth:`Client.wait_for` to wait for an event in-line instead of having to define a separate event listener.
+What is ``wait_for``?
+~~~~~~~~~~~~~~~~~~~~~
+Similar to callbacks decorated with `@client.event`, :meth:`Client.wait_for` waits for the first event it finds tied to the passed event name (without the `on_` prefix!) and returns the event's arguments.
```
AbstractUmbra, line 11:
```suggestion
Similar to callbacks decorated with ``@client.event``, :meth:`Client.wait_for` waits for the first event it finds tied to the passed event name (without the `on_` prefix!) and returns the event's arguments.
```
Inline codeblocks require double backticks, unless you meant to only italicize, in which case disregard this suggestion.

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,86 @@
...
+        await msg.reply('hello!')
+
+We pass the timeout in seconds to the ``timeout`` keyword-argument. Timeouts behave like :meth:`asyncio.wait_for`, so we do the usual `try` and `except`, catching :class:`asyncio.TimeoutError`:
+.. warning::
+
+    Avoid calling :meth:`Client.wait_for` within a loop to catch events. Due to the nature of async IO, events may be fired between loops, causing the loop to miss events.
+
+Examples
+~~~~~~~~~~~~~~~~~~~~
+Wait for reaction
```
AbstractUmbra, line 68: Likely nitpicky but I feel this could link to the documentation pages for `on_reaction_add` and mention the raw variant too.

No need for an explanation of raw vs not, just linking them would be good so people can "avoid" the message needing to be in cache to catch a full Reaction.

AbstractUmbra (changes requested):
Changes supplied in the review. :)

Some other thoughts are, could we add a bit more complex examples? Like waiting for some non-common event (so not message or reaction, but maybe something like a typing event?) that showcases the difference in passed check parameters and whatnot, and maybe showcase a raw one to mention the raw typed payload(s) in comparison to complete objects?

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,86 @@
...
+    await message.channel.send('say hi!')
+    try:
+        msg: discord.Message = await client.wait_for('message', timeout=5.0) # timeout in seconds.
+    except asyncio.TimeoutError: 
+        await message.channel.send('You did not respond :(')
+    else:
+        await msg.reply('hello!')
+
+We pass the timeout in seconds to the ``timeout`` keyword-argument. Timeouts behave like :meth:`asyncio.wait_for`, so we do the usual `try` and `except`, catching :class:`asyncio.TimeoutError`:
+.. warning::
```
AbstractUmbra, line 62: Sorry missed this one in my earlier review.

```suggestion
We pass the timeout in seconds to the ``timeout`` keyword-argument. Timeouts behave like :meth:`asyncio.wait_for`, so we do the usual `try` and `except`, catching :class:`asyncio.TimeoutError`:

.. warning::
```

Newline needed here for warning block parsing.

AbstractUmbra:
Oh, one last thing for now.

Could the file be moved to `docs/guide/topics/wait_for.rst` to match the rest?

docs/topics/wait_for.rst
tailoric, line 9: Missing empty line above causes header to fail 
image

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,86 @@
...
+    def check(m: discord.Message):
+        # check that the author is the one who sent the original message
+        return m.author == message.author 
+
+    msg: discord.Message = await client.wait_for('message', check=check)
+    await msg.reply('hello!')
+
+In this example, ``wait_for`` will only accept the incoming message when the predicate returns ``True``.
+The arguments passed to the predicate match the signature of the corresponding event.
+For example, when defining a message event handler we do `async def on_message(message)` - a predicate would similarly be defined with `def check(message)`.
```
tailoric, line 45: wouldn't it be better to have these short code parts be in a inline code style (two backticks) instead of italic like they are now?
image

docs/topics/wait_for.rst
tailoric, line 62: Another missing empty line above causes another rendering problem of not being a correct displayed warning
image

docs/topics/wait_for.rst
```diff
@@ -0,0 +1,86 @@
...
+
+    reaction, user = await client.wait_for('reaction_add', check=check)
+    await message.channel.send(f'You reacted with {reaction.emoji}!')
+
+Notice the ``reaction_add`` event, unlike the ``message`` event, takes 2 arguments. Thus, the check function takes the same arguments as that, ``reaction, user``.
+
+
+Closing Remarks
+~~~~~~~~~~~~~~~~~~~~
+``wait_for`` is a powerful tool used often to wait for responses in code. The examples above show only 2 types of ``wait_for``, reactions and messages, but you can wait for any event! A full list of events can be seen here: :ref:`event reference <discord-api-events>`.
```
tailoric, line 84: While they are powerful I think that it should be mentioned that Buttons and Modals can be also used for similar functionality that provide better UX for users instead and could/should be considered of waiting for a message or reaction. 

However I can also understand that this may be out of scope for the `wait_for` guide and probably better to be added to the views/modal guide instead.

Also I am a bit struggling to find good use cases on waiting for other events than reaction or `on_message` events, especially ones that aren't superseded by UI components. Because I think the examples should show some functionality that is not possible with those, however again that is just a major personal nitpick mostly but I wanted to mention it.

Only one example I can think of could be having a `wait_for` for the `voice_state_update` event to join with a user who invoked a command that needs someone to be in voice before doing something. Roughly working like so:

1. Command should play something in the voice channel the user is in
2. check if user is in voice if yes join and play sound
3. else send a message asking them to join a voice channel
4. wait_for user to join a voice channel with timeout
5. play sound / file after user joined.

It is a bit convoluted but also a bit more of an involved example that could show off the capability of `wait_for`

tailoric (commented):
Mostly some formatting problems that I've noticed when building the docs for myself.

PR #7722, opened 2022-03-19
Add support for specifying the type of a generic discord.Object

bijij:
## Summary

This PR adds a new `type` parameter to `discord.Object` allowing for use inside sets and dictionaries of known types.

For example `{client.get_role(882127116123111435), discord.Object(id=882127116123111435, type=discord.Role)}` assuming the get_role call succeeds will store only one object in the set.

if a type is not specified discord.Object will behave like before.

## Checklist

- [x] This PR fixes an issue.


Rapptz:
Already denied this on Discord.

Rapptz:
I'll think about this one a bit more even though right now I'm still negative on it.

discord/object.py
```diff
@@ -77,19 +79,35 @@ class Object(Hashable):
...
-    def __init__(self, id: SupportsIntCast):
+    def __init__(self, id: SupportsIntCast, *, type: Type[abc.Snowflake] = MISSING):
         try:
             id = int(id)
         except ValueError:
             raise TypeError(f'id parameter must be convertable to int not {id.__class__!r}') from None
-        else:
-            self.id = id
+        self.id: int = id
+        self.type: Type[abc.Snowflake] = type or self.__class__
```
bijij, line 99: Similarly to `id` I would like this to be validated, however without moving `abc.Snowflake` into here, or its own file, it can't be done as `abc` calls `Object.__init__`.

Rapptz (approved):
After *a lot* of discussion and other attempts it seems this is probably the sanest approach to do this.

PR #7724, opened 2022-03-19
[commands] Add tree_cls to Bot constructor

bijij:
## Summary

This PR allows for setting a custom CommandTree subclass in the commands extension.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -156,6 +156,7 @@ class BotBase(GroupMixin[None]):
         self,
         command_prefix: PrefixType[BotT],
         help_command: Optional[HelpCommand[Any]] = _default,
+        tree_cls: Type[app_commands.CommandTree] = app_commands.CommandTree,
         description: Optional[str] = None,
         **options: Any,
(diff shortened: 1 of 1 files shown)


Rapptz:
I denied this one originally since it requires HKT for proper typing and I wasn't even sure why anyone would do this other than implementing `on_error`, which is why I just added the decorator or tell people to explicitly assign it.

discord/ext/commands/bot.py
```diff
@@ -151,15 +151,15 @@ class BotBase(GroupMixin[None]):
     def __init__(
         self,
         command_prefix: PrefixType[BotT],
+        tree_cls: Type[app_commands.CommandTree[Self]] = app_commands.CommandTree,  # type: ignore
```
Rapptz, line 154: Is the `Self` generic necessary here?

Rapptz (commented):
This has conflicts now.

I think going forward it might make sense to provide this even without HKT support in Python since in the future I need to do global checks via the tree.

PR #7730, opened 2022-03-20
Only send allowed_mentions if content changed

BeatButton:
## Summary

Fixes a problem where you could not edit another user's message to suppress embeds (something that is meant to/used to be possible).

## Checklist

- [x] This PR fixes an issue.


BeatButton:
Sorry, I did it wrong two times, but now I believe everything is correct.

I don't think the docstring needs to change, since the context of allowed_mentions is messages that the bot has sent itself.

PR #7733, opened 2022-03-20
Fix typos in some View methods' docstrings

quackbarc:
## Summary

This fixes whitespace typos that introduced inconsistency in docstrings and caused docs of View.add_item to look slightly bugged.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ui/view.py
+++ b/discord/ui/view.py
@@ -256,6 +256,7 @@ class View:
         """Adds an item to the view.
         This function returns the class instance to allow for fluent-style
         chaining.
+
         Parameters
         -----------
         item: :class:`Item`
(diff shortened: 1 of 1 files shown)


discord/ui/view.py
```diff
@@ -256,6 +256,7 @@ def add_item(self, item: Item[Any]) -> Self:
         """Adds an item to the view.
```
nerdguyahmad, line 256: A whitespace under this line should also be added to stay consistent with rest of the docstring's.

nerdguyahmad (changes requested):

discord/ui/view.py
quackbarc: Seems to be the case with `View.remove_item` and `View.clear_items` too. I'll add one to those as well.

nerdguyahmad (approved):

PR #7743, opened 2022-03-22
style: add typing metadatas

SaidBySolo:
### Before

image

### After
image

I am using pylance strict option.

After applying that change, it will no longer complain about not being able to find the stub file.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,8 +3,5 @@ include LICENSE
 include requirements.txt
 include discord/bin/*.dll
 include discord/py.typed
-include discord/app_commands/py.typed
 include discord/ext/commands/py.typed
 include discord/ext/tasks/py.typed
-include discord/ui/py.typed
-include discord/webhook/py.typed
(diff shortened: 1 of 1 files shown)


Rapptz:
Please note that `discord.ext.commands.bot` is not a publicly exported namespace.

SaidBySolo:
A problem also occurs in the following cases:
image

Do I have to use it in the following way unconditionally?
```py

bot = commands.Bot("")
bot.run()
```
I prefer the absolute import.

Rapptz:
> Do I have to use it in the following way unconditionally?

Yes, this is the only supported export.

Ideally, `from discord.ext.commands import Bot` would work too but the `discord.ext.commands.bot` module is technically private.

SaidBySolo:
``from discord.ext.commands import Bot``
I want the following format, but if py.typed is applied, as mentioned above, unintended problems may occur, so I will close this PR.

PR #7750, opened 2022-03-23
Typing fixes

bryanforbes:
## Summary

* Make some Cog methods positional-only
* Fix maybe_coroutine and MaybeCoroFunc
* Mark stub functions that can be coroutines as returning MaybeCoro

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/transformers.py
+++ b/discord/app_commands/transformers.py
@@ -48,7 +48,7 @@ from .errors import AppCommandError, TransformerError
 from .models import AppCommandChannel, AppCommandThread, Choice
 from ..channel import StageChannel, StoreChannel, VoiceChannel, TextChannel, CategoryChannel
 from ..enums import AppCommandOptionType, ChannelType
-from ..utils import MISSING, maybe_coroutine
+from ..utils import MISSING, MaybeCoro, maybe_coroutine
 from ..user import User
 from ..role import Role
 from ..member import Member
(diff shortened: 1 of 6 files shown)


Rapptz:
> Mark stub functions that can be coroutines as returning MaybeCoro

I don't think I agree with this change. Is there a reason it was done?

Rapptz:
The set of users who use strict is incredibly small and generally aren't the type of people I'm trying to please in the library. The incompatible method error for example will always trigger when a generic is passed as bound even though it's technically okay.

A big issue I have with this is usability. A lot of tooling helpfully autocomplete methods when overriding and this one will give users a bizarre signature to override that includes a private alias type.

bryanforbes:
> A big issue I have with this is usability. A lot of tooling helpfully autocomplete methods when overriding and this one will give users a bizarre signature to override that includes a private alias type.

This is a really good concern to have. Let me know what you'd like me to do with this PR. I can split it into two: one for fixing `maybe_coroutine()` and one for the positional-only params.

Rapptz:
I think just the maybe_coroutine fixes are okay. The positional-only parameter change seems like an unnecessary breaking change in this case.

bryanforbes:
This has stagnated and the `maybe_coroutine()` changes have been handled in 06c257760bdedd39c37a7eb12f0338ac60b48c20. Closing this.

PR #7751, opened 2022-03-23
Remove outdated note and example.

nerdguyahmad:
## Summary

Removes an outdated note in `query_members` and example in `fetch_members` doc.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -1897,11 +1897,6 @@ class Guild(Hashable):
 
             async for member in guild.fetch_members(limit=150):
                 print(member.name)
-
-        Flattening into a list ::
-
-            members = await guild.fetch_members(limit=150).flatten()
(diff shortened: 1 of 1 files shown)


discord/guild.py
```diff
@@ -3438,8 +3433,6 @@ async def query_members(
         Request members that belong to this guild whose username starts with
         the query given.
 
-        This is a websocket operation and can be slow.
```
Rapptz, line 3441: You can just remove the "can be slow" part.

Rapptz (changes requested):

discord/guild.py
nerdguyahmad: Done

Rapptz (approved):

PR #7769, opened 2022-03-26
Update parent reference of `Group` children

bryanforbes:
## Summary

Take the following example using `master`:

```python
import discord

class MyGroup(app_commands.Group, name='stuff'):
    @app_commands.command()
    async def things(self, interaction: discord.Interaction) -> None:
        ...

g = MyGroup()
copy = g._copy_with_binding({})  # This is what `Cog` does

MyGroup.__discord_app_commands_group_children__[0].parent is g  # True
g.things.parent is g  # True
copy.things.parent is copy  # False
copy.things.parent is g  # True
```

Currently, when a `Group` is copied via `_copy_with_binding()`, the children copies do not have their `parent` reference updated to the new copy of `Group` and the copy does not have its attributes updated to the new child copies. Further, the routine in `Group.__init_subclass__()` changes the `parent` reference of the `Command` stored at the class level to the latest created group subclass.

This PR fixes the above so that `MyGroup.__discord_app_commands_group_children__[0].parent is g` returns `False` and `copy.things.parent is copy` returns `True`.

## Checklist

- [x] This PR fixes an issue.


discord/app_commands/commands.py
```diff
@@ -934,8 +934,8 @@ def __init__(
         self._children: Dict[str, Union[Command, Group]] = {}
 
         for child in self.__discord_app_commands_group_children__:
-            child.parent = self
             child = child._copy_with_binding(self) if not cls.__discord_app_commands_skip_init_binding__ else child
+            child.parent = self
```
Rapptz, line 938: I originally had these lines like this but I swapped them because the child would not copy the parent if it's done after the `_copy_with_binding` call. What's the reason for swapping it back?

In my mental model with a setup like `/foo`, `/foo bar`, `/foo bar baz`

Old order:

- `/foo` gets copied. 
- Parent for `bar` is set to `/foo`. 
- `bar` is copied (parent properly assigned)
- Bug: Children are copied (but parent is not properly assigned)

New order:

- `/foo` gets copied
- `bar` parent is the *old* `/foo`
- `bar` is copied
- children are copied and their parent is set to the copy

Unless I'm missing something I think the first layer for `/foo bar` points to the old `/foo`?

Rapptz (commented):
I think while we're here, there's another bug with parent bindings that isn't caught with this patch:

Given something like:

```py
class TestCog(commands.Cog):
    group = app_commands.Group(name='test', description='...')
    foo = app_commands.Group(name='foo', description='...', parent=group)

    @foo.command()
    async def bar(self, interaction):
        pass


cog = TestCog()
print('TestCog.bar.parent', cog.bar.parent)
print('TestCog.foo', cog.foo)
print('Equal', cog.bar.parent is cog.foo)
print('TestCog.foo.parent', cog.foo.parent)
print('TestCog.group', cog.group)
print('Equal', cog.group is cog.foo.parent)
```

This prints `True`/`False`. This is true currently as well.

discord/app_commands/commands.py
bryanforbes: > What's the reason for swapping it back?

In the case where a copy is made on line 937, setting the parent before the copy is made is updating the class child's parent reference, so `MyGroup.__discord_app_commands_group_children__[0].parent` becomes the latest instance of `MyGroup`.

I'm going to put together a test to make sure all of the scenarios are properly covered by this new code because I think you're right that this new order misses something.

bryanforbes:
I created some tests that check a whole bunch of scenarios and then came up with a solution (that's admittedly a bit hacky) to make sure all of the tests pass. Let me know how you want to proceed.

bryanforbes:
This is ready for another review

PR #7771, opened 2022-03-27
Allow context menus have '?' and '! 'in their name

Puncher1:
## Summary

This pull request is for allowing context menus to have `?` and `!` in their names.

For context menus ([`@app_commands.context_menu()`](https://discordpy.readthedocs.io/en/master/interactions/api.html?highlight=app_commands%20context_menu#discord.app_commands.context_menu)) it throws an error (`ValueError: context menu names must be between 1-32 characters`) when a character is in the string which isn't ...
* from `a` to `Z`
* from `0` to `9`
* the underscore `_`
* the dash `-`
* the white space

See the current code in `master`
```py
VALID_SLASH_COMMAND_NAME = re.compile(r'^[\w-]{1,32}$')
VALID_CONTEXT_MENU_NAME = re.compile(r'^[\w\s-]{1,32}$')

# ...

def validate_name(name: str) -> str:
    match = VALID_SLASH_COMMAND_NAME.match(name)
    if match is None:
        raise ValueError('names must be between 1-32 characters')

    # Ideally, name.islower() would work instead but since certain characters
    # are Lo (e.g. CJK) those don't pass the test. I'd use `casefold` instead as
    # well, but chances are the server-side check is probably something similar to
    # this code anyway.
    if name.lower() != name:
        raise ValueError('names must be all lower case')
    return name


def validate_context_menu_name(name: str) -> str:
    if VALID_CONTEXT_MENU_NAME.match(name) is None:
        raise ValueError('context menu names must be between 1-32 characters')
    return name
```

Proof that it should work with `?`:
unknown (1)

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -125,7 +125,7 @@ else:
 
 
 CheckInputParameter = Union['Command[Any, ..., Any]', 'ContextMenu', CommandCallback, ContextMenuCallback]
-VALID_SLASH_COMMAND_NAME = re.compile(r'^[?!\w-]{1,32}$')
+VALID_SLASH_COMMAND_NAME = re.compile(r'^[\w-]{1,32}$')
 VALID_CONTEXT_MENU_NAME = re.compile(r'^[?!\w\s-]{1,32}$')
 CAMEL_CASE_REGEX = re.compile(r'(?<!^)(?=[A-Z])')
 
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -125,8 +125,8 @@ else:
 
 
 CheckInputParameter = Union['Command[Any, ..., Any]', 'ContextMenu', CommandCallback, ContextMenuCallback]
-VALID_SLASH_COMMAND_NAME = re.compile(r'^[\w-]{1,32}$')
-VALID_CONTEXT_MENU_NAME = re.compile(r'^[\w\s-]{1,32}$')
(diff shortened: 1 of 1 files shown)


Rapptz:
I don't believe the API allows `!` and `?` as per [the documented regex](https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-naming).

pandabear189:
https://images-ext-1.discordapp.net/external/b0GkBhNWcsgj7VaDxhEDVc-NQ8nx5oGpX_QEbX2vD2c/https/media.discordapp.net/attachments/932066545738350640/957428255248494603/unknown.png 
This was done on pycord, i believe you can use them

Rapptz:
That's for context menus, which I guess is fair. I'll have to ask what the validation regex is like there. This PR however is still wrong since slash commands can't have ? or !.

Puncher1:
@Rapptz Yeah oh true. But maybe the error can be improved though because it says it has to be 1-32 chars long even it is 1-32 chars long.

Space-Turtle0:
Slash Commands don't allow for `?` and `!` but since context menu's do, maybe it's best to revert changes on the regex for slash commands and keep the PR open for the change on context menu's.

Puncher1:
Updated the PR to only allow context menus to have `?` and `!` in their name.

Rapptz (approved):
Thanks.

PR #7779, opened 2022-03-27
Update package metadata

ofek:
## Checklist

- [ ] This PR fixes an issue.

## Background

Hello there! The Python packaging ecosystem has standardized on the interface for build backends ([PEP 517](https://www.python.org/dev/peps/pep-0517/)/[PEP 660](https://www.python.org/dev/peps/pep-0660/)) and the format for metadata declaration ([PEP 621](https://www.python.org/dev/peps/pep-0621/)/[PEP 631](https://www.python.org/dev/peps/pep-0631/)). As a result, the use of `setup.py` files is now [heavily discouraged](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).

So, I'm spending my free time updating important projects so that they are modernized and set an example for others 😄

## Summary of changes

This implements [PEP 621](https://www.python.org/dev/peps/pep-0621/), obviating the need for `setup.py` and `MANIFEST.in`. The build backend `hatchling` is quite stable and actively maintained, the only reason the version is 0.x is because 1.0.0 will drop support for Python 2. It's also on the major distribution channels such as [Debian](https://salsa.debian.org/python-team/packages/hatchling), [Fedora](https://src.fedoraproject.org/rpms/python-hatchling), [Arch Linux](https://archlinux.org/packages/community/any/python-hatchling/), [conda-forge](https://anaconda.org/conda-forge/hatchling), [Nixpkgs](https://github.com/NixOS/nixpkgs/blob/staging/pkgs/development/python-modules/hatchling/default.nix), etc.

Other projects have done this such as [pipx](https://github.com/pypa/pipx/blob/2ecc668acf472ad6956cc682499c077c1130d17e/pyproject.toml), [tox](https://github.com/tox-dev/tox/blob/dbbcb37a816b0fbc72280ef26aeca7dece8b2858/pyproject.toml), all [Datadog Agent integrations](https://github.com/DataDog/integrations-core/pulls?q=is%3Apr+author%3Aofek+in%3Atitle+Add+pyproject.toml+file), etc.

In comparison to `setuptools`, Hatchling has reproducible builds, supports the modern [PEP 660](https://peps.python.org/pep-0660/) editable installation standard, has a more intuitive plugin system, and the codebase is an order of magnitude smaller thus making it more auditable and less prone to bugs (`setuptools` is based on a [deprecated](https://peps.python.org/pep-0632/) standard library module that it now must vendor).

## Notes

- The Git version logic previously in `setup.py` is replaced by a plugin for the well-known [setuptools_scm](https://github.com/pypa/setuptools_scm) (which despite the legacy naming is actually now decoupled from `setuptools`)


.github/workflows/test.yml
```diff
@@ -26,8 +26,11 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip setuptools wheel "coverage[toml]" pytest pytest-asyncio pytest-cov pytest-mock
-          pip install -U -r requirements.txt
+          python -m pip install --upgrade "coverage[toml]" pytest pytest-asyncio pytest-cov pytest-mock
+
+      - name: Install package
+        run: |
+          pip install -e .
```
JelNiSlaw, line 33: You can (should?) use single line strings here

ofek: Either work; would you like me to change it?

JelNiSlaw: I'm not really someone that can decide on such things, but a normal string would look nicer here

ofek: It seems all of the other steps use this format

jack1142:
You should probably be more clear about `hatch` being a project made by *you* because you're obviously going to be biased due to this.

With the goal you have set to achieve, I feel like moving to `setup.cfg` or setuptools's new `pyproject.toml` support would make much more sense for this project rather than going with such a radical move when the tool (setuptools) that this project already uses is a perfectly fine build backend and there is no need to introduce a new tool here.

ofek:
> going to be biased due to this

Ah that is true, thank you. Though the only subjective thing I mentioned was the `more intuitive plugin system` bit; the other enumerated benefits are descriptively true.

> With the goal you have set to achieve, I feel like moving to setup.cfg or setuptools's new pyproject.toml support would make much more sense for this project rather than going with such a radical move

I do understand that critique but it's actually not that radical of a move if you read https://peps.python.org/pep-0517/#abstract. The packaging PEPs have made it so the ecosystem is free from `setuptools` lock-in. Now there is no standard tool but rather any tool is valid as long as the standards are implemented.

> not a good way to get projects to use it

My main motivation is rather the inverse: to move projects _away_ from `setup.py`/`setuptools`. Would you prefer a different build backend?

jack1142:
> Ah that is true, thank you. Though the only subjective thing I mentioned was the `more intuitive plugin system` bit; the other enumerated benefits are descriptively true.

Sure but I more meant that the bias comes from you suggesting to switch to *your* backend even though there are other build backends that are equally as valid and that's why I feel it is important to mention that you maintain `hatch`.

> I do understand that critique but it's actually not that radical of a move if you read [peps.python.org/pep-0517/#abstract](https://peps.python.org/pep-0517/#abstract). The packaging PEPs have made it so the ecosystem is free from `setuptools` lock-in. Now there is no standard tool but rather any tool is valid as long as the standards are implemented.

PEP 517's goal was to allow other build backends to exist by standardizing backends, it was not the intention to discourage usage of setuptools (emphasis mine):
> The goal of this PEP is get distutils-sig out of the business of being a gatekeeper for Python build systems. **If you want to use distutils, great**; if you want to use something else, then that should be easy to do using standardized methods.


> My main motivation is rather the inverse: to move projects away from setup.py/setuptools. Would you prefer a different build backend?

Personally (I'm not the maintainer of this project), I think that flit, poetry, hatch, etc. are all good build backends **but** setuptools is also a good build backend, and switching to a different backend doesn't really achieve much if there is no clear benefit to this change and it causes some churn for the project.
I'm aware that you have listed various reasons in the PR description, though as I read through it, the only one that is meaningful to an established project like this could be reproducible builds. I'm not sure if that's going to be enough for the project maintainer to switch to a different build backend.
The plugin system is a very nice thing *but* unless the project maintainer is actually struggling with some problem that could be solved by a plugin, it doesn't really give a reason to switch *now*.

ofek:
Definitely fair. I think this PR is ready for review since the CI is passing on my fork.

> The plugin system is a very nice thing but unless the project maintainer is actually struggling with some problem that could be solved by a plugin, it doesn't really give a reason to switch now.

Oh BTW if this does get merged I was planning on making a follow-up PR introducing [a plugin](https://github.com/ofek/hatch-mypyc) for compiling code with [Mypyc](https://github.com/mypyc/mypyc) to speed things up.

Gobot1234:
This repo doesn't target mypy (it targets pyright), using mypyc to speed it up would probably be a bad idea

.github/workflows/test.yml
nihaals: I agree that it's better as a single line string but the repo uses multiline strings for all other `run`s including single line ones so it makes sense to keep it consistent

ofek:
> This repo doesn't target mypy (it targets pyright), using mypyc to speed it up would probably be a bad idea

All type checkers are pretty much the same except for error messages. If there are differences it's usually that one has implemented a new PEP faster than another. I think one supports recursive types whereas the rest do not.

mikeshardmind:
> Oh BTW if this does get merged I was planning on making a follow-up PR introducing [a plugin](https://github.com/ofek/hatch-mypyc) for compiling code with [Mypyc](https://github.com/mypyc/mypyc) to speed things up.

> All type checkers are pretty much the same except for error messages. If there are differences it's usually that one has implemented a new PEP faster than another. I think one supports recursive types whereas the rest do not.

This repo contains several things which no currently existing type checkers fully understand(yet?), and has helpful lies in type checker via if type_checking blocks (to ensure the types are correct from the consumer standpoint where the type checkers currently don't have a means of understanding something used in it's true form (for example, the custom enum class)). mypyc would not function well here.

Rapptz:
Hello, thanks for all the effort that went into this PR.

As it stands right now, I have no plans to switch from setuptools to another build system. Rather, in the future when PEP-621 becomes more widely available and setuptools supports it (which, as far as I'm aware happened with the 0.61.0 release of setuptools) I'd move some of the package metadata over to that similar to this PR but without switching from setuptools.

I'm overall similar in the sentiments that @jack1142 described above.

> As a result, the use of setup.py files is now [heavily discouraged](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).

To my knowledge, `setup.py` itself is not deprecated. Its user-level invocation is. This distinction is important, IMO. It also doesn't make sense to deprecate a file that is such a widespread cornerstone in the ecosystem. Build scripts such as `setup.py` will always be necessary for certain complicated use cases that can't be appropriately described with bare metadata in a declarative manner.

Note that in the future it is not entirely unlikely that discord.py will make use of native modules written in either C++ or Rust, of which setuptools is the tool of choice I'm familiar with when this transition will happen.

> The Git version logic previously in setup.py is replaced by a plugin for the well-known [setuptools_scm](https://github.com/pypa/setuptools_scm) (which despite the legacy naming is actually now decoupled from setuptools)

I generally want to avoid unnecessary dependencies, especially ones where it can be simply done with a few lines of code such as it was done before.

> All type checkers are pretty much the same except for error messages. If there are differences it's usually strictness or that one has implemented a new PEP faster than another. I think one supports recursive types whereas the rest do not.

This is strictly not true. Mypy, for example, has a lot of limitations because it does not understand Python expressions or metaclasses at the same level as pyright does. For example, mypy hardcodes `X[Y]` expressions to have a specific meaning even though in Python the meaning the `X[Y]` differs depending on the context. Pyright understands these different contexts, while mypy will always fallback to `Any` when it does not. 

There are a lot of fundamental design differences that make one preferably over the other when actually developing that go far beyond the error messages.

ofek:
Sounds good, take care!

ofek:
Just FYI Hatch was adopted by the Python Packaging Authority:

- https://github.com/pypa/hatch
- https://mail.python.org/archives/list/pypa-committers@python.org/thread/J2BL5XTTOCXUCELUE7L5P5HHJ7NXZGZ5/

@Rapptz Would a switch be acceptable to you in the future, given any criteria you decide?

PR #7781, opened 2022-03-27
Add support for int() to return ID for AppCommand and AppCommandChannel

ImAnAgent:
## Summary

Adding the ability to get the ID of an AppCommand or AppCommandChannel using int()

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
Using `__int__` to get the ID is an unsupported operation in all models, so it makes little sense to do that here.

Note that these are logically not integers, so the cast makes no logical sense to support.

PR #7782, opened 2022-03-27
Update comments after `# type: ignore` to be compatible with PEP 484

jack1142:
## Summary

Without this, mypy outright refuses to run when d.py is part of dependencies:
image
[Per PEP 484](https://peps.python.org/pep-0484/#type-comments):
> In some cases, linting tools or other comments may be needed on the same line as a type comment. In these cases, the type comment should be before other comments and linting markers:
> > ```py
> > # type: ignore # <comment or other marker>
> > ```

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -511,7 +511,7 @@ class GuildChannel:
 
         If there is no category then this is ``None``.
         """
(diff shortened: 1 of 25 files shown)


Rapptz (approved):
Thanks.

PR #7787, opened 2022-03-28
refactor(rename): refactor attribute access

SaidBySolo:
## Summary

Mypy reported that #type: ignore is invalid.

This is a refactoring that solves this.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -1347,11 +1347,11 @@ def rename(**parameters: str) -> Callable[[T], T]:
         if isinstance(inner, Command):
             _populate_renames(inner._params, parameters)
         else:
-            try:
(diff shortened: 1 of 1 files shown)


SaidBySolo:
> I think the ideal solution here is actually to change the format of the comment like #7782 did

I think it's better to fix it than to ignore it.

However, if you mean to incorporate code style, we will roll back without delay and follow your comments.

Rapptz:
Yes this should use `#` instead of `-`.

SaidBySolo:
> Yes this should use `#` instead of `-`.
> 
> > I think it's better to refactor it than to ignore it.
> 
> This is not ignoring it. The type checker does not understand guarding with AttributeError, which is very much handling the error but the control flow does not understand it. Using global `getattr`/`setattr` functions to accomplish something Python already natively does just to appease the type checker seems bizarre to me.


Thanks for your comments!

PR #7789, opened 2022-03-28
Update types to use `Awaitable` where possible

bryanforbes:
## Summary

This updates some types in the library to use `typing.Awaitable` where possible

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -31,6 +31,7 @@ from typing import (
     AsyncIterator,
     Awaitable,
     Callable,
+    Coroutine,
     Dict,
     ForwardRef,
     Generic,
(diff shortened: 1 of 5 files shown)


discord/utils.py
```diff
@@ -396,11 +392,11 @@ def find(predicate: Callable[[T], Any], iterable: Iterable[T], /) -> Optional[T]
 
 
 @overload
-def find(predicate: Callable[[T], Any], iterable: AsyncIterable[T], /) -> Coro[Optional[T]]:
+def find(predicate: Callable[[T], Any], iterable: AsyncIterable[T], /) -> Awaitable[Optional[T]]:
```
Gobot1234, line 395: Surely this is better off using Coro so the return type is more specific?

bryanforbes: Good call. I'll revert get() as well.

PR #7797, opened 2022-03-29
use f strings wherever possible

r-priyam:
## Summary

Since 98% of library code uses `f-string`, so this PR aims to add `f-string` over string concatenation

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
Thanks but I'm not actually against string concatenation and a lot of these are kind of a personal preference thing.

PR #7815, opened 2022-03-31
Rework File.spoiler to synchronise with the filename

bijij:
## Summary

This PR reworks `File.filename` and `File.spoiler` to ensure the two match the internal state.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/file.py
+++ b/discord/file.py
@@ -38,11 +38,9 @@ __all__ = (
 
 
 def _strip_spoiler(filename: str) -> Tuple[str, bool]:
-    spoiler = False
-    while filename.startswith('SPOILER_'):
-        spoiler = True
-        filename = filename[8:]
-    return filename, spoiler
+    stripped = filename.lstrip('SPOILER_')
+    spoiler = stripped != filename
+    return stripped, spoiler
 
 
 class File:
--- a/discord/file.py
+++ b/discord/file.py
@@ -124,7 +124,7 @@ class File:
         If this is not given then it defaults to ``fp.name`` or if ``fp`` is
         a string then the ``filename`` will default to the string given.
         """
-        return 'SPOILER_' * self.spoiler + self._filename
+        return 'SPOILER_' + self._filename if self.spoiler else self._filename
 
     @filename.setter
     def filename(self, value: str) -> None:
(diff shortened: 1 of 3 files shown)


discord/file.py
```diff
@@ -23,18 +23,28 @@
...
 )
 # fmt: on
 
 
+def _strip_spoiler(filename: str) -> Tuple[str, bool]:
+    spoiler = False
+    while filename.startswith('SPOILER_'):
+        spoiler = True
+        filename = filename[8:]
+    return filename, spoiler
```
Rapptz, line 45:
```suggestion
def _strip_spoiler(filename: str) -> Tuple[str, bool]:
    stripped = filename.lstrip('SPOILER_')
    spoiler = stripped != filename
    return stripped, spoiler
```

discord/file.py
```diff
@@ -100,18 +107,30 @@ def __init__(
...
         self.description: Optional[str] = description
 
+    @property
+    def filename(self) -> str:
+        """:class:`str`:
+        The filename to display when uploading to Discord.
+        If this is not given then it defaults to ``fp.name`` or if ``fp`` is
+        a string then the ``filename`` will default to the string given.
+        """
+        return 'SPOILER_' * self.spoiler + self._filename
```
Rapptz, line 128:
```suggestion
        return 'SPOILER_' + self._filename if self.spoiler else self._filename
```

Don't be too clever.

discord/file.py
```diff
@@ -100,18 +107,30 @@ def __init__(
...
-        self.spoiler: bool = spoiler or (self.filename is not None and self.filename.startswith('SPOILER_'))
+        self.spoiler: bool = spoiler
         self.description: Optional[str] = description
 
+    @property
+    def filename(self) -> str:
+        """:class:`str`:
+        The filename to display when uploading to Discord.
+        If this is not given then it defaults to ``fp.name`` or if ``fp`` is
+        a string then the ``filename`` will default to the string given.
```
Rapptz, line 126:
```suggestion
        """:class:`str`: The filename to display when uploading to Discord.
        If this is not given then it defaults to ``fp.name`` or if ``fp`` is
        a string then the ``filename`` will default to the string given.
```

Rapptz (changes requested):
Needs a migrating entry.

Rapptz (approved):

PR #7833, opened 2022-04-03
[ext.commands] Implement commands.parameter

Gobot1234:
## Summary

Adds `commands.parameter` and `commands.Parameter` to handle metadata on a parameter for a command.
Adds support for:
- Late binding default arguments (includes `Author`, `CurrentChannel` and `CurrentGuild` from #1849)
- Work around for typecheckers not liking
  ```py
  class Converter(commands.Converter[int]): ...
  @commands.command()
  async def foo(ctx, bar: Converter):
      bar.as_integer_ratio()  # Converter has no attribute as_integer_ratio
  ```
- A custom string in the signature of a command.

## Checklist

- [ ] This PR fixes an issue.

Co-authored-by: khazhyk <khazhy@gmail.com>
Supersedes #1849


discord/ext/commands/parameters.py
```diff
@@ -0,0 +1,264 @@
...
+param = parameter
+"""param(*, converter=..., default=..., displayed_default=...)
+
+An alias for :func:`parameter`.
+"""
+
+# some handy defaults
+Author: Union[Member, User] = parameter(
+    default=attrgetter("author"),
+    displayed_default="<you>",
```
jack1142, line 241: Shouldn't this be `[you]` since the presence of the default makes this optional?

Gobot1234: I thought it should be <> as it's meant to represent a custom object you can't construct similar to https://discordpy.readthedocs.io/en/master/ext/commands/api.html#discord.ext.commands.Context.send_help and others, interested as to what @Rapptz thinks.

jack1142: Never mind, I got confused a bit... The usage ends up being `[p]command [argument=<you>]`, I don't know why I thought it would be `[p]command <you>`. Sorry about that 😄

discord/ext/commands/parameters.py
```diff
@@ -0,0 +1,269 @@
...
+    Because this is such a common use-case, the library provides :obj:`.Author`, :obj:`.CurrentChannel` and
+    :obj:`.CurrentGuild`, armed with this we can simplify ``wave`` to:
+
+    .. code-block:: python3
+
+        @bot.command()
+        async def wave(to: discord.User = commands.Author):
+            await ctx.send(f'Hello {to.mention} :wave:')
+
+    :obj:`.Author` and co also have other benefits like having the displayed default being filled.
```
Rapptz, line 214: I think this belongs in the prose docs instead (commands.rst) and an example can be added here.

discord/ext/commands/parameters.py
```diff
@@ -0,0 +1,269 @@
...
+An alias for :func:`parameter`.
+"""
+
+# some handy defaults
+Author: Union[Member, User] = parameter(
+    default=attrgetter("author"),
+    displayed_default="<you>",
+    converter=Union[converter.MemberConverter, converter.UserConverter],
+)
+"""A default :class:`Parameter` which returns the :attr:`~.Context.author` for this context."""
```
Rapptz, line 249: Can these floating docstrings be removed from the defaults since they're already explicitly enumerated in the API documentation?

Rapptz (changes requested):
In the future please don't sort imports. It causes unnecessary diffs and isn't even part of CI or the development flow.

This library also uses single quotes not double quotes.

That aside, the type changes from inspect.Parameter -> Parameter need to be mentioned in the migrating.rst

Rapptz (changes requested):
The library uses single quotes not double quotes.

discord/ext/commands/parameters.py
```diff
@@ -0,0 +1,238 @@
...
+    )
+    from .context import Context
+
+__all__ = (
+    "Parameter",
+    "parameter",
+    "param",
+    "Author",
+    "CurrentChannel",
+    "CurrentGuild",
```
Rapptz, line 52:
```suggestion
    'Parameter',
    'parameter',
    'param',
    'Author',
    'CurrentChannel',
    'CurrentGuild',
```

discord/ext/commands/parameters.py
```diff
@@ -0,0 +1,238 @@
...
+    A custom default can be used to have late binding behaviour.
+
+    .. code-block:: python3
+
+        @bot.command()
+        async def wave(to: discord.User = commands.parameter(default=lambda ctx: ctx.author)):
+            await ctx.send(f'Hello {to.mention} :wave:')
+
+    .. versionadded:: 2.0.0
+
```
Rapptz, line 183:
```suggestion
```

discord/ext/commands/parameters.py
```diff
@@ -0,0 +1,238 @@
...
+def parameter(
+    *,
+    converter: Any = empty,
+    default: Any = empty,
+    displayed_default: str = empty,
+) -> Any:
+    r"""parameter(\*, converter=..., default=..., displayed_default=...)
+
+    A way to assign custom metadata for a :class:`Command`\'s parameter.
+
```
Rapptz, line 171:
```suggestion

    ..versionadded:: 2.0

```

discord/ext/commands/parameters.py
```diff
@@ -0,0 +1,238 @@
...
+empty: Any = inspect.Parameter.empty
+
+
+def _gen_property(name: str) -> property:
+    return property(
+        attrgetter(f"_{name}"),
+        lambda self, value: setattr(self, "_{name}", value),
+        doc="The parameter's {name}.",
+    )
```
Rapptz, line 72:
```suggestion
    attribute = f'_{name}'
    return property(
        attrgetter(attribute),
        lambda self, value: setattr(self, attribute, value),
        doc=f"The parameter's {name}.",
    )
```

discord/ext/commands/parameters.py
```diff
@@ -0,0 +1,238 @@
...
+            default=default,
+            annotation=annotation,
+            displayed_default=displayed_default,
+        )
+
+    if not TYPE_CHECKING:  # this is to prevent anything breaking if inspect internals change
+        name = _gen_property("name")
+        kind = _gen_property("kind")
+        default = _gen_property("default")
+        annotation = _gen_property("annotation")
```
Rapptz, line 123:
```suggestion
        name = _gen_property('name')
        kind = _gen_property('kind')
        default = _gen_property('default')
        annotation = _gen_property('annotation')
```

discord/ext/commands/parameters.py
```diff
@@ -0,0 +1,238 @@
...
+
+        return self.__class__(
+            name=name,
+            kind=kind,
+            default=default,
+            annotation=annotation,
+            displayed_default=displayed_default,
+        )
+
+    if not TYPE_CHECKING:  # this is to prevent anything breaking if inspect internals change
```
Rapptz, line 119: I think as a consequence this is storing double the amount of attributes? The internal `inspect.Parameter` attributes and then the recreations in `Parameter`? So this would make Command objects take more memory. I'm sure this doesn't matter since most people aren't storing that many of these objects but it is unfortunate.

Maybe `inspect.Parameter` is slotted and this could use the same treatment? Unsure.

discord/ext/commands/parameters.py
```diff
@@ -0,0 +1,238 @@
...
+
+def _gen_property(name: str) -> property:
+    return property(
+        attrgetter(f"_{name}"),
+        lambda self, value: setattr(self, "_{name}", value),
+        doc="The parameter's {name}.",
+    )
+
+
+class Parameter(inspect.Parameter):
```
Rapptz, line 75: This has no docstring so the documentation for it is empty.

Rapptz (changes requested):

discord/ext/commands/parameters.py
Gobot1234: This only storing the instance attributes from Parameter once is it not (Parameter has them called _name, _default etc)? Adding slots seems like a good idea

Rapptz (approved):

Gobot1234:
Thank you for all the effort reviewing this :)

Rapptz:
Good job on this.

PR #7834, opened 2022-04-03
Add autocomplete method to Transformers

bijij:
## Summary

This pr adds a new Transformer.autocomplete method which will automatically assign the autocomplete function on commands which use the transformer.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/transformers.py
+++ b/discord/app_commands/transformers.py
@@ -258,7 +258,9 @@ class Transformer:
 
     @classmethod
     @_not_overridden
-    async def autocomplete(cls, interaction: Interaction, value: Union[int, float, str]) -> List[Choice[Union[int, float, str]]]:
+    async def autocomplete(
+        cls, interaction: Interaction, value: Union[int, float, str]
+    ) -> List[Choice[Union[int, float, str]]]:
         """|coro|
 
         An autocomplete prompt handler to be automatically used by options using this transformer.
--- a/discord/app_commands/transformers.py
+++ b/discord/app_commands/transformers.py
@@ -265,7 +265,7 @@ class Transformer:
 
         .. note::
 
-            Autocomplete is only supported for options with :attr:`~discord.app.commands.Transformer.type`
+            Autocomplete is only supported for options with :meth:`~discord.app.commands.Transformer.type`
             :attr:`~discord.AppCommandOptionType.string`, :attr:`~discord.AppCommandOptionType.integer`,
             and :attr:`~discord.AppCommandOptionType.number`.
 
--- a/discord/app_commands/transformers.py
+++ b/discord/app_commands/transformers.py
@@ -265,8 +265,8 @@ class Transformer:
 
         .. note::
 
-            Autocomplete is only supported for options with :attr:`~discord.app.commands.Transformer.type` 
-            :attr:`~discord.AppCommandOptionType.string`, :attr:`~discord.AppCommandOptionType.integer`, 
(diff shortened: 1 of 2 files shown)


discord/app_commands/transformers.py
```diff
@@ -679,4 +714,12 @@ def annotation_to_parameter(annotation: Any, parameter: inspect.Parameter) -> Co
     if parameter.kind in (parameter.POSITIONAL_ONLY, parameter.VAR_KEYWORD, parameter.VAR_POSITIONAL):
         raise TypeError(f'unsupported parameter kind in callback: {parameter.kind!s}')
 
+    if not hasattr(inner.autocomplete.__func__, '__transformer_autocomplete_not_overridden__'):
+        from .commands import _validate_auto_complete_callback
+
+        async def func(*args, **kwargs):
+            return await inner.autocomplete(*args, **kwargs)
+
+        result.autocomplete = _validate_auto_complete_callback(func, skip_binding=True)
```
Rapptz, line 723: This doesn't seem proper. This validation doesn't do anything.

discord/app_commands/transformers.py
```diff
@@ -249,6 +256,34 @@ async def transform(cls, interaction: Interaction, value: Any) -> Any:
         """
         raise NotImplementedError('Derived classes need to implement this.')
 
+    @classmethod
+    @_not_overridden
+    async def autocomplete(cls, interaction: Interaction, value: ChoiceT) -> List[Choice[ChoiceT]]:
```
Rapptz, line 261: In public interfaces for inherited members, the type hints get auto-filled in when using certain tools so these need to be publicly named types.
```suggestion
    async def autocomplete(cls, interaction: Interaction, value: Union[int, float, str]) -> List[Choice[Union[int, float, str]]]:
```

discord/app_commands/transformers.py
```diff
@@ -249,6 +256,34 @@ async def transform(cls, interaction: Interaction, value: Any) -> Any:
...
+        """|coro|
+
+        An autocomplete prompt handler to be automatically used by options using this transformer.
+
+        .. note::
+
+            Autocomplete is only supported for options with :meth:`~discord.app_commands.Transformer.type`
+            :attr:`~discord.AppCommandOptionType.string`, :attr:`~discord.AppCommandOptionType.integer`,
+            and :attr:`~discord.AppCommandOptionType.number`.
```
Rapptz, line 270:
```suggestion
            Autocomplete is only supported for options with a :meth:`~discord.app_commands.Transformer.type`
            of :attr:`~discord.AppCommandOptionType.string`, :attr:`~discord.AppCommandOptionType.integer`,
            or :attr:`~discord.AppCommandOptionType.number`.
```

Rapptz (changes requested):

discord/app_commands/transformers.py
bijij: Yeah I'm not sure how to avoid it though, since no matter what it's going to try setting `.requires_binding` on the callback passed, which unless we always take `__func__` out, is going to fail.

Could rework it such that only if requires_binding is True do we set the attribute, and use getattr when checking it?

Rapptz:
```py
defined_autocomplete = inner.__dict__['autocomplete']
if isinstance(defined_autocomplete, (classmethod, staticmethod)):
    result.autocomplete = _validate_auto_complete_callback(inner.autocomplete, skip_binding=True)
else:
    raise TypeError('autocomplete method must be a classmethod or a staticmethod')
```

Rapptz (approved):

PR #7866, opened 2022-04-10
[commands] Add option ending_note to DefaultHelpCommand

TobiWan54:
## Summary

This simply allows users to easily use their own ending note with the default help command, by adding ending_note='a string' to the constructor.

This is my first ever pull request so let me know if it's dumb.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
Hello. Thanks for the PR.

I think for this type of stuff it's better to subclass. Most of the attributes that are available right now was before subclassing was possible and it's a relic of the old design. Going forward though it makes sense to just tell people to subclass for the behaviour they want to achieve.

PR #7867, opened 2022-04-10
Add `jump_url` property to Channels

Pysics:
## Summary

- `TextChannel`
- `VoiceChannel`
- `StageChannel`
- `ForumChannel`
- `DMChannel`
- `GroupChannel`

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -228,6 +228,13 @@ class TextChannel(discord.abc.Messageable, discord.abc.GuildChannel, Hashable):
         """
         return [thread for thread in self.guild._threads.values() if thread.parent_id == self.id]
 
+    @property
+    def jump_url(self) -> str:
+        """:class:`str`: Returns a URL that allows the client to jump to the channel.
+        .. versionadded:: 2.0
+        """
+        return f'https://discord.com/channels/{self.guild.id}/{self.id}'
+
     def is_nsfw(self) -> bool:
         """:class:`bool`: Checks if the channel is NSFW."""
         return self.nsfw
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -1066,6 +1066,13 @@ class VoiceChannel(discord.abc.Messageable, VocalGuildChannel):
         """
         return self._state._get_message(self.last_message_id) if self.last_message_id else None
 
+    @property
+    def jump_url(self) -> str:
+        """:class:`str`: Returns a URL that allows the client to jump to the channel.
+        .. versionadded:: 2.0
+        """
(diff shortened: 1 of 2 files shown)


LostLuma:
This is missing the property on `ForumChannel`, although I'm unsure this will be merged as the feature has been denied previously (see #2355).

Vioshim:
> This is missing the property on `ForumChannel`, although I'm unsure this will be merged as the feature has been denied previously (see #2355).

To be fair, back then we didn't have discord components, this jump_url feature in channels would be very useful to redirect users through URL Buttons in views.

Gobot1234:
This is also missing support for DMChannel is it not?

Pysics:
For now, I have made this PR a draft.
I will look into implementing more features tomorrow.

dolfies:
> This is missing the property on `ForumChannel`, although I'm unsure this will be merged as the feature has been denied previously (see #2355).

Copy link is an option in the client's context menu now, so it isn't a "hack" anymore.

Pysics:
I have added more support and this PR is ready for review.

discord/channel.py
```diff
@@ -1059,6 +1066,13 @@ def last_message(self) -> Optional[Message]:
         """
         return self._state._get_message(self.last_message_id) if self.last_message_id else None
 
+    @property
+    def jump_url(self) -> str:
+        """:class:`str`: Returns a URL that allows the client to jump to the channel.
+        .. versionadded:: 2.0
```
Rapptz, line 1072: These need a newline between them.

```suggestion

        .. versionadded:: 2.0
```

Rapptz (changes requested):
Rather than copying this implementation everywhere in every guild channel type you probably want to do it inside abc.GuildChannel.

Pysics:
I have not added support for `CategoryChannel`.
So, I can't add the `jump_url` property to `abc.GuildChannel`.

Rapptz:
> I have not added support for `CategoryChannel`. So, I can't add the `jump_url` property to `abc.GuildChannel`.

I'd rather it be in all channels than duplicate all of this. Plus, this would play better with type checkers.

Pysics:
@Rapptz I have made the requested changes.

Rapptz:
This is failing type checking.

You don't want to add this jump_url property to the abc.PrivateChannel protocol. That one will have to be done manually to the things that meet the protocol since it isn't an actual runtime base class.

Rapptz (changes requested):
Needs to fix type checking concern.

Rapptz (approved):
Thanks. LGTM.

PR #7879, opened 2022-04-13
Correct order of arguments in View.on_error

Snaptraks:
## Summary

According to [this message](https://discord.com/channels/336642139381301249/381965829857738772/958229416536858684) the order for arguments of callbacks inside Views should be `interaction, item`. This corrects the order in the View.on_error.

If I missed something or it was intentional, let me know!

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ui/view.py
+++ b/discord/ui/view.py
@@ -381,7 +381,7 @@ class View:
         """
         pass
 
-    async def on_error(self, error: Exception, interaction: Interaction, item: Item[Any]) -> None:
+    async def on_error(self, interaction: Interaction, error: Exception, item: Item[Any]) -> None:
         """|coro|
 
         A callback that is called when an item's callback or :meth:`interaction_check`
@@ -391,10 +391,10 @@ class View:
 
         Parameters
         -----------
-        error: :class:`Exception`
-            The exception that was raised.
         interaction: :class:`~discord.Interaction`
             The interaction that led to the failure.
+        error: :class:`Exception`
+            The exception that was raised.
         item: :class:`Item`
             The item that failed the dispatch.
         """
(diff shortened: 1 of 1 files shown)


discord/ui/view.py
```diff
@@ -381,7 +381,7 @@ async def on_timeout(self) -> None:
         """
         pass
 
-    async def on_error(self, error: Exception, item: Item[Any], interaction: Interaction) -> None:
+    async def on_error(self, error: Exception, interaction: Interaction, item: Item[Any]) -> None:
```
Rapptz, line 384:
```suggestion
    async def on_error(self, interaction: Interaction, error: Exception, item: Item[Any]) -> None:
```

Interaction should always be first.

Rapptz (changes requested):
It is amazing to me that I somehow missed this. Thanks, there's one error though.

discord/ui/view.py
Snaptraks: Should `error` be between `interaction` and `item`, or be last?

Rapptz: Amusingly, `Tree.on_error` does `(interaction, command, error)`.

I am not a fan of three parameter callbacks. For that one in particular it might be worth removing `command` altogether since there's `Interaction.command` now.

Snaptraks: This is, I think, beyond the scope of this PR, as here Interaction was clearly not before Item. I've applied the suggested changes.

Rapptz: Yeah, it is outside of the scope of the PR. It's just a relevant musing.

Rapptz (approved):
Thanks.

PR #7885, opened 2022-04-14
feat: filename and description kwargs for to_file

DenverCoder1:
## Summary

Adds kwargs to `Asset.to_file` and `Attachment.to_file` for overriding the `filename` or `description` when creating the `File`.

Fixes to docstrings

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/asset.py
+++ b/discord/asset.py
@@ -125,7 +125,13 @@ class AssetMixin:
             with open(fp, 'wb') as f:
                 return f.write(data)
 
-    async def to_file(self, *, spoiler: bool = False) -> File:
+    async def to_file(
+        self,
+        *,
+        filename: Optional[str] = None,
+        description: Optional[str] = None,
+        spoiler: bool = False,
+    ) -> File:
         """|coro|
 
         Converts the asset into a :class:`File` suitable for sending via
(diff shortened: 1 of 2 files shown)


Rapptz (commented):
I'm not sure I get the point of this considering you can just set the filename and description afterwards, plus these are conversion routines that are meant to transform something from type A to type B with the same information.

The actual changeset looks okay to me, I'm just not sure of the merit of the feature being proposed and I'm not sure it's popular enough to warrant inclusion, especially when there's already a viable and readable alternative.

DenverCoder1:
Yeah, you can set it afterwards and that's what I said originally when it was suggested by kal on the Discord server.

I do think it makes a simpler to be able to do it with 1 line though.

```py
await ctx.send(file=ctx.author.display_avatar.with_format("png").to_file(filename="avatar.png"))
```
vs.
```py
file = ctx.author.display_avatar.with_format("png").to_file()
file.filename = "avatar.png"
await ctx.send(file=file)
```

If you disagree, I could disallow overriding parameters that it is able to get from the Asset or Attachment.

Rapptz:
I generally don't like adding things just to encourage things being done in 1-line since I think that can be more unreadable than not. For example, your example is nearing high density for my taste. That aside, I gave it some thought and I guess there's no real harm in adding it.

Rapptz (approved):
Thanks.

PR #7886, opened 2022-04-14
add self_deaf and self_mute params to Connectable.connect and related…

### LINKED ISSUE #7805: Add self_deaf and self_mute to Voice(Channel/Client/Protocol).connect()
### Summary

`VoiceClient.connect()` calls `Guild.change_voice_state()` to connect to the voice channel so it could set self_deaf/mute kwargs if they were passable to connect()

The core library

### The Problem

With the current flow, you have to first call `voice_channel.connect()` and only then you can self-deafen/mute with a call to `guild.change_voice_state()`. When using VoiceProtocol, I can choose to call `guild.change_voice_state()` in `voice_client.connect()` with a different default but I can't choose this at the time of the call to `voice_channel.connect()` as it's the one that calls `voice_client.connect()`.

### The Ideal Solution

Add `self_deaf` and `self_mute` keyword arguments to `abc.Connectable.connect()` that will be passed to `VoiceProtocol.connect()` so that it can use them when connecting.

### The Current Solution

factory function that would instantiate a subclass of `VoiceProtocol` with such arguments and `voice_client.connect()` could base its behavior on that. I think this isn't really that great as these arguments are specific to connect action and should be passed when `connect()` method is called, not by changing the instance's attributes whenever you want to have different behavior for connecting.

### Additional Context

_No response_

ImVaskel:
… classes

## Summary

Adds support for passing ``self_deaf`` and ``self_mute`` to Connectable.connect (and VoiceProtocol) to have the client self deafen /mute when it joins. 

Closes #7805.

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -1806,8 +1806,12 @@ class Connectable(Protocol):
             Defaults to :class:`~discord.VoiceClient`.
         self_mute: :class:`bool`
             Indicates if the client should be self-muted.
+
+            .. versionadded: 2.0
         self_deaf: :class:`bool`
             Indicates if the client should be self-deafened.
+
+            .. versionadded: 2.0
 
         Raises
         -------
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -1783,6 +1783,8 @@ class Connectable(Protocol):
         timeout: float = 60.0,
         reconnect: bool = True,
         cls: Callable[[Client, Connectable], T] = MISSING,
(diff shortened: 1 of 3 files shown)


Rapptz (changes requested):
Needs versionadded

jack1142:
This needs update to migrating docs as well as VoiceProtocol.connect is called with new keyword arguments

Rapptz (approved):

PR #7895, opened 2022-04-17
fixed Guild.fetch_member docstring

lightningyx:
## Summary

Guild.fetch_member raises a .NotFound error when member_id is invalid
and the docs were missing this information. This was fixed

If you want to reproduce the error, just call Guild.fetch_member with an invalid id.

For the error explanation I went for `Invalid Member ID.`, but `This member was not found.` could also work.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -1957,6 +1957,8 @@ class Guild(Hashable):
             You do not have access to the guild.
         HTTPException
             Fetching the member failed.
+        NotFound
+            Invalid Member ID.
 
         Returns
         --------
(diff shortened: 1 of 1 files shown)


discord/guild.py
```diff
@@ -1957,6 +1957,8 @@ async def fetch_member(self, member_id: int, /) -> Member:
             You do not have access to the guild.
         HTTPException
             Fetching the member failed.
+        NotFound
+            Invalid Member ID.
```
Rapptz, line 1961:
```suggestion
            The member could not be found.
```

PR #7897, opened 2022-04-17
Add some slash commands FAQs

PythonCoderAS:
## Summary

It adds some documentation that starts to list some common questions and answers for slash commands.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -418,4 +418,4 @@ My bot's commands are not showing up!
 1. Did you :meth:`~.CommandTree.sync` your command? Commands need to be synced before they will appear.
 2. Did you invite your bot with the correct permissions? Bots need to be invited with the ``application.commands``
    scope in addition to the ``bot`` scope. For example, invite the bot with the following URL:
-   https://discord.com/oauth2/authorize?client_id=<client id>&scope=applications.commands+bot
+   ``https://discord.com/oauth2/authorize?client_id=<client id>&scope=applications.commands+bot``
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -406,3 +406,16 @@ Example: ::
         await ctx.send(f'Pushing to {remote} {branch}')
 
 This could then be used as ``?git push origin master``.
+
+Slash Commands
+--------------
+
+Questions regarding discord's new Slash Commands.
+
+My bot's commands are not showing up!
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+1. Did you :meth:`~.CommandTree.sync` your command? Commands need to be synced before they will appear.
+2. Did you invite your bot with the correct permissions? Bots need to be invited with the ``application.commands``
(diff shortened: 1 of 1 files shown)


ChesterWOV-DPY (changes requested):

docs/faq.rst
```diff
@@ -406,3 +406,16 @@ Example: ::
...
+
+Questions regarding Discord's new Application Commands.
+
+My bot's commands are not showing up!
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+1. Did you :meth:`~.CommandTree.sync` your command? Commands need to be synced before they will appear.
+2. Did you invite your bot with the correct permissions? Bots need to be invited with the ``application.commands``
+   scope in addition to the ``bot`` scope. For example, invite the bot with the following URL:
+   https://discord.com/oauth2/authorize?client_id=<client id>&scope=applications.commands+bot
```
Rapptz, line 421:
```suggestion
   ``https://discord.com/oauth2/authorize?client_id=<client id>&scope=applications.commands+bot``
```

This should also probably mention that `utils.oauth_url` helps create this URL too.

docs/faq.rst
```diff
@@ -406,3 +406,16 @@ Example: ::
         await ctx.send(f'Pushing to {remote} {branch}')
 
 This could then be used as ``?git push origin master``.
+
+Application Commands
+--------------------
+
+Questions regarding Discord's new Application Commands.
```
Rapptz, line 413:
```suggestion
Questions regarding Discord's new application commands, frequently known as "slash commands" or "context menu commands".
```

Rapptz (changes requested):
Just a few wording nits.

docs/faq.rst
PythonCoderAS: Slash commands and context menu commands are not alternate names for application commands, rather they are types of application commands.

Rapptz: I don't see the relevance for the note. Most users will call them slash commands or context menu commands, that's what that note is saying. The API calls them application commands, yes, but users and developers just call them something else.

PythonCoderAS: The problem is that **slash commands** and **context menu commands** _are not the same thing_.

Rapptz: Sure. Except this has nothing to do with the note which is there to help users understand what an application command is from their perspective. This type of pedantry is unnecessary and is just holding back the merging of this PR.

PythonCoderAS: That's a valid point, but I was worried about someone calling context menu commands "slash commands" or vice versa.

PythonCoderAS:
Additionally, it might be time to change the default scopes for `oauth_url` to include the application commands scope as well.

Rapptz:
That change was made a few weeks ago by now.

PythonCoderAS:
> That change was made a few weeks ago by now.

Sorry, I was looking at a stale version of the docs, my bad.

Rapptz (approved):

PR #7913, opened 2022-04-21
[commands] Allow subtracting multiple tokens from cooldown at once

MinnDevelopment:
## Summary

This adds a keyword argument to `Cooldown.update_rate_limit` to allow subtracting multiple tokens from the cooldown at once. The idea is to use this for weighted cooldowns or simply avoiding loops. Note that this also allows to go into negative tokens, which would always result in a retry-after response as well.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/checks.py
+++ b/discord/app_commands/checks.py
@@ -112,7 +112,8 @@ class Cooldown:
         if not current:
             current = time.time()
 
-        tokens = self._tokens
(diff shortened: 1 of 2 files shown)


Rapptz (commented):
Maybe instead of swapping the operations you can leave the condition first as-is and when subtracting you can do `self._tokens = max(self._tokens - tokens, 0)`

PR #7922, opened 2022-04-23
Fix Client.fetch_guilds not working correctly, increase default limit

LostLuma:
## Summary

Fixes fetch_guilds yielding duplicate values, as well as only returning one `limit` worth of results when no other parameters are specified.

The endpoint gives results starting from the lowest ID, meaning the default strategy to get more results before the previous ones would yield no results and the `before` or `after` variable were set to a wrong value respectively.

The maximum limit for this endpoint was also increased to 200 when nitro users got access to the higher guild limit, so I've thrown that change in as well.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -1194,7 +1194,7 @@ class Client:
     async def fetch_guilds(
         self,
         *,
-        limit: Optional[int] = 100,
+        limit: Optional[int] = 200,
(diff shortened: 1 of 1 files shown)


Rapptz (approved):
Thanks. I found this while working on the ban pagination but I forgot to actually fix it. Glad it's been fixed now.

PR #7923, opened 2022-04-23
Add raw typing event, fix typing event not working for DMs

### LINKED ISSUE #7432: [v2]: DM typing and Thread typing events are not dispatched.
### Summary

DM typing events are not dispatched.

### Reproduction Steps

Start typing in a dm or a thread with an on_typing listener. The on_typing listener will never be dispatched.

### Minimal Reproducible Code

```python
import discord
import logging

logging.getLogger("discord").setLevel(logging.DEBUG)


    async def on_ready(self):
        print(f"Logged in as {self.user} (ID: {self.user.id})")
        print("------")

    async def on_typing(self, channel, user, _):
        # we do not want the bot to reply to itself
        if user.id == self.user.id:
            return

        # these never gets called in a DM channel or Thread.
        print(f"{user} is typing in {channel}")
        await channel.trigger_typing()

    async def on_socket_raw_receive(self, msg):
        print(msg)


client = MyClient(intents=discord.Intents(typing=True), enable_debug_events=True)
client.run("token")
```

### Expected Results

The on_typing event should be dispatched to the client for dm typing events.

### Actual Results

The on_typing event is not dispatched to the client for dm typing events.

### Intents

typing=True ; the issue occurs even with all intents

### System Information

- Python v3.8.10-final
- discord.py v2.0.0-alpha
    - discord.py pkg_resources: v2.0.0a3495+gb6590d7f
- aiohttp v3.7.4.post0
- system info: Linux 5.10.43.3-microsoft-standard-WSL2 #1 SMP Wed Jun 16 23:47:55 UTC 2021

### Checklist

### Additional Context

https://canary.discord.com/channels/815369174096412692/815369174096412695/878394332640591922

LostLuma:
## Summary

Fixes #7432 for users that are cached.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/raw_models.py
+++ b/discord/raw_models.py
@@ -330,7 +330,7 @@ class RawTypingEvent(_RawReprMixin):
     ----------
     channel_id: :class:`int`
         The ID of the channel the user started typing in.
-    user_id: :class:`int:
+    user_id: :class:`int`
         The ID of the user that started typing.
     user: Optional[Union[:class:`discord.User`, :class:`discord.Member`]]
         The user that started typing, if they could be found in the internal cache.
--- a/discord/raw_models.py
+++ b/discord/raw_models.py
@@ -332,7 +332,7 @@ class RawTypingEvent(_RawReprMixin):
         The ID of the channel the user started typing in.
     user_id: :class:`int:
         The ID of the user that started typing.
-    user: Union[:class:`discord.User`, :class:`discord.Member`]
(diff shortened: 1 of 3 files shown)


Rapptz:
This will probably need an provisional on_raw_typing event too.

LostLuma:
Can do, I was not sure whether I should do that.

PR #7925, opened 2022-04-23
Change event system to use only a single parameter for all events

lightningyx:
## Summary

This is still a draft and should implement this card:
https://github.com/Rapptz/discord.py/projects/3#card-58336471
> Change event system to use only a single parameter for all events. Potentially only in the core library since discord.ext.commands would be in a weird spot.

I started by adding classes for the first 36 events and noticed that [on_raw_integration_delete](https://discordpy.readthedocs.io/en/master/api.html#discord.on_raw_integration_delete) already receives an `RawIntegrationDeleteEvent` as payload, so wrapping it in another event feels kinda redundant.

## Todo

- [ ] Create classes for all events
- [ ] Use these classes in the gateway / dispatch methods
- [ ] Ensure no redundancy
- [ ] Update the docs

### DIFF
--- a/discord/event.py
+++ b/discord/event.py
@@ -1,0 +1,273 @@
+"""
+The MIT License (MIT)
+
+Copyright (c) 2015-present Rapptz
+
+Permission is hereby granted, free of charge, to any person obtaining a
+copy of this software and associated documentation files (the "Software"),
+to deal in the Software without restriction, including without limitation
+the rights to use, copy, modify, merge, publish, distribute, sublicense,
+and/or sell copies of the Software, and to permit persons to whom the
+Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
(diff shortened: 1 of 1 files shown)


discord/event.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+@dataclass(frozen=True)
+class SocketRawReceive(Event):
+    msg: str
+
+
+@dataclass(frozen=True)
+class SocketRawSend(Event):
+    payload: str | bytes
```
JelNiSlaw, line 171: This is a Python 3.10+ syntax, while discord.py 2.0 targets Python 3.9

AbstractUmbra: 3.8, actually! :)

JelNiSlaw: Oh I did it again, sorry for that!

XuaTheGrate: By using ``from __future__ import annotations`` this syntax is actually valid, however it's inconsistent with the rest of the library and should be replaced with a ``Union[str, bytes]`` regardless.

Rapptz:
Hey, thanks for the PR.

This is one of those design things that are in the maybe pile for more than a few reasons and the way this PR is envisioning it does not match the way I'm envisioning it. I know this is a draft but there is a lot of work that needs to be discussed in the discord.py server before even beginning to actually make a PR. I'm also not even sure if I want to do this in the v2 timeline or in the v3 timeline, hence the hesitancy in even implementing it.

PR #7928, opened 2022-04-24
Separately delay ready event for each shard

### LINKED ISSUE #5857: when using intents members get cached 1 shard at a time
## The Problem

When using the new intents on 1.5.0a and running multiple shards, waits until all guilds from the first shard are cached before requesting any guild members from other chunks

## The Ideal Solution

requesting member chunks for all shards at the same time, or using maybe moving rotating through shards when one is ratelimited to start working on the next one as it usually has to wait a full minute before continuing on that shard.

## The Current Solution

There currently is no way to achieve this speedup

## Summary

Make member caching faster for bots running multiple shards

LostLuma:
## Summary

Fixes #5857. As this allows simplifying the delay logic a lot I've mostly copied the implementation from the regular ConnectionState.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -484,7 +484,6 @@ class Client:
         self.loop = loop
         self.http.loop = loop
         self._connection.loop = loop
-        await self._connection.async_setup()
 
         self._ready = asyncio.Event()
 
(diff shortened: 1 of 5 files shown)


Rapptz (approved):
This looks fine at first glance, might be worth trialing it and seeing how it goes.

PR #7930, opened 2022-04-25
Tweak wording for "less than" Range error

diceroll123:
## Summary

Keeping it consistent!

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
If we're going to be tweaking these then I think these labels are technically swapped. It should probably be `X must be less than Y` and `X must be more than Y`.

diceroll123:
Hmm, but then we'd have to add and subtract 1 to the max and min respectively to make it strictly `less than`, or we'd have to go with `less than or equal to`. Your call!

Another thing that'd be nice is having the name of the variable in question here. `"item_id" value must be ...`

PR #7934, opened 2022-04-26
Update api.rst documentation to add timeout to on_member_update

killakej:
## Summary

Just updating the API reference doc to note that on_member_update is also called when communication_disabled_until (aka timeout) is updated (so, when users are timed out) - which I just denoted as timeout. Unsure if using communication_disable_until (as in discord docs) or timeout is preferable.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -683,6 +683,7 @@ Members
     - nickname
     - roles
     - pending
+    - timeout
 
     This requires :attr:`Intents.members` to be enabled.
 


docs/api.rst
```diff
@@ -683,6 +683,7 @@ Members
     - nickname
     - roles
     - pending
+    - timeout
```
Rapptz, line 686:
```suggestion
    - timeout

    Due to a Discord limitation, this event is not dispatched when a member's timeout expires.

```

Rapptz (changes requested):

docs/api.rst
```diff
@@ -683,6 +683,10 @@ Members
...
+
+    Due to a Discord limitation, this event is not dispatched when a member's timeout expires.
+
```
Rapptz, line 689:
```suggestion
```

Rapptz (approved):
Thanks.

PR #7939, opened 2022-04-27
Add intents parameter to basic example

adwaithkrishna:
## Summary

Add the intents parameter which is now required to the basic example in the intro docs

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -99,6 +99,8 @@ A quick example to showcase how events work:
 
 .. code-block:: python3
 
+    # This example requires the 'message_content' intent.
+
     import discord
 
     class MyClient(discord.Client):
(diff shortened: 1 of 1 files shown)


docs/intro.rst
```diff
@@ -99,6 +99,8 @@ A quick example to showcase how events work:
 
 .. code-block:: python3
 
+    # This example requires the 'message_content' intent.
+
```
nerdguyahmad, line 103:
```suggestion
```

`intents.message_content = True` already indicates this so this isn't really needed and to stay consistent with other examples across the documentation or GitHub/PyPi readme's.

nerdguyahmad (changes requested):

docs/intro.rst
adwaithkrishna: I have looked at other examples like the one in [Quickstart](https://github.com/Rapptz/discord.py/blob/master/docs/quickstart.rst#a-minimal-bot) and noticed that the same is being used there. So I guess it is actually telling the user to toggle it in the developer portal.

Rapptz (approved):
Thanks.

Issue #7949, opened 2022-04-28
[commands] issue with typing of commands.check_any


mikeshardmind:
### Summary

commands.check_any's typing is incompatible with the library's checks

### Reproduction Steps

used the commands.check_any decorator as documented.

### Minimal Reproducible Code

```python
from discord.ext import commands


@commands.check_any(commands.is_owner(), commands.is_nsfw())
@commands.command()
async def example(ctx: commands.Context):
    pass
```

### Expected Results

This to work at runtime, and be type checked properly (for the minimal example, I picked 2 checks from the library, but this holds for any use of the decorator as documented) 

### Actual Results

While this works at runtime, pyright reports:
```
Argument of type "(MaybeCoro[bool] | Context[Any]) -> (MaybeCoro[bool] | Context[Any])" cannot be assigned to parameter "checks" of type "Check[ContextT@check_any]" in function "check_any"
  Type "(MaybeCoro[bool] | Context[Any]) -> (MaybeCoro[bool] | Context[Any])" cannot be assigned to type "Check[ContextT@check_any]"
    Function return type "MaybeCoro[bool] | Context[Any]" is incompatible with type "MaybeCoro[bool]"
      Type "MaybeCoro[bool] | Context[Any]" cannot be assigned to type "MaybeCoro[bool]"
        Type "Context[Any]" cannot be assigned to type "MaybeCoro[bool]"
          "Context[Any]" is incompatible with "bool"
          "Context[Any]" is incompatible with "Coro[bool]"
```
(twice, once for each check)

This also breaks the inferred type of the decorated command.

### Intents

N/A

### System Information

- Python v3.10.2-final
- discord.py v2.0.0-alpha
    - discord.py pkg_resources: v2.0.0a4184+g10af1ea2
- aiohttp v3.8.1
- system info: Windows 10 10.0.19043

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.

### Additional Context

I can go ahead and PR a fix for this later, but there are really multiple possible fixes.

1. Change the typing of commands.check to return a protocol and accept that protocol in check_any
2. Change the typing of check_any to have less detailed type information.

I'd prefer the 1st route but figured this was worth running by others first.

PR #7950, opened 2022-04-29
[commands] Fix typing of checks

mikeshardmind:
I'm still working on this, but given the design space, I wanted to make sure I wasn't going too far with a direction that might not work out. I've done what I can to keep the user interface the same and avoid breaking both currently working code and currently correctly typed code given concerns about churn expressed in this Discord server.

A few things which could further simplify this, but are probably too much churn:

- Only allowing coroutines as user-provided predicates
- Enforcing that command decorators are the first decorator applied to callbacks

## Summary

see: #7949

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -60,7 +60,7 @@ if TYPE_CHECKING:
 
     from discord.message import Message
 
-    from ._types import BotT, Check, ContextT, Coro, CoroFunc, Error, Hook
+    from ._types import BotT, Check, ContextT, Coro, CoroFunc, Error, Hook, UserCheck
 
 
 __all__ = (
@@ -378,7 +378,7 @@ class Command(_BaseCommand, Generic[CogT, P, T]):
         except AttributeError:
             checks = kwargs.get('checks', [])
 
-        self.checks: List[Check[ContextT]] = checks
+        self.checks: List[UserCheck[ContextT]] = checks
 
         try:
             cooldown = func.__commands_cooldown__
@@ -458,7 +458,7 @@ class Command(_BaseCommand, Generic[CogT, P, T]):
 
         self.params: Dict[str, Parameter] = get_signature_parameters(function, globalns)
 
-    def add_check(self, func: Check[ContextT], /) -> None:
+    def add_check(self, func: UserCheck[ContextT], /) -> None:
         """Adds a check to the command.
 
         This is the non-decorator interface to :func:`.check`.
@@ -477,7 +477,7 @@ class Command(_BaseCommand, Generic[CogT, P, T]):
 
         self.checks.append(func)
 
-    def remove_check(self, func: Check[ContextT], /) -> None:
+    def remove_check(self, func: UserCheck[ContextT], /) -> None:
         """Removes a check from the command.
 
         This function is idempotent and will not raise an exception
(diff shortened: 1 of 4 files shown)


mikeshardmind:
So, other than there being types not in ._types (locality while working on it, will clean that up before marking this ready) fixing this decorator seems to allow other issues to be seen

The below is a pre-existing failure case that was obscured by other type issues fixed in this so far:

```py
from typing_extensions import reveal_type


@commands.check_any(commands.is_owner(), commands.is_nsfw())
@commands.command()
async def example(ctx: commands.Context):
    pass


reveal_type(example)  # Type of "example" is "Command[Unknown, Unknown, Unknown]"


@commands.check_any(commands.is_owner(), commands.is_nsfw())
@commands.group()
async def example_group(ctx: commands.Context):
    pass


reveal_type(example_group)  # Type of "example_group" is "Command[Unknown, Unknown, Unknown]"


@example_group.command()  # Cannot access member "command" for type "Command[Unknown, Unknown, Unknown]" Member "command" is unknown
async def new_failure(ctx: commands.Context):
    ...

**Comment by mikeshardmind** [comment]
I went with the simpler approach that was brought up in Discord of just typing it with a protocol rather than a concrete type.

This required distinguishing between user-provided types and what the decorator returns, as well as some minor lies to the type-checker since dynamic attribute access, even in a highly predictable fashion isn't ever going to be supported. I'm not necessarily a fan of lying to the type checker, but as far as making it work for users, pyright happily allows the examples of chaining predicates via .predicate now, as well as not having the decorator eat the type information of what is under it due to the decorator type being partially unknown.

File: discord/ext/commands/core.py [hunk 1]
```diff
@@ -1911,10 +1911,10 @@ async def predicate(ctx: Context[BotT]) -> bool:
         # if we're here, all checks failed
         raise CheckAnyFailure(unwrapped, errors)
 
-    return check(predicate)
+    return check(predicate)  # type: ignore
 
 
-def has_role(item: Union[int, str], /) -> Callable[[T], T]:
+def has_role(item: Union[int, str], /) -> Check:
```
Rapptz, line 1917: For `--verifytypes` these need to have their generics passed, something like `Check[Any]` works fine to silence it.

Rapptz (changes requested):
This looks fine to me, just a minor nit and a question. With your changes what would `reveal_type` be for this?

```py
@commands.check(lambda ctx: True)
async def foo(ctx: commands.Context, x: int, y: str):
    ...

reveal_type(foo)
```

discord/ext/commands/core.py
mikeshardmind: I've gone ahead and added the generic Any where applicable.

```py
from typing_extensions import reveal_type


@commands.check(lambda ctx: True)
async def foo(ctx: commands.Context, x: int, y: str):
    ...

reveal_type(foo)  # Type of "foo" is "(ctx: Context[Unknown], x: int, y: str) -> Coroutine[Any, Any, None]"


@commands.check(lambda ctx: True)
@commands.command()
async def foo_com(ctx: commands.Context, x: int, y: str):
    ...


reveal_type(foo_com)  # Type of "foo_com" is "Command[None, (x: int, y: str), None]"
```

This is the typing of what you were interested in there, as well as the corresponding command (in case you wanted more information or inadvertently left a line out of what you wanted to see)

I also did a quick re-check after updating pyright, and noticed that

1. One of my local fixes (Specifically, an issue with the return type of `__call__` in the protocol) was not in the commit pushed. (what I tested had not reflected the commit 1:1)
   
3. The Generic type of a protocol needs to be covariant. (There can be arguments over that, but that's what the spec currently is, and this still works for subclassed context here)

The requested change as well as those two fixes have been rebased in, along with updating the base. I've made the commit message contain more details about the specific changes as well.

Rapptz (approved):

PR #7954, opened 2022-04-29
Add basic slash command example

### LINKED ISSUE #7941: Should Add application_commands usage in examples
### Summary

will you add appcommands usage in examples 

The core library

### The Problem

it's difficult to make a slash commands in discord.py for beginners, because docs are not well documented, may be discord.py guide will be release soon but for now please add some basic or advance example of slash commands in examples folder, for getting an idea of how to make it....

### The Ideal Solution

By adding that, it will be easy for those who are reverting back to discord.py from other 3rd party libraries, or else make a guide for all application commands, discord.ui Components, like other 3rd party libraries

it is available by another developers as a gist, but it may be not accessible to all

### The Current Solution

_No response_

### Additional Context

_No response_

NarmyOnDiscord:
## Summary

Adding app_command examples. Preferred structure:
```
examples/
  app_commands/
    cog.py
    groups.py
    basic.py
    context_menu.py
    checks.py
    hybrid.py
```

## Checklist

- [x] This PR fixes #7941


examples/app_commands/basic.py
```diff
@@ -0,0 +1,58 @@
+import discord
+from discord import app_commands
+from discord.ext import commands
+
+class MyBot(commands.Bot):
+
+    # In this basic example, we just synchronize the app commands to one guild,
+    # without requiring to set the guild to every command individually by just coping them over.
```
RGBCube, line 8:
```suggestion
    # without requiring to set the guild to every command individually by just coping them over.
```
typo

RGBCube (changes requested):

examples/app_commands/basic.py
NarmyOnDiscord: Thanks

examples/app_commands/basic.py
```diff
@@ -0,0 +1,58 @@
+import discord
+from discord import app_commands
+from discord.ext import commands
+
+class MyBot(commands.Bot):
```
Rapptz, line 5: For this example, the bare bones basic, I think it makes sense to just use a regular `discord.Client` instance instead of `commands.Bot` since we don't want to complicate it. Other examples can use `commands.Bot` such as `cog.py`.

examples/app_commands/basic.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # In this basic example, we just synchronize the app commands to one guild,
+    # without requiring to set the guild to every command individually by just copying them over.
+    # By doing so we don't have to wait up to 1hr until they are shown for the end-user.
+
+    # If you want to synchronize your app commands to every guild your bot is in, 
+    # remove the following setup_hook and uncomment the one below.
```
Rapptz, line 12: I think this is a bit strange to ask. It should probably just be a comment on top of the `tree.sync` and `copy_global_to` calls.

examples/app_commands/basic.py
```diff
@@ -0,0 +1,58 @@
...
+    async def setup_hook(self):
+        # Replace ... with the guild id you want your commands to be synced with.
+        guild = discord.Object(...)
```
Rapptz, line 15: It might be better to make this a global constant, e.g. `MY_GUILD = discord.Object(id=0)  # replace with your ID here` to steer people into the right direction.

examples/app_commands/basic.py
```diff
@@ -0,0 +1,58 @@
...
+        self.tree.copy_global_to(guild=guild)
+        synced_app_commands = await self.tree.sync(guild=guild)
+        print(f'Synced the following commands to {guild.id}: {[c.name for c in synced_app_commands]}')
```
Rapptz, line 18: I don't think there's much of a point to make use of the output here. Might as well leave it as a bare sync.

examples/app_commands/basic.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # async def setup_hook(self):
+    #     synced_app_commands = await self.tree.sync()
+    #     print(f'Synced the following commands to every guild: {[c.name for c in synced_app_commands]}')
```
Rapptz, line 22:
```suggestion
```

examples/app_commands/basic.py
```diff
@@ -0,0 +1,58 @@
...
+@app_commands.describe(member='The member you want to get the joined date from, defaults to the user who uses the command')
+async def joined(interaction: discord.Interaction, member: discord.Member = None):
+    """Says when a member joined."""
+    # By doing this we use the member that was provided, 
+    # tho if this is None we use the user that has invoked the command
+    member = member or interaction.user
+
+    await interaction.response.send_message(f'{member} joined in {member.joined_at}')
+
+bot.run('token')
```
Rapptz, line 58: Needs a newline in the end.

examples/app_commands/basic.py
```diff
@@ -0,0 +1,58 @@
...
+    """Adds two numbers together."""
+    await interaction.response.send_message(f'{first_value} + {second_value} = {first_value + second_value}')
+
+# To make an argument optional, you can make it a default argument
+@bot.tree.command()
+@app_commands.describe(member='The member you want to get the joined date from, defaults to the user who uses the command')
+async def joined(interaction: discord.Interaction, member: discord.Member = None):
+    """Says when a member joined."""
+    # By doing this we use the member that was provided, 
+    # tho if this is None we use the user that has invoked the command
```
Rapptz, line 53:
```suggestion
    # If no member is explicitly provided then we use the command user here
```

examples/app_commands/basic.py
```diff
@@ -0,0 +1,58 @@
...
+    """Says hello!"""
+    await interaction.response.send_message(f'Hi, {interaction.user.mention}')
+
+@bot.tree.command()
+@app_commands.describe(first_value='The first value you want to add something to', second_value='The value you want to add to the first value')
+async def add(interaction: discord.Interaction, first_value: int, second_value: int):
+    """Adds two numbers together."""
+    await interaction.response.send_message(f'{first_value} + {second_value} = {first_value + second_value}')
+
+# To make an argument optional, you can make it a default argument
```
Rapptz, line 47:
```suggestion
# To make an argument optional, you can either give it a supported default argument
# or you can mark it as Optional from the typing library. This example does both.
```

examples/app_commands/basic.py
```diff
@@ -0,0 +1,58 @@
...
+@bot.tree.command()
+@app_commands.describe(member='The member you want to get the joined date from, defaults to the user who uses the command')
+async def joined(interaction: discord.Interaction, member: discord.Member = None):
```
Rapptz, line 50:
```suggestion
async def joined(interaction: discord.Interaction, member: Optional[discord.Member] = None):
```

Might as well play nice with type checkers.

examples/app_commands/basic.py
```diff
@@ -0,0 +1,58 @@
...
+    print(f'Logged in as {bot.user} (ID: {bot.user.id})')
+    print('------')
+
+@bot.tree.command()
+async def hello(interaction: discord.Interaction):
+    """Says hello!"""
+    await interaction.response.send_message(f'Hi, {interaction.user.mention}')
+
+@bot.tree.command()
+@app_commands.describe(first_value='The first value you want to add something to', second_value='The value you want to add to the first value')
```
Rapptz, line 42: This line's pretty long so you need to run `black` on it.

Rapptz (changes requested):
I think it's better to stick with a single example so things are easier to digest and review for me.

Also, you leaked a test file with a token.

examples/app_commands/basic.py
```diff
@@ -0,0 +1,58 @@
+import discord
+from discord import app_commands
+from discord.ext import commands
+
+class MyBot(commands.Bot):
+
+    # In this basic example, we just synchronize the app commands to one guild,
+    # without requiring to set the guild to every command individually by just copying them over.
+    # By doing so we don't have to wait up to 1hr until they are shown for the end-user.
```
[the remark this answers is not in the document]
NarmyOnDiscord: oops, resetted the token

examples/app_commands/basic.py
NarmyOnDiscord: Done already. :)

examples/app_commands/basic.py
```diff
@@ -0,0 +1,64 @@
...
+from discord import app_commands
+
+MY_GUILD = discord.Object(id=0)  # replace with your guild id
+
+
+class MyClient(discord.Client):
+    def __init__(self, *, intents: discord.Intents, application_id: int):
+        super().__init__(intents=intents, application_id=application_id)
+
+        self.tree = app_commands.CommandTree(self)
```
Rapptz, line 13:
```suggestion
        # A CommandTree is a special type that holds all the application command
        # state required to make it work. This is a separate class because it
        # allows all the extra state to be opt-in.
        # Whenever you want to work with application commands, your tree is used
        # to store it and work with it.
        # Note: When using commands.Bot instead of discord.Client, the bot will
        # maintain its own tree instead.
        self.tree = app_commands.CommandTree(self)
```

examples/app_commands/basic.py
```diff
@@ -0,0 +1,64 @@
...
+
+@client.tree.command()
+async def hello(interaction: discord.Interaction):
+    """Says hello!"""
+    await interaction.response.send_message(f'Hi, {interaction.user.mention}')
+
+
+@client.tree.command()
+@app_commands.describe(
+    first_value='The first value you want to add something to', second_value='The value you want to add to the first value'
```
Rapptz, line 45:
```suggestion
    first_value='The first value you want to add something to', 
    second_value='The value you want to add to the first value',
```

Magic commas are better for readability.

Rapptz (changes requested):
Looks good, just a minor nit.

examples/app_commands/basic.py
```diff
@@ -0,0 +1,71 @@
...
+        # allows all the extra state to be opt-in.
+        # Whenever you want to work with application commands, your tree is used
+        # to store it and work with it.
+        # Note: When using commands.Bot instead of discord.Client, the bot will
+        # maintain its own tree instead.
+        self.tree = app_commands.CommandTree(self)
+
+    # In this basic example, we just synchronize the app commands to one guild,
+    # without requiring to set the guild to every command individually by just copying them over.
+    # By doing so we don't have to wait up to an hour until they are shown to the end-user.
```
Rapptz, line 23:
```suggestion
    # In this basic example, we just synchronize the app commands to one guild.
    # Instead of specifying a guild to every command, we copy over our global commands instead.
    # By doing so we don't have to wait up to an hour until they are shown to the end-user.
```

examples/app_commands/basic.py
```diff
@@ -0,0 +1,71 @@
...
+    async def setup_hook(self):
+        # This copies the global commands over to your guild.
+        self.tree.copy_global_to(guild=MY_GUILD)
+        await self.tree.sync(guild=MY_GUILD)
+
+
+intents = discord.Intents.default()
+
+# In order to use a basic synchronization of the app commands in the setup_hook,
+# you have replace ... with your bots application_id you find in the developer portal.
```
Rapptz, line 33:
```suggestion
# you have replace the 0 with your bots application_id you find in the developer portal.
```

Rapptz (approved):
Thanks!

PR #7955, opened 2022-04-30
Add error decorator to app_commands.Group

HeroicosHM:
## Summary

This pull request fixes an inconsistency in the `app_commands.Group` class, compared to the `app_commands.Command`, `app_commands.ContextMenu`, and `app_commands.CommandTree` classes by adding a `@error` decorator for registering error handlers at the group level.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -1282,8 +1282,7 @@ class Group:
     def error(self, coro: ErrorFunc) -> ErrorFunc:
         """A decorator that registers a coroutine as a local error handler.
 
-        The local error handler is called whenever an exception is raised in the body
-        of the command or during handling of the command. The error handler must take
+        The local error handler is called whenever an exception is raised in a child command. The error handler must take
         2 parameters, the interaction and the error.
 
         The error passed will be derived from :exc:`AppCommandError`.
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -1282,8 +1282,8 @@ class Group:
     def error(self, coro: ErrorFunc) -> ErrorFunc:
         """A decorator that registers a coroutine as a local error handler.
 
-        The local error handler is called whenever an exception is raised in a child command. The error handler must take
-        2 parameters, the interaction and the error.
+        The local error handler is called whenever an exception is raised in a child command.
+        The error handler must take 2 parameters, the interaction and the error.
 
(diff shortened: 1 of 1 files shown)


Rapptz (changes requested):
I don't think this implementation is sound. I would suggest keeping the code as it used to be, with a base class and an overridden method but then adding a decorator that just overrides that method. See CommandTree.on_error and CommandTree.error for how this should be done.

discord/app_commands/commands.py
```diff
@@ -70,6 +70,11 @@
     # reference the other to prevent type checking errors in callbacks
     from discord.ext.commands import Cog
 
+    ErrorFunc = Callable[
+        [Interaction, AppCommandError],
+        Coroutine[Any, Any, Any],
+    ]
```
Rapptz, line 76:
```suggestion
    ErrorFunc = Callable[[Interaction, AppCommandError], Coroutine[Any, Any, None]]
```

I prefer to keep this in one line. This also satisfies the type checker

Rapptz (changes requested):

discord/app_commands/commands.py
```diff
@@ -1274,6 +1276,35 @@ async def on_error(self, interaction: Interaction, error: AppCommandError) -> No
...
+        """
+
+        if not inspect.iscoroutinefunction(coro):
+            raise TypeError('The error handler must be a coroutine.')
+
+        params = inspect.signature(coro).parameters
+        if len(params) != 2:
+            raise TypeError('The error handler must have 2 parameters.')
+
+        self.on_error = coro
```
Rapptz, line 1305:
```suggestion
        self.on_error = coro  # type: ignore
```

HeroicosHM: Oh that's what I was missing. Thank you. Is there some way that I could run the linter for myself before asking for review on a PR? I feel bad asking for review 20 times for tiny updates.

Rapptz: CI is mainly `black` (`pip install -U black` and then `black discord examples` in the root of this project) and `pyright`. You can install `pyright` using `npm`. `npm install -g pyright@1.1.242` for the one used for CI and then use `pyright` in the root of the project.

Rapptz (approved):
Thanks.

PR #7965, opened 2022-04-30
Add an example for app_commands.rename()

NarmyOnDiscord:
## Summary

Add a simple command with a renamed argument to showcase this feature.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/app_commands/basic.py
+++ b/examples/app_commands/basic.py
@@ -56,6 +56,14 @@ async def add(interaction: discord.Interaction, first_value: int, second_value: 
     await interaction.response.send_message(f'{first_value} + {second_value} = {first_value + second_value}')
 
 
+@client.tree.command()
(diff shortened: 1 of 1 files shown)


examples/app_commands/basic.py
```diff
@@ -56,6 +56,14 @@ async def add(interaction: discord.Interaction, first_value: int, second_value:
     await interaction.response.send_message(f'{first_value} + {second_value} = {first_value + second_value}')
 
 
+@client.tree.command()
```
Rapptz, line 59:
```suggestion
# The rename decorator allows us to change the display of the parameter on Discord
# In this example, even though we use `text_to_send` in the code, the client will use `text` instead.
# Note that other decorators will still refer to it as `text_to_send` in the code.
@client.tree.command()
```

Rapptz (approved):
LGTM.

PR #7966, opened 2022-04-30
Fix minor convention mistakes in basic examples

GoogleGenius:
## Summary
This PR just fixes some minor convention errors in the [app_command basic example](https://github.com/Rapptz/discord.py/blob/master/examples/app_commands/basic.py) as well as the [basic bot example](https://github.com/Rapptz/discord.py/blob/master/examples/basic_bot.py).

## Checklist
- [ ] This PR fixes an issue.
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed).
- [x] This PR is **not** a code change (e.g. documentation, README, etc.).

### DIFF
--- a/examples/app_commands/basic.py
+++ b/examples/app_commands/basic.py
@@ -67,7 +67,7 @@ async def joined(interaction: discord.Interaction, member: Optional[discord.Memb
     member = member or interaction.user
 
     # The format_dt function formats the date time into a human readable representation in the official client
-    await interaction.response.send_message(f'{member} joined at {discord.utils.format_dt(member.joined_at)}')
+    await interaction.response.send_message(f'{member} joined {discord.utils.format_dt(member.joined_at)}')
 
 
 # A Context Menu command is an app command that can be run on a member or on a message by
--- a/examples/app_commands/basic.py
+++ b/examples/app_commands/basic.py
@@ -67,7 +67,7 @@ async def joined(interaction: discord.Interaction, member: Optional[discord.Memb
     member = member or interaction.user
 
     # The format_dt function formats the date time into a human readable representation in the official client
-    await interaction.response.send_message(f'{member} joined on {discord.utils.format_dt(member.joined_at)}')
+    await interaction.response.send_message(f'{member} joined at {discord.utils.format_dt(member.joined_at)}')
 
 
 # A Context Menu command is an app command that can be run on a member or on a message by
@@ -78,7 +78,7 @@ async def joined(interaction: discord.Interaction, member: Optional[discord.Memb
 @client.tree.context_menu(name='Show Join Date')
 async def show_join_date(interaction: discord.Interaction, member: discord.Member):
     # The format_dt function formats the date time into a human readable representation in the official client
-    await interaction.response.send_message(f'{member} joined on {discord.utils.format_dt(member.joined_at)}')
+    await interaction.response.send_message(f'{member} joined at {discord.utils.format_dt(member.joined_at)}')
 
 
 # This context menu command only works on messages
--- a/examples/app_commands/basic.py
+++ b/examples/app_commands/basic.py
@@ -3,6 +3,7 @@ from typing import Optional
 import discord
 from discord import app_commands
 
+
 MY_GUILD = discord.Object(id=0)  # replace with your guild id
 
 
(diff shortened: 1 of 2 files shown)


NarmyOnDiscord (commented):
Well done. But if we change L81 here, we also have to change it [here](https://github.com/Rapptz/discord.py/blob/master/examples/basic_bot.py#L60)

GoogleGenius:
> Well done, but if we change L81 here, we also have to change it [here](https://github.com/Rapptz/discord.py/blob/master/examples/basic_bot.py#L60).

I'll update the PR and change that to improve consistency.

examples/app_commands/basic.py
```diff
@@ -57,16 +58,16 @@ async def add(interaction: discord.Interaction, first_value: int, second_value:
...
-@app_commands.describe(member='The member you want to get the joined date from, defaults to the user who uses the command')
+@app_commands.describe(member='The member you want to get the joined date from; defaults to the user who uses the command')
 async def joined(interaction: discord.Interaction, member: Optional[discord.Member] = None):
     """Says when a member joined."""
     # If no member is explicitly provided then we use the command user here
     member = member or interaction.user
 
     # The format_dt function formats the date time into a human readable representation in the official client
-    await interaction.response.send_message(f'{member} joined in {discord.utils.format_dt(member.joined_at)}')
+    await interaction.response.send_message(f'{member} joined on {discord.utils.format_dt(member.joined_at)}')
```
GoogleGenius, line 70: Actually, now that I think about it, should it be joined "at" or joined "on?" "At" is used for precise time, for example, "at 5:32PM." "On" is used for general days, for example, "on April 30." However this is a combination of datetime, for example, "April 30, 2022 8:22AM." Either it would be best to omit the "at" or "on" keyword completely, or perhaps follow the library's choice of "joined_at."

examples/app_commands/basic.py
```diff
@@ -77,7 +78,7 @@ async def joined(interaction: discord.Interaction, member: Optional[discord.Memb
 @client.tree.context_menu(name='Show Join Date')
 async def show_join_date(interaction: discord.Interaction, member: discord.Member):
     # The format_dt function formats the date time into a human readable representation in the official client
-    await interaction.response.send_message(f'{member} joined in {discord.utils.format_dt(member.joined_at)}')
+    await interaction.response.send_message(f'{member} joined on {discord.utils.format_dt(member.joined_at)}')
```
GoogleGenius, line 81: Same question here.

examples/app_commands/basic.py
NarmyOnDiscord: Ealier this did not use format_dt (as seen in the basic_bot.py example) and used `in`, I just copied it over from the basic bot example. But since english is not my native language, I cannot decide here.

GoogleGenius: I think I will use "at" just to be consistent with the library's `joined_at` function. Honestly, it's an example so the significance isn't the highest.

NarmyOnDiscord: Seems fine.

examples/app_commands/basic.py
```diff
@@ -77,7 +78,7 @@ async def joined(interaction: discord.Interaction, member: Optional[discord.Memb
 @client.tree.context_menu(name='Show Join Date')
 async def show_join_date(interaction: discord.Interaction, member: discord.Member):
     # The format_dt function formats the date time into a human readable representation in the official client
-    await interaction.response.send_message(f'{member} joined in {discord.utils.format_dt(member.joined_at)}')
+    await interaction.response.send_message(f'{member} joined at {discord.utils.format_dt(member.joined_at)}')
```
Rapptz, line 81: I think this can go either way.

The resulting string is e.g. "Danny#0007 joined in|at May 16, 2016 13:40"

Gobot1234: I personally think on goes better than either of these

GoogleGenius: The nuances of the English language: 
"Joined at 13:40" would make sense.
"Joined in 2016" would make sense.
"Joined on May 16" would make sense.

But since this is combined datetime, including only 1 of the `at | in | on` would not as it is only correct for part of the datetime string. 

So I'm thinking thinking to omit the keyword completely, for example, "Danny#007 joined `May 16, 2016 1:40PM`" or like it is now, `at` to match the lib's `joined_at`.

Please let me know what you think.

Rapptz: Just omit it since this is reaching excessive bikeshedding territory.

examples/basic_bot.py
```diff
@@ -57,7 +57,7 @@ async def repeat(ctx, times: int, content='repeating...'):
 @bot.command()
 async def joined(ctx, member: discord.Member):
     """Says when a member joined."""
-    await ctx.send(f'{member.name} joined in {member.joined_at}')
+    await ctx.send(f'{member.name} joined {member.joined_at}')
```
Rapptz, line 60:
```suggestion
    await ctx.send(f'{member.name} joined {discord.utils.format_dt(member.joined_at)}')
```

PR #7968, opened 2022-04-30
Try Again: Add initial OAuth2 extension implementation.

FrostiiWeeb:
## Checklist

- [ ] This PR fixes an issue.


discord/ext/oauth/token.py
```diff
@@ -0,0 +1,20 @@
+from dataclasses import dataclass
+from datetime import datetime
+
+from ...utils import utcnow
+
+
+@dataclass
```
AbstractUmbra, line 7: A note here is that the library doesn't use dataclasses to represent Discord models generally. Whilst this may not be an issue, it may be wiser to follow our current standard.

We create payload types with a `typing.TypedDict` and store their definition in `discord.types.<item>` and we create a regular class instead.

AbstractUmbra (changes requested):
This is also missing documentation entries. Perhaps you could emulate what was needed to document ext.tasks?

PR #7971, opened 2022-05-01
[commands] fix signature when displayed_default is missing

darthshittious:
## Summary
This makes it so that passing an empty string (or MISSING) to the displayed_default of commands.Parameter not show up as [name=] or [name=...]

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
I don't think this works for the case where you just do `x: int = 10`?

darthshittious:
I don't think it is? right now it looks like this: 

This fix is for cases where you assign, say, a callable to default, and don't want to display anything.

PR #7972, opened 2022-05-01
Add application command permissions to audit log

z03h:
## Summary

I am probably missing something.

- Add `PartialIntegration`
  - `entry.target` will be `PartialIntegration` when updating general app command permissions or `AppCommand` when updating a specific app commands permissions or Object with matching ID if neither could be matched. A little more rich than `Object` vs `AppCommand`.
  - Can also be used for integration actions. Untested with non discord integrations and don't have any to test.
- Add `app_commands.AllChannels` as special value for all channels. 
- Add `AuditLogDiff.app_command_permissions` that is a `List[Tuple[target, bool]]`
  - target = `Union[Member, abc.GuildChannel, Role, AllChannels, Object]`
- Add `AuditLogAction.app_command_permission_update`
- Set `entry.extra` to a`PartialIntegration` or Object with id=application id for `app_command_permission_update`
- Fixed bug where `guild.audit_logs` with a limit>100 fetched too many entries
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/models.py
+++ b/discord/app_commands/models.py
@@ -78,6 +78,8 @@ class AllChannels:
 
     Attributes
     -----------
+    id: :class:`int`
+        The guilds id - 1.
     guild: :class:`~discord.Guild`
         The guild the application command permission is for.
     """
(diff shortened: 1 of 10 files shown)


Rapptz (approved):
Thanks for all the work on this. A shame that Discord messed up the representation of the `changes` dictionary while adding this.

PR #7990, opened 2022-05-04
Allow enabling and disabling all items in a `View`

Pysics:
## Checklist

- [ ] This PR fixes an issue.


Rapptz:
This feature doesn't make sense since not all items are guaranteed to be disabled. For example, Modal is a subclass of View and TextInput is a subclass of Item yet TextInput cannot be disabled making this method not type safe. Not to mention future components might not be possible to disable. Therefore this method is unsound.

PR #7994, opened 2022-05-04
feat: return embed from `clear_fields` and `remove_field`

DenverCoder1:
## Summary

For consistency with all other Embed methods (eg. `add_field`, `remove_author`, `set_footer`, etc.), this makes `clear_fields` and `remove_field` return `self`.

### Example usage:

```py
await ctx.send(embed=embed.add_field(name="Test", value=f"field").remove_field(0))

await ctx.send(embed=embed.clear_fields().add_field(name="Test", value=f"field"))
```

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -602,18 +602,27 @@ class Embed:
 
         return self
 
-    def clear_fields(self) -> None:
-        """Removes all fields from this embed."""
+    def clear_fields(self) -> Self:
+        """Removes all fields from this embed.
+        
+        This function returns the class instance to allow for fluent-style
+        chaining.
+        """
         try:
             self._fields.clear()
         except AttributeError:
             self._fields = []
+            
+        return self
 
-    def remove_field(self, index: int) -> None:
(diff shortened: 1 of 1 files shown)


Rapptz:
Hm. In other programming languages that use this pattern only additive functions are fluent while destructive ones tend to not be. I never really gave it much thought as to why, but this made sense to me in some way.

DenverCoder1:
`remove_author` and `remove_footer` already return `self`.

I think it makes more sense to add it to the remaining methods for consistency, rather than removing functionality from others.

Rapptz:
Yeah I guess the ship there has sailed. Your methods are missing versionchanged however.

DenverCoder1:
Is `versionchanged` necessary when there are no changed side effects or new parameters? I see it as more of a notice of things that one might need to potentially update in their code, whereas this shouldn't break any existing code.

Rapptz:
versionchanged is not used for solely breaking changes but to notify of any type of behavioural changes between releases.

PR #8003, opened 2022-05-07
Add more detailed description to accent_colour

rperson1508:
## Summary

Added a short description to explain what an accent colour is

## Checklist

- [ ] This PR fixes an issue.


Rapptz (commented):
I guess while we're here you should probably mention that it's only given if the user explicitly sets a colour.

rperson1508:
> I guess while we're here you should probably mention that it's only given if the user explicitly sets a colour.

I may be wrong but doesn't Discord automatically take the colour from their avatar?

Rapptz:
Yes but that's done client side, the API does not return that colour. It does this when the `accent_color` is `null`.

rperson1508:
Alright am sorry for replying this late, adding that in now.

PR #8018, opened 2022-05-09
add the color white

galaxine-senpai:
## Summary

This PR adds the color white to the colour.py file that for some reason did not exist before and had frustrated me as I looked through the docs so I added it myself
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/colour.py
+++ b/discord/colour.py
@@ -396,6 +396,14 @@ class Colour:
         .. versionadded:: 2.0
         """
         return cls(0xFEE75C)
+    
+       @classmethod
+    def white(cls) -> Self:
+        """A factory method that returns a :class:`Colour` with a value of ``0xFFFFFF``.
+
+        .. versionadded:: 2.0
+        """
+        return cls(0xFFFFFF)
 
 
 Color = Colour
--- a/discord/colour.py
+++ b/discord/colour.py
@@ -397,7 +397,7 @@ class Colour:
         """
         return cls(0xFEE75C)
     
-       @classmethod
+    @classmethod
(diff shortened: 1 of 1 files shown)


GoogleGenius:
That's surprising that it was missing lol

discord/colour.py
```diff
@@ -396,6 +396,14 @@ def yellow(cls) -> Self:
         .. versionadded:: 2.0
         """
         return cls(0xFEE75C)
+    
+       @classmethod
```
GoogleGenius, line 400:
```suggestion
   @classmethod
```

GoogleGenius (changes requested):

discord/colour.py
GoogleGenius: I think that was wrongly indented or something

galaxine-senpai: Ah, I didn't catch that thanks

wasi-master:
There is probably a bug or something like that with the pure white color if I recall correctly. People used to suggest using a color close to pure white to mitigate the bug, I forgot the details though.

Rapptz:
White used to be special cased by the client to be a theme specific colour that is either black or white depending on the theme the user did. On top of that, the colours used have always been historically colours that Discord uses within the client and their branding, that way the classmethods don't spiral out of control with a bunch of different colours. On top of that, black (0x000000) and white (0xFFFFFF) are pretty commonly known so they're not there.

GoogleGenius:
Makes sense 👍

galaxine-senpai:
Alright, thanks!

PR #8022, opened 2022-05-10
Fix aiohttp dependency vulnerability

Faholan:
## Summary

This pulls request fixes a vulnerability in aiohttp versions < 3.7.4.
https://github.com/aio-libs/aiohttp/security/advisories/GHSA-v6wp-4m6f-gcjg
Even tho, as far as I know, discord.py makes no use of middlewares, and is thus not impacted by this vulnerability, I feel that leaving a vulnerability of any kind in any dependency, while it can be easily fixed, isn't a good idea.

## Checklist

- [x] This PR fixes an issue.


Rapptz (approved):
Thanks.

PR #8034, opened 2022-05-12
[commands]Add **attrs parameter to load_extension

aindrigo0:
## Summary
This adds support for custom parameters when calling load_extension. For example:

Extension file:
```py
async def setup(bot: commands.Bot, text: str):
    print(text)
```
Bot file:
```py
await bot.load_extension("asdf", text="Hello")
```

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -908,7 +908,7 @@ class BotBase(GroupMixin[None]):
                 if _is_submodule(name, module):
                     del sys.modules[module]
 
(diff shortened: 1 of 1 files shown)


RGBCube:
pretty sure this was rejectee before

Rapptz:
Hello,

Thanks for the PR. Unfortunately this has been rejected before because I don't consider this the proper way of doing it. It's easy to make it fail when in a loop since some extensions might not expect extraneous keyword arguments. A proper solution would need some form of namespaced keys  in order to prevent other extensions from hijacking the configuration of another extension.

PR #8035, opened 2022-05-13
[commands] Fix FlagConverters not raising BadFlagArgument

z03h:
## Summary

`BadFlagArgument` would almost never be raised since `run_converters` raised some subclass of `CommandError`.

Add attributes to `BadFlagArgument` for the original error and current flag argument for better error handling.
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/errors.py
+++ b/discord/ext/commands/errors.py
@@ -1138,14 +1138,22 @@ class BadFlagArgument(FlagError):
     -----------
     flag: :class:`~discord.ext.commands.Flag`
         The flag that failed to convert.
+    argument: :class:`str`
+        The argument supplied by the caller that was not able to be converted.
+    original: :class:`Exception`
+        The original exception that was raised. You can also get this via
+        the ``__cause__`` attribute.
     """
 
-    def __init__(self, flag: Flag) -> None:
+    def __init__(self, flag: Flag, argument: str, original: Exception) -> None:
         self.flag: Flag = flag
         try:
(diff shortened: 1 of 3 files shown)


discord/ext/commands/errors.py
```diff
@@ -1138,15 +1138,23 @@ class BadFlagArgument(FlagError):
...
-    def __init__(self, flag: Flag) -> None:
+    def __init__(self, flag: Flag, argument: str, original: Exception) -> None:
         self.flag: Flag = flag
         try:
             name = flag.annotation.__name__
         except AttributeError:
             name = flag.annotation.__class__.__name__
 
+        self.argument = argument
+        self.original = original
```
Rapptz, line 1156:
```suggestion
        self.argument: str = argument
        self.original: Exception = original
```

Rapptz (commented):
Unsure how I feel about this one. CommandError are usually never wrapped. If someone makes a converter and uses a custom exception and then they use it generally speaking just handling the custom exception is enough to deal with it. With this, they have to actually unwrap it to handle it just because it's in a flag.

discord/ext/commands/errors.py
z03h: Wanted to have a way to get the flag and input that failed since they're not accessible from ctx

Rapptz (commented):
I think since this behaviour deviates from the usual it needs to be documented in commands.rst

PR #8037, opened 2022-05-15
Resolved #8002, added new constant.

### LINKED ISSUE #8002: Provide Constants for Max Files and Max Embeds When Sending a Message
### Summary

Provide constants for the max number of files and of embeds a message can be sent with.

The core library

### The Problem

When sending a message, the number of embeds and files are limited to 10, and an exception is thrown if you exceed that. So we have to make sure we don't go over the limit.

### The Ideal Solution

I would like there to be a constant I can refer to that has the limit so I don't have to hardcode it myself and remember to update it if it ever changes.

### The Current Solution

The current solution is to hardcode in the limits and hope they don't change (which, to be fair, I don't think they ever have).

### Additional Context

_No response_

farkon00:
## Summary

Resolves issue #8002, asking for new constant.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -57,7 +57,6 @@ from .file import File
 from .mentions import AllowedMentions
 from . import __version__, utils
 from .utils import MISSING
-print(__name__)
 
 _log = logging.getLogger(__name__)
 
--- a/discord/http.py
+++ b/discord/http.py
@@ -50,13 +50,14 @@ import weakref
 
 import aiohttp
 
-from .client import Client
 from .errors import HTTPException, Forbidden, NotFound, LoginFailure, DiscordServerError, GatewayNotFound
+from .embeds import MAX_MESSAGE_EMBEDS
 from .gateway import DiscordClientWebSocketResponse
 from .file import File
 from .mentions import AllowedMentions
 from . import __version__, utils
 from .utils import MISSING
+print(__name__)
 
 _log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


discord/http.py
```diff
@@ -51,11 +51,13 @@
 import aiohttp
 
 from .errors import HTTPException, Forbidden, NotFound, LoginFailure, DiscordServerError, GatewayNotFound
+from .embeds import MAX_MESSAGE_EMBEDS
 from .gateway import DiscordClientWebSocketResponse
 from .file import File
 from .mentions import AllowedMentions
 from . import __version__, utils
 from .utils import MISSING
+print(__name__)
```
DJJ05, line 60: Did you forget to remove this?

farkon00: @DJJ05 Sorry, yes, I forgot, I will fix it, one moment.

farkon00: Fixed

DJJ05 (approved):

discord/http.py
DJJ05: 👍

GoogleGenius:
Might want to fix the title by the way :+1:

farkon00:
@GoogleGenius Its just the same thing.

> the number of embeds and files are limited to 10

GoogleGenius:
> @GoogleGenius Its just the same thing.
> 
> > the number of embeds and files are limited to 10

Ah alright thanks for letting me knw!

Rapptz:
Heya,

Thanks for the PR. Unfortunately I'm not really sure if I want to expose a constant for this. I'm also not really sure I want to expose it in this manner too. I figure hypothetical constants would belong in their own separate submodule with proper documentation. However I'm still not sure on how useful such as hypothetical submodule would be.

Issue #8046, opened 2022-05-17
OpusError on macOS arm64 platform (only)


m13253:
### Summary

When creating an opus.Encoder() instance, an exception of discord.opus.OpusError: invalid argument is thrown

### Reproduction Steps

First, install either Python 3.9 or Python 3.10.
```bash
$ brew install python
$ python3 -m pip install -U "discord.py[voice]"
```
Then, execute the code in the section below.

### Minimal Reproducible Code

```python
import discord.opus
opus_encoder = discord.opus.Encoder()
```

### Expected Results

Should load `libopus.dylib` and create an Opus encoder instance.

### Actual Results

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/homebrew/lib/python3.10/site-packages/discord/opus.py", line 290, in __init__
    self.set_bitrate(128)
  File "/opt/homebrew/lib/python3.10/site-packages/discord/opus.py", line 308, in set_bitrate
    _lib.opus_encoder_ctl(self._state, CTL_SET_BITRATE, kbps * 1024)
  File "/opt/homebrew/lib/python3.10/site-packages/discord/opus.py", line 92, in _err_lt
    raise OpusError(result)
discord.opus.OpusError: invalid argument
```

### Intents

Not applicable

### System Information

```
- Python v3.10.4-final
- discord.py v1.7.3-final
- aiohttp v3.7.4.post0
- system info: Darwin 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:46:32 PDT 2022; root:xnu-8020.101.4~15/RELEASE_ARM64_T6000
```

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.

#### [1/4] This bug only happens on arm64 macOS?

Yes, only on arm64 macOS.

Confirmed no issue on x86_64 macOS or Windows or Linux.

#### [2/4] Why can't I use Windows or Linux instead?

It's because my music production software is on this platform.

I'm willing to run test code snippets on my machine if you wish.

#### [3/4] Some more check on libopus

I try to check if libopus is correctly loaded.
**Result:** yes.
```python
>>> import discord.opus
>>> discord.opus._load_default()
True
>>> discord.opus._lib
<CDLL '/opt/homebrew/lib/libopus.dylib', handle 209560680 at 0x104f380a0>
[...]

PR #8052, opened 2022-05-18
Specify required argument types for FFI functions

m13253:
## Summary

Fix issue #8046

On arm64 macOS, variadic arguments have special calling convention rules.
Therefore, we must specify the required argument types for ctypes to pass the arguments correctly.

Apple documentation: https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Update-Code-that-Passes-Arguments-to-Variadic-Functions

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/discord/opus.py
+++ b/discord/opus.py
@@ -142,7 +142,7 @@ def _err_ne(result: T, func: Callable, args: List) -> T:
 # The fourth is the error handler.
 exported_functions: List[Tuple[Any, ...]] = [
     # Generic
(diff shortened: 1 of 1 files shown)


m13253:
FYI:
I already tested this patch on both **arm64 macOS** and **x86_64 Windows**,
to make sure the patch works on both platforms.

(I haven't tested Linux but I think this patch shouldn't break on Linux.)

PR #8059, opened 2022-05-23
Add support for extension setup variables for providing arguments from within `setup()`

Mega-JC:
## Summary

This PR adds a unified way to safely provide setup variables to extensions. It makes use of  `ContextVar` objects to store these variables as a dictionary. Extensions can access these variables by using the new `Bot.extension_vars` property from within their `setup()` function. To pass these setup variables, one can call `Bot.(re)load_extension("extname", package=..., extension_vars=dict(var1=1, var2=2, ...))`

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -33,6 +33,7 @@ import importlib.util
 import sys
 import traceback
 import types
+from contextvars import ContextVar
 from typing import (
     Any,
     Callable,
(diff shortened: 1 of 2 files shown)


kaylynn234:
I'm not sure how I feel about this approach. Using contextvars for this purpose feels pretty unnecessary and I don't really find this API particularly ergonomic.

Rapptz:
Thanks for the PR.

I'm not really a big fan of approaches like these that require modifying the signature of e.g. `setup` or `load_extension` (and its derivatives). These things tend to bleed. This approach tried to remedy this by using a contextvar but that's a lot of added overhead for what is essentially a simple key-value pairing. On top of that, I'm not sure I get many use cases where people would want to do this in the first place. If you control your bot and the extensions then it's easy to keep a mental mapping of the configuration by assigning them as attributes to the bot and accessing them within the setup function.

The only valid use case I've seen for this type of setup configuration is for third party extensions that don't control the bot instance themselves. I'm more of a fan of Sphinx's approach for configuration for things like that, though it might need some tidying up for it to work with type checkers (if at all).

PR #8061, opened 2022-05-23
Update basic_bot.py

ArjunSharda:
## Summary

Different way of printing in on_ready

This pull request will instead of having four lines for printing in on_ready, just having one using formatting and \n.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
Thanks but the previous code was fine.

PR #8096, opened 2022-06-02
Add on_raw_thread_update event dispatch

3nws:
## Summary

Add on_raw_thread_update event dispatch and updated documentation accordingly.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/state.py
+++ b/discord/state.py
@@ -874,16 +874,14 @@ class ConnectionState:
             _log.debug('THREAD_UPDATE referencing an unknown guild ID: %s. Discarding', guild_id)
             return
 
-        thread_id = int(data['id'])
         raw = RawThreadUpdateEvent(data)
-        thread = guild.get_thread(thread_id)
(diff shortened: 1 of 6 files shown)


discord/state.py
```diff
@@ -875,12 +875,15 @@ def parse_thread_update(self, data: gw.ThreadUpdateEvent) -> None:
...
         thread_id = int(data['id'])
+        raw = RawThreadUpdateEvent(data)
         thread = guild.get_thread(thread_id)
+        raw.thread = thread
         if thread is not None:
             old = copy.copy(thread)
             thread._update(data)
             if thread.archived:
                 guild._remove_thread(thread)
+            self.dispatch('raw_thread_update', raw)
```
LostLuma, line 886: This dispatch call should be outside of the `if` block as otherwise the event is only dispatched when the thread is cached.

Rapptz (approved):
Thanks.

PR #8098, opened 2022-06-03
examples: increase readability and add cogs example

Dark-Light007:
This PR adds typehints and return annotations for examples.

    - [ ] No Changes have to be documented..
- [ ] This PR adds up to readability (e.g. new method or parameters).
- [ ] This PR is **not** a breaking change (e.g. methods or parameters removed/renamed)

### DIFF
--- a/examples/app_commands/basic.py
+++ b/examples/app_commands/basic.py
@@ -8,7 +8,7 @@ MY_GUILD = discord.Object(id=0)  # replace with your guild id
 
 
 class MyClient(discord.Client):
-    def __init__(self, *, intents: discord.Intents):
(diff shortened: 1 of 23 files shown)


Rapptz:
Thanks for the PR.

A lot of these are wrong (e.g. `__init__` returns `None`, not `Self`). Likewise, I don't think overly annotating code that can be inferred really improves readability much. The library does it because it needs it for `--verifytypes` and in case there's divergence between inference algorithms but most actual programs have no such issues since they use one type checker and not many. Some of these are fine, but considering a lot of these muddy it I'm gonna have to close this PR.

PR #8103, opened 2022-06-03
[commands] Add Context.permissions and Context.bot_permissions

zephyrkul:
## Summary

Adds `Context.permissions` and `Context.bot_permissions` properties

The former returns the resolved permissions for the invoking user, which falls through to `Interaction.permissions` or `ctx.channel.permissions_for` depending on if the command is interaction-based or not. Essentially, this is a shortcut to skip overwrites resolution since `Interaction.permissions` is documented to already do that.

The latter returns the "effective" permissions for the bot when acting through `ctx.send` and the like, to reflect the fact that hybrid commands' app commands would be going through a webhook. When in a message command, this would also just fall back to `ctx.channel.permissions_for`.

This also includes changes to `ext.commands` checks to take advantage of these new properties.

Note that, as of the current state of this PR, `Context.permissions` may include voice permissions in text channels and the like due to how `Interaction.permissions` works. I can add in a check for the channel type and forcibly deny unrelated permissions if desired.

## Checklist

- [ ] This PR fixes an issue.


discord/ext/commands/context.py
```diff
@@ -452,6 +452,63 @@ def me(self) -> Union[Member, ClientUser]:
...
+        .. note::
+
+            For interaction-based commands, this will reflect the effective permissions
+            for :class:`Context` calls, which may differ from calls through
+            other :class:`.abc.Messageable` endpoints, like :attr:`channel`.
+
+            Notably, sending messages, embedding links, and attaching files are always
+            permitted, while reading messages might not be. Mass mentioning and using
+            external emojis and stickers also inherits their permissions from the
+            default guild role.
```
Rapptz, line 482: You can probably not use the admonition here.

discord/ext/commands/context.py
```diff
@@ -452,6 +452,63 @@ def me(self) -> Union[Member, ClientUser]:
...
         return self.guild.me if self.guild is not None else self.bot.user  # type: ignore
 
+    @discord.utils.cached_property
+    def permissions(self) -> Permissions:
+        """:class:`.Permissions`: Returns the resolved permissions for the invoking user in this channel.
+        Shorthand for :meth:`.abc.GuildChannel.permissions_for` or :attr:`.Interaction.permissions`.
+
+        .. versionadded:: 2.0
+        """
+        if self.channel.type == ChannelType.private:
```
Rapptz, line 462:
```suggestion
        if self.channel.type is ChannelType.private:
```

Rapptz (changes requested):
> Note that, as of the current state of this PR, Context.permissions may include voice permissions in text channels and the like due to how Interaction.permissions works. I can add in a check for the channel type and forcibly deny unrelated permissions if desired.

This is a bit tricky because this deviation in behaviour, combined with the fact that you changed the checks means that in hybrid_commands the checks will pass or fail depending on whether it's an interaction or not. For example, historically things like `@has_permissions(mute_members=True)` have always failed but now it depends on the context. I consider this to be a massive gotcha and a weird inconsistency on top of the pre-existing surprise.

As you've noted, there are a few solutions to this problem.
1. We keep the functions as-is and just document the inconsistency more prominently
2. We modify the behaviour of `Interaction.permissions` to fit within the confines of the library.

I have no strong opinion on here except that for consistency sake checks should stick to using the regular `permissions_for` method because this deviation would be too surprising. I suspect the easiest route would be to just document the discrepancy with a relevant example (such as e.g. `connect` not being available in text channels).

discord/ext/commands/context.py
zephyrkul: Do you mean that I should remove the `.. note::` directive and leave the content of the directive in the docstring, or rather that I should remove the entire block?

Rapptz: I meant removing the admonition not the content

Rapptz:
If you want to go that route, there are more implicit permissions, such as `connect` in voice channels being disabled means that most permissions get denied. 

https://github.com/Rapptz/discord.py/blob/ae51d5849a946699362372fb9d7f637bb81186f3/discord/channel.py#L959-L969

zephyrkul:
Thanks for the catch, I've added that case in. I believe all that isn't supported here is group DM channels, which bots can't join and which likely would be impossible to differentiate from a DM channel anyway.

Rapptz:
Now that `Interaction.app_permissions` is a thing, this PR might need to be updated.

discord/ext/commands/context.py
```diff
@@ -456,6 +456,84 @@ def me(self) -> Union[Member, ClientUser]:
...
+        """:class:`.Permissions`: Returns the resolved permissions for the bot in this channel.
+
+        For interaction-based commands, this will reflect the effective permissions
+        for :class:`Context` calls, which may differ from calls through
+        other :class:`.abc.Messageable` endpoints, like :attr:`channel`.
+
+        Notably, sending messages, embedding links, and attaching files are always
+        permitted, while reading messages might not be. Mass mentioning and using
+        external emojis and stickers also inherits their permissions from the
+        default guild role.
```
Rapptz, line 496: This will actually no longer be true in a month.

discord/ext/commands/context.py
```diff
@@ -456,6 +456,84 @@ def me(self) -> Union[Member, ClientUser]:
...
+            base.value &= ~denied.value
+        return base
+
+    @discord.utils.cached_property
+    def bot_permissions(self) -> Permissions:
+        """:class:`.Permissions`: Returns the resolved permissions for the bot in this channel.
+
+        For interaction-based commands, this will reflect the effective permissions
+        for :class:`Context` calls, which may differ from calls through
+        other :class:`.abc.Messageable` endpoints, like :attr:`channel`.
```
Rapptz, line 491: Might need to mention this uses app_permissions.

discord/ext/commands/context.py
Rapptz: Yeah I think forward compatibility is better here

zephyrkul: Since the planned change to ATTACH_FILES and EMBED_LINKS was reverted, I've left those in, but otherwise I've now updated this property with the planned 28 July change.

PR #8104, opened 2022-06-03
Use `static_format` over `format` if present

### LINKED ISSUE #7345: Asset replacing breaks sometimes when using 'gif' as `format`
### Summary

Asset replacing breaks sometimes when using 'gif' as `format`

### Reproduction Steps

- Use discord.py 2.0 (the alpha version, anyways).
- Get an avatar asset of a user (in this case the bot, though this can apply to other types of avatars and possibly other types of assets as well).
- Use the `replace` function to get either an animated GIF or a static PNG of said icon depending on what's provided.
  - In this specific use case and error, the icon needs to be static for the error to happen.
  - Also worth of note is that I get the URL right after by using `str`, though I don't think that matters too much.
  - In the example below I'm using the bot's icon so this is *technically* wasteful, but I only did it this way because I just wanted to use the same line of code for all avatar assets.

### Minimal Reproducible Code

```python
ctx.guild.me.avatar.replace(format='gif',static_format='png',size=128) # should do the trick in a command
```

### Expected Results

It to get the new asset just fine.

### Actual Results

```python
  File "PATH_TO_PYTHON\discord\ext\commands\core.py", line 125, in wrapped
  File "PATH_TO_COG", line 123, in about
    icon_url=f"{str(ctx.guild.me.avatar.replace(format='gif',static_format='png',size=128))}",
  File "PATH_TO_PYTHON\discord\asset.py", line 307, in replace
    raise InvalidArgument(f'format must be one of {VALID_STATIC_FORMATS}')
discord.errors.InvalidArgument: format must be one of frozenset({'jpeg', 'webp', 'png', 'jpg'})
```

### Intents

`guilds=True, members=True, emojis=True, messages=True, reactions=True`

### System Information

```
- Python v3.8.9-final
- discord.py v2.0.0-alpha
    - discord.py pkg_resources: v2.0.0a3453+g58ca9e99
- aiohttp v3.7.4.post0
```

### Checklist

Soheab:
## Summary

Fixes #7345 and resolved [card-81287372](https://github.com/Rapptz/discord.py/projects/3#card-81287372)

This PR uses `static_format` if `format` is not in VALID_STATIC_FORMATS and static_format is.
Not sure if this is the right way to do this.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/asset.py
+++ b/discord/asset.py
@@ -387,7 +387,12 @@ class Asset(AssetMixin):
                     raise ValueError(f'format must be one of {VALID_ASSET_FORMATS}')
             else:
                 if format not in VALID_STATIC_FORMATS:
-                    raise ValueError(f'format must be one of {VALID_STATIC_FORMATS}')
+                    # don't bother raising if static_format is set to a valid static format.
+                    if static_format is not MISSING and static_format in VALID_STATIC_FORMATS:
+                        format = static_format
+                    else:
+                        raise ValueError(f'format must be one of {VALID_STATIC_FORMATS}')
+
             url = url.with_path(f'{path}.{format}')
 
         if static_format is not MISSING and not self._animated:
--- a/discord/asset.py
+++ b/discord/asset.py
@@ -355,7 +355,7 @@ class Asset(AssetMixin):
 
 
         .. versionchanged:: 2.0
-            ``static_format`` is now preferred over ``format`` 
+            ``static_format`` is now preferred over ``format``
             if both are present and the asset is not animated.
 
         .. versionchanged:: 2.0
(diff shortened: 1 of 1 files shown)


ajsharda17 (approved):

Rapptz (changes requested):
I think since the `static_format` branch will run regardless, it might make sense to delay the error checking?

Ergo, change the original raising condition from `if format not in VALID_STATIC_FORMATS` to `if static_format is MISSING and format not in VALID_STATIC_FORMATS`.

Soheab:
> I think since the `static_format` branch will run regardless, it might make sense to delay the error checking?
> 
> Ergo, change the original raising condition from `if format not in VALID_STATIC_FORMATS` to `if static_format is MISSING and format not in VALID_STATIC_FORMATS`.

Yep, that works nicely.

Rapptz (approved):
Thanks.

PR #8129, opened 2022-06-09
Fix sticker limit in `_PREMIUM_GUILD_LIMITS`

sarthhh:
## Summary

`Guild.sticker_limit` was returning 0 for guilds with no boost tiers, which should actually be 5.
* Made changes in  `_PREMIUM_GUILD_LIMITS` for it to return 5 instead of 0.

## Checklist

- [ ] This PR fixes an issue.


Rapptz (approved):
This is live now, thanks.

PR #8132, opened 2022-06-09
Add `type` property to `app_commands.Command` and `app_commands.Group`

Soheab:
## Summary

This PR adds a `type` property to `app_commands.Command` and `app_commands.Group`.

For `app_commands.Group`, the new type property is used in `.to_dict()`.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
Oh right. I just remembered why this isn't a thing, because it uses a union type which is mostly useless. Can I ask what the use case is for exposing this anyway?

Rapptz:
I'm deciding against this feature, sorry.

PR #8134, opened 2022-06-09
Change to simpler pip install command

froggal:
## Summary
`python3 -m pip install -U discord.py` to `pip3 install discord.py`

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/README.ja.rst
+++ b/README.ja.rst
@@ -30,20 +30,20 @@ discord.py は機能豊富かつモダンで使いやすい、非同期処理に
 .. code:: sh
 
     # Linux/OS X
-    python3 -m pip install -U discord.py
+    pip3 install discord.py
 
     # Windows
-    py -3 -m pip install -U discord.py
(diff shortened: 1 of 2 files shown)


froggal:
+ Ja README

froggal:
Not related to this PR, but changing OS X to macOS in ja

ImVaskel:
generally speaking, it's preferred to be verbose with pip, as it could point to a different installation than intended.

sarthhh:
there are many cases where `pip3` isn't directly accessible through the path, while `python3 -m pip` always works. so whats the point of this?

pikaninja:
this also kinda just... doesnt work. If you remove the quotes around `discord.py[voice]` it breaks on macos

Rapptz:
Thanks for the PR. Usage of `-m pip` is actually recommended since it gives you the appropriate pip for the interpreter being used.

PR #8136, opened 2022-06-10
Select values contextvar

sgtlaggy:
## Summary

This fixes race conditions that can occur when you don't immediately copy `.values` in a `Select`'s callback.

Using this code for testing:
```py
class MyView(View):
    @select(options=[SelectOption(label='one'), SelectOption(label='two')])
    async def myselect(self, itx: Interaction, select: Select):
        before_sleep = select.values[:]
        await itx.response.defer()
        await asyncio.sleep(5)
        after_sleep = select.values[:]
        await itx.followup.send(f'{before_sleep = }\n{after_sleep = }')


class TestCog(commands.Cog):
    @commands.command()
    async def test(self, ctx: commands.Context):
        await ctx.send(view=MyView())
```
Select one option then select the other within 5 seconds.
Without the patch, `after_sleep` will be the second option selected.
With this patch, `after_sleep` will remain the same as `before_sleep`.

Excuse compact mode, just want to keep these short.
First without patch, second with:
imageimage

Selects still work as expected in modals.

Only foreseen issue is that if you use selects as a way to "store" choices, like with a "confirm" button, those choices would not be available if accessed from the button's callback. The select callback would need to cache the choices. This is also true currently if the view message is not ephemeral, as anyone could affect the Select's values.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ui/select.py
+++ b/discord/ui/select.py
@@ -54,7 +54,7 @@ if TYPE_CHECKING:
 
 V = TypeVar('V', bound='View', covariant=True)
 
-selected_values: ContextVar[List[str]] = ContextVar('selected_values')
+selected_values: ContextVar[Optional[List[str]]] = ContextVar('selected_values', default=None)
 
 
 class Select(Item[V]):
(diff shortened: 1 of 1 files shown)


Rapptz:
I'm unsure if this is going to be the way to go or not. I'm 50/50 torn on it.

There is an alternative approach that doesn't involve storing `Select.values` which is to change callbacks to take in the values instead. I figured this approach would be more future proof when other select types get added that deal with users, roles, channels, and mentionables. The way those selects are implemented in the API will require a lot of duplication too... 

Anyway, those are my thoughts on this. I feel like something needs to be done about this but I don't know what.

PR #8138, opened 2022-06-11
Fix AppCommand models

Soheab:
## Summary
Apologies for the title, don't know how to describe this otherwise.

I have done the following:
- Document `options` for AppCommand
- Remove `choices` and `required` from AppCommandGroup, why were these a thing?
- Rename `arguments` to `options` since it can include `AppCommand` and `AppCommandGroup` ~~and groups can't have arguments~~

This mainly fixes `AppCommandGroup` not supporting another AppCommandGroup + its arguments.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/models.py
+++ b/discord/app_commands/models.py
@@ -145,6 +145,8 @@ class AppCommand(Hashable):
         The application command's name.
     description: :class:`str`
         The application command's description.
+    options: List[Union[:class:`AppCommand`, :class:`AppCommandGroup`]]
+        A list of options.
     default_member_permissions: Optional[:class:`~discord.Permissions`]
         The default member permissions that can run this command.
     dm_permission: :class:`bool`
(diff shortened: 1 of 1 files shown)


Rapptz:
All of these idiosyncrasies are mainly because, as mentioned in the d.py guild, I rewrote the slash command code around ~4 times total between the time it was in a private beta for a few developers to the time it was actually published in the library. The required bool was there because at the time, the model had a `default` key which had a similar function but got removed. The `arguments` key is because originally the models were read-write which meant that users could edit them as well.

A lot of the other legacy was also from the class-based design that I was experimenting with. For example, the `to_dict()` method as a whole is unused and is just there because it had some use in a private implementation.

Obviously a lot of this legacy doesn't make sense with the actual final design so it'd seem bizarre to run into it.

PR #8141, opened 2022-06-12
Remove Command.usage in favor of Command.signature

sgtlaggy:
## Summary

The original implementation of `Command.usage` felt a bit rushed and seemed to often cause confusion.
Setting `signature` itself directly makes more sense and should lead to less confusion.

The behavior remains the same, `usage` is essentially renamed/merged into `signature`.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -274,8 +274,6 @@ class Command(_BaseCommand, Generic[CogT, P, T]):
         The long help text for the command.
     brief: Optional[:class:`str`]
         The short help text for the command.
-    usage: Optional[:class:`str`]
-        A replacement for arguments in the default help text.
     aliases: Union[List[:class:`str`], Tuple[:class:`str`]]
         The list of aliases the command can be invoked under.
     enabled: :class:`bool`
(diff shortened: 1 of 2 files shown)


rtk-rnjn:
I believe `Command.usage` is still "fine" :thinking:

Rapptz:
I get the intent behind this one but I think doing `cmd.signature = None` followed by `cmd.signature` and it giving you something else is more confusing than the previous set up.

sgtlaggy:
Maybe that could be documented as restoring the original POSIX-like signature?
I added the setter in case it was ever used after defining the command, as rarely as that may be done.
`signature` is still documented as just `str`, so maybe it should only accept `str` instead of `Optional[str]`.

If we stick with `usage`, changing the description to something like "An override for the default signature." and adding a note in the `signature` description similar to `short_doc` may be better, though not sure how I'd word that.

mikeshardmind:
> Maybe that could be documented as restoring the original POSIX-like signature? I added the setter in case it was ever used after defining the command, as rarely as that may be done. `signature` is still documented as just `str`, so maybe it should only accept `str` instead of `Optional[str]`.
> 
> If we stick with `usage`, changing the description to something like "An override for the default signature." and adding a note in the `signature` description similar to `short_doc` may be better, though not sure how I'd word that.

This could be made to accept a `str` instead of `Optional[str]`, and you could add a property deleter as well that is documented as resetting it.

Rapptz:
I've decided not to pursue this breaking change.

PR #8143, opened 2022-06-12
Handle special case of the default role in Role.mention

LeoCx1000:
## Summary

Make pinging the guild's default role return ``@everyone`` instead of ``<@&role_id>``, so it actually pings.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/role.py
+++ b/discord/role.py
@@ -341,7 +341,14 @@ class Role(Hashable):
 
     @property
     def mention(self) -> str:
-        """:class:`str`: Returns a string that allows you to mention a role."""
+        """:class:`str`: Returns a string that allows you to mention a role.
+
+        .. versionchanged:: 2.0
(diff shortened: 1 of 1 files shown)


discord/role.py
```diff
@@ -341,7 +341,14 @@ def created_at(self) -> datetime.datetime:
 
     @property
     def mention(self) -> str:
-        """:class:`str`: Returns a string that allows you to mention a role."""
+        """:class:`str`: Returns a string that allows you to mention a role.
+
+        .. versionchanged:: 2.0
```
dolfies, line 346: Is this necessary? The original behavior of "Returns a string that allows you to mention a role." hasn't changed.

LeoCx1000: It has changed.

dolfies: No, I meant it still "Returns a string that allows you to mention a role."

PR #8147, opened 2022-06-12
Make `app_commands.Command` and others comparable with `AppCommand(Group)`

### LINKED ISSUE #7669: compare AppCommand and Command
### Summary

Add posibility to compare `AppCommand` - API object and `Command` - runtime one

The core library

### The Problem

It would be easier to determine if command synchronization is required.

### The Ideal Solution

```py
synced_commands = await bot.tree.fetch_commands()
existing_commands = bot.tree.get_commands()

if synced_commands != existing_commands:
    await bot.tree.sync()
```

### The Current Solution

In turn, compare first the type of command, then the options if any.

### Additional Context

_No response_

Soheab:
## Summary

Fixes #7669 and resolves [card-81287257](https://github.com/Rapptz/discord.py/projects/3#card-81287257)

This PR makes it possible to compare the following classes to the `AppCommand(Group)` modals and itself:

- app_commands.Command
- app_commands.ContextMenu
- app_commands.Group

I renamed the `AppCommand`'s `.to_dict()` method to `.to_comparison_dict()` [as discussed](https://discord.com/channels/336642139381301249/603069307286454290/984789577602592829) in the [discord.py server](https://discord.com/invite/dpy)

I have tested this thoughtfully but not with every possibility but it should be fine...

## Checklist

    - [] I have updated the documentation to reflect the changes.
- [x] This PR fixes an issue.
- [] This PR adds something new (e.g. new method or parameters).

### DIFF
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -542,7 +542,6 @@ class Command(Generic[GroupT, P, T]):
             # always returns True if the command is a guild command
             # so we need to change it to match the behavior of the API
             # for this comparison.
-            # otherwise it is handled correctly.
             cls_to_dict = self.to_dict()
             if self._guild_ids is not None:
                 cls_to_dict['dm_permission'] = True
@@ -551,8 +550,8 @@ class Command(Generic[GroupT, P, T]):
         else:
             return False
 
-    def __ne__(self, other: object) -> bool:
-        return not self.__eq__(other)
+    def __hash__(self) -> int:
+        return hash(self.to_dict().values())
 
     def __set_name__(self, owner: Type[Any], name: str) -> None:
         self._attr = name
@@ -1000,8 +999,8 @@ class ContextMenu:
         else:
             return False
 
-    def __ne__(self, other: object) -> bool:
-        return not self.__eq__(other)
+    def __hash__(self) -> int:
+        return hash(self.to_dict().values())
 
     @property
     def callback(self) -> ContextMenuCallback:
@@ -1288,8 +1287,8 @@ class Group:
         else:
             return False
 
-    def __ne__(self, other: object) -> bool:
-        return not self.__eq__(other)
+    def __hash__(self) -> int:
+        return hash(self.to_dict().values())
 
     def __set_name__(self, owner: Type[Any], name: str) -> None:
         self._attr = name
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -49,7 +49,7 @@ from textwrap import TextWrapper
 import re
 
 from ..enums import AppCommandOptionType, AppCommandType
-from .models import Choice
+from .models import AppCommand, Choice
 from .transformers import annotation_to_parameter, CommandParameter, NoneType
 from .errors import AppCommandError, CheckFailure, CommandInvokeError, CommandSignatureMismatch, CommandAlreadyRegistered
 from ..message import Message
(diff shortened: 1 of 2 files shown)


discord/app_commands/commands.py
```diff
@@ -534,6 +534,26 @@ def __init__(
...
+            if self._guild_ids is not None:
+                cls_to_dict['dm_permission'] = True
+
+            return cls_to_dict == other.to_comparison_dict()
+        else:
+            return False
+
+    def __ne__(self, other: object) -> bool:
+        return not self.__eq__(other)
+
```
Rapptz, line 556: This is already implied

```suggestion
```

discord/app_commands/commands.py
```diff
@@ -972,6 +992,17 @@ def __init__(
...
+        if isinstance(other, self.__class__):
+            return self.to_dict() == other.to_dict()
+        elif isinstance(other, AppCommand):
+            return self.to_dict() == other.to_comparison_dict()
+        else:
+            return False
+
+    def __ne__(self, other: object) -> bool:
+        return not self.__eq__(other)
+
```
Rapptz, line 1005: This is already implied

```suggestion
```

discord/app_commands/models.py
```diff
@@ -201,15 +201,22 @@ def _from_data(self, data: ApplicationCommandPayload) -> None:
...
             'options': [opt.to_dict() for opt in self.options],
-        }  # type: ignore # Type checker does not understand this literal.
+            'nsfw': self.nsfw,
+            'default_member_permissions': self.default_member_permissions and self.default_member_permissions.value,
+            'dm_permission': self.dm_permission,
+        }
+        # context menus don't support this.
+        if self.type in (AppCommandType.user, AppCommandType.message):
+            del base['options']
+            del base['description']
```
Rapptz, line 217: Invert the condition so keys are added rather than removed.

discord/app_commands/models.py
```diff
@@ -201,15 +201,22 @@ def _from_data(self, data: ApplicationCommandPayload) -> None:
         self.dm_permission: bool = dm_permission
         self.nsfw: bool = data.get('nsfw', False)
 
-    def to_dict(self) -> ApplicationCommandPayload:
-        return {
-            'id': self.id,
+    def to_comparison_dict(self) -> ApplicationCommandPayload:
```
Rapptz, line 204:
```suggestion
    def to_comparison_dict(self) -> Dict[str, Any]:
```

discord/app_commands/commands.py
```diff
@@ -1249,6 +1280,17 @@ def __init__(
...
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, self.__class__):
+            return self.to_dict() == other.to_dict()
+        elif isinstance(other, AppCommand):
+            return self.to_dict() == other.to_comparison_dict()
+        else:
+            return False
+
+    def __ne__(self, other: object) -> bool:
+        return not self.__eq__(other)
```
Rapptz, line 1292:
```suggestion
```

Rapptz (changes requested):
Note that `api_model == user_command` will fail. You need to make sure the `__eq__` and `__ne__` of the app command models return `NotImplemented` when the types differ so it can propagate it back to the user facing types.

That aside, I'm a bit worried about tying the comparison behaviour to `to_dict()`, especially when localisations come into play.

Soheab:
> Note that `api_model == user_command` will fail. You need to make sure the `__eq__` and `__ne__` of the app command models return `NotImplemented` when the types differ so it can propagate it back to the user facing types.

Done, thanks.

> That aside, I'm a bit worried about tying the comparison behaviour to `to_dict()`, especially when localisations come into play.

Doing it with a dict seems like the easiest way IMO as it includes all the values we need to compare and localisations stuff should be the same afaik.

> Also note that by defining `__eq__` you break the hashability of the objects since now there's no `__hash__`, causing test failures.

Fixed. There are still some tests failing but I don't think those are in the scope of this PR?

Rapptz:
As discussed in the server, I think this feature has kind of become infeasible due to localisation. The original request was also kind of dubious since the server would know nothing changed making the sync call have no side effects.

PR #8159, opened 2022-06-17
first implementation of AutoMod

AbstractUmbra:
## Summary

This PR implements Discord's AutoMod features.

Summary of additions/changes:-
- Add AutoMod related types, gateway events and payload types.
- Addition of flags.ArrayFlags to support the new data format type.
- Documentation.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/audit_logs.py
+++ b/discord/audit_logs.py
@@ -510,7 +510,7 @@ class AuditLogEntry(Hashable):
             elif self.action is enums.AuditLogAction.auto_moderation_block_message:
                 self.extra = _AuditLogProxyAutoModAction(
                     auto_moderation_rule_name=extra['auto_moderation_rule_name'],
-                    auto_moderatio_rule_trigger_type=extra['auto_moderation_rule_trigger_type'],
+                    auto_moderation_rule_trigger_type=extra['auto_moderation_rule_trigger_type'],
                     channel_id=extra['channel_id'],
                 )
 
--- a/discord/audit_logs.py
+++ b/discord/audit_logs.py
@@ -276,6 +276,7 @@ class AuditLogChanges:
         'preferred_locale':              (None, _enum_transformer(enums.Locale)),
         'image_hash':                    ('cover_image', _transform_cover_image),
         'app_command_permission_update': ('app_command_permissions', _transform_app_command_permissions),
+        'trigger_type':                  (None, _enum_transformer(enums.AutoModRuleTriggerType)),
     }
     # fmt: on
 
--- a/discord/audit_logs.py
+++ b/discord/audit_logs.py
@@ -507,7 +507,7 @@ class AuditLogEntry(Hashable):
                     channel=self.guild.get_channel_or_thread(channel_id) or Object(id=channel_id),
                     message_id=int(extra['message_id']),
                 )
-            elif self.action is enums.AuditLogAction.auto_moderation_block_message:
+            elif self.action is enums.AuditLogAction.automod_block_message:
                 self.extra = _AuditLogProxyAutoModAction(
                     auto_moderation_rule_name=extra['auto_moderation_rule_name'],
                     auto_moderation_rule_trigger_type=extra['auto_moderation_rule_trigger_type'],
--- a/discord/audit_logs.py
+++ b/discord/audit_logs.py
@@ -394,6 +394,12 @@ class _AuditLogProxyMessageBulkDelete(_AuditLogProxy):
     count: int
 
 
+class _AuditLogProxyAutoModAction(_AuditLogProxy):
+    auto_moderation_rule_name: str
+    auto_moderation_rule_trigger_type: str
+    channel_id: str
+
+
 class AuditLogEntry(Hashable):
     r"""Represents an Audit Log entry.
 
@@ -469,6 +475,7 @@ class AuditLogEntry(Hashable):
             _AuditLogProxyPinAction,
             _AuditLogProxyStageInstanceAction,
             _AuditLogProxyMessageBulkDelete,
+            _AuditLogProxyAutoModAction,
             Member, User, None, PartialIntegration,
             Role, Object
         ] = None
@@ -500,6 +507,13 @@ class AuditLogEntry(Hashable):
                     channel=self.guild.get_channel_or_thread(channel_id) or Object(id=channel_id),
                     message_id=int(extra['message_id']),
                 )
+            elif self.action is enums.AuditLogAction.auto_moderation_block_message:
+                self.extra = _AuditLogProxyAutoModAction(
+                    auto_moderation_rule_name=extra['auto_moderation_rule_name'],
+                    auto_moderatio_rule_trigger_type=extra['auto_moderation_rule_trigger_type'],
+                    channel_id=extra['channel_id'],
+                )
+
             elif self.action.name.startswith('overwrite_'):
                 # the overwrite_ actions have a dict with some information
                 instance_id = int(extra['id'])
--- a/discord/audit_logs.py
+++ b/discord/audit_logs.py
@@ -508,10 +508,13 @@ class AuditLogEntry(Hashable):
                     message_id=int(extra['message_id']),
                 )
             elif self.action is enums.AuditLogAction.automod_block_message:
+                channel_id = int(extra['channel_id'])
                 self.extra = _AuditLogProxyAutoModAction(
                     auto_moderation_rule_name=extra['auto_moderation_rule_name'],
-                    auto_moderation_rule_trigger_type=extra['auto_moderation_rule_trigger_type'],
-                    channel_id=extra['channel_id'],
+                    auto_moderation_rule_trigger_type=enums.try_enum(
+                        enums.AutoModRuleTriggerType, extra['auto_moderation_rule_trigger_type']
+                    ),
+                    channel=self.guild.get_channel_or_thread(channel_id) or Object(id=channel_id),
                 )
 
             elif self.action.name.startswith('overwrite_'):
--- a/discord/audit_logs.py
+++ b/discord/audit_logs.py
@@ -395,9 +395,9 @@ class _AuditLogProxyMessageBulkDelete(_AuditLogProxy):
 
 
 class _AuditLogProxyAutoModAction(_AuditLogProxy):
-    auto_moderation_rule_name: str
-    auto_moderation_rule_trigger_type: str
-    channel_id: str
+    automod_rule_name: str
+    automod_rule_trigger_type: str
+    channel: Union[abc.GuildChannel, Thread]
 
 
 class AuditLogEntry(Hashable):
@@ -510,8 +510,8 @@ class AuditLogEntry(Hashable):
             elif self.action is enums.AuditLogAction.automod_block_message:
                 channel_id = int(extra['channel_id'])
                 self.extra = _AuditLogProxyAutoModAction(
-                    auto_moderation_rule_name=extra['auto_moderation_rule_name'],
-                    auto_moderation_rule_trigger_type=enums.try_enum(
+                    automod_rule_name=extra['auto_moderation_rule_name'],
+                    automod_rule_trigger_type=enums.try_enum(
                         enums.AutoModRuleTriggerType, extra['auto_moderation_rule_trigger_type']
                     ),
                     channel=self.guild.get_channel_or_thread(channel_id) or Object(id=channel_id),
@@ -677,3 +677,6 @@ class AuditLogEntry(Hashable):
 
     def _convert_target_integration_or_app_command(self, target_id: int) -> Union[PartialIntegration, AppCommand, Object]:
         return self._get_integration_by_app_id(target_id) or self._get_app_command(target_id) or Object(target_id)
+
+    def _convert_target_auto_moderation(self, target_id: int) -> Union[Member, Object]:
+        return self.guild.get_member(target_id) or Object(target_id)
--- a/discord/automod.py
+++ b/discord/automod.py
@@ -88,7 +88,7 @@ class AutoModRuleAction:
             raise ValueError('channel_id must be set if type is send_alert_message.')
 
     @classmethod
-    def from_data(cls: Type[Self], data: AutoModerationActionPayload) -> Self:
+    def from_data(cls, data: AutoModerationActionPayload) -> Self:
         type_ = try_enum(AutoModRuleActionType, data['type'])
         if data['type'] == AutoModRuleActionType.block_message.value:
             return cls(type=type_)
@@ -101,14 +101,11 @@ class AutoModRuleAction:
         return cls(type=type_)
 
     def to_dict(self) -> AutoModerationActionPayload:
-        ret: AutoModerationActionPayload = {'type': self.type.value}
+        ret: AutoModerationActionPayload = {'type': self.type.value, 'metadata': {}}
         if self.type is AutoModRuleActionType.timeout:
             ret['metadata'] = {'duration_seconds': self.duration_seconds}  # type: ignore # guarded by type check
-            return ret
         elif self.type is AutoModRuleActionType.send_alert_message:
             ret['metadata'] = {'channel_id': str(self.channel_id)}  # type: ignore # guarded by type check
-            return ret
-        ret['metadata'] = {}
         return ret
 
 
@@ -164,7 +161,7 @@ class AutoModTrigger:
 
 
 class AutoModRule:
-    """Represents a auto moderation rule.
+    """Represents an auto moderation rule.
 
     Attributes
     -----------
--- a/discord/automod.py
+++ b/discord/automod.py
@@ -1,0 +1,400 @@
+"""
+The MIT License (MIT)
+
+Copyright (c) 2015-present Rapptz
+
+Permission is hereby granted, free of charge, to any person obtaining a
+copy of this software and associated documentation files (the "Software"),
+to deal in the Software without restriction, including without limitation
+the rights to use, copy, modify, merge, publish, distribute, sublicense,
+and/or sell copies of the Software, and to permit persons to whom the
+Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Optional, List, Union, Type
+from typing_extensions import Self
+
+
+from .enums import AutoModRuleTriggerType, AutoModRuleActionType, AutoModRuleEventType, try_enum
+from .flags import AutoModPresets
+from . import utils
+
+if TYPE_CHECKING:
+    from .abc import GuildChannel
+    from .threads import Thread
+    from .guild import Guild
+    from .member import Member
+    from .state import ConnectionState
+    from .types.automod import (
+        AutoModerationRule as AutoModerationRulePayload,
+        AutoModerationTriggerMetadata as AutoModerationTriggerMetadataPayload,
+        AutoModerationAction as AutoModerationActionPayload,
+        AutoModerationActionExecution as AutoModerationActionExecutionPayload,
+    )
+    from .user import User
+    from .role import Role
+
+__all__ = (
+    'AutoModRuleAction',
+    'AutoModTrigger',
+    'AutoModRule',
+    'AutoModRuleExecution',
+)
+
+
+class AutoModRuleAction:
+    """Represents a rule action.
+
+    Attributes
+    -----------
+    type: :class:`AutoModRuleActionType`
+        The type of action to take.
+    channel_id: Optional[int]
+        The channel to send the alert message to, if any.
+    duration_seconds: Optional[int]
+        The duration of the timeout to apply, if any.
+
+    Raises
+    -------
+    :exc:`ValueError`
+        Provided arguments do not match the needed types.
+    """
+
+    __slots__ = ('type', 'channel_id', 'duration_seconds')
+
+    def __init__(
+        self, *, type: AutoModRuleActionType, channel_id: Optional[int] = None, duration_seconds: Optional[int] = None
+    ) -> None:
+        self.type: AutoModRuleActionType = type
+        self.channel_id: Optional[int] = channel_id
+        self.duration_seconds: Optional[int] = duration_seconds
+
+        if self.type is AutoModRuleActionType.timeout and self.duration_seconds is None:
+            raise ValueError('duration_seconds must be set if type is timeout.')
+        elif self.type is AutoModRuleActionType.send_alert_message and self.channel_id is None:
+            raise ValueError('channel_id must be set if type is send_alert_message.')
+
+    @classmethod
+    def from_data(cls: Type[Self], data: AutoModerationActionPayload) -> Self:
+        type_ = try_enum(AutoModRuleActionType, data['type'])
+        if data['type'] == AutoModRuleActionType.block_message.value:
+            return cls(type=type_)
+        elif data['type'] == AutoModRuleActionType.timeout.value:
+            duration_seconds = data['metadata']['duration_seconds']
+            return cls(type=type_, duration_seconds=duration_seconds)
+        elif data['type'] == AutoModRuleActionType.send_alert_message.value:
+            channel_id = int(data['metadata']['channel_id'])
+            return cls(type=type_, channel_id=channel_id)
+        return cls(type=type_)
+
+    def to_dict(self) -> AutoModerationActionPayload:
+        ret: AutoModerationActionPayload = {'type': self.type.value}
+        if self.type is AutoModRuleActionType.timeout:
+            ret['metadata'] = {'duration_seconds': self.duration_seconds}  # type: ignore # guarded by type check
+            return ret
+        elif self.type is AutoModRuleActionType.send_alert_message:
+            ret['metadata'] = {'channel_id': str(self.channel_id)}  # type: ignore # guarded by type check
+            return ret
+        ret['metadata'] = {}
+        return ret
+
+
+class AutoModTrigger:
+    """Represents a trigger for an auto moderation rule.
+
+    Attributes
+    -----------
+    type: :class:`AutoModRuleTriggerType`
+        The type of trigger.
+    keyword_filter: Optional[List[:class:`str`]]
+        The list of strings that will trigger the keyword filter.
+    presets: Optional[List[:class:`int`]]
+        The list of presets used with the preset keyword filter.
+
+    .. versionadded:: 2.0
+    """
+
+    __slots__ = ('type', 'keyword_filter', 'presets')
+
+    def __init__(
+        self,
+        *,
+        type: AutoModRuleTriggerType,
+        keyword_filter: Optional[List[str]] = None,
+        presets: Optional[AutoModPresets] = None,
+    ) -> None:
+        self.type: AutoModRuleTriggerType = type
+        self.keyword_filter: Optional[List[str]] = keyword_filter
+        self.presets: Optional[AutoModPresets] = presets
+
+        if self.type is AutoModRuleTriggerType.keyword and self.keyword_filter is None:
+            raise ValueError('keyword_filter must be set if type is keyword.')
+        elif self.type is AutoModRuleTriggerType.keyword_preset and self.presets is None:
+            raise ValueError('presets must be set if type is keyword_preset.')
+
+    @classmethod
+    def from_data(cls: Type[Self], type: int, data: Optional[AutoModerationTriggerMetadataPayload]) -> Self:
+        type_ = try_enum(AutoModRuleTriggerType, type)
+        if type_ is AutoModRuleTriggerType.keyword:
+            return cls(type=type_, keyword_filter=data['keyword_filter'])  # type: ignore # unable to typeguard due to outer payload
+        elif type_ is AutoModRuleTriggerType.keyword_preset:
+            return cls(type=type_, presets=AutoModPresets._from_value(data['presets']))  # type: ignore # unable to typeguard due to outer payload
+        return cls(type=type_)
+
+    def to_metadata_dict(self) -> AutoModerationTriggerMetadataPayload:
+        if self.type is AutoModRuleTriggerType.keyword:
+            return {'keyword_filter': self.keyword_filter}  # type: ignore # guarded by the type check
+        elif self.type is AutoModRuleTriggerType.keyword_preset:
+            return {'presets': self.presets.value}  # type: ignore # guarded by the type check
+
+        return {}
+
+
+class AutoModRule:
+    """Represents a auto moderation rule.
+
+    Attributes
+    -----------
+    id: :class:`int`
+        The ID of the rule.
+    guild: :class:`Guild`
+        The guild the rule is for.
+    name: :class:`str`
+        The name of the rule.
+    creator_id: :class:`int`
+        The ID of the user that created the rule.
+    trigger: :class:`AutoModRuleTrigger`
+        The rule's trigger.
+    enabled: :class:`bool`
+        Whether the rule is enabled.
+    raw_exempt_roles: List[:class:`int`]
+        The IDs of the roles that are exempt from the rule.
+    raw_exempt_channels: List[:class:`int`]
+        The IDs of the channels that are exempt from the rule.
+
+    .. versionadded:: 2.0
+    """
+
+    __slots__ = (
+        '_state',
+        'id',
+        'guild',
+        'name',
+        'creator_id',
+        'event_type',
+        'trigger',
+        'enabled',
+        'raw_exempt_roles',
+        'raw_exempt_channels',
+        '_actions',
+    )
+
+    def __init__(self, *, data: AutoModerationRulePayload, guild: Guild, state: ConnectionState) -> None:
+        self._state: ConnectionState = state
+        self.guild: Guild = guild
+        self.id: int = int(data['id'])
+        self.name: str = data['name']
+        self.creator_id = int(data['creator_id'])
+        self.event_type: AutoModRuleEventType = try_enum(AutoModRuleEventType, data['event_type'])
+        self.trigger: AutoModTrigger = AutoModTrigger.from_data(data['trigger_type'], data=data.get('trigger_metadata'))
+        self.enabled: bool = data['enabled']
+        self.raw_exempt_roles: List[int] = [int(role_id) for role_id in data['exempt_roles']]
+        self.raw_exempt_channels: List[int] = [int(channel_id) for channel_id in data['exempt_channels']]
+        self._actions: List[AutoModerationActionPayload] = data['actions']
+
+    def __repr__(self) -> str:
+        return f'<AutoModRule id={self.id} name={self.name!r}>'
+
+    def to_dict(self) -> AutoModerationRulePayload:
+        ret: AutoModerationRulePayload = {
+            'id': str(self.id),
+            'guild_id': str(self.guild.id),
+            'name': self.name,
+            'creator_id': str(self.creator_id),
+            'event_type': self.event_type.value,
+            'trigger_type': self.trigger.type.value,
+            'trigger_metadata': self.trigger.to_metadata_dict(),
+            'actions': [action.to_dict() for action in self.actions],
+            'enabled': self.enabled,
+            'exempt_roles': [str(role_id) for role_id in self.raw_exempt_roles],
+            'exempt_channels': [str(channel_id) for channel_id in self.raw_exempt_channels],
+        }  # type: ignore # trigger types break the flow here.
+
+        return ret
+
+    @property
+    def creator(self) -> Optional[User]:
+        """Optional[:class:`User`]: The user that created this rule."""
+        return self._state.get_user(self.creator_id)
+
+    @property
+    def exempt_roles(self) -> List[Role]:
+        """List[:class:`Role`]: The roles that are exempt from this rule."""
+        it = [role for role in (self.guild.get_role(role_id) for role_id in self.raw_exempt_roles) if role is not None]
+        return utils._unique(it)
+
+    @property
+    def exempt_channels(self) -> List[Union[GuildChannel, Thread]]:
+        """List[Union[:class:`GuildChannel`, :class:`Thread`]]: The channels that are exempt from this rule."""
+        it = filter(None, map(self.guild._resolve_channel, self.raw_exempt_channels))
+        return utils._unique(it)
+
+    @property
+    def actions(self) -> List[AutoModRuleAction]:
+        """List[:class:`AutoModRuleAction`]: The actions that are taken when this rule is triggered."""
+        return [AutoModRuleAction.from_data(action) for action in self._actions]
+
+    async def edit(
+        self,
+        *,
+        name: Optional[str] = None,
+        event_type: Optional[AutoModRuleEventType] = None,
+        actions: List[AutoModRuleAction],
+        enabled: Optional[bool] = None,
+        exempt_roles: Optional[List[Role]] = None,
+        exempt_channels: Optional[List[Union[GuildChannel, Thread]]] = None,
+    ) -> Self:
+        """|coro|
+
+        This method will alter an existing auto moderation rule.
+
+        Parameters
+        -----------
+        name: Optional[:class:`str`]
+            The name to change to, if any.
+        event_type: Optional[:class:`AutoModRuleEventType`]
+            The event type to change to, if any.
+        actions: List[:class:`AutoModRuleAction`]
+            The rule actions to update, if any.
+        enabled: Optional[:class:`bool`]
+            Whether the rule should be enabled or not.
+        exempt_roles: Optional[List[:class:`Role`]]
+            The roles to exempt from the rule, if any.
+        exempt_channels: Optional[List[Union[:class:`GuildChannel`], :class:`Thread`]]
+            The channels to exempt from the rule, if any.
+
+        .. versionadded:: 2.0
+
+        Raises
+        -------
+        :exc:`Forbidden`
+            You are not authorised to perform this action.
+        :exc:`NotFound`
+            The rule was not found.
+
+        Returns
+        --------
+        :class:`AutoModRule`
+            The updated auto moderation rule.
+        """
+        transformed_actions = None
+        if actions:
+            transformed_actions = [action.to_dict() for action in actions]
+
+        data = await self._state.http.modify_auto_moderation_rule(
+            self.guild.id,
+            self.id,
+            name=name,
+            event_type=event_type,
+            actions=transformed_actions,
+            enabled=enabled,
+            exempt_roles=exempt_roles,
+            exempt_channels=exempt_channels,
+        )
+
+        return self.__class__(data=data, guild=self.guild, state=self._state)
+
+    async def delete(self) -> None:
+        """|coro|
+
+        This method will delete the auto moderation rule.
+
+        Raises
+        -------
+        :exc:`Forbidden`
+            You are not authorised to perform this action.
+        :exc:`NotFound`
+            The rule was not found.
+        """
+        await self._state.http.delete_auto_moderation_rule(self.guild.id, self.id)
+
+
+class AutoModRuleExecution:
+    """Represents an auto moderation rule execution.
+
+    Attributes
+    -----------
+    """
+
+    __slots__ = (
+        '_state',
+        '_guild_id',
+        '_channel_id',
+        'action',
+        'rule_id',
+        'rule_trigger_type',
+        'user_id',
+        'message_id',
+        'alert_system_message_id',
+        'content',
+        'matched_keyword',
+        'matched_content',
+    )
+
+    def __init__(self, *, data: AutoModerationActionExecutionPayload, state: ConnectionState) -> None:
+        self._state: ConnectionState = state
+        self._guild_id: int = int(data['guild_id'])
+        self._channel_id: Optional[int] = utils._get_as_snowflake(data, 'channel_id')
+        self.message_id: Optional[int] = utils._get_as_snowflake(data, 'message_id')
+        self.action: AutoModRuleAction = AutoModRuleAction.from_data(data['action'])
+        self.rule_id: int = int(data['rule_id'])
+        self.rule_trigger_type: AutoModRuleTriggerType = try_enum(AutoModRuleTriggerType, data['rule_trigger_type'])
+        self.user_id: int = int(data['user_id'])
+        self.alert_system_message_id: Optional[int] = utils._get_as_snowflake(data, 'alert_system_message_id')
+        self.content: str = data['content']
+        self.matched_keyword: Optional[str] = data['matched_keyword']
+        self.matched_content: Optional[str] = data['matched_content']
+
+    @property
+    def guild(self) -> Guild:
+        """Optional[:class:`Guild`]: The guild this rule was executed in."""
+        return self._state._get_guild(self._guild_id)  # type: ignore # this should never be None here
+
+    @property
+    def channel(self) -> Optional[Union[GuildChannel, Thread]]:
+        """Optional[Union[:class:`GuildChannel`, :class:`Thread`]]: The channel this rule was executed in."""
+        if self._channel_id:
+            return self.guild and self.guild.get_channel(self._channel_id)
+        return None
+
+    @property
+    def member(self) -> Optional[Member]:
+        """Optional[:class:`Member`]: The member this rule was executed on/who triggered this rule."""
+        return self.guild and self.guild.get_member(self.user_id)
+
+    async def fetch_rule(self) -> AutoModRule:
+        """|coro|
+
+        This method will fetch the rule that was executed.
+
+        Returns
+        --------
+        :class:`AutoModRule`
+            The rule that was executed.
+        """
+
+        data = await self._state.http.get_auto_moderation_rule(self.guild.id, self.rule_id)
+        return AutoModRule(data=data, guild=self.guild, state=self._state)
--- a/discord/automod.py
+++ b/discord/automod.py
@@ -25,7 +25,6 @@ DEALINGS IN THE SOFTWARE.
 from __future__ import annotations
 
 from typing import TYPE_CHECKING, Optional, List, Union, Type
-from typing_extensions import Self
 
 
 from .enums import AutoModRuleTriggerType, AutoModRuleActionType, AutoModRuleEventType, try_enum
@@ -33,6 +32,7 @@ from .flags import AutoModPresets
 from . import utils
 
 if TYPE_CHECKING:
+    from typing_extensions import Self
     from .abc import GuildChannel
     from .threads import Thread
     from .guild import Guild
--- a/discord/automod.py
+++ b/discord/automod.py
@@ -301,7 +301,7 @@ class AutoModRule:
         exempt_channels: Sequence[:class:`abc.Snowflake`]
             The new channels to exempt from the rule.
         reason: :class:`str`
-            The reason for updating this rule.
+            The reason for updating this rule. Shows up on the audit log.
 
         Raises
         -------
@@ -341,7 +341,7 @@ class AutoModRule:
         Parameters
         -----------
         reason: :class:`str`
-            The reason for deleting this rule.
+            The reason for deleting this rule. Shows up on the audit log.
 
         Raises
         -------
--- a/discord/automod.py
+++ b/discord/automod.py
@@ -176,7 +176,7 @@ class AutoModRule:
         The name of the rule.
     creator_id: :class:`int`
         The ID of the user that created the rule.
-    trigger: :class:`AutoModRuleTrigger`
+    trigger: :class:`AutoModTrigger`
         The rule's trigger.
     enabled: :class:`bool`
         Whether the rule is enabled.
@@ -248,7 +248,7 @@ class AutoModRule:
 
     @property
     def exempt_channels(self) -> List[Union[GuildChannel, Thread]]:
-        """List[Union[:class:`GuildChannel`, :class:`Thread`]]: The channels that are exempt from this rule."""
+        """List[Union[:class:`abc.GuildChannel`, :class:`Thread`]]: The channels that are exempt from this rule."""
         it = filter(None, map(self.guild._resolve_channel, self.raw_exempt_channels))
         return utils._unique(it)
 
@@ -283,10 +283,8 @@ class AutoModRule:
             Whether the rule should be enabled or not.
         exempt_roles: Optional[List[:class:`Role`]]
             The roles to exempt from the rule, if any.
-        exempt_channels: Optional[List[Union[:class:`GuildChannel`], :class:`Thread`]]
+        exempt_channels: Optional[List[Union[:class:`abc.GuildChannel`], :class:`Thread`]]
             The channels to exempt from the rule, if any.
-
-        .. versionadded:: 2.0
 
         Raises
         -------
@@ -375,7 +373,7 @@ class AutoModRuleExecution:
 
     @property
     def channel(self) -> Optional[Union[GuildChannel, Thread]]:
-        """Optional[Union[:class:`GuildChannel`, :class:`Thread`]]: The channel this rule was executed in."""
+        """Optional[Union[:class:`abc.GuildChannel`, :class:`Thread`]]: The channel this rule was executed in."""
         if self._channel_id:
             return self.guild and self.guild.get_channel(self._channel_id)
         return None
--- a/discord/automod.py
+++ b/discord/automod.py
@@ -257,7 +257,7 @@ class AutoModRule:
         *,
         name: Optional[str] = None,
         event_type: Optional[AutoModRuleEventType] = None,
-        actions: List[AutoModRuleAction],
+        actions: Optional[List[AutoModRuleAction]] = None,
         enabled: Optional[bool] = None,
         exempt_roles: Optional[List[Role]] = None,
         exempt_channels: Optional[List[Union[GuildChannel, Thread]]] = None,
@@ -272,7 +272,7 @@ class AutoModRule:
             The new name to change to.
         event_type: Optional[:class:`AutoModRuleEventType`]
             The new event type to change to.
-        actions: List[:class:`AutoModRuleAction`]
+        actions: Optional[List[:class:`AutoModRuleAction`]]
             The new rule actions to update.
         enabled: Optional[:class:`bool`]
             Whether the rule should be enabled or not.
--- a/discord/automod.py
+++ b/discord/automod.py
@@ -338,7 +338,7 @@ class AutoModRule:
             name=name,
             event_type=event_type,
             actions=transformed_actions,
-            trigger_metadata=trigger.to_metadata_dict(),
+            trigger_metadata=trigger.to_metadata_dict() if trigger else None,
             enabled=enabled,
             exempt_roles=exempt_roles,
             exempt_channels=exempt_channels,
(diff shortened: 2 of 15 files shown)


discord/automod.py
```diff
@@ -0,0 +1,398 @@
...
+        if self.type is AutoModRuleTriggerType.keyword:
+            return {'keyword_filter': self.keyword_filter}  # type: ignore # guarded by the type check
+        elif self.type is AutoModRuleTriggerType.keyword_preset:
+            return {'presets': self.presets.value}  # type: ignore # guarded by the type check
+
+        return {}
+
+
+class AutoModRule:
+    """Represents a auto moderation rule.
```
scrazzz, line 167:
```suggestion
    """Represents an auto moderation rule.
```

discord/automod.py
```diff
@@ -0,0 +1,398 @@
...
+        name: Optional[str] = None,
+        event_type: Optional[AutoModRuleEventType] = None,
+        actions: List[AutoModRuleAction],
+        enabled: Optional[bool] = None,
+        exempt_roles: Optional[List[Role]] = None,
+        exempt_channels: Optional[List[Union[GuildChannel, Thread]]] = None,
+    ) -> Self:
+        """|coro|
+
+        This method will alter an existing auto moderation rule.
```
scrazzz, line 272:
```suggestion
        This method will edit an existing auto moderation rule.
```

scrazzz (changes requested):

discord/automod.py
```diff
@@ -0,0 +1,398 @@
...
+        self.channel_id: Optional[int] = channel_id
+        self.duration_seconds: Optional[int] = duration_seconds
+
+        if self.type is AutoModRuleActionType.timeout and self.duration_seconds is None:
+            raise ValueError('duration_seconds must be set if type is timeout.')
+        elif self.type is AutoModRuleActionType.send_alert_message and self.channel_id is None:
+            raise ValueError('channel_id must be set if type is send_alert_message.')
+
+    @classmethod
+    def from_data(cls: Type[Self], data: AutoModerationActionPayload) -> Self:
```
Gobot1234, line 91: You don't need Type[Self] as an annotation here
```suggestion
    def from_data(cls, data: AutoModerationActionPayload) -> Self:
```

discord/audit_logs.py
```diff
@@ -500,6 +507,13 @@ def _from_data(self, data: AuditLogEntryPayload) -> None:
                     channel=self.guild.get_channel_or_thread(channel_id) or Object(id=channel_id),
                     message_id=int(extra['message_id']),
                 )
+            elif self.action is enums.AuditLogAction.auto_moderation_block_message:
+                self.extra = _AuditLogProxyAutoModAction(
+                    auto_moderation_rule_name=extra['auto_moderation_rule_name'],
+                    auto_moderatio_rule_trigger_type=extra['auto_moderation_rule_trigger_type'],
```
NCPlayz, line 513:
```suggestion
                    auto_moderation_rule_trigger_type=extra['auto_moderation_rule_trigger_type'],
```

discord/automod.py
```diff
@@ -0,0 +1,398 @@
...
+    def to_dict(self) -> AutoModerationActionPayload:
+        ret: AutoModerationActionPayload = {'type': self.type.value}
+        if self.type is AutoModRuleActionType.timeout:
+            ret['metadata'] = {'duration_seconds': self.duration_seconds}  # type: ignore # guarded by type check
+            return ret
+        elif self.type is AutoModRuleActionType.send_alert_message:
+            ret['metadata'] = {'channel_id': str(self.channel_id)}  # type: ignore # guarded by type check
+            return ret
+        ret['metadata'] = {}
+        return ret
```
NCPlayz, line 112: Should have metadata as a default key and remove the extra returns

discord/enums.py
```diff
@@ -347,6 +350,10 @@ class AuditLogAction(Enum):
     thread_update                 = 111
     thread_delete                 = 112
     app_command_permission_update = 121
+    auto_moderation_rule_create   = 140
+    auto_moderation_rule_update   = 141
+    auto_moderation_rule_delete   = 142
+    auto_moderation_block_message = 143
```
NCPlayz, line 356: Maybe set the category of these too

discord/flags.py
```diff
@@ -1104,6 +1108,32 @@ def guild_scheduled_events(self):
...
+        return 1 << 20
+
+    @flag_value
+    def auto_moderation_execution(self):
+        """:class:`bool`: Whether auto moderation configuration related events are enabled.
+
+        This corresponds to the following events:
+        - :func:`on_auto_moderation_rule_create`
+        - :func:`on_auto_moderation_rule_update`
+        - :func:`on_auto_moderation_rule_delete`
```
Herbstblatt, line 1131:
```suggestion
        """:class:`bool`: Whether auto moderation execution related events are enabled.

        This corresponds to the following events:
        - :func:`on_auto_moderation_action_execution`
```

discord/flags.py
```diff
@@ -1436,3 +1466,107 @@ class ChannelFlags(BaseFlags):
...
+        ret = []
+        for idx, (_, value) in enumerate(self, start=1):
+            if value:
+                ret.append(idx)
+
+        return ret
+
+
+@fill_with_flags()
+class AutoModPresets(ArrayFlags):
```
Herbstblatt, line 1489: Don't see this class mentioned anywhere in api.rst
Also flags (`profanity`, `sexual_content`, `slurs`) should probably have descriptions?

AbstractUmbra: Good catch, thanks

discord/automod.py
```diff
@@ -0,0 +1,413 @@
...
+
+        if self.type is AutoModRuleActionType.timeout and self.duration_seconds is None:
+            raise ValueError('duration_seconds must be set if type is timeout.')
+        elif self.type is AutoModRuleActionType.send_alert_message and self.channel_id is None:
+            raise ValueError('channel_id must be set if type is send_alert_message.')
+
+    @classmethod
+    def from_data(cls, data: AutoModerationActionPayload) -> Self:
+        type_ = try_enum(AutoModRuleActionType, data['type'])
+        if data['type'] == AutoModRuleActionType.block_message.value:
```
NCPlayz, line 93: Can remove this statement

discord/automod.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+class AutoModRuleAction:
+    """Represents a rule action.
+
+    Attributes
+    -----------
+    type: :class:`AutoModRuleActionType`
+        The type of action to take.
+    channel_id: Optional[int]
```
NCPlayz, line 65:
```suggestion
    channel_id: Optional[:class:`int`]
```

discord/automod.py
```diff
@@ -0,0 +1,413 @@
...
+        The channel to send the alert message to, if any.
+    duration_seconds: Optional[int]
```
NCPlayz, line 67:
```suggestion
    duration_seconds: Optional[:class:`int`]
```

discord/automod.py
NCPlayz: We use edit everywhere else so it mostly makes sense to use it, in my opinion

discord/automod.py
```diff
@@ -0,0 +1,413 @@
...
+        exempt_roles: Optional[List[Role]] = None,
+        exempt_channels: Optional[List[Union[GuildChannel, Thread]]] = None,
+    ) -> Self:
+        """|coro|
+
+        This method will alter an existing auto moderation rule.
+
+        Parameters
+        -----------
+        name: Optional[:class:`str`]
```
NCPlayz, line 273: Since these are edit parameters I think it would be better to say `The new X` instead of `, if any` and for parameters that have `Optional` `Could be None to remove X`

discord/automod.py
```diff
@@ -0,0 +1,413 @@
...
+    async def edit(
+        self,
+        *,
+        name: Optional[str] = None,
+        event_type: Optional[AutoModRuleEventType] = None,
+        actions: List[AutoModRuleAction],
+        enabled: Optional[bool] = None,
+        exempt_roles: Optional[List[Role]] = None,
+        exempt_channels: Optional[List[Union[GuildChannel, Thread]]] = None,
+    ) -> Self:
```
NCPlayz, line 266: Does this and delete support reason?

discord/automod.py
```diff
@@ -0,0 +1,413 @@
...
+            self.id,
+            name=name,
+            event_type=event_type,
+            actions=transformed_actions,
+            enabled=enabled,
+            exempt_roles=exempt_roles,
+            exempt_channels=exempt_channels,
+        )
+
+        return self.__class__(data=data, guild=self.guild, state=self._state)
```
NCPlayz, line 313: Explicit constructor?

discord/guild.py
```diff
@@ -3833,3 +3835,97 @@ async def change_voice_state(
         ws = self._state._get_websocket(self.id)
         channel_id = channel.id if channel else None
         await ws.voice_state(self.id, channel_id, self_mute, self_deaf)
+
+    async def fetch_automod_rule(self, automod_rule_id: int) -> AutoModRule:
```
NCPlayz, line 3839:
```suggestion
    async def fetch_automod_rule(self, automod_rule_id: int, /) -> AutoModRule:
```

discord/automod.py
AbstractUmbra: How exactly would you handle this one?

discord/automod.py
AbstractUmbra: Hilariously, none of them do. I've removed it from create since I assumed it would have it.

discord/automod.py
NCPlayz: I don't see anything inheriting this (unless we're allowing users to inherit it) so can't we just `AutoModRule(...)`?

AbstractUmbra: Huh, I thought that would NameError. Good to know, thanks.

discord/automod.py
```diff
@@ -0,0 +1,411 @@
...
+        self.rule_trigger_type: AutoModRuleTriggerType = try_enum(AutoModRuleTriggerType, data['rule_trigger_type'])
+        self.user_id: int = int(data['user_id'])
+        self.alert_system_message_id: Optional[int] = utils._get_as_snowflake(data, 'alert_system_message_id')
+        self.content: str = data['content']
+        self.matched_keyword: Optional[str] = data['matched_keyword']
+        self.matched_content: Optional[str] = data['matched_content']
+
+    @property
+    def guild(self) -> Guild:
+        """Optional[:class:`Guild`]: The guild this rule was executed in."""
```
diceroll123, line 384: The previous and next lines say this should never be None

AbstractUmbra: Good catch, thanks

lol219:
Not bad

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+        if self.type is AutoModRuleTriggerType.keyword and self.keyword_filter is None:
+            raise ValueError('keyword_filter must be set if type is keyword.')
+        elif self.type is AutoModRuleTriggerType.keyword_preset and self.presets is None:
+            raise ValueError('presets must be set if type is keyword_preset.')
+
+    def __repr__(self) -> str:
+        return f"<AutoModTrigger type={self.type}>"
+
+    @classmethod
+    def from_data(cls: Type[Self], type: int, data: Optional[AutoModerationTriggerMetadataPayload]) -> Self:
```
Rapptz, line 150:
```suggestion
    def from_data(cls, type: int, data: Optional[AutoModerationTriggerMetadataPayload]) -> Self:
```

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+            return cls(type=type_, keyword_filter=data['keyword_filter'])  # type: ignore # unable to typeguard due to outer payload
+        elif type_ is AutoModRuleTriggerType.keyword_preset:
+            return cls(type=type_, presets=AutoModPresets._from_value(data['presets']))  # type: ignore # unable to typeguard due to outer payload
+        return cls(type=type_)
+
+    def to_metadata_dict(self) -> AutoModerationTriggerMetadataPayload:
+        if self.type is AutoModRuleTriggerType.keyword:
+            return {'keyword_filter': self.keyword_filter}  # type: ignore # guarded by the type check
+        elif self.type is AutoModRuleTriggerType.keyword_preset:
+            return {'presets': self.presets.value}  # type: ignore # guarded by the type check
```
Rapptz, line 162:
```suggestion
        if self.keyword_filter is not None:
            return {'keyword_filter': self.keyword_filter}
        elif self.presets is not None:
            return {'presets': self.presets.value}
```

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+        The guild the rule is for.
+    name: :class:`str`
+        The name of the rule.
+    creator_id: :class:`int`
+        The ID of the user that created the rule.
+    trigger: :class:`AutoModTrigger`
+        The rule's trigger.
+    enabled: :class:`bool`
+        Whether the rule is enabled.
+    raw_exempt_roles: List[:class:`int`]
```
Rapptz, line 184:
```suggestion
    exempt_role_ids: List[:class:`int`]
```

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+        The IDs of the roles that are exempt from the rule.
+    raw_exempt_channels: List[:class:`int`]
```
Rapptz, line 186:
```suggestion
    exempt_channel_ids: List[:class:`int`]
```

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+        '_state',
+        'id',
+        'guild',
+        'name',
+        'creator_id',
+        'event_type',
+        'trigger',
+        'enabled',
+        'raw_exempt_roles',
+        'raw_exempt_channels',
```
Rapptz, line 202:
```suggestion
        'exempt_role_ids',
        'exempt_channel_ids',
```

I know this was done for consistency with `Message.raw_` methods but the context is different.

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+            'exempt_roles': [str(role_id) for role_id in self.raw_exempt_roles],
+            'exempt_channels': [str(channel_id) for channel_id in self.raw_exempt_channels],
+        }  # type: ignore # trigger types break the flow here.
+
+        return ret
+
+    @property
+    def creator(self) -> Optional[User]:
+        """Optional[:class:`User`]: The user that created this rule."""
+        return self._state.get_user(self.creator_id)
```
Rapptz, line 242: Surely this should have an upgrade flow for Member considering there's a guild attached?

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+
+    @property
+    def exempt_roles(self) -> List[Role]:
+        """List[:class:`Role`]: The roles that are exempt from this rule."""
+        it = [role for role in (self.guild.get_role(role_id) for role_id in self.raw_exempt_roles) if role is not None]
+        return utils._unique(it)
```
Rapptz, line 248:
```suggestion
        result = []
        get_role = self.guild.get_role
        for role_id in self.exempt_role_ids:
            role = get_role(role_id)
            if role is not None:
                result.append(role)

        return result
```

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+        self._state: ConnectionState = state
+        self.guild: Guild = guild
+        self.id: int = int(data['id'])
+        self.name: str = data['name']
+        self.creator_id = int(data['creator_id'])
+        self.event_type: AutoModRuleEventType = try_enum(AutoModRuleEventType, data['event_type'])
+        self.trigger: AutoModTrigger = AutoModTrigger.from_data(data['trigger_type'], data=data.get('trigger_metadata'))
+        self.enabled: bool = data['enabled']
+        self.raw_exempt_roles: List[int] = [int(role_id) for role_id in data['exempt_roles']]
+        self.raw_exempt_channels: List[int] = [int(channel_id) for channel_id in data['exempt_channels']]
```
Rapptz, line 216:
```suggestion
        self.exempt_role_ids: List[int] = [int(role_id) for role_id in data['exempt_roles']]
        self.exempt_channel_ids: List[int] = [int(channel_id) for channel_id in data['exempt_channels']]
```

I wonder if casting these into a set makes more sense than list here since it allows for quicker O(1) checking. It also allows there to be a quick `AutoModRule.is_exempt` utility method.

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+    @property
+    def exempt_roles(self) -> List[Role]:
+        """List[:class:`Role`]: The roles that are exempt from this rule."""
+        it = [role for role in (self.guild.get_role(role_id) for role_id in self.raw_exempt_roles) if role is not None]
+        return utils._unique(it)
+
+    @property
+    def exempt_channels(self) -> List[Union[GuildChannel, Thread]]:
+        """List[Union[:class:`abc.GuildChannel`, :class:`Thread`]]: The channels that are exempt from this rule."""
+        it = filter(None, map(self.guild._resolve_channel, self.raw_exempt_channels))
```
Rapptz, line 253:
```suggestion
        it = filter(None, map(self.guild._resolve_channel, self.exempt_channel_ids))
```

discord/automod.py
Rapptz: How did you check if it was unsupported? The client notoriously doesn't render reasons despite it being supported by the API.

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+
+    async def edit(
+        self,
+        *,
+        name: Optional[str] = None,
+        event_type: Optional[AutoModRuleEventType] = None,
+        actions: Optional[List[AutoModRuleAction]] = None,
+        enabled: Optional[bool] = None,
+        exempt_roles: Optional[List[Role]] = None,
+        exempt_channels: Optional[List[Union[GuildChannel, Thread]]] = None,
```
Rapptz, line 269: Unless `None` is supported as a way of "clearing" or has specific meaning, none of these should be `Optional` and should be using `MISSING` instead.

Also `exempt_roles` and `exempt_channels` should be taking `Sequence[abc.Snowflake]` instead.

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+            The new roles to exempt from the rule.
+        exempt_channels: Optional[List[Union[:class:`abc.GuildChannel`], :class:`Thread`]]
+            The new channels to exempt from the rule.
+
+        Raises
+        -------
+        :exc:`Forbidden`
+            You are not authorised to perform this action.
+        :exc:`NotFound`
+            The rule was not found.
```
Rapptz, line 295:
```suggestion
        Forbidden
            You do not have permissions to edit the rule.
        HTTPException
            Editing the rule failed.
```

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+            exempt_roles=exempt_roles,
+            exempt_channels=exempt_channels,
+        )
+
+        return AutoModRule(data=data, guild=self.guild, state=self._state)
+
+    async def delete(self) -> None:
+        """|coro|
+
+        This method will delete the auto moderation rule.
```
Rapptz, line 322:
```suggestion
        Deletes the auto moderation rule.
```

What permissions are needed to delete the rule? This should be documented.

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+        name: Optional[str] = None,
+        event_type: Optional[AutoModRuleEventType] = None,
+        actions: Optional[List[AutoModRuleAction]] = None,
+        enabled: Optional[bool] = None,
+        exempt_roles: Optional[List[Role]] = None,
+        exempt_channels: Optional[List[Union[GuildChannel, Thread]]] = None,
+    ) -> Self:
+        """|coro|
+
+        This method will edit an existing auto moderation rule.
```
Rapptz, line 273:
```suggestion
        Edit an existing auto moderation rule.
```

What permissions are needed to edit the rule? This should be documented.

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+        """|coro|
+
+        This method will delete the auto moderation rule.
+
+        Raises
+        -------
+        :exc:`Forbidden`
+            You are not authorised to perform this action.
+        :exc:`NotFound`
+            The rule was not found.
```
Rapptz, line 329:
```suggestion
        Forbidden
            You do not have permissions to delete the rule.
        HTTPException
            Deleting the rule failed.
```

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+    trigger: :class:`AutoModTrigger`
+        The rule's trigger.
+    enabled: :class:`bool`
+        Whether the rule is enabled.
+    raw_exempt_roles: List[:class:`int`]
+        The IDs of the roles that are exempt from the rule.
+    raw_exempt_channels: List[:class:`int`]
+        The IDs of the channels that are exempt from the rule.
+
+    .. versionadded:: 2.0
```
Rapptz, line 189: This goes before Attributes.

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+            You are not authorised to perform this action.
+        :exc:`NotFound`
+            The rule was not found.
+        """
+        await self._state.http.delete_auto_moderation_rule(self.guild.id, self.id)
+
+
+class AutoModRuleExecution:
+    """Represents an auto moderation rule execution.
+
```
Rapptz, line 336:
```suggestion
    """Represents an auto moderation rule execution.

    .. versionadded:: 2.0

```

What, exactly, is an auto moderation rule execution? I feel this needs to be described.

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+    rule_id: :class:`int`
+        The ID of the rule that was triggered.
+    rule_trigger_type: :class:`AutoModRuleTriggerType`
+        The trigger type of the rule that was triggered.
+    user_id: :class:`int`
+        The ID of the user that triggered the rule.
+    alert_system_message_id: Optional[:class:`int`]
+        The ID of the system message that was sent to the predefined alert channel.
+    content: :class:`str`
+        The content of the message that triggered the rule.
```
Rapptz, line 352: Probably mention that this requires the message content intent otherwise it'll always be an empty string.

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+        self.rule_id: int = int(data['rule_id'])
+        self.rule_trigger_type: AutoModRuleTriggerType = try_enum(AutoModRuleTriggerType, data['rule_trigger_type'])
+        self.user_id: int = int(data['user_id'])
+        self.alert_system_message_id: Optional[int] = utils._get_as_snowflake(data, 'alert_system_message_id')
+        self.content: str = data['content']
+        self.matched_keyword: Optional[str] = data['matched_keyword']
+        self.matched_content: Optional[str] = data['matched_content']
+
+    def __repr__(self) -> str:
+        return f"<AutoModRuleExecution rule_id={self.rule_id} action={self.action!r}>"
```
Rapptz, line 389:
```suggestion
        return f'<AutoModRuleExecution rule_id={self.rule_id} action={self.action!r}>'
```

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+
+    @property
+    def guild(self) -> Guild:
+        """:class:`Guild`: The guild this rule was executed in."""
+        return self._state._get_guild(self._guild_id)  # type: ignore # this should never be None here
```
Rapptz, line 394:
```suggestion
        return self._state._get_or_create_unavailable_guild(self._guild_id)
```

This new method is type safe and returns an unavailable Guild instead of None.

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+        'alert_system_message_id',
+        'content',
+        'matched_keyword',
+        'matched_content',
+    )
+
+    def __init__(self, *, data: AutoModerationActionExecutionPayload, state: ConnectionState) -> None:
+        self._state: ConnectionState = state
+        self._guild_id: int = int(data['guild_id'])
+        self._channel_id: Optional[int] = utils._get_as_snowflake(data, 'channel_id')
```
Rapptz, line 377: Why are these two private?

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+    def guild(self) -> Guild:
+        """:class:`Guild`: The guild this rule was executed in."""
+        return self._state._get_guild(self._guild_id)  # type: ignore # this should never be None here
+
+    @property
+    def channel(self) -> Optional[Union[GuildChannel, Thread]]:
+        """Optional[Union[:class:`abc.GuildChannel`, :class:`Thread`]]: The channel this rule was executed in."""
+        if self._channel_id:
+            return self.guild and self.guild.get_channel(self._channel_id)
+        return None
```
Rapptz, line 401:
```suggestion
        if self._channel_id:
            return self.guild.get_channel(self._channel_id)
        return None
```

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+
+    @property
+    def member(self) -> Optional[Member]:
+        """Optional[:class:`Member`]: The member this rule was executed on/who triggered this rule."""
+        return self.guild and self.guild.get_member(self.user_id)
```
Rapptz, line 406:
```suggestion
        return self.guild.get_member(self.user_id)
```

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+
+    async def fetch_rule(self) -> AutoModRule:
+        """|coro|
+
+        This method will fetch the rule that was executed.
```
Rapptz, line 411:
```suggestion
        Fetch the rule that was executed.
```

Permissions, raises? The documentation for this needs to be fleshed out.

discord/enums.py
Rapptz, line 356: Maybe these should be shortened for consistency.

```suggestion
    automod_rule_create           = 140
    automod_rule_update           = 141
    automod_rule_delete           = 142
    automod_block_message         = 143
```

discord/flags.py
```diff
@@ -1104,6 +1109,30 @@ def guild_scheduled_events(self):
         """
         return 1 << 16
 
+    @flag_value
+    def auto_moderation_configuration(self):
+        """:class:`bool`: Whether auto moderation configuration related events are enabled.
+
+        This corresponds to the following events:
```
Rapptz, line 1116:
```suggestion
        This corresponds to the following events:

```

discord/flags.py
```diff
@@ -1436,3 +1465,110 @@ class ChannelFlags(BaseFlags):
     def pinned(self):
         """:class:`bool`: Returns ``True`` if the thread is pinned to the forum channel."""
         return 1 << 1
+
+
+class ArrayFlags(BaseFlags):
```
Rapptz, line 1470: I'm not really big on the changes done to flags.py, sorry. Please keep `.value` as `.value` and do not hide it under `_value`. 

You can keep ArrayFlags mostly as-is but just add a method for the actual AutoModPresets class to transform it into an actual list instead of modifying everything to fit the change.

discord/flags.py
```diff
@@ -1436,3 +1465,110 @@ class ChannelFlags(BaseFlags):
...
+        .. describe:: hash(x)
+
+            Return the flag's hash.
+        .. describe:: iter(x)
+
+            Returns an iterator of ``(name, value)`` pairs. This allows it
+            to be, for example, constructed as a dict or a list of pairs.
+            Note that aliases are not shown.
+
+    .. versionadded:: 2.0
```
Rapptz, line 1536: This goes above the operators container.

discord/guild.py
```diff
@@ -3833,3 +3835,93 @@ async def change_voice_state(
         ws = self._state._get_websocket(self.id)
         channel_id = channel.id if channel else None
         await ws.voice_state(self.id, channel_id, self_mute, self_deaf)
+
+    async def fetch_automod_rule(self, automod_rule_id: int, /) -> AutoModRule:
+        """|coro|
+
+        This method will fetch an active automod rule from the guild.
```
Rapptz, line 3842:
```suggestion
        Fetch an active automod rule from the guild.
```

Needs versionadded, permissions required, and raises section.

discord/guild.py
```diff
@@ -3833,3 +3835,93 @@ async def change_voice_state(
...
+        """
+
+        data = await self._state.http.get_auto_moderation_rule(self.id, automod_rule_id)
+
+        return AutoModRule(data=data, guild=self, state=self._state)
+
+    async def fetch_automod_rules(self) -> List[AutoModRule]:
+        """|coro|
+
+        This method will fetch all automod rules from the guild.
```
Rapptz, line 3862:
```suggestion
        Fetch all automod rules from the guild.
```

Needs versionadded, permissions required, and raises section.

discord/guild.py
```diff
@@ -3833,3 +3835,93 @@ async def change_voice_state(
...
+        name: str,
+        event_type: AutoModRuleEventType,
+        trigger: AutoModTrigger,
+        actions: List[AutoModRuleAction],
+        enabled: Optional[bool] = None,
+        exempt_roles: Optional[List[Role]] = None,
+        exempt_channels: Optional[List[Union[GuildChannel, Thread]]] = None,
+    ) -> AutoModRule:
+        """|coro|
+        This method will create an automod rule in the guild.
```
Rapptz, line 3885:
```suggestion

        Create an automod rule in the guild.
```

Needs the versionadded directive and permissions required.

discord/guild.py
```diff
@@ -3833,3 +3835,93 @@ async def change_voice_state(
...
+    async def create_automod_rule(
+        self,
+        *,
+        name: str,
+        event_type: AutoModRuleEventType,
+        trigger: AutoModTrigger,
+        actions: List[AutoModRuleAction],
+        enabled: Optional[bool] = None,
+        exempt_roles: Optional[List[Role]] = None,
+        exempt_channels: Optional[List[Union[GuildChannel, Thread]]] = None,
```
Rapptz, line 3882: Unless `None` is valid these should be using `MISSING`.

`exempt_roles` and` exempt_channels` needs to take a `Sequence[abc.Snowflake]`.

discord/guild.py
```diff
@@ -3833,3 +3835,93 @@ async def change_voice_state(
...
+            Discord will default to ``False``.
+        exempt_roles: Optional[List[:class:`Role`]]
+            A list of roles that will be exempt from the automod rule.
+        exempt_channels: Optional[List[Union[:class:`abc.GuildChannel`, :class:`Thread`]]]
+            A list of channels that will be exempt from the automod rule.
+
+        Raises
+        -------
+        :exc:`Forbidden`
+            You do not have permissions to create an automod rule.
```
Rapptz, line 3908:
```suggestion
        Forbidden
            You do not have permissions to create an auto moderation rule.
        HTTPException
            Creating the auto moderation rule failed.
```

discord/state.py
```diff
@@ -1079,6 +1082,46 @@ def parse_guild_stickers_update(self, data: gw.GuildStickersUpdateEvent) -> None
...
 
+    def parse_auto_moderation_rule_create(self, data: AutoModerationRule) -> None:
+        guild = self._get_guild(int(data['guild_id']))
+        if guild is None:
+            _log.debug('AUTO_MODERATION_RULE_CREATE referencing an unknown guild ID: %s. Discarding.', data['guild_id'])
+            return
+
+        rule = AutoModRule(data=data, guild=guild, state=self)
+
+        self.dispatch('auto_moderation_rule_create', rule)
```
Rapptz, line 1093:
```suggestion
        self.dispatch('automod_rule_create', rule)
```

Might as well be consistent.

discord/state.py
```diff
@@ -1079,6 +1082,46 @@ def parse_guild_stickers_update(self, data: gw.GuildStickersUpdateEvent) -> None
...
+
+    def parse_auto_moderation_rule_update(self, data: AutoModerationRule) -> None:
+        guild = self._get_guild(int(data['guild_id']))
+        if guild is None:
+            _log.debug('AUTO_MODERATION_RULE_UPDATE referencing an unknown guild ID: %s. Discarding.', data['guild_id'])
+            return
+
+        rule = AutoModRule(data=data, guild=guild, state=self)
+
+        self.dispatch('auto_moderation_rule_update', rule)
```
Rapptz, line 1103:
```suggestion
        self.dispatch('automod_rule_update', rule)
```

discord/state.py
```diff
@@ -1079,6 +1082,46 @@ def parse_guild_stickers_update(self, data: gw.GuildStickersUpdateEvent) -> None
...
+
+    def parse_auto_moderation_rule_delete(self, data: AutoModerationRule) -> None:
+        guild = self._get_guild(int(data['guild_id']))
+        if guild is None:
+            _log.debug('AUTO_MODERATION_RULE_DELETE referencing an unknown guild ID: %s. Discarding.', data['guild_id'])
+            return
+
+        rule = AutoModRule(data=data, guild=guild, state=self)
+
+        self.dispatch('auto_moderation_rule_delete', rule)
```
Rapptz, line 1113:
```suggestion
        self.dispatch('automod_rule_delete', rule)
```

discord/state.py
```diff
@@ -1079,6 +1082,46 @@ def parse_guild_stickers_update(self, data: gw.GuildStickersUpdateEvent) -> None
...
+
+    def parse_auto_moderation_action_execution(self, data: AutoModerationActionExecution) -> None:
+        guild = self._get_guild(int(data['guild_id']))
+        if guild is None:
+            _log.debug('AUTO_MODERATION_ACTION_EXECUTION referencing an unknown guild ID: %s. Discarding.', data['guild_id'])
+            return
+
+        execution = AutoModRuleExecution(data=data, state=self)
+
+        self.dispatch('auto_moderation_action_execution', execution)
```
Rapptz, line 1123:
```suggestion
        self.dispatch('automod_action_execution', execution)
```

I'm still not big on the name but whatever.

docs/api.rst
```diff
@@ -1300,6 +1300,41 @@ Voice
     :param after: The voice state after the changes.
     :type after: :class:`VoiceState`
 
+AutoMod
+~~~~~~~
+
+.. function:: on_auto_moderation_rule_create(rule)
```
Rapptz, line 1306:
```suggestion
.. function:: on_automod_rule_create(rule)
```

docs/api.rst
```diff
@@ -1300,6 +1300,41 @@ Voice
...
+
+    Called when a :class:`AutoModRule` is created.
+
+    This requires :attr:`Intents.auto_moderation_configuration` to be enabled.
```
Rapptz, line 1310:
```suggestion
    This requires :attr:`Intents.auto_moderation_configuration` to be enabled.

    .. versionadded:: 2.0

```

docs/api.rst
```diff
@@ -1300,6 +1300,41 @@ Voice
...
+    :param rule: The rule that was created.
+    :type rule: :class:`AutoModRule`
+
+.. function:: on_auto_moderation_rule_update(rule)
```
Rapptz, line 1314:
```suggestion
.. function:: on_automod_rule_update(rule)
```

docs/api.rst
```diff
@@ -1300,6 +1300,41 @@ Voice
...
+
+    Called when a :class:`AutoModRule` is updated.
+
+    This requires :attr:`Intents.auto_moderation_configuration` to be enabled.
```
Rapptz, line 1318:
```suggestion
    This requires :attr:`Intents.auto_moderation_configuration` to be enabled.

    .. versionadded:: 2.0

```

docs/api.rst
```diff
@@ -1300,6 +1300,41 @@ Voice
...
+    :param rule: The rule that was updated.
+    :type rule: :class:`AutoModRule`
+
+.. function:: on_auto_moderation_rule_delete(rule)
```
Rapptz, line 1322:
```suggestion
.. function:: on_automod_rule_delete(rule)
```

docs/api.rst
```diff
@@ -1300,6 +1300,41 @@ Voice
...
+
+    Called when a :class:`AutoModRule` is deleted.
+
+    This requires :attr:`Intents.auto_moderation_configuration` to be enabled.
```
Rapptz, line 1326:
```suggestion
    This requires :attr:`Intents.auto_moderation_configuration` to be enabled.

    .. versionadded:: 2.0

```

docs/api.rst
```diff
@@ -1300,6 +1300,41 @@ Voice
...
+    :param rule: The rule that was deleted.
+    :type rule: :class:`AutoModRule`
+
+.. function:: on_auto_moderation_action_execution(execution)
+
+    Called when a :class:`AutoModRuleExecution` is created/performed.
+
+    This requires :attr:`Intents.auto_moderation_execution` to be enabled.
```
Rapptz, line 1334:
```suggestion
    This requires :attr:`Intents.auto_moderation_execution` to be enabled.

    .. versionadded:: 2.0

```

Rapptz (changes requested):
Good work. This is my first pass at reviewing this. 

This is missing documentation for the audit logs in api.rst. Please revert a lot of the flags.py changes with `_value`.

`AutoModRuleAction` and `AutoModTrigger` are called Data Classes because they can be constructed by users. I think this construction could be made easier to prevent pit of failures but the current design is pretty error prone since you have to keep a few things in mind rather than pushing the user to do the right thing by default.

I didn't do suggestions for all changes so you'll have to do another pass to ensure that things don't raise exceptions, sorry.

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+    """Represents a rule action.
+
+    Attributes
+    -----------
+    type: :class:`AutoModRuleActionType`
+        The type of action to take.
+    channel_id: Optional[:class:`int`]
+        The channel to send the alert message to, if any.
+    duration_seconds: Optional[:class:`int`]
+        The duration of the timeout to apply, if any.
```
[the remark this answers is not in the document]
AbstractUmbra: For this you'd rather the end user passed a `timedelta`? Or either the `int` or `timedelta`?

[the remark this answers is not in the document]
Rapptz: timedelta

discord/automod.py
```diff
@@ -0,0 +1,450 @@
...
+    """
+
+    __slots__ = ('type', 'channel_id', 'duration')
+
+    def __init__(self, *, channel_id: Optional[int] = None, duration: Optional[datetime.timedelta] = None) -> None:
+        self.channel_id: Optional[int] = channel_id
+        self.duration: Optional[datetime.timedelta] = duration
+        if channel_id and duration:
+            raise ValueError("Please provide only one of ``channel`` or ``duration``")
```
apple502j, line 78: Shouldn't this use singlequote?

discord/automod.py
```diff
@@ -0,0 +1,450 @@
...
+    def __init__(
+        self,
+        *,
+        keyword_filter: Optional[List[str]] = None,
+        presets: Optional[AutoModPresets] = None,
+    ) -> None:
+        self.keyword_filter: Optional[List[str]] = keyword_filter
+        self.presets: Optional[AutoModPresets] = presets
+        if keyword_filter and presets:
+            raise ValueError("Please pass only one of keyword_filter or presets.")
```
apple502j, line 136: Same here

discord/automod.py
```diff
@@ -0,0 +1,459 @@
...
+            'trigger_metadata': self.trigger.to_metadata_dict(),
+            'actions': [action.to_dict() for action in self.actions],
+            'enabled': self.enabled,
+            'exempt_roles': [str(role_id) for role_id in self.exempt_role_ids],
+            'exempt_channels': [str(channel_id) for channel_id in self.exempt_channel_ids],
+        }  # type: ignore # trigger types break the flow here.
+
+        return ret
+
+    @property
```
apple502j, line 232: I think the properties here should use cached_slot_property.

discord/automod.py
```diff
@@ -0,0 +1,459 @@
...
+        it = filter(None, map(self.guild._resolve_channel, self.exempt_channel_ids))
+        return utils._unique(it)
+
+    @property
+    def actions(self) -> List[AutoModRuleAction]:
+        """List[:class:`AutoModRuleAction`]: The actions that are taken when this rule is triggered."""
+        return [AutoModRuleAction.from_data(action) for action in self._actions]
+
+    def is_exempt(self, object: Union[GuildChannel, Thread, Role], /) -> bool:
+        """This method will check if an object is exempt from the automod rule.
```
apple502j, line 261: Method description should start with a verb.

discord/automod.py
```diff
@@ -0,0 +1,459 @@
...
+
+        Parameters
+        -----------
+        object: Union[:class:`abc.GuildChannel`, :class:`Thread`, :class:`Role`]
+            The object which we are checking.
```
apple502j, line 266: "The object to check"? Docs usually avoid first-person pronouns.

discord/automod.py
```diff
@@ -0,0 +1,459 @@
...
+        return object.id in self.exempt_channel_ids or object.id in self.exempt_role_ids
+
+    async def edit(
+        self,
+        *,
+        name: str = MISSING,
+        event_type: AutoModRuleEventType = MISSING,
+        actions: List[AutoModRuleAction] = MISSING,
+        enabled: bool = MISSING,
+        exempt_roles: List[Role] = MISSING,
```
apple502j, line 281: Why does this take List and not Sequence, unlike `exempt_channels`?

discord/automod.py
```diff
@@ -0,0 +1,459 @@
...
+        exempt_channels: Sequence[Snowflake] = MISSING,
+        reason: str = MISSING,
+    ) -> Self:
+        """|coro|
+
+        Edit an existing auto moderation rule. All parameters of this method are optional.
```
apple502j, line 287: "Edit*s* this auto moderation rule" (see other edit methods), no need to mention params being optional

discord/automod.py
```diff
@@ -0,0 +1,459 @@
...
+            exempt_channels=exempt_channels,
+            reason=reason,
+        )
+
+        return AutoModRule(data=data, guild=self.guild, state=self._state)
+
+    async def delete(self, reason: str = MISSING) -> None:
+        """|coro|
+
+        Delete the auto moderation rule.
```
apple502j, line 339: Delete*s*

discord/automod.py
```diff
@@ -0,0 +1,459 @@
...
+            actions=transformed_actions,
+            enabled=enabled,
+            exempt_roles=exempt_roles,
+            exempt_channels=exempt_channels,
+            reason=reason,
+        )
+
+        return AutoModRule(data=data, guild=self.guild, state=self._state)
+
+    async def delete(self, reason: str = MISSING) -> None:
```
apple502j, line 336: `reason` should be keyword-only.

discord/automod.py
```diff
@@ -0,0 +1,459 @@
...
+
+class AutoModAction:
+    """Represents an action that was taken as the result of a moderation rule.
+
+    Attributes
+    -----------
+    action: :class:`AutoModRuleAction`
+        The action that was taken.
+    message_id: Optional[:class:`int`]
+        The message ID that was triggered the reaction.
```
apple502j, line 364: I think "was" is unnecessary here? Also this isn't a reaction.

discord/automod.py
```diff
@@ -0,0 +1,459 @@
...
+
+    @property
+    def member(self) -> Optional[Member]:
+        """Optional[:class:`Member`]: The member this action was taken against /who triggered this rule."""
+        return self.guild.get_member(self.user_id)
+
+    async def fetch_rule(self) -> AutoModRule:
+        """|coro|
+
+        Fetch the rule whose from which the action was taken.
```
apple502j, line 441: "Fetches the rule whose action was taken"?

discord/guild.py
```diff
@@ -3833,3 +3835,122 @@ async def change_voice_state(
         ws = self._state._get_websocket(self.id)
         channel_id = channel.id if channel else None
         await ws.voice_state(self.id, channel_id, self_mute, self_deaf)
+
+    async def fetch_automod_rule(self, automod_rule_id: int, /) -> AutoModRule:
+        """|coro|
+
+        Fetch an active automod rule from the guild.
```
apple502j, line 3842: Fetches?

discord/guild.py
```diff
@@ -3833,3 +3835,122 @@ async def change_voice_state(
...
+        """
+
+        data = await self._state.http.get_auto_moderation_rule(self.id, automod_rule_id)
+
+        return AutoModRule(data=data, guild=self, state=self._state)
+
+    async def fetch_automod_rules(self) -> List[AutoModRule]:
+        """|coro|
+
+        Fetch all automod rules from the guild.
```
apple502j, line 3871: Fetches

discord/guild.py
```diff
@@ -3833,3 +3835,122 @@ async def change_voice_state(
...
+
+    async def create_automod_rule(
+        self,
+        *,
+        name: str,
+        event_type: AutoModRuleEventType,
+        trigger: AutoModTrigger,
+        actions: List[AutoModRuleAction],
+        enabled: bool = MISSING,
+        exempt_roles: List[Role] = MISSING,
```
apple502j, line 3899: Shouldn't they take `Sequence`?

discord/guild.py
```diff
@@ -3833,3 +3835,122 @@ async def change_voice_state(
...
+        exempt_channels: List[Union[GuildChannel, Thread]] = MISSING,
+        reason: str = MISSING,
+    ) -> AutoModRule:
+        """|coro|
+
+        Create an automod rule in the guild.
```
apple502j, line 3905: Creates; "in the guild" unnecessary.

discord/guild.py
```diff
@@ -3833,3 +3835,122 @@ async def change_voice_state(
...
+
+        Requires the :attr:`Permissions.manage_guild` permission to use this.
```
apple502j, line 3907: You must have the...

apple502j (changes requested):

discord/flags.py
```diff
@@ -1436,3 +1462,105 @@ class ChannelFlags(BaseFlags):
...
+
+
+    .. container:: operations
+
+        .. describe:: x == y
+
+            Checks if two AutoMod preset flags are equal.
+        .. describe:: x != y
```
Willy-C, line 1490:
```suggestion

        .. describe:: x != y
```

discord/automod.py
```diff
@@ -0,0 +1,459 @@
...
+        actions: List[:class:`AutoModRuleAction`]
+            The new rule actions to update.
+        enabled: :class:`bool`
+            Whether the rule should be enabled or not.
+        exempt_roles: Sequence[:class:`abc.Snowflake`]
+            The new roles to exempt from the rule.
+        exempt_channels: Sequence[:class:`abc.Snowflake`]
+            The new channels to exempt from the rule.
+        reason: :class:`str`
+            The reason for updating this rule.
```
Willy-C, line 304:
```suggestion
            The reason for updating this rule. Shows up on the audit log.
```

discord/automod.py
```diff
@@ -0,0 +1,459 @@
...
+
+    async def delete(self, *, reason: str = MISSING) -> None:
+        """|coro|
+
+        Deletes the auto moderation rule.
+
+        Parameters
+        -----------
+        reason: :class:`str`
+            The reason for deleting this rule.
```
Willy-C, line 344:
```suggestion
            The reason for deleting this rule. Shows up on the audit log.
```

discord/guild.py
```diff
@@ -3833,3 +3835,122 @@ async def change_voice_state(
...
+            The actions that will be taken when the automod rule is triggered.
+        enabled: :class:`bool`
+            Whether the automod rule is enabled.
+            Discord will default to ``False``.
+        exempt_roles: Sequence[:class:`abc.Snowflake`]
+            A list of roles that will be exempt from the automod rule.
+        exempt_channels: Sequence[:class:`abc.Snowflake`]
+            A list of channels that will be exempt from the automod rule.
+        reason: :class:`str`
+            The reason for creating this automod rule.
```
Willy-C, line 3929:
```suggestion
            The reason for creating this automod rule. Shows up on the audit log.
```

Willy-C (commented):
To be consistent with other methods' `reason` documentation
image

discord/flags.py
```diff
@@ -1436,3 +1462,106 @@ class ChannelFlags(BaseFlags):
     def pinned(self):
         """:class:`bool`: Returns ``True`` if the thread is pinned to the forum channel."""
         return 1 << 1
+
+
+class ArrayFlags(BaseFlags):
+    @classmethod
+    def _from_value(cls: Type[Self], value: List[int]) -> Self:
+        self = cls.__new__(cls)
+        self.value = sum((1 << i - 1 for i in value))
```
Rapptz, line 1471:
```suggestion
        self.value = reduce(lambda a, b: a | (1 << b - 1), value)
```

discord/automod.py
```diff
@@ -0,0 +1,463 @@
...
+__all__ = (
+    'AutoModRuleAction',
+    'AutoModTrigger',
+    'AutoModRule',
+    'AutoModAction',
+)
+
+
+class AutoModRuleAction:
+    """Represents an auto moderation's rule action.
```
Rapptz, line 60: Missing versionadded.

discord/automod.py
```diff
@@ -0,0 +1,420 @@
...
+    Attributes
+    -----------
+    type: :class:`AutoModRuleTriggerType`
+        The type of trigger.
+    keyword_filter: Optional[List[:class:`str`]]
+        The list of strings that will trigger the keyword filter.
+    presets: Optional[:class:`AutoModPresets`]
+        The presets used with the preset keyword filter.
+
+    .. versionadded:: 2.0
```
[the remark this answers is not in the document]
Rapptz: Don't know why this was resolved since it wasn't fixed.

discord/automod.py
```diff
@@ -0,0 +1,463 @@
...
+            'exempt_roles': [str(role_id) for role_id in self.exempt_role_ids],
+            'exempt_channels': [str(channel_id) for channel_id in self.exempt_channel_ids],
+        }  # type: ignore # trigger types break the flow here.
+
+        return ret
+
+    @cached_slot_property('_cs_creator')
+    def creator(self) -> Optional[Member]:
+        """Optional[:class:`Member`]: The member that created this rule."""
+        return self.guild.get_member(self.creator_id)
```
Rapptz, line 239: I don't think this should be cached.

discord/automod.py
```diff
@@ -0,0 +1,463 @@
...
+        Parameters
+        -----------
+        object: Union[:class:`abc.GuildChannel`, :class:`Thread`, :class:`Role`]
+            The object to check.
+
+        Returns
+        --------
+        :class:`bool`
+        """
+        return object.id in self.exempt_channel_ids or object.id in self.exempt_role_ids
```
Rapptz, line 276:
```suggestion
    def is_exempt(self, obj: Snowflake, /) -> bool:
        """Check if an object is exempt from the automod rule.

        Parameters
        -----------
        obj: :class:`abc.Snowflake`
            The role, channel, or thread to check.

        Returns
        --------
        :class:`bool`
            Whether the object is exempt from the automod rule.
        """
        return obj.id in self.exempt_channel_ids or obj.id in self.exempt_role_ids
```

discord/automod.py
```diff
@@ -0,0 +1,463 @@
...
+        event_type: AutoModRuleEventType = MISSING,
+        actions: List[AutoModRuleAction] = MISSING,
+        enabled: bool = MISSING,
+        exempt_roles: Sequence[Snowflake] = MISSING,
+        exempt_channels: Sequence[Snowflake] = MISSING,
+        reason: str = MISSING,
+    ) -> Self:
+        """|coro|
+
+        Edits this auto moderation rule.
```
Rapptz, line 291: Needs documentation on permissions required.

discord/automod.py
```diff
@@ -0,0 +1,463 @@
...
+            exempt_channels=exempt_channels,
+            reason=reason,
+        )
+
+        return AutoModRule(data=data, guild=self.guild, state=self._state)
+
+    async def delete(self, *, reason: str = MISSING) -> None:
+        """|coro|
+
+        Deletes the auto moderation rule.
```
Rapptz, line 343: Needs documentation on permissions required.

discord/automod.py
```diff
@@ -0,0 +1,463 @@
...
+    content: :class:`str`
+        The content of the message that triggered the rule.
+    matched_keyword: Optional[:class:`str`]
+        The matched keyword from the triggering message.
+    matched_content: Optional[:class:`str`]
+        The matched content from the triggering message.
+
+
+    .. note::
+        The :attr:`content` requires the :attr:`Intents.message_content` or it will always return an empty string.
```
Rapptz, line 390: This shouldn't be an admonition and it should be located in the paragraph in the fields that it applies to (which now includes `matched_content`).

discord/automod.py
```diff
@@ -0,0 +1,463 @@
...
+        Forbidden
+            You do not have permissions to delete the rule.
+        HTTPException
+            Deleting the rule failed.
+        """
+        await self._state.http.delete_auto_moderation_rule(self.guild.id, self.id, reason=reason)
+
+
+class AutoModAction:
+    """Represents an action that was taken as the result of a moderation rule.
```
Rapptz, line 361: Missing versionadded.

discord/state.py
```diff
@@ -49,6 +49,7 @@
 
 import os
 
+
```
Rapptz, line 52:
```suggestion
```

discord/automod.py
```diff
@@ -0,0 +1,463 @@
...
+            return cls(duration=datetime.timedelta(seconds=duration_seconds))
+        elif data['type'] == AutoModRuleActionType.send_alert_message.value:
+            channel_id = int(data['metadata']['channel_id'])
+            return cls(channel_id=channel_id)
+        return cls()
+
+    def to_dict(self) -> Dict[str, Any]:
+        ret = {'type': self.type.value, 'metadata': {}}
+        if self.type is AutoModRuleActionType.timeout:
+            ret['metadata'] = {'duration_seconds': self.duration.seconds}  # type: ignore # duration cannot be None here
```
Rapptz, line 104:
```suggestion
            ret['metadata'] = {'duration_seconds': int(self.duration.total_seconds())}  # type: ignore # duration cannot be None here
```

discord/automod.py
```diff
@@ -0,0 +1,463 @@
...
+    """Represents an auto moderation's rule action.
+
+    Attributes
+    -----------
+    type: :class:`AutoModRuleActionType`
+        The type of action to take.
+    channel_id: Optional[:class:`int`]
+        The ID of the channel to send the alert message to, if any.
+    duration: Optional[:class:`datetime.timedelta`]
+        The duration of the timeout to apply, if any.
```
Rapptz, line 69: Should probably mention this is a maximum of 28 days.

discord/http.py
```diff
@@ -2039,6 +2040,63 @@ def edit_application_command_permissions(
...
+            'exempt_channels',
+        )
+
+        payload = {k: v for k, v in payload.items() if k in valid_keys and v is not None}
+
+        return self.request(
+            Route('POST', '/guilds/{guild_id}/auto-moderation/rules', guild_id=guild_id), json=payload, reason=reason
+        )
+
+    def modify_auto_moderation_rule(
```
Rapptz, line 2071:
```suggestion
    def edit_auto_moderation_rule(
```

discord/automod.py
```diff
@@ -0,0 +1,463 @@
...
+        Returns
+        --------
+        :class:`AutoModRule`
+            The updated auto moderation rule.
+        """
+        transformed_actions = None
+        if actions:
+            transformed_actions = [action.to_dict() for action in actions]
+
+        data = await self._state.http.modify_auto_moderation_rule(
```
Rapptz, line 326:
```suggestion
        data = await self._state.http.edit_auto_moderation_rule(
```

Rapptz (changes requested):
Mostly just documentation nits now.

1. As I mentioned earlier, `AutoModRuleAction` and `AutoModTrigger` are both Data Classes and should be in that heading instead of Discord Models due to the fact they can be user created.
2. This is still missing the documentation on the audit log changes.

docs/api.rst
```diff
@@ -1300,6 +1300,53 @@ Voice
     :param after: The voice state after the changes.
     :type after: :class:`VoiceState`
 
+AutoMod
+~~~~~~~
```
Chiggy-Playz, line 1304: The automod category should be on the top rather than the bottom since the events are sorted in alphabetical order.

Heurican:
this is super cool and i can't wait for the docs to use it.

discord/automod.py
```diff
@@ -0,0 +1,468 @@
...
+        The guild the rule is for.
+    name: :class:`str`
+        The name of the rule.
+    creator_id: :class:`int`
+        The ID of the user that created the rule.
+    trigger: :class:`AutoModTrigger`
+        The rule's trigger.
+    enabled: :class:`bool`
+        Whether the rule is enabled.
+    exempt_role_ids: List[:class:`int`]
```
Rapptz, line 182:
```suggestion
    exempt_role_ids: Set[:class:`int`]
```

discord/automod.py
```diff
@@ -0,0 +1,468 @@
...
+        The IDs of the roles that are exempt from the rule.
+    exempt_channel_ids: List[:class:`int`]
```
Rapptz, line 184:
```suggestion
    exempt_channel_ids: Set[:class:`int`]
```

discord/automod.py
```diff
@@ -0,0 +1,468 @@
...
+        actions: List[AutoModRuleAction] = MISSING,
+        enabled: bool = MISSING,
+        exempt_roles: Sequence[Snowflake] = MISSING,
+        exempt_channels: Sequence[Snowflake] = MISSING,
+        reason: str = MISSING,
+    ) -> Self:
+        """|coro|
+
+        Edits this auto moderation rule.
+        Requires the :attr:`Permissions.manage_guild` to use.
```
Rapptz, line 295:
```suggestion

        You must have :attr:`~Permissions.manage_guild` to edit rules.
```

discord/automod.py
```diff
@@ -0,0 +1,468 @@
...
+            reason=reason,
+        )
+
+        return AutoModRule(data=data, guild=self.guild, state=self._state)
+
+    async def delete(self, *, reason: str = MISSING) -> None:
+        """|coro|
+
+        Deletes the auto moderation rule.
+        Requires the :attr:`Permissions.manage_guild` to use.
```
Rapptz, line 348:
```suggestion

        You must have :attr:`~Permissions.manage_guild` to delete rules.
```

discord/automod.py
```diff
@@ -0,0 +1,468 @@
...
+    def member(self) -> Optional[Member]:
+        """Optional[:class:`Member`]: The member this action was taken against /who triggered this rule."""
+        return self.guild.get_member(self.user_id)
+
+    async def fetch_rule(self) -> AutoModRule:
+        """|coro|
+
+        Fetch the rule whose action was taken.
+
+        You need the :attr:`Permissions.manage_guild` permission to use this.
```
Rapptz, line 452:
```suggestion
        You must have the :attr:`Permissions.manage_guild` permission to use this.
```

The "must have" wording is consistent (to my knowledge) across all methods in the library.

Rapptz (changes requested):
You requested help on what to do for the audit logs documentation so.

1. Every new key that's added in the `changes` dict is documented under `AuditLogDiff`
2. Every new action type is added to `AuditLogAction`
3. The action type mentions what the type of `target` is and what is expected.
4. Those new action types document every possible `AuditLogDiff` value you can receive from it to the best of your ability.
5. If that action type has a special `extras` then it gets documented there.

There are a lot of pre-existing examples you can just copy paste and adapt from.

lol219:
> You requested help on what to do for the audit logs documentation so.
> 
> 1. Every new key that's added in the `changes` dict is documented under `AuditLogDiff`
> 2. Every new action type is added to `AuditLogAction`
> 3. The action type mentions what the type of `target` is and what is expected.
> 4. Those new action types document every possible `AuditLogDiff` value you can receive from it to the best of your ability.
> 5. If that action type has a special `extras` then it gets documented there.
> 
> There are a lot of pre-existing examples you can just copy paste and adapt from.
> 
> 

Yeah

Heurican:
> After this is resolved I'll move to the latest commit discord.py because stable version is kinda outdated and customers want up-to-date bots

agreed- my current client is super picky- doesn't code but they're asking for all these complex and new features that haven't been implemented. i kind of mentally cringe every time i have to tell them it can't happen

muddyfish:
`AutoModRule.edit` is missing the `trigger_metadata` field.

discord/automod.py
```diff
@@ -334,6 +338,7 @@ async def edit(
             name=name,
             event_type=event_type,
             actions=transformed_actions,
+            trigger_metadata=trigger.to_metadata_dict(),
```
muddyfish, line 341: This crashes if you pass None/MISSING

muddyfish:
`edit` crashes when you call it without passing every parameter as it defaults to `MISSING` which isn't JSON serialisable.
```
TypeError: Object of type _MissingSentinel is not JSON serializable
```

discord/automod.py
```diff
@@ -328,21 +328,33 @@ async def edit(
...
+        if name:
+            payload['name'] = name
+
+        if event_type:
+            payload['event_type'] = event_type
+
+        if trigger:
+            payload['trigger_metadata'] = trigger.to_metadata_dict()
+
+        if enabled:
```
muddyfish, line 344: Does not trigger when enabled is False. Same for exempt_roles and exempt_channels

lol219: Works fine for me

lol219:
Nice

Rapptz (approved):
Good job.

PR #8173, opened 2022-06-22
Passing websocket RESUME data in and out of discord.py

mniip:
## Summary

Addresses https://github.com/Rapptz/discord.py/projects/3#card-60313188

We export two new types: `ResumeState` and `ShardedResumeState` that can be converted to and from `str`. `Client` methods such as `connect`, `start`, `run` accept and return these states. Additionally `Client.close` takes a `resumable` parameter to close websockets in a way that allows future RESUME.

I've also refactored the disconnection exception logic a little bit in hopes of deduplicating code between `Client` and `AutoShardedClient`.

## Checklist

    - Tested a little bit but not sure how to test this adequately. There are many code paths.
        - TODO
- [ ] This PR fixes an issue.


Armaan533:
What does this ``Resume`` state do in particular?
Is it like disconnect and reconnect to some saved state of bot?

Rapptz:
Surely it makes more sense for close to return Optional[ResumeState]?

Armaan533:
> Surely it makes more sense for close to return Optional[ResumeState]?

Yes ofcourse.
If it doesn't return how will we get the state out of the function.
Or maybe it stores the state somewhere else like in some binary file

mniip:
> Surely it makes more sense for close to return Optional[ResumeState]?

I don't think I agree here. I mean I see why you'd think that. You'd probably want a self-contained command that calls `Client.close()` and in the same procedure transfers control to a different process. However:

- Calling exec right in the middle of the command means no cleanup for anything in the bot process. Doing exec without fork makes more sense at the end of main. 
- The alternative implementation is to actually exit the process, and have some sort of watchdog restart it. Unless you're using `os._exit` (you really shouldn't) this will also reach the end of main.
- To load the previous bot's state you still need a gadget in the main file that does so. Makes sense to do the converse there as well.

I think the implicitly suggested solution here is that `connect` should save the resume state inside the `Client` object and then `close` can pick it up. So here's the *actual* reason I dislike the `.close` idea, and other ideas similar to it: 
- I think resume states should be part of the *control flow* inside the `connect` call, as opposed to actual *state* of a `Client` object. Superficially, you'd think that the resume state is just `Client.ws.resume_state()` which is sort-of always available to look at. But what is `Client.ws` really? On the same superficial level you'd say "the websocket connected to the gateway", but as you start to unravel what that really means in context of how the control flow of `connect` affects it, you end up with something more like "the websocket that is as of now connected to the gateway, unless we're currently reconnecting, in which case it is the most recently disconnected websocket that has been successfully handshaken at some point, unless this is our first time in which case it is None". With that in mind it still makes sense to have a `Client.ws` because it makes sense in the "steady state" when you're connected which is supposed to be 99% of the time. But the same reasoning cannot be extended to resume states because they are pretty much the opposite, being only relevant outside of the steady state.

> Or maybe it stores the state somewhere else like in some binary file
That is certainly outside the scope of the library.

Rapptz:
The `exec` functions should not be used at all with a Discord bot so that's a non-starter. Right now this interface makes it very awkward to use the feature and mostly unnatural since all usage of resume states, whether for a single shard or all of them, has to be entirely located in the blocking call of the bot which inhibits the ability to use these resume states with the AutoShardedClient without bringing down the entire bot.

xhamst:
> ## Summary
> 
> Addresses https://github.com/Rapptz/discord.py/projects/3#card-60313188
> 
> We export two new types: `ResumeState` and `ShardedResumeState` that can be converted to and from `str`. `Client` methods such as `connect`, `start`, `run` accept and return these states. Additionally `Client.close` takes a `resumable` parameter to close websockets in a way that allows future RESUME.
> 
> I've also refactored the disconnection exception logic a little bit in hopes of deduplicating code between `Client` and `AutoShardedClient`.
> 
> 
> 
> ## Checklist
> 
> 
> 
> - [ ] If code changes were made then they have been tested.
>     - Tested a little bit but not sure how to test this adequately. There are many code paths.
>     - [x] I have updated the documentation to reflect the changes.
> - [ ] This PR fixes an issue.
> - [x] This PR adds something new (e.g. new method or parameters).
> - [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
> - [ ] This PR is **not** a code change (e.g. documentation, README, ...)
>

PR #8175, opened 2022-06-22
Update documentation for `discord.Message`

emisdumb:
## Summary

Add a line to indicate that `Message.content` requires `Intents.message_content`.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -1311,7 +1311,8 @@ class Message(PartialMessage, Hashable):
         private channel or the user has the left the guild, then it is a :class:`User` instead.
     content: :class:`str`
         The actual contents of the message.
-        This requires :attr:`Intents.message_content` to be enabled.
+        If :attr:`Intents.message_content` is not enabled this will always be an empty string
+        unless the bot is mentioned or the message is a direct message.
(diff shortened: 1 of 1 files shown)


dolfies:
You probably mean `Intents.message_content`. Either way, this requirement is not just for `Message.content`.

emisdumb:
> You probably mean `Intents.message_content`. Either way, this requirement is not just for `Message.content`.

Fixed.

Rapptz (changes requested):
If you want to do this then you need to do it for all the other attributes that apply (like `attachments`, `components`, and `embeds`).

However, I don't really see the value of doing it since the dedicated intents documentation already does it and repeating it here doesn't seem that useful.

emisdumb:
> If you want to do this then you need to do it for all the other attributes that apply (like `attachments`, `components`, and `embeds`).
> 
> However, I don't really see the value of doing it since the dedicated intents documentation already does it and repeating it here doesn't seem that useful.

I've implemented the requested changes. I do find it useful that someone looking at the `discord.Message` documentation should have this information mentioned somewhere. Although, this is up to your discretion.

Rapptz (approved):
Thanks.

PR #8192, opened 2022-06-29
Support unicode emojis in PartialEmojiConverter

DenverCoder1:
## Summary

Previously only custom emojis worked with `PartialEmojiConverter`. An exception would be raised:

`discord.ext.commands.errors.PartialEmojiConversionFailure: Couldn't convert "⏹️" to PartialEmoji.`

This adds support for all kinds of unicode emojis.

Note: The regex is a bit loose but all emojis are covered by the regex.

Test code:

```py
@bot.command()
async def test(ctx: commands.Context[commands.Bot], emoji: discord.PartialEmoji):
    await ctx.message.add_reaction(emoji)
```

image

The regex has been tested for correctness using a third-party [emoji](https://pypi.org/project/emoji/) package:

```py
from emoji import UNICODE_EMOJI
import re
# output should be empty
print(*["" if re.match(r"^[\u0023-\u0039]?[\u00ae\u00a9\U00002000-\U0010ffff]+$", argument) else f"{argument} {[hex(ord(c)) for c in argument]}\n" for argument in UNICODE_EMOJI['en']], sep="")
```

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -771,8 +771,13 @@ class PartialEmojiConverter(Converter[discord.PartialEmoji]):
 
     This is done by extracting the animated flag, name and ID from the emoji.
 
+    If the emoji is a unicode emoji, then the name is the unicode character.
+
     .. versionchanged:: 1.5
          Raise :exc:`.PartialEmojiConversionFailure` instead of generic :exc:`.BadArgument`
+
+    .. versionchanged:: 2.0
+        Add support for converting unicode emojis
     """
 
     async def convert(self, ctx: Context[BotT], argument: str) -> discord.PartialEmoji:
@@ -786,6 +791,10 @@ class PartialEmojiConverter(Converter[discord.PartialEmoji]):
             return discord.PartialEmoji.with_state(
                 ctx.bot._connection, animated=emoji_animated, name=emoji_name, id=emoji_id
             )
+
+        # If the emoji is a unicode emoji, then the name is the unicode character.
+        if re.match(r'^[\u0023-\u0039]?[\u00ae\u00a9\U00002000-\U0010ffff]+$', argument):
+            return discord.PartialEmoji.with_state(ctx.bot._connection, name=argument)
 
         raise PartialEmojiConversionFailure(argument)
 
--- a/docs/migrating.rst
+++ b/docs/migrating.rst
@@ -1445,6 +1445,7 @@ Miscellaneous Changes
 - :meth:`ext.commands.Cog.cog_load` has been added as part of the :ref:`migrating_2_0_commands_extension_cog_async` changes.
 - :meth:`ext.commands.Cog.cog_unload` may now be a :term:`coroutine` due to the :ref:`migrating_2_0_commands_extension_cog_async` changes.
 - :attr:`ext.commands.Command.clean_params` type now uses a custom :class:`inspect.Parameter` to handle defaults.
+- :class:`ext.commands.PartialEmojiConverter` can now return a :class:`PartialEmoji` representing a unicode emoji.
 
 .. _migrating_2_0_tasks:
 


rtk-rnjn:
Annotate `emoji` as `Union[PartialEmoji, str]`. Unicode emojis are basically string.

muddyfish:
This is a breaking change for those who use `.url` or `.read()` on the partial emojis returned

DenverCoder1:
> Annotate `emoji` as `Union[PartialEmoji, str]`. Unicode emojis are basically string.

The unicode emoji string is contained within a `PartialEmoji` as the name attribute. So it's still a `PartialEmoji`.

DenverCoder1:
> This is a breaking change for those who use `.url` or `.read()` on the partial emojis returned

Those methods already check `if self.is_unicode_emoji():`.

`PartialEmoji` already supports unicode emojis, this PR simply allows the command converter to convert unicode emojis.

muddyfish:
> > This is a breaking change for those who use `.url` or `.read()` on the partial emojis returned
> 
> Those methods already check `if self.is_unicode_emoji():`.
> 
> `PartialEmoji` already supports unicode emojis, this PR simply allows the command converter to convert unicode emojis.

Previously when using a PartialEmojiConverter, these methods always succeeded and gave useful results. With this change they won't

DenverCoder1:
> Previously when using a PartialEmojiConverter, these methods always succeeded and gave useful results. With this change they won't

I'd say that converting unicode emojis is the expected behaviour, so while it's true that someone _might_ be catching PartialEmojiConversionFailure to determine that an argument is not a custom emoji, I find it to be unlikely and a bit of a stretch.

If you want we can call it a breaking change and I'll add it to migrating for now anyway.

ArjunSharda (approved):

Rapptz (changes requested):
I don't consider this to be a strict breaking change, however, I don't think this is something I should do because the range of emoji tends to be volatile between unicode versions and that's a maintenance burden historically I didn't want to deal with. That aside, the range in this regex is incredibly wide to the point of being meaningless for its purpose.

Consider a sentence in Japanese:

```py
>>> import re
>>> regex = re.compile(r'^[\u0023-\u0039]?[\u00ae\u00a9\U00002000-\U0010ffff]+$')
>>> regex.match('お前は何してんだ？')
<re.Match object; span=(0, 9), match='お前は何してんだ？'>
```

The reason why this regex works with everything is because it casts such a wide net, basically a large amount of the BMP is being included here. A more proper solution would be to take into consideration the actual ranges that a unicode emoji falls under, and this is very much a best-guess basis. The known ranges, from looking around, are more or less:

- `U+1F1E6` to `U+1F1FF`: Flags
- `U+1F300` to `U+1F5FF`: Misc Symbols and Pictographs
- `U+1F600` to `U+1F64F`: Emoticons
- `U+1F680` to `U+1F6FF`: Transport and Map
- `U+1F900` to `U+1F9FF`: Supplemental Symbols and Pictographs
- `U+2600` to `U+26FF`: Misc Symbols
- `U+2700` to `U+27BF`: Dingbats
- `U+FE00` to `U+FE0F`: Variation Selectors

These ranges are inclusive.

DenverCoder1:
> * `U+1F1E6` to `U+1F1FF`: Flags
> * `U+1F300` to `U+1F5FF`: Misc Symbols and Pictographs
> * `U+1F600` to `U+1F64F`: Emoticons
> * `U+1F680` to `U+1F6FF`: Transport and Map
> * `U+1F900` to `U+1F9FF`: Supplemental Symbols and Pictographs
> * `U+2600` to `U+26FF`: Misc Symbols
> * `U+2700` to `U+27BF`: Dingbats
> * `U+FE00` to `U+FE0F`: Variation Selectors

This misses more than half the existing emoji, and Unicode could always be extended.

Using a regex to match emojis with few false positives is very complicated and would be an extremely complicated, long regex.

[This one](https://github.com/mathiasbynens/emoji-regex/blob/main/index.js) for example is over 3,000 characters long.

The simplest, most reliable option would be to just use an external lib method such as [emoji](https://pypi.org/project/emoji/).is_emoji(), but that would require adding dependencies and it seems like it's preferred to avoid that.

Alternatively, it could just be assumed that anything that's not a custom emoji is a unicode emoji (similar to PartialEmoji.from_string) and never raise the error.

Or we could use a very loose regex that will for sure cover everything without being unnecessarily complicated.

Rapptz:
Right, my point is that this wasn't added because it is complicated to do and is a maintenance burden. A lot of the regex are generated, mainly out of laziness and because they don't want to deal with the post-validity. The post-validity in this case would just be Discord rejecting it, but ideally a regex could be much shorter and cast a smaller net. However, the current regex is far too wide (it's quite literally U+2000 until the end of the unicode range) so that's not going to be very helpful. 

Thanks for the PR but I think I'm just not going to go forward with it.

PR #8196, opened 2022-06-30
Add built-in check for Interaction.app_permissions

Luc1412:
## Summary

This pr adds a check for the new `Interaction.app_permissions` attribute.

Am not sure about the naming. `app_has_permissions` was a bit weird and didn't stated the use-case clear enough. Feel free to suggest any changes related to naming.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
This check is just duplicating what `bot_has_permissions` does. An alternative approach is just to port `bot_has_permissions` to use it instead, though this has the same gotchas mentioned in #8103

Issue #7918, opened 2022-04-22
Application commands not using proxy


CoolSpring8:
### Summary

Application commands are not using proxy

### Reproduction Steps

1. Build a `discord.ext.commands.Bot` with one `discord.app_commands` and one `discord.ext.commands.Command`, and pass a proxy URL to `proxy` parameter on creation (the bot runs in a network where all requests should go through the proxy);
2. In Discord, call the bot with normal commands and application (slash) commands.

### Expected Results

Bot responds to both normal commands and application commands.

### Actual Results

Bot responds to normal commands. But for slash commands, interaction shows "The application did not respond", and python throws `aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host discord.com:443 ssl:default`.

### Intents

intents = discord.Intents.default(); intents.message_content = True

### System Information

- Python v3.9.12-final
- discord.py v2.0.0-alpha
    - discord.py pkg_resources: v2.0.0a4122+gbc07a02
- aiohttp v3.7.4.post0
- system info: Darwin 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:47:26 PDT 2022; root:xnu-8020.101.4~15/RELEASE_ARM64_T8101

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.

### Additional Context

Turns out the bot is using proxy for normal commands, but not for application commands.

Looking into the code, I find out the probable reason is that the `aiohttp.ClientSession` which is reused by `webhook/async_.py` to send interaction response, does not store the `proxy` info in its internal state, instead `proxy` parameter has to be passed into every `session.request()` call. This limitation has been tracked in https://github.com/aio-libs/aiohttp/issues/2580, but no recent fixing signal there.

Suggestion: in `webhook/async_.py`, pass down `proxy` parameter to `session.request()`, or use `discord.HTTPClient` instead of `aiohttp.ClientSession`?

FeifanChang:
same bug

PR #8198, opened 2022-07-01
Fix passing proxy to interactions and webhooks

CoolSpring8:
## Summary

Hi, thanks for the fix https://github.com/Rapptz/discord.py/commit/4dd5cc2bc9115d26df8f5d80a9c56d07c62d66a0 for #7918, but I think there's one line of code left out though. Tested in my case it works after this PR.

## Checklist

- [x] This PR fixes an issue.


Rapptz (approved):
Thanks.

PR #8212, opened 2022-07-07
Add advanced startup example

mikeshardmind:
## Summary

shows setup_hook and async context manager use with examples
approximating real world use cases

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/advanced_startup.py
+++ b/examples/advanced_startup.py
@@ -2,9 +2,10 @@
 
 import asyncio
 import logging
+import logging.handlers
 import os
 
-from typing import Optional
+from typing import List, Optional
 
 import asyncpg  # asyncpg is not a dependency of the discord.py, and is only included here for illustrative purposes.
 import discord
@@ -16,7 +17,7 @@ class CustomBot(commands.Bot):
     def __init__(
         self,
         *args,
-        initial_extensions: list,
+        initial_extensions: List[str],
         db_pool: asyncpg.Pool,
         web_client: ClientSession,
         testing_guild_id: Optional[int] = None,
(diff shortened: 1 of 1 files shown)


examples/advanced_startup.py
```diff
@@ -0,0 +1,74 @@
...
+
+async def main():
+
+    # When taking over how the bot process is run, you become responsible for a few additional things.
+
+    # 1. logging
+
+    # for this example, we're just going to use some basic defaults provided by python
+    # for more info on setting up logging, see https://docs.python.org/3/howto/logging.html
+    logging.basicConfig(level=logging.INFO)
```
Rapptz, line 57: d.py has a dedicated logging page which is decently in-depth for those who just want quick examples, might want to refer to it too and use one of the snippets.

examples/advanced_startup.py
```diff
@@ -0,0 +1,74 @@
...
+import discord
+from discord.ext import commands
+from aiohttp import ClientSession
+
+
+class CustomBot(commands.Bot):
+    def __init__(
+        self,
+        *args,
+        initial_extensions: list,
```
Rapptz, line 19:
```suggestion
        initial_extensions: List[str],
```

examples/advanced_startup.py
```diff
@@ -0,0 +1,74 @@
+# This example covers advanced startup options and uses some real world examples for why you may need them.
+
+import asyncio
+import logging
+import os
+
+from typing import Optional
```
Rapptz, line 7:
```suggestion
from typing import Optional, List
```

Rapptz (changes requested):

examples/advanced_startup.py
mikeshardmind: Sure, this is a good place to advocate for devs to do more with logging than the basics anyway once they step into needing more fine-grained control over start.

Rapptz (approved):
Seems fine, thanks.

PR #8217, opened 2022-07-09
Add AppCmdPermissions update event

z03h:
## Summary

adds new event
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/models.py
+++ b/discord/app_commands/models.py
@@ -869,8 +869,8 @@ class AppCommandPermissions:
 
     Attributes
     -----------
-    guild_id: :class:`int`
-        The ID of the guild associated with this permission.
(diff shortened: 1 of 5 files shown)


Heurican:
what does "adds an event" mean? could you be more specific?

Rapptz:
This stomps all over the previous work of the previous PR that made `guild` non-optional by using an unavailable fallback. I suggest changing it and fixing it so it doesn't reintroduce issues brought up from the earlier PR.

Rapptz (changes requested):
I don't want to shorten "command" to "cmd", even if it leads to extra long event and class names here.

Rapptz (approved):
Looks good, thanks.

PR #8253, opened 2022-07-19
[commands] Change cooldowns to take context instead of message

sgtlaggy:
## Summary

This makes cooldown callables take context instead of message.
For most cases, context should be a drop-in replacement without causing issues.
Any cooldowns using `BucketType` or using `CooldownMapping` directly without the `@cooldown`/`@dynamic_cooldown` decorator remain unaffected.
This will only cause breaking changes in decorator-based cooldowns where the `type` is a function that uses message-exclusive attributes.

This also makes CooldownMapping a generic for more useful typing.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/cooldowns.py
+++ b/discord/ext/commands/cooldowns.py
@@ -25,7 +25,7 @@ DEALINGS IN THE SOFTWARE.
 from __future__ import annotations
 
 
-from typing import Any, Callable, Deque, Dict, Optional, TYPE_CHECKING
+from typing import Any, Callable, Deque, Dict, Optional, Union, Generic, TypeVar, TYPE_CHECKING
 from discord.enums import Enum
 import time
 import asyncio
@@ -33,12 +33,14 @@ from collections import deque
 
 from ...abc import PrivateChannel
 from .errors import MaxConcurrencyReached
+from .context import Context
 from discord.app_commands import Cooldown as Cooldown
 
 if TYPE_CHECKING:
     from typing_extensions import Self
 
     from ...message import Message
+    from ._types import BotT
 
 __all__ = (
     'BucketType',
@@ -47,6 +49,8 @@ __all__ = (
     'DynamicCooldownMapping',
     'MaxConcurrency',
 )
+
+T = TypeVar('T')
 
 
 class BucketType(Enum):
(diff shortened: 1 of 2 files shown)


discord/ext/commands/cooldowns.py
```diff
@@ -48,6 +50,8 @@
     'MaxConcurrency',
 )
 
+T = TypeVar('T')
```
Rapptz, line 53: This feels like it needs a bound, though I can't test right now whether it'd work as I expected it to with it.

Rapptz (changes requested):
Need to explain the type ignore comments

discord/ext/commands/cooldowns.py
sgtlaggy: What type would it be bound to? As written, it allows specifying any type to be used, as opposed to strictly limiting to `Message` and `Context`.

Rapptz: Yes, and that's the problem

sgtlaggy:
> Need to explain the type ignore comments

Should I explain as an extra comment on the line?

They're there because `BucketType` works for both `Context` and `Message`, but Pyright isn't a fan, regardless of whether the typevar is covariant or not.
```
Cannot assign member "_buckets" for type "Command[CogT@Command, P@Command, T@Command]"
  Expression of type "CooldownMapping[Message | Context[BotT@__call__]] | CooldownMapping[Unknown]" cannot be assigned to member "_buckets" of class "Command[CogT@Command, P@Command, T@Command]"
    Type "CooldownMapping[Message | Context[BotT@__call__]] | CooldownMapping[Unknown]" cannot be assigned to type "CooldownMapping[Context[Unknown]]"
      TypeVar "T@CooldownMapping" is invariant
        Type "Message | Context[BotT@__call__]" cannot be assigned to type "Context[Unknown]"
          "Message" is incompatible with "Context[Unknown]"
```
Is there some other way to solve this?

Rapptz:
You can either do `# type: ignore  # <explanation>` or put the explanation comment above it if it gets too long.

That message is because one of the paths for `CooldownMapping` (the one right above with `isinstance`) probably becomes `CooldownMapping[Unknown]`.

discord/ext/commands/cooldowns.py
sgtlaggy: I don't see a problem in that. If it were bound and someone wanted to use it to ratelimit reactions or interactions, they'd need to `# type: ignore` the assignment.

Rapptz (approved):
Guess we'll trial this. Thanks.

PR #8271, opened 2022-07-26
Implemented Embed.__eq__

Vioshim:
## Summary

The method allows to have consistency in comparison between Embeds.

```py
e1 = Embed().set_footer(text="A")
e2 = Embed().set_footer(text="A")

assert e1.footer == e2.footer
assert e1 == e2
```

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -50,7 +50,7 @@ class EmbedProxy:
 
     def __getattr__(self, attr: str) -> None:
         return None
-    
+
     def __eq__(self, other: object) -> bool:
         return isinstance(other, EmbedProxy) and self.__dict__ == other.__dict__
 
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -50,6 +50,9 @@ class EmbedProxy:
 
     def __getattr__(self, attr: str) -> None:
         return None
+    
+    def __eq__(self, other: object) -> bool:
+        return isinstance(other, EmbedProxy) and self.__dict__ == other.__dict__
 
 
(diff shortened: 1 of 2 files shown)


discord/embeds.py
```diff
@@ -281,6 +290,24 @@ def __bool__(self) -> bool:
             )
         )
 
+    def __eq__(self, other: Embed) -> bool:
+        return isinstance(other, Embed) and all(
+            (
+                self.title == other.title,
```
Rapptz, line 296:
```suggestion
                self.type == other.type,
                self.title == other.title,
```

Rapptz (changes requested):
This is a breaking change because `Embed` and `EmbedProxy` lost their `__hash__` since `__eq__` is now defined which means that it can no longer be used in mappings or sets. So this needs to be documented in migrating.rst somewhere. 

Ideally the `__eq__` would use `to_dict()` but `to_dict` would need to be changed to handle false-y values so it might not be worth making the change for this PR.

docs/migrating.rst
```diff
@@ -883,22 +883,28 @@ The following properties were changed to return a sequence instead of a list:
 This change should be transparent, unless you are modifying the sequence by doing things such as ``list.append``.
 
 
-Removal of ``Embed.Empty``
+Embed Changes
 ---------------------------
```
Rapptz, line 887: You should remove the extraneous `-`. It only needs to match the title's length (+1 preferably)

Rapptz (approved):
Thanks

PR #8280, opened 2022-07-30
Remove documentation for on_group_join/leave

Middledot:
## Summary
Remove group channel events that weren't removed with the deprecation of user-bot features.

## Checklist
- [ ] This PR fixes an issue.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -288,16 +288,6 @@ Channels
     :param after: The updated guild channel's new info.
     :type after: :class:`abc.GuildChannel`
 
-.. function:: on_group_join(channel, user)
-              on_group_remove(channel, user)
-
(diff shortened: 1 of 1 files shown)


Rapptz:
These events are technically possible to dispatch for a very small minority of bots.

Middledot:
These events aren't referenced anywhere and aren't dispatched in the library itself, also how is it possible?

SebbyLaw:
> These events aren't referenced anywhere and aren't dispatched in the library itself, also how is it possible?

Several years ago "bots" were just user accounts. When they were converted to bot accounts, they remained in DM groups.

Rapptz:
> These events aren't referenced anywhere and aren't dispatched in the library itself, also how is it possible?

Actually, you're right. I seemingly removed them.

PR #8283, opened 2022-07-30
Fix `CooldownMapping` generic typing and ensure other public methods/types have proper generics

bryanforbes:
## Summary

The title says it all

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/cooldowns.py
+++ b/discord/ext/commands/cooldowns.py
@@ -40,7 +40,6 @@ if TYPE_CHECKING:
     from typing_extensions import Self
 
     from ...message import Message
-    from ._types import BotT
 
 __all__ = (
     'BucketType',
@@ -50,7 +49,7 @@ __all__ = (
     'MaxConcurrency',
 )
 
-T = TypeVar('T')
+T_contra = TypeVar('T_contra')
 
 
 class BucketType(Enum):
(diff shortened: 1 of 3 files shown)


discord/ext/commands/cooldowns.py
```diff
@@ -50,7 +49,7 @@
     'MaxConcurrency',
 )
 
-T = TypeVar('T')
+T_contra = TypeVar('T_contra')
```
Gobot1234, line 52: Seems like you're missing contravariant=True here

bryanforbes: Ugh, I took it out to test something... _sigh_

Rapptz:
Why make this contravariant?

bryanforbes:
> Why make this contravariant?

If the type is only being used in a parameter, it can be either invariant or contravariant. Since `commands.Command._bucket` is typed as `CooldownMapping[Context[Any]]` (so `get_bucket()` will accept a `Context` or subclass), an invariant generic for `CooldownMapping` technically disallows assigning the mapping from `CooldownMapping.from_cooldown()` with a `BucketType` because that results in `CooldownMapping[Message | Context[Any]]`. Contravariance is the solution to that problem.

PR #8296, opened 2022-08-03
add InteractionResponse.type property

sgtlaggy:
## Summary

This adds `InteractionResponse.type` to indicate how the interaction was responded to.
This is particularly useful for error handlers, so you can know whether to edit the original response or send a followup to notify the user an error has occurred.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/interactions.py
+++ b/discord/interactions.py
@@ -514,13 +514,18 @@ class InteractionResponse:
 
     def __init__(self, parent: Interaction):
         self._parent: Interaction = parent
-        self._responded: bool = False
+        self._responded: Optional[InteractionResponseType] = None
 
     def is_done(self) -> bool:
         """:class:`bool`: Indicates whether an interaction response has been done before.
 
         An interaction can only be responded to once.
         """
+        return bool(self._responded)
+
+    @property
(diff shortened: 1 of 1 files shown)


discord/interactions.py
```diff
@@ -514,13 +514,18 @@ class InteractionResponse:
 
     def __init__(self, parent: Interaction):
         self._parent: Interaction = parent
-        self._responded: bool = False
+        self._responded: Optional[InteractionResponseType] = None
```
Rapptz, line 517: Rename this attribute to something clearer, like `_response_type`.

discord/interactions.py
```diff
@@ -514,13 +514,18 @@ class InteractionResponse:
...
 
     def is_done(self) -> bool:
         """:class:`bool`: Indicates whether an interaction response has been done before.
 
         An interaction can only be responded to once.
         """
+        return bool(self._responded)
```
Rapptz, line 524:
```suggestion
        return self._response_type is not None
```

Rapptz (changes requested):

Rapptz (approved):
Thanks.

PR #8298, opened 2022-08-03
Bump Pyright to 1.1.265, fix type errors, and remove unnecessary ignores

bryanforbes:
## Summary

The title says it all

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/tasks/__init__.py
+++ b/discord/ext/tasks/__init__.py
@@ -29,8 +29,8 @@ import datetime
 import logging
 from typing import (
     Any,
-    Awaitable,
     Callable,
+    Coroutine,
     Generic,
     List,
     Optional,
@@ -56,10 +56,10 @@ __all__ = (
 # fmt: on
 
 T = TypeVar('T')
-_func = Callable[..., Awaitable[Any]]
+_func = Callable[..., Coroutine[Any, Any, Any]]
 LF = TypeVar('LF', bound=_func)
 FT = TypeVar('FT', bound=_func)
-ET = TypeVar('ET', bound=Callable[[Any, BaseException], Awaitable[Any]])
+ET = TypeVar('ET', bound=Callable[[Any, BaseException], Coroutine[Any, Any, Any]])
 
 
 def is_ambiguous(dt: datetime.datetime) -> bool:
@@ -562,7 +562,7 @@ class Loop(Generic[LF]):
             raise TypeError(f'Expected coroutine function, received {coro.__class__.__name__!r}.')
 
         self._before_loop = coro
-        return coro  # type: ignore
+        return coro
 
     def after_loop(self, coro: FT) -> FT:
         """A decorator that registers a coroutine to be called after the loop finishes running.
@@ -590,7 +590,7 @@ class Loop(Generic[LF]):
             raise TypeError(f'Expected coroutine function, received {coro.__class__.__name__!r}.')
 
         self._after_loop = coro
-        return coro  # type: ignore
+        return coro
 
     def error(self, coro: ET) -> ET:
         """A decorator that registers a coroutine to be called if the task encounters an unhandled exception.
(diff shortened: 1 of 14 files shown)


discord/ext/tasks/__init__.py
```diff
@@ -562,7 +562,7 @@ def before_loop(self, coro: FT) -> FT:
             raise TypeError(f'Expected coroutine function, received {coro.__class__.__name__!r}.')
 
         self._before_loop = coro
-        return coro
+        return coro  # type: ignore
```
Rapptz, line 565: Can we add an explanation comment for this and similar ones?

I feel like this one's a pyright thing since this is just an identity function...

discord/state.py
```diff
@@ -247,7 +247,7 @@ def __init__(
         self._command_tree: Optional[CommandTree] = None
 
         if not intents.members or cache_flags._empty:
-            self.store_user = self.store_user_no_intents  # type: ignore # This reassignment is on purpose
+            self.store_user = self.store_user_no_intents  # This reassignment is on purpose
```
Rapptz, line 250:
```suggestion
            self.store_user = self.store_user_no_intents
```

discord/state.py
```diff
@@ -1329,7 +1329,7 @@ def parse_guild_integrations_update(self, data: gw.GuildIntegrationsUpdateEvent)
             _log.debug('GUILD_INTEGRATIONS_UPDATE referencing an unknown guild ID: %s. Discarding.', data['guild_id'])
 
     def parse_integration_create(self, data: gw.IntegrationCreateEvent) -> None:
-        guild_id = int(data.pop('guild_id'))
+        guild_id = int(data.pop('guild_id'))  # type: ignore
```
Rapptz, line 1332:
```suggestion
        guild_id = int(data['guild_id'])
```

discord/state.py
```diff
@@ -1339,7 +1339,7 @@ def parse_integration_create(self, data: gw.IntegrationCreateEvent) -> None:
             _log.debug('INTEGRATION_CREATE referencing an unknown guild ID: %s. Discarding.', guild_id)
 
     def parse_integration_update(self, data: gw.IntegrationUpdateEvent) -> None:
-        guild_id = int(data.pop('guild_id'))
+        guild_id = int(data.pop('guild_id'))  # type: ignore
```
Rapptz, line 1342:
```suggestion
        guild_id = int(data['guild_id'])
```

Rapptz (changes requested):
Thanks a lot for getting to this! I've been meaning to do this for so long but never had the time.

Though, when editing on VSCode there are quite a few more errors (e.g. in `hybrid.py`) that I don't see are touched on here. I wonder if the newest version fixed these and Pylance is a bit behind.

bryanforbes:
> Though, when editing on VSCode there are quite a few more errors (e.g. in `hybrid.py`) that I don't see are touched on here. I wonder if the newest version fixed these and Pylance is a bit behind.

Pyright 1.1.265 was just released today, so I imagine a new Pylance release is coming with this latest version of Pyright. We can wait to merge this until that happens if you'd like.

discord/ext/tasks/__init__.py
bryanforbes: I did a bit of digging here and `inspect.iscoroutinefunction()` changes the type to `(...) -> types.CoroutineType[Any, Any, Any]` which is not compatible with `FT` (bound to an `Awaitable[]` return). There are two ways to fix this depending on what the goal is:

* Change `LF`, `FT`, and `ET` to be bound to a callable that returns `Coroutine[Any, Any, Any]`. If we want the code to stay the same as it is right now, this is the correct solution.
* Change the task code to check if the return value of the functions is awaitable rather than checking if the function itself is a coroutine function. If we want `Loop` to be able to use functions returning awaitables, this is the correct solution.

Let me know what you'd like me to do.

Rapptz: The first one is the correct solution. These are decorators so passing in sync functions that return awaitables is not really the intended use case since it'd be hard to check for this in a clean manner (some awaitables can only be awaited once or may return a type that causes some kind of limitation in terms of use).

bryanforbes:
Pylance 2022.8.11 has been released which includes Pyright 1.1.265. We should be good to go to merge this PR.

Rapptz (approved):

PR #8313, opened 2022-08-07
Add `description` kwarg to command.Parameter and show in help command

Soheab:
## Summary

This PR adds a way to set a description for an ext.commands/hybrid command argument.
and reformats the `DefaultHelpCommand` to show that description, MinimalHelpCommand is left unchanged.

Also added `arguments_heading` and `default_argument_description` keyword arguments to `DefaultHelpCommand`.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/help.py
+++ b/discord/ext/commands/help.py
@@ -1148,8 +1148,8 @@ class DefaultHelpCommand(HelpCommand):
         :attr:`indent` spaces, padded to ``max_size`` using :meth:`~HelpCommand.get_max_size`
         followed by the argument's :attr:`~.commands.Parameter.description` or
         :attr:`.default_argument_description` and then shortened
-        to fit into the :attr:`width` and  then the
-        :attr:`~.commands.Parameter.displayed_default` between () if one is present.
+        to fit into the :attr:`width` and then :attr:`~.commands.Parameter.displayed_default`
+        between () if one is present after that.
 
         .. versionadded:: 2.0
 
--- a/discord/ext/commands/help.py
+++ b/discord/ext/commands/help.py
@@ -1038,7 +1038,7 @@ class DefaultHelpCommand(HelpCommand):
         self.dm_help_threshold: int = options.pop('dm_help_threshold', 1000)
         self.arguments_heading: str = options.pop('arguments_heading', "Arguments:")
         self.commands_heading: str = options.pop('commands_heading', 'Commands:')
-        self.default_argument_description: str = options.pop('default_argument_description', 'No description given.')
+        self.default_argument_description: str = options.pop('default_argument_description', 'No description given')
         self.no_category: str = options.pop('no_category', 'No Category')
         self.paginator: Paginator = options.pop('paginator', None)
         self.show_parameter_descriptions: bool = options.pop('show_parameter_descriptions', True)
@@ -1164,15 +1164,9 @@ class DefaultHelpCommand(HelpCommand):
         for argument in arguments:
             name = argument.name
             width = max_size - (get_width(name) - len(name))
-            entry = f'{self.indent * " "}{name:<{width}}'
-            if argument.description is not None:
-                entry += f' - {argument.description}'
-            else:
-                entry += f' - {self.default_argument_description}'
-
+            entry = f'{self.indent * " "}{name:<{width}} {argument.description or self.default_argument_description}'
             # we do not want to shorten the default value, if any.
             entry = self.shorten_text(entry)
-
             if argument.displayed_default is not None:
                 entry += f' (default: {argument.displayed_default})'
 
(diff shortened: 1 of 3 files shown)


discord/ext/commands/parameters.py
```diff
@@ -152,6 +158,11 @@ def converter(self) -> Any:
 
         return self.annotation
 
+    @property
+    def description(self) -> Optional[str]:
+        """Optional[:class:`str`]: The description of this parameter."""
+        return self._description if self._description is not empty else None
```
Gobot1234, line 164: This shouldn't be empty right?

```suggestion
        return self._description if self._description is not MISSING else None
```

Gobot1234: Sorry please ignore this

lol219 (approved):

discord/ext/commands/help.py
```diff
@@ -1088,9 +1087,60 @@ def add_indented_commands(
...
-            entry = f'{self.indent * " "}{name:<{width}} {command.short_doc}'
+            entry = f'{self.indent * " "}{name:<{width}} - {command.short_doc}'
             self.paginator.add_line(self.shorten_text(entry))
 
+    def add_command_arguments(
+        self, command: Command[Any, ..., Any], /, *, heading: str, max_size: Optional[int] = None
+    ) -> None:
+        """Indents a list of command arguments after the specified heading.
+
+        The formatting is added to the :attr:`paginator`.
```
Rapptz, line 1098:
```suggestion
```

discord/ext/commands/help.py
```diff
@@ -1005,9 +992,19 @@ class DefaultHelpCommand(HelpCommand):
...
+
+        .. versionadded:: 2.0
     commands_heading: :class:`str`
         The command list's heading string used when the help command is invoked with a category name.
         Useful for i18n. Defaults to ``"Commands:"``
+    default_argument_description: :class:`str`
+        The default argument description string used when the argument's :attr:`~.commands.Parameter.description` is ``None``.
+        Useful for i18n. Defaults to ``"No description given."``
+
+        .. versionadded:: 2.0
```
Rapptz, line 1007: Surely if no description is given then it should just be empty?

discord/ext/commands/help.py
```diff
@@ -1088,9 +1087,60 @@ def add_indented_commands(
         for command in commands:
             name = command.name
             width = max_size - (get_width(name) - len(name))
-            entry = f'{self.indent * " "}{name:<{width}} {command.short_doc}'
+            entry = f'{self.indent * " "}{name:<{width}} - {command.short_doc}'
             self.paginator.add_line(self.shorten_text(entry))
 
+    def add_command_arguments(
+        self, command: Command[Any, ..., Any], /, *, heading: str, max_size: Optional[int] = None
```
Rapptz, line 1094:
```suggestion
        self, command: Command[Any, ..., Any], /
```

This interface doesn't make much sense. The heading is already an attribute and the `max_size` doesn't need to be specified.

Rapptz (changes requested):
I feel like there should probably be a way to disable this argument heading thing for those who don't want it.

discord/ext/commands/help.py
Soheab: Correct but then the default (if any) looks out of place.
```
  name (default: ...)
  name - Description 
  name - Description (default: ...)
  name (default: ...)
```

discord/ext/commands/help.py
Soheab: Ah my bad, that was copied from the signature of `DefaultHelpCommand.add_indented_commands`
https://github.com/Rapptz/discord.py/blob/584c22f642cfb63b921467a901c8a47c721923b3/discord/ext/commands/help.py#L1051-L1053

Rapptz: For that one, the heading used is dynamic so it makes a bit more sense. The documentation nit is still relevant there though...

discord/ext/commands/help.py
```diff
@@ -1049,6 +1067,34 @@ def get_ending_note(self) -> str:
...
+        ------------
+        command: :class:`Command`
+            The command to get the signature of.
+
+        Returns
+        --------
+        :class:`str`
+            The signature for the command.
+        """
+        if self.show_parameter_descriptions is False:
```
Rapptz, line 1088:
```suggestion
        if not self.show_parameter_descriptions:
```

discord/ext/commands/help.py
```diff
@@ -1124,6 +1208,9 @@ def add_command_formatting(self, command: Command[Any, ..., Any], /) -> None:
                     self.paginator.add_line(line)
                 self.paginator.add_line()
 
+        if self.show_parameter_descriptions is True:
```
Rapptz, line 1211:
```suggestion
        if self.show_parameter_descriptions:
```

Rapptz (approved):
Thanks.

PR #8314, opened 2022-08-07
HybridCommand with_command not used + with_app_command docs adjustion

Puncher1:
## Summary

This pull request is for some changes for `with_command` and in addition `with_app_command`'s documentation.
At the moment `with_command` is an argument of `HybridCommand` but it's neither documented nor used in code. For that it got removed. 

In addition the documentation for the argument `with_app_command` got changed to a more understandable sentence. As I understand it right, the `HybridCommand` is basically a regular command with the feature through the argument `with_app_command` to **also** register it as an application command. The sentence has been adjusted accordingly.

## Checklist

    - [x] I have updated the documentation to reflect the changes. -> `with_command` wasn't even in the docs.
- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/hybrid.py
+++ b/discord/ext/commands/hybrid.py
@@ -500,12 +500,8 @@ class HybridCommand(Command[CogT, P, T]):
 
         super().__init__(func, **kwargs)
         self.with_app_command: bool = kwargs.pop('with_app_command', True)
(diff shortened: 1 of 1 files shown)


Rapptz:
Yeah this was left-over code from a previous implementation that I never removed. Good catch.

PR #8323, opened 2022-08-10
Fix ui.Select.values in modals

### LINKED ISSUE #8319: Multiple selects error
### Summary

master commit breaks multiple selects

### Reproduction Steps

this [commit](https://github.com/Rapptz/discord.py/tree/d826f4f3a8d4d97d24499f5de5c51e11e640e98f) provides global state for `Select` object 

### Minimal Reproducible Code

_No response_

### Expected Results

No global state

### Actual Results

global state

### Intents

...

### System Information

...

### Checklist

### Additional Context

_No response_

sgtlaggy:
## Summary

This fixes a bug in #8136 where `.values` of every `Select` in a modal would return those of the last `Select`.
resolves #8319 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ui/modal.py
+++ b/discord/ui/modal.py
@@ -174,9 +174,11 @@ class Modal(View):
                     continue
                 item._refresh_state(component)  # type: ignore
 
-    async def _scheduled_task(self, interaction: Interaction):
+    async def _scheduled_task(self, interaction: Interaction, components: List[ModalSubmitComponentInteractionDataPayload]):
         try:
             self._refresh_timeout()
+            self._refresh(components)
+
             allow = await self.interaction_check(interaction)
             if not allow:
                 return
@@ -189,8 +191,10 @@ class Modal(View):
             # In the future, maybe this will require checking if we set an error response.
             self.stop()
 
-    def _dispatch_submit(self, interaction: Interaction) -> None:
-        asyncio.create_task(self._scheduled_task(interaction), name=f'discord-ui-modal-dispatch-{self.id}')
+    def _dispatch_submit(
+        self, interaction: Interaction, components: List[ModalSubmitComponentInteractionDataPayload]
+    ) -> None:
+        asyncio.create_task(self._scheduled_task(interaction, components), name=f'discord-ui-modal-dispatch-{self.id}')
 
     def to_dict(self) -> Dict[str, Any]:
         payload = {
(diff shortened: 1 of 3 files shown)


Rapptz:
Have we checked the memory impact of this at all on a long running bot

sgtlaggy:
I have not, though there shouldn't be long-term lingering values as ContextVars *should* clean up after themselves in an async environment as implied by the [example in the docs](https://docs.python.org/3.10/library/contextvars.html#asyncio-support).

While writing this, I realized I missed a case when testing this outside modals and reintroduced the race condition with regular views, so back to the drawing board.

z03h:
> While writing this, I realized I missed a case when testing this outside modals and reintroduced the race condition with regular views, so back to the drawing board.

The contxtvar gets set in item/modal refresh which is not in the callback task so the contextvar value gets overwritten.

Moving `item._refresh_state`/`modal._refresh` inside `view/modal._scheduled_task` seems to fix this.

sgtlaggy:
I think best course of action would be to close this PR and revert the original commit if you care for a quick fix to modals, despite it not being actually supported by Discord.
I accidentally scuffed this PR with a rebase, and really have no ideas on how to fix this.

sgtlaggy:
Unscuffed commits, moved the refresh calls as early as possible in `_scheduled_task` to avoid a breaking change if someone were to access the component(s) in `interaction_check`. Tested, works for both modals and views (didn't miss that same case this time).

Rapptz:
One fundamental issue with this approach is the same as the other one, except this one kind of exacerbates the problem IMO. It's the fact that more select types are coming soon, and I wanted to commit to a solution that would work with those without resorting to repetition. As it stands right now the footprint for those other selects is going to end up creating N contextvars for every select type there is going to be in the future (channel, user, role, mentionable, text) which just does not work very well.

sgtlaggy:
I'm not sure I understand? This solution uses a single contextvar with a new token per interaction context, as the original commit does. It doesn't make the contextvar an attribute of `Select`, as was my initial thought for solving this. 
If Discord adds more types of selects, it should just need changing the typehint of `List[str]` to `List[Union[str, User, ...]]` and maybe mix in a typevar for `choices` and `values`.

Rapptz:
I was talking about this PR vs just reverting the commit.

Rapptz (changes requested):
Can we add a fallback for the last accessed value similar to how it was before the original PR was merged as discussed in https://canary.discord.com/channels/336642139381301249/812600992116899861/1006990647724613773

Basically `Select._values` is always the last value and if the contextvar returns not found then return that instead.

Rapptz (approved):
Thanks.

PR #8327, opened 2022-08-12
Mark extensions as typed.

AiroPi:
## Summary

This pull request aims to fix an error between pyright and the discord.py extensions (contained in `discord/ext/`).

Indeed, when we import an element from an extension (and not the entire extension) such as: `from discord.ext.commands import Cog`, then pyright (from the pylance extension of vs-code) report us a problem:

Screenshot 2022-08-12 at 13 01 05

Adding the `py.typed` files thus announces that these modules are correctly typed and do not need stub files.

## Checklist

- [x] This PR fixes an issue.


Rapptz:
Duplicate of https://github.com/Rapptz/discord.py/pull/7743

jack1142:
I believe this file won't get included in the distribution because `MANIFEST.in` only includes `discord/py.typed`.

PR #8330, opened 2022-08-12
Add "Why does ``get_x`` and ``utils.get`` return ``None``?" to FAQ

Puncher1:
## Summary

Since this is a very frequently asked question (this repo has even a pinned issue for that) and Danny also wants a better documentation, I added this to the FAQ section. 

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -98,7 +98,7 @@ Why does ``get_x`` and ``utils.get`` return ``None``?
 
 - Bot not logged in, and trying to grab objects from cache
 
-  - Subvariant, using two client objects. :class:`Bot` has all the functionality of :class:`Client` in addition to a full command system.
+  - Subvariant, using two client objects. :class:`~ext.commands.Bot` has all the functionality of :class:`Client` in addition to a full command system.
 
 - Bot cannot "see" the object.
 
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -84,6 +84,29 @@ Where can I find usage examples?
 Example code can be found in the `examples folder <https://github.com/Rapptz/discord.py/tree/master/examples>`_
 in the repository.
 
+Why does ``get_x`` and ``utils.get`` return ``None``?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+``get_x``  and :meth:`utils.get` both return ``None`` if they can't find anything matching.
```
Gobot1234, line 89:
```suggestion
``get_x`` and :func:`utils.get` both return ``None`` if they can't find anything matching.
```

Rapptz:
I'm not really keen on taking tags wholesale and just making them into FAQ entries. Mainly, this question has an immediate knee-jerk answer ("because it's documented to return that when something is not found"). So it should be better worded to not have that reaction. But also the style of this question doesn't match the other ones in the FAQ. This is definitely a common problem but the people who are affected by it are probably not the type to read the FAQ anyway, since almost all of these things are part of the documentation and the type hints. 

If I wanted to add a question like this in the FAQ it'd have to be reworked.

PR #8343, opened 2022-08-16
Add mention_spam automod trigger and change default values

z03h:
## Summary

Adds mention_spam trigger type
Changes trigger defaults values from None

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/audit_logs.py
+++ b/discord/audit_logs.py
@@ -190,24 +190,29 @@ def _transform_automod_trigger_metadata(
 ) -> Optional[AutoModTrigger]:
 
     if isinstance(entry.target, AutoModRule):
-        # type of trigger cannot be changed
-        # try to get trigger type from the resolved rule
+        # Trigger type cannot be changed, so type should be the same before and after updates.
+        # Avoids checking which keys are in data to guess trigger type
+        # or returning None if data is empty.
         try:
             return AutoModTrigger.from_data(type=entry.target.trigger.type.value, data=data)
         except Exception:
             pass
 
-    # if no trigger type from resolved rule, cannot determine trigger type if data is empty
+    # If cannot get trigger type from the rule and data is empty, then cannot determine trigger type
     if not data:
         return None
 
-    # try to infer trigger type from data
+    # Try to infer trigger type from available keys in data
     if 'presets' in data:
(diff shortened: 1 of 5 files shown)


discord/audit_logs.py
```diff
@@ -188,15 +188,26 @@ def _transform(entry: AuditLogEntry, data: Optional[str]) -> Optional[Asset]:
...
     entry: AuditLogEntry, data: AutoModerationTriggerMetadata
 ) -> Optional[AutoModTrigger]:
+
+    if isinstance(entry.target, AutoModRule):
+        # type of trigger cannot be changed
+        # try to get trigger type from the resolved rule
+        try:
+            return AutoModTrigger.from_data(type=entry.target.trigger.type.value, data=data)
+        except Exception:
+            pass
```
Rapptz, line 198: What is this doing?

z03h: The trigger type can't be changed, so if the AutoModRule is available the trigger type should be the same instead of having to check which keys are in the data.

Found trying to edit with invalid trigger metadata such as `{'bogus': 'value'}` was accepted and the data from audit log could be `{}`, not sure if fetched versions behave the same. This can still get the trigger type and returns a trigger vs None.

Rapptz: Probably should document that better in the comments

discord/automod.py
```diff
@@ -114,27 +114,47 @@ def to_dict(self) -> Dict[str, Any]:
...
+    r"""Represents a trigger for an auto moderation rule.
+
+    The following table illustrates relevant attributes for each :class:`AutoModRuleTriggerType`:
+
+    +-----------------------------------------------+--------------------------------------+
+    |                    Type                       |              Attributes              |
+    +===============================================+======================================+
+    | :attr:`AutoModRuleTriggerType.keyword`        | :attr:`keyword_filter`               |
+    +-----------------------------------------------+--------------------------------------+
+    | :attr:`AutoModRuleTriggerType.harmful_link`   |                                      |
```
Rapptz, line 126: This trigger type is apparently deprecated/removed and is on by default according to DAPI docs so it might be worth removing this before pypi release

Rapptz (approved):
Thanks

PR #8345, opened 2022-08-16
Document that docstrings can be used to document parameters

Bluesy1:
## Summary

Currently it is not documented that Google, Sphinx, or Numpy style docstrings can be used instead of the `app_commands.describe` decorator. This changes that.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -2102,7 +2102,8 @@ def describe(**parameters: Union[str, locale_str]) -> Callable[[T], T]:
 
         @app_commands.command()
         async def ban(interaction: discord.Interaction, member: discord.Member):
-            '''
+            """Bans a member
+
             Parameters
             -----------
             member: discord.Member
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -2102,13 +2102,13 @@ def describe(**parameters: Union[str, locale_str]) -> Callable[[T], T]:
 
         @app_commands.command()
         async def ban(interaction: discord.Interaction, member: discord.Member):
-            '''Bans a member
+            \"\"\"Bans a member
 
             Parameters
             -----------
             member: discord.Member
                 the member to ban
-            '''
+            \"\"\"
             await interaction.response.send_message(f'Banned {member}')
 
     Parameters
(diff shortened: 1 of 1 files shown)


discord/app_commands/commands.py
```diff
@@ -2094,6 +2094,22 @@ def describe(**parameters: Union[str, locale_str]) -> Callable[[T], T]:
...
+    .. code-block:: python3
+
+        @app_commands.command()
+        async def ban(interaction: discord.Interaction, member: discord.Member):
+            '''
+            Parameters
+            -----------
+            member: discord.Member
+                the member to ban
+            '''
```
nerdguyahmad, line 2110:
```suggestion
            """
```

discord/app_commands/commands.py
```diff
@@ -2094,6 +2094,22 @@ def describe(**parameters: Union[str, locale_str]) -> Callable[[T], T]:
...
 
+    Alternatively, you can describe parameters using Google, Sphinx, or Numpy style docstrings.
+
+    Example:
+
+    .. code-block:: python3
+
+        @app_commands.command()
+        async def ban(interaction: discord.Interaction, member: discord.Member):
+            '''
```
nerdguyahmad, line 2105: The first line of docstring is used as the command's description. You should also add it.
```suggestion
            """Bans a member

```

nerdguyahmad (changes requested):
`"` should be used for docstrings.

discord/app_commands/commands.py
Bluesy1: If i use `"""` it will interfere with the outer docstring because that uses `r"""` to open

nerdguyahmad: You can escape the inner docstring's quotes by adding a backslash before each `"`

Bluesy1: Done

Rapptz (approved):
Thanks. I think this would be better suited in a prose page but we're out of time for these at the moment.

sgtlaggy:
image
This doesn't look very good in the docs. Maybe `describe`'s docstring should be changed to use single quotes instead, so the example can use double quotes.

PR #8364, opened 2022-08-21
Mark public callbacks as positional-only

bryanforbes:
## Summary

Marks several public-facing callbacks as positional-only. These methods are only ever called with positional arguments. This PR allows developers to use strict mode typing and name these parameters any name they would like.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -1745,7 +1745,7 @@ class Group:
             if isinstance(command, Group):
                 yield from command.walk_commands()
 
-    async def on_error(self, interaction: Interaction, error: AppCommandError) -> None:
+    async def on_error(self, interaction: Interaction, error: AppCommandError, /) -> None:
         """|coro|
 
         A callback that is called when a child's command raises an :exc:`AppCommandError`.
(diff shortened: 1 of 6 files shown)


Rapptz:
This one is really bordering on the breaking change definition here.

bryanforbes:
Kind of, but not really based on how they are called. As it stands for these methods, the API contract requires the user to have named parameters for those methods even though the library doesn't call them that way. By making them positional-only, a user can use positional-only or named parameters and also rename their parameter names. This just updates the contract to what the library actually requires. This shouldn't break any users.

Rapptz:
It's strictly breaking because these can technically be called outside of the callback context. e.g.

```py
TRANSFORMER = SomethingTransformer()
async def command(interaction, thing: Transform[Something, TRANSFORMER]): 
   # this is used as a callback...

# this one is not..
async def arbitrary_user_function(interaction, value):
   try:
        return await TRANSFORMER.transform(interaction, value=value)  # silly but doable
   except:
       return some_fallback
```

The question is just whether this breakage is necessary and if it's too late. Technically a lot of bug fixes fall under the breaking change umbrella but then the question becomes is this really a bug fix?

PR #8388, opened 2022-08-27
Add initialize argument for create view instance.

peco2282:
## Summary

This commit allows the Item to be added when creating a new View instance.

Example:
```py
button1 = Button()
button2 = Button()
button3 = Button()

view = View(button1, button2, button3, timeout=50)
```

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ui/view.py
+++ b/discord/ui/view.py
@@ -178,7 +178,7 @@ class View:
             children.append(item)
         return children
 
-    def __init__(self, *, timeout: Optional[float] = 180.0):
+    def __init__(self, *items: Item, timeout: Optional[float] = 180.0):
         self.__timeout = timeout
         self._children: List[Item[Self]] = self._init_children()
         self.__weights = _ViewWeights(self._children)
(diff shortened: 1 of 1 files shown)


discord/ui/view.py
```diff
@@ -178,7 +178,7 @@ def _init_children(self) -> List[Item[Self]]:
             children.append(item)
         return children
 
-    def __init__(self, *, timeout: Optional[float] = 180.0):
+    def __init__(self, *items: Item, timeout: Optional[float] = 180.0):
```
Puncher1, line 181:
```suggestion
    def __init__(self, *items: Optional[Item], timeout: Optional[float] = 180.0):
```
Also edit the docs at line 150:
```
    *items: Optional[:class:`Item`]
        The items to add to the view.

```

Puncher1 (approved):

discord/ui/view.py
bijij: This doesn't look like the right use of `Optional`.
It should however be annotated as `Item[Any]`

Puncher1: Updated

Rapptz:
Sorry, I deliberately don't want to hold all variadic arguments for something like this and is unlike anything in the library. I also don't think this is common enough to warrant the change. Thanks though!

PR #8390, opened 2022-08-27
Add support for toggling discoverable and invites_disabled features

bijij:
## Summary

This PR adds support for the DISCOVERABLE and INVITES_DISABLED guild feature.

Note, INVITES_DISABLED is not yet stable.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -245,6 +245,7 @@ class Guild(Hashable):
         - ``VERIFIED``: Guild is a verified server.
         - ``VIP_REGIONS``: Guild has VIP voice regions.
         - ``WELCOME_SCREEN_ENABLED``: Guild has enabled the welcome screen.
+        - ``INVITES_DISABLED``: Guild has disabled invites.
 
     premium_tier: :class:`int`
         The premium tier for this guild. Corresponds to "Nitro Server" in the official UI.
@@ -1698,6 +1699,7 @@ class Guild(Hashable):
         rules_channel: Optional[TextChannel] = MISSING,
         public_updates_channel: Optional[TextChannel] = MISSING,
         premium_progress_bar_enabled: bool = MISSING,
+        invites_disabled: bool = MISSING,
     ) -> Guild:
         r"""|coro|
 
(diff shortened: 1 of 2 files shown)


discord/guild.py
```diff
@@ -1906,17 +1913,26 @@ async def edit(
 
             fields['system_channel_flags'] = system_channel_flags.value
 
+        features = []
```
bijij, line 1916: might be worth at this point defaulting features to all the features the guild has, and adding/removing as appropriate.

bijij: Thinking more on this, I'm pretty sure that's necessary and the only way to safely do this.

Since if a new editable feature not supported by dpy is added it'll be blown away every time someone updates other features if we keep this as is.

Rapptz: Using prior state is always going to be buggy. For features in particular though it doesn't allow most features to be in the patch request so I don't think your suggestion is the way to go.

bijij: It seems the client just sends the entire feature list now? Unless I missed something.

Rapptz: If they finally changed it so it allows you to pass in the entire list that's good though it also feels finicky. I'll have to ask for clarification later.

Rapptz: I asked and passing the entire list is the way to go now, so might as well make that change.

bijij: Changed to pass the entire list.

discord/guild.py
```diff
@@ -1906,17 +1916,32 @@ async def edit(
...
+                features.discard('DISCOVERABLE')
+
+        if invites_disabled is not MISSING:
+            if invites_disabled:
+                features.add('INVITES_DISABLED')
+            else:
+                features.discard('INVITES_DISABLED')
 
-            fields['features'] = features
+        fields['features'] = list(features)
```
Rapptz, line 1944: Might be better to only pay for the cost of this cast if the features are set. As it stands right now everyone pays for both casts to `set` and back to `list`.

Rapptz (approved):
This PR looks fine, but I'm putting it on hold because apparently `discoverable` might be changing.

PR #8443, opened 2022-09-07
Add note about renamed options in the Namespace

NarmyOnDiscord:
## Summary

Adds a note that renamed options will also be renamed in the namespace of the interaction

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/namespace.py
+++ b/discord/app_commands/namespace.py
@@ -72,7 +72,9 @@ class Namespace:
     then ``None`` is returned rather than an attribute error.
     
     .. note::
-        Options that were renamed using ``@app_commands.rename()`` will also ne renamed in the namespace.
+
+        The key names come from the raw Discord data, which means that if a parameter was renamed then the
+        renamed key is used instead of the function parameter name.
 
     .. versionadded:: 2.0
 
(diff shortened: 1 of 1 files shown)


discord/app_commands/namespace.py
```diff
@@ -70,6 +70,9 @@ class Namespace:
     key-pair mapping. These attributes can be accessed using dot notation. For example, an option
     with the name of ``example`` can be accessed using ``ns.example``. If an attribute is not found,
     then ``None`` is returned rather than an attribute error.
+    
+    .. note::
+        Options that were renamed using ``@app_commands.rename()`` will also ne renamed in the namespace.
```
Rapptz, line 75:
```suggestion
    .. note::

        The key names come from the raw Discord data, which means that if a parameter was renamed then the
        renamed key is used instead of the function parameter name.
```

discord/app_commands/namespace.py
```diff
@@ -70,6 +70,11 @@ class Namespace:
     key-pair mapping. These attributes can be accessed using dot notation. For example, an option
     with the name of ``example`` can be accessed using ``ns.example``. If an attribute is not found,
     then ``None`` is returned rather than an attribute error.
+    
+    .. note::
```
Rapptz, line 74:
```suggestion

    .. warning::
```

This is failing black due to the whitespace here. Also, I think a warning is more appropriate than a note.

NarmyOnDiscord: Ah, already wondered why it failed

Rapptz (approved):
Thanks!

PR #8447, opened 2022-09-08
fix Range and Transform to allow 3.10 Union syntax

lmaotrigine:
## Summary

This PR allows 3.10 Union syntax to work for `app_commands.Range` and `app_commands.Transform` as was done in #8446

## Checklist

- [x] This PR fixes an issue.


Rapptz:
These should already work with unions.

PR #8453, opened 2022-09-10
Normalize type formatting in Type Errors

ionite34:
Currently, there remain some mixed usages of `obj.__class__` in reporting TypeErrors which results in unnecessary angled brackets and formatting:
```py
[Bot] TypeError: expected token to be a str, received <class 'NoneType'> instead
```
whereas, if sticking with the standard form of `obj.__class__.__name__`, as used by most Python standard library TypeErrors:
```py
[Bot] TypeError: expected token to be a str, received NoneType instead
```
## Summary

Whilst most Type Errors within the project already use this form; there remain a few instances where this is not the case. This PR normalizes print formatting of incorrect types in TypeErrors to use `{obj.__class__.__name__}` 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -1513,7 +1513,7 @@ class Messageable:
             reference_dict = MISSING
 
         if view and not hasattr(view, '__discord_ui_view__'):
-            raise TypeError(f'view parameter must be View not {view.__class__!r}')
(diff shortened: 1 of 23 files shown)


Rapptz (approved):
Thanks! I'm not really sure which one I prefer but this one is fine.

PR #8458, opened 2022-09-10
Fix default guild parameter in tree.clear_commands

Daste745:
## Summary

This fixes the default `guild` argument in `CommandTree.clear_commands`. It's an optional, but didn't have a `= None` default.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/tree.py
+++ b/discord/app_commands/tree.py
@@ -458,7 +458,12 @@ class CommandTree(Generic[ClientT]):
             key = (command, guild_id, type.value)
             return self._context_menus.pop(key, None)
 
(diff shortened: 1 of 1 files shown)


AbstractUmbra:
This is actually by design. A parameter marked as `Optional[T]` accepts `T` or `None`, not that it defaults to `None`.

This parameter explicitly has no default so the user must clarify which they are clearing as it's a volatile operation.

Daste745:
Oh, I didn't realise that this is by design. For me it seemed like a mistake, but now I see that more commands have parameters are like this.

PR #8473, opened 2022-09-17
fix: library stub not found

SaidBySolo:
## Summary

```py
# sample.py

app_commands.command()
```
When I check the file with mypy, it says that the stub cannot be found.

``discord\app_commands\commands.py:89: error: Cannot find implementation or library stub for module named "discord.ext.commands"``

https://github.com/Rapptz/discord.py/pull/7743#issuecomment-1074682871
Perhaps it has something to do with the issue comment.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -69,7 +69,7 @@ if TYPE_CHECKING:
     # Generally, these two libraries are supposed to be separate from each other.
     # However, for type hinting purposes it's unfortunately necessary for one to
     # reference the other to prevent type checking errors in callbacks
-    from discord.ext.commands import Cog
+    from discord.ext import commands
 
     ErrorFunc = Callable[[Interaction, AppCommandError], Coroutine[Any, Any, None]]
 
@@ -105,7 +105,7 @@ Error = Union[
     UnboundError,
 ]
 Check = Callable[['Interaction'], Union[bool, Coro[bool]]]
-Binding = Union['Group', 'Cog']
+Binding = Union['Group', 'commands.Cog']
 
 
 if TYPE_CHECKING:


mikeshardmind:
I can't reproduce the issue this intends to fix.

```
(.venv) PS > Get-Content .\scratch.py

import discord


@app_commands.command()
async def test(interaction: discord.Interaction):
    ...
(.venv) PS > mypy --namespace-packages .\scratch.py 
Success: no issues found in 1 source file
```

You might need to add namespace-packages to your mypy config

```
(.venv) PS > mypy --no-namespace-packages .\scratch.py    
.venv\lib\site-packages\discord\app_commands\commands.py:72: error: Cannot find implementation or library stub for module named "discord.ext.commands"
.venv\lib\site-packages\discord\app_commands\commands.py:72: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 1 source file)
```

which is something the error actually points you to

SaidBySolo:
```
sample/
├─ __init__.py
├─ main.py
```

```py
# main.py
import discord

@app_commands.command()
async def main(interaction: discord.Interaction) -> None:
    ...
```

`` mypy ./sample --strict``

This is a reproducible example.

Incidentally, as noted in the issue comments, I don't think this is an ideal import.

mikeshardmind:
As pointed out previously, you can resolve this as a configuration change. mypy supports changing the behavior of import discovery. Trimming what I had above to only what's needed to show that, and including the error from mypy which *also* informed you of this:

```
(.venv) PS > mypy --strict --namespace-packages .\scratch.py 
Success: no issues found in 1 source file
```
vs
```
(.venv) PS > mypy --strict .\scratch.py
.venv\lib\site-packages\discord\app_commands\commands.py:72: error: Cannot find implementation or library stub for module named "discord.ext.commands"
.venv\lib\site-packages\discord\app_commands\commands.py:72: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 1 source file)
```

SaidBySolo:
Actually, I don't know.

I mostly use strict type hints.

However, when I checked using the following command ``mypy sample``, I was able to reproduce that there was no problem.

There seems to be no problem in normal use.

Can you tell me what your opinion is? @mikeshardmind

SaidBySolo:
@mikeshardmind 
Changing the mypy configuration in my project is a bit tricky because of one discord.py.

Are there any problems with these code changes?

mikeshardmind:
> I don't like changing the configuration of my mypy because of one discord.py.

mypy's error there literally tells you to do this if this is the reason for it not detecting the import, see the link in the error message. Beyond this, you can change it on a per-package basis. you don't need to change the behavior globally.

> Are there any problems with these code changes?
Possibly (such as inconsistencies in import, churn for no tangible gain, etc) but I wasn't exactly reviewing the code, just pointing out that this shouldn't be an issue.

SaidBySolo:
In my opinion, considering the issues mentioned above, I think these changes are meaningful changes to maintain consistency.

Rapptz (commented):
Thanks. Though I can't make any promises that I'll care about this for internal code. My comment pertained to external code. What @mikeshardmind said is correct, considering discord.py makes extensive use of namespace packages you need to enable the feature for it to work as expected on mypy.

richfromm:
Fwiw, I am able to reproduce this problem with discord.py 2.0.1 and mypy 0.982 and python 3.10.7, but in a bizarre way, it alternates between failing and passing:
https://github.com/bryanforbes/discord.py-stubs/issues/171#issuecomment-1270945026
https://github.com/bryanforbes/discord.py-stubs/issues/171#issuecomment-1270949957

But from inspection (ony) I'm wondering if the fix is correct or not:
https://github.com/Rapptz/discord.py/commit/e1961676a979e0684833bd90f1b7f74fe766677c#r86069248
https://github.com/Rapptz/discord.py/commit/e1961676a979e0684833bd90f1b7f74fe766677c#r86069343

Rapptz:
Are you running mypy with `--namespace-packages`?

PR #8484, opened 2022-09-22
Added Mobile Status

bl-d:
this is kinda useless but my cat is black and white so


Rapptz:
The library is not supposed to spoof its properties. Likewise, the fact that Discord renders this differently should be considered a bug and I don't know why they don't fix it.

PR #8958, opened 2022-09-25
Add privacy_level to create_scheduled_event docstring

DJJ05:
## Summary

Previously, the `privacy_level` kwarg was absent from the `Guild.create_scheduled_event` docstring. It is now added and described.

## Checklist

- [x] This PR fixes an issue.


Rapptz:
This field is actually unused in the code so it doesn't do anything when you set it. I don't think it can be set to anything other than `guild_only`.

PR #8964, opened 2022-09-28
Add repr to the SequenceProxy

jack1142:
## Summary

Fixes lack of proper repr on inspected proxied sequences.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -224,6 +224,9 @@ class SequenceProxy(Sequence[T_co]):
             self.__proxied = list(self.__proxied)
         return self.__proxied
 
+    def __repr__(self) -> str:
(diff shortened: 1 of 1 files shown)


jack1142:
Looking a bit more through possible sequences used here, the shown repr might not be ideal in some cases:
```py
>>> guild.threads
dict_values([<Thread id=845443120635314196 name='ssss' parent=public owner_id=176070082584248320 locked=False archived=False>])
```

Should I do something like `f"SequenceProxy({self.__proxied!r})"` perhaps?

Rapptz:
I didn't really include it on purpose cause of this. Realistically `SequenceProxy({self.__proxied!r})` is probably ok

PR #8985, opened 2022-10-06
register guild_only commands as guild commands

odrling:
## Summary

This PR allows guild_only commands to share the name of a global command.
It was an issue when migrating to discord.py v2 from a personal fork as we were using that for one of the commands of our bot.
We've been running that for about a month without any issues, but I'm not sure if this would be the expected behaviour for this parameter.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/tree.py
+++ b/discord/app_commands/tree.py
@@ -44,7 +44,7 @@ from typing import (
     Union,
     overload,
 )
-from collections import Counter
+from collections import ChainMap, Counter
 
 
 from .namespace import Namespace, ResolveKey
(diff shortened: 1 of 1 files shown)


Rapptz:
Thanks but this isn't the point of the `guild_only` decorator and I'm not really sure I agree with the storage changes this would enable. It should also be noted that on top of being a decorator this is also a boolean parameter. 

Can I ask what the use case for this is? If you wanted to have a global command with the same name as a guild command you could just attach it to a guild ID and work from there.

odrling:
Well the point was to avoid hard-coding guild IDs in the code.
The guild command complements the global command (they have different subcommands) with the only exception of not being available in DM with the bot.

Rapptz:
You can use a named constant instead.

The point of `guild_only` is to implement the `dm_permission` toggle. Anything more than that is way outside the scope of the feature. Especially guild commands that aren't actually bound to any specific guild but all guilds. 

Thanks for the PR, sorry it could not be accepted at this time.

odrling:
Hmm actually, would it be fine to use just set the guilds parameter of the
command to a special value without changing the semantics of guild_only ?
With my current changes I suppose we could define a command like 
```python
@command()
@guilds(None)
async def my_command(...):
  ...
```
But ideally it would be a special value instead of `None`.

Rapptz:
The IDs aren't validated in any form so you can use whatever sentinel you want.

PR #8991, opened 2022-10-08
[commands] Add Cog.has_app_error_handler

Predeactor:
## Summary

This pull request adds `discord.ext.commands.Cog.has_app_error_handler` in order to know if the cog implements the method `cog_app_command_error`.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/cog.py
+++ b/discord/ext/commands/cog.py
@@ -519,6 +519,13 @@ class Cog(metaclass=CogMeta):
         """
         return not hasattr(self.cog_command_error.__func__, '__cog_special_method__')
 
+    def has_app_error_handler(self) -> bool:
+        """:class:`bool`: Checks whether the cog has an app error handler.
+
+        .. versionadded:: 2.1
+        """
+        return not hasattr(self.cog_app_command_error.__func__, '__cog_special_method__')
+
     @_cog_special_method
     async def cog_load(self) -> None:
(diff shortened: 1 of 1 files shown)


Rapptz:
Thanks for the PR.

I don't really like the name of this method, as it doesn't really make sense to me on first read. On top of that, the behaviour is a bit ambiguous when taking into consideration things like `GroupCog` but I'm not sure what the situation is between `Group.on_error` and `Cog.cog_app_command_error` right now.

Predeactor:
Hi, thanks for the feedback.

I've been thinking on how to add this method, and got suggested making it `Cog.cog_app_command_error`, but I wanted to rather make it `Cog.app_commands.cog_command_error`, but now thinking about it, it seems to make even less sense. Feel free to tell me which you would prefer, however, and how could the method be renamed to be more clear on its purpose.

Also, I'd like to point out I am new to 2.0 (Coming from Red), I don't know anything about the new objects the library has added since then and how I should *correctly* implement the following, as such, do feel free to give me indication in what to takes in account before making any new changes.

Rapptz:
At the bare minimum the name should be something like `has_app_command_error_handler()` since `app_error_handler` is too short and unused elsewhere. The ambiguity problem I'm unsure about. Chances are it's probably fine as-is.

PR #8993, opened 2022-10-10
allow setting guild commands to all available guilds

odrling:
## Summary

Update to #8985.
Allows to register a command to all the guilds the bot is in by setting the ALL_GUILDS value in `@guilds` or the `guild_ids` parameter.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/tree.py
+++ b/discord/app_commands/tree.py
@@ -44,7 +44,7 @@ from typing import (
     Union,
     overload,
 )
-from collections import Counter
+from collections import ChainMap, Counter
 
 
 from .namespace import Namespace, ResolveKey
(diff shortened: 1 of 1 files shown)


Rapptz:
Hello,

Sorry. I'm still not a fan of this. I'm not really sure why you'd do this over just using global commands. It's also undocumented. Realistically the only thing that'd make me be okay with this is if a way to copy guild commands over to another guild in the tree. That would be marginally more useful than greenlighting a hacky sentinel.

odrling:
I swear, I don't see how I would use global commands in this case...
If I could set `guild_only` to True for some subcommands I'd be happy with it but that doesn't seem possible (https://discordpy.readthedocs.io/en/stable/interactions/api.html?highlight=guild_only#discord.app_commands.Command.guild_only).

PR #9013, opened 2022-10-21
Implement New Select Types

NextChai:
## Summary

As of October 13th, Discord has released their [New Select Menu Components](https://discord.com/developers/docs/change-log#new-select-menu-components) to the API. This allows for a better selection of roles, members, and channels within a select. This PR tackles that implementation.

### Simple Example
```python
class MyView(discord.ui.View):
    @discord.ui.select(
        cls=discord.ui.ChannelSelect,
        channel_types=[discord.ChannelType.public_thread, discord.ChannelType.text],
        placeholder='Select a channel',
        max_values=10,
    )
    async def channel_select(self, interaction: discord.Interaction, select: discord.ui.ChannelSelect[Self]) -> None:
        return await interaction.response.send_message(
            f'You selected: {", ".join([c.mention for c in select.values])}'
        )
```

## Checklist

- [x] If code changes were made, they were tested.
- [ ] This PR fixes an issue.


discord/enums.py
```diff
@@ -575,8 +575,12 @@ def __int__(self) -> int:
 class ComponentType(Enum):
     action_row = 1
     button = 2
-    select = 3
+    string_select = 3
```
LostLuma, line 578: You shouldn't rename existing enum members since this is a breaking change, at most add an alias for the new name.

discord/ui/select.py
```diff
@@ -251,77 +390,416 @@ def append_option(self, option: SelectOption) -> None:
...
+        regardless of the :attr:`discord.Intents.members` intent.
+        """
+        return super().values
+
+
+class ChannelSelect(BaseSelect[V]):
+    """Represents a UI "channel" select menu with a list of predefined options representing channels in the guild.
+    It is possible to filter the channels that are shown per type by passing the ``channel_types`` parameter.
+
+    Please note that if you use this in a direct message with a user, no channels will be displayed to the user
```
Rapptz, line 602:
```suggestion
    Please note that if you use this in a private message with a user, no channels will be displayed to the user
```

Might as well be consistent.

discord/ui/select.py
```diff
@@ -251,77 +390,416 @@ def append_option(self, option: SelectOption) -> None:
...
+
+    .. versionadded:: 2.1
+
+    Parameters
+    ------------
+    custom_id: :class:`str`
+        The ID of the select menu that gets received during an interaction.
+        If not given then one is generated for you.
+    channel_types: List[:class:`~discord.ChannelType`]
+        The types of channels to show in the select menu. If not given then all channel types are shown.
```
Rapptz, line 613:
```suggestion
        The types of channels to show in the select menu. Defaults to all channels.
```

discord/ui/select.py
```diff
@@ -251,77 +390,416 @@ def append_option(self, option: SelectOption) -> None:
...
+) -> SelectCallbackDecorator[V, BaseSelectT]:
     """A decorator that attaches a select menu to a component.
 
     The function being decorated should have three parameters, ``self`` representing
     the :class:`discord.ui.View`, the :class:`discord.Interaction` you receive and
-    the :class:`discord.ui.Select` being used.
-
-    In order to get the selected items that the user has chosen within the callback
-    use :attr:`Select.values`.
+    the :class:`BaseSelect` being used.
```
Rapptz, line 767: Implementation detail.

Rapptz (changes requested):
Due to a flight and small vacation I don't have time to review this right now. This is a preliminary review from a very first initial look.

discord/components.py
```diff
@@ -248,41 +248,48 @@ class SelectMenu(Component):
...
         self.placeholder: Optional[str] = data.get('placeholder')
         self.min_values: int = data.get('min_values', 1)
         self.max_values: int = data.get('max_values', 1)
         self.options: List[SelectOption] = [SelectOption.from_dict(option) for option in data.get('options', [])]
         self.disabled: bool = data.get('disabled', False)
-
-    @property
-    def type(self) -> Literal[ComponentType.select]:
-        """:class:`ComponentType`: The type of component."""
-        return ComponentType.select
```
Rapptz, line 275: Forgot to mention that this is now technically undocumented since the docstring was removed.

discord/components.py
```diff
@@ -248,41 +248,48 @@ class SelectMenu(Component):
...
-            'options': [op.to_dict() for op in self.options],
             'disabled': self.disabled,
         }
+        options = getattr(self, 'options', None)
+        if options:
+            payload['options'] = [op.to_dict() for op in options]
+
+        channel_types = getattr(self, 'channel_types', None)
+        if channel_types:
+            payload['channel_types'] = [t.value for t in channel_types]
```
Rapptz, line 292: This code definitely needs work. You should take advantage of the truth-y aspect instead since right now this just looks dirty with attributes not existing that should just because of an internal detail. Using an empty list as the options or the channel_types accomplishes the same exact thing for the underlying type.

discord/ui/select.py
```diff
@@ -251,77 +383,415 @@ def append_option(self, option: SelectOption) -> None:
...
+    @property
+    def values(self) -> List[Union[AppCommandChannel, AppCommandThread]]:
+        """List[Union[:class:`~discord.app_commands.AppCommandChannel`, :class:`~discord.app_commands.AppCommandThread`]]: A list of channels selected by the user."""
+        return super().values  # type: ignore
+
+
+@overload
 def select(
     *,
+    cls: Type[SelectT],
```
Zephyrkul, line 662:
```suggestion
    cls: Type[SelectT] = ...,
```

NextChai: This caused issues actually: 
```python
@overload
def select(
    *,
    cls: Type[SelectT] = ...,
    ...
)
```
```python
Type variable "SelectT" may go unsolved if caller supplies no argument for parameter "cls"
  Provide an overload that specifies the return type when the argument is not supplied
```

The appropriate solution I discovered was to set the default in this overload to `Select`. In my testing this caused no additional issues and solved the issue 👍
```python
@overload
def select(
    *,
    cls: Type[SelectT] = Select,
    ...
)

File: discord/ui/select.py [hunk 8]
```diff
@@ -171,6 +194,113 @@ def max_values(self) -> int:
...
+        )
+
+    @property
+    def values(self) -> List[str]:
+        """List[:class:`str`]: A list of values that have been selected by the user."""
+        return super().values  # type: ignore
+
+    @property
+    def type(self) -> Literal[ComponentType.string_select]:
+        """:class:`.ComponentType`: The discord type of this component."""
```
Rapptz, line 301:
```suggestion
        """:class:`.ComponentType`: The type of this component."""
```

Same for the rest of them.

discord/ui/select.py
```diff
@@ -340,25 +812,37 @@ def select(
...
         Whether the select is disabled or not. Defaults to ``False``.
     """
 
-    def decorator(func: ItemCallbackType[V, Select[V]]) -> ItemCallbackType[V, Select[V]]:
+    def decorator(func: ItemCallbackType[V, BaseSelectT]) -> ItemCallbackType[V, BaseSelectT]:
         if not inspect.iscoroutinefunction(func):
             raise TypeError('select function must be a coroutine function')
+        if not issubclass(cls, BaseSelect):
+            raise TypeError(
+                f'cls must be one of {", ".join(c.__name__ for c in BaseSelect.__subclasses__())} or a subclass of one of them, not {cls!r}.'
```
Rapptz, line 829: `__subclasses__` is a CPython implementation detail, and is not actually documented anywhere to my knowledge so I try to avoid using it. Not to mention that join inside an f-string looks incredibly busy.

discord/ui/select.py
```diff
@@ -21,68 +21,81 @@
...
+    from typing_extensions import TypeAlias, Self
 
     from .view import View
     from ..types.components import SelectMenu as SelectMenuPayload
-    from ..types.interactions import (
-        MessageComponentInteractionData,
-    )
+    from ..types.interactions import SelectMessageComponentInteractionData
+    from ..app_commands import AppCommandChannel, AppCommandThread
+    from discord import Interaction, Member, Role, User
```
Rapptz, line 57: Don't import from `discord` like this, only the `ext` packages do this.

Rapptz (changes requested):

discord/ui/select.py
mikeshardmind: For future reference, this one is [documented](https://docs.python.org/3/library/stdtypes.html#class.__subclasses__) but I don't think using it here is a real benefit.

PR #9020, opened 2022-10-23
docs: Fix usage of single backticks to links and code-blocks

DenverCoder1:
## Summary

Fixes over 25 occurrences of single-backticks to reference links or double-backticks depending on the case.

This helps with readability since single-backtick phrases are italicized rather than set in inline-codeblocks.

image

image

## Checklist

- [ ] This PR fixes an issue.


Rapptz (approved):
Thanks. I think in the future I'll just set the default role to be code so this isn't an issue going forward.

PR #9027, opened 2022-10-28
Add __int__ method to BaseFlags

rdrescher909:
## Summary

Adds a __int__ method to BaseFlags. Allows you to get the value using int(obj)

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/flags.py
+++ b/discord/flags.py
@@ -159,6 +159,9 @@ class BaseFlags:
 
     def __repr__(self) -> str:
         return f'<{self.__class__.__name__} value={self.value}>'
+    
+    def __int__(self) -> int:
+        return self.value
 
     def __iter__(self) -> Iterator[Tuple[str, bool]]:
         for name, value in self.__class__.__dict__.items():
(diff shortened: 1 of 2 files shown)


Rapptz:
Hey thanks for the PR.

This is deliberately not implemented since nothing else really does any type of int <-> object conversion. What's the use case of this over just using .value? 

Conceptually a flag may be represented as an integer but that's not strictly true now, especially with the automod flags.

rdrescher909:
I suppose there isn't really a beneficial use case over just using .value apart from it being a shortcut. This originally spawned out of a help channel question regarding how to get the int representation of a Permissions (they had tried int(Permissions) and it didn't work).

After thinking about it I guess it just kind of made sense to me that `int(Permissions)` would give you the integer representation of the Permissions you had. Felt like an intuitive way to get the value and I couldn't think of any explicit downsides.

I figured adding it to the rest of the Flags based objects for the sake of consistency would be better than just adding it onto Permissions specifically.

PR #9047, opened 2022-11-05
improve app_commands.Range parameter type checking and documentation

HexCodeFFF:
## Summary

the values passed for `min` and `max` for the Range are now properly checked against the given Range type. the original code just checked if the types of `min` and `max` were the same which doesn't work if both types are wrong and erroneously makes ints not work as `min`/`max` for float ranges.

I also improved the documentation as the original documentation didn't show examples for floats or strings.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/transformers.py
+++ b/discord/app_commands/transformers.py
@@ -599,6 +599,7 @@ else:
             else:
                 raise TypeError(f'expected int, float, or str as range type, received {obj_type!r} instead')
 
+            # ensure min and max types are correct
             if obj_type is int or obj_type is str:
                 if min is None or type(min) != int:
                     raise TypeError(f'expected min to be int, got {type(min)}')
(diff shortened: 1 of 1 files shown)


HexCodeFFF:
sidenote: might be useful to add an `async def convert` method to these for compatibility with hybrid/classic commands. i may do it myself in a PR later

Rapptz:
> erroneously makes ints not work as min/max for float ranges.

That's by design.

HexCodeFFF:
In my testing they work just fine. I realize it’s not hard to add “.0” after a number but I don’t feel it’s necessary

HexCodeFFF:
If this pull is merged, I think similar changes should be applied to `commands.Range` as well
I'll update the pull later

discord/app_commands/transformers.py
```diff
@@ -548,6 +548,9 @@ class Range:
         """A type annotation that can be applied to a parameter to require a numeric or string
         type to fit within the range provided.
 
+        for ints or floats, the parameter is restricted between the minimum and maximum value.
+        for strings, the length of the string is restricted between the minimum and maximum value.
```
numbermaniac, line 552:
```suggestion
        For ints or floats, the parameter is restricted between the minimum and maximum value.
        For strings, the length of the string is restricted between the minimum and maximum value.
```

PR #9055, opened 2022-11-10
Persistent views: Change to get_event_loop()

julien777z:
## Summary

I've run into an issue on my bot where defining a persistent view caused an error saying no event loop was running, so while this may be an edge case on my bot causing it to fail, I changed it to use get_event_loop() instead of get_running_loop(), and it worked, and this should not introduce any breaking changes. So, I believe get_event_loop() is a better choice here.

## Checklist

- [x] This PR fixes an issue.


Rapptz:
The asyncio initialisation in discord.py changed to use `asyncio.run` which creates a new event loop, so using `get_event_loop` before this happens would cause a the view to be attached to a different event loop than the one the library is using.

PR #9068, opened 2022-11-15
added setter method for ChannelSelect.channel_types

cibere:
## Summary

When ##9013 was added, it brought the new select menus that discord added. Though ChannelSelect.channel_types did not have a setter method while Select.options does, so I've added the setter method to it 

## Checklist

### DIFF
--- a/discord/ui/select.py
+++ b/discord/ui/select.py
@@ -657,6 +657,11 @@ class ChannelSelect(BaseSelect[V]):
 
     @channel_types.setter
     def channel_types(self, value: List[ChannelType]) -> None:
+        if not isinstance(value, list): 
+            raise TypeError('channel_types must be a list of ChannelType')
+        if not all(isinstance(obj, ChannelType) for obj in value):
+            raise TypeError('all list items must be a ChannelType')
+            
         self._underlying.channel_types = value
     
     @property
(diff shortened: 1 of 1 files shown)


Rapptz (changes requested):
In the `ui` module, setters are usually type checked to ensure the invariant is upheld.

cibere:
> In the `ui` module, setters are usually type checked to ensure the invariant is upheld.

Though the decorator doesn't type check it, so wouldn't inconsistent?

JDJGInc:
> > In the `ui` module, setters are usually type checked to ensure the invariant is upheld.
> 
> Though the decorator doesn't type check it, so wouldn't that make it inconsistent?

I have confirmed this as it seems that all the other setters in the file look practically the same

Rapptz:
> Though the decorator doesn't type check it, so wouldn't that make it inconsistent?

Kind of. The decorator doesn't do it but setters do.

> I have confirmed this as it seems that all the other setters in the file look practically the same

Confirmed what? Here's a setter from the `Select` itself:

https://github.com/Rapptz/discord.py/blob/029ed6c46d388c1e07a42b7905f3c17df59bdb9d/discord/ui/select.py#L312-L319

Note that it does type checking to ensure you're not passing it garbage.

JDJGInc:
Danny, cibere has updated it
 could you please run the checks and check the pull request please?
Thanks

Rapptz (approved):
Thanks.

PR #9075, opened 2022-11-17
Add async def convert to RangeTransformer for compatability with classic/hybrid Converters

HexCodeFFF:
## Summary

I added an `async def convert` method to `RangeTransformer` so that when used in a classic/hybrid command, the expected range value is enforced. However, I wasn't able to import `discord.ext.commands.errors.BadArgument` so I raise `discord.app_commands.errors.TransformerError` instead. I'm not sure how to fix this without making more changes than I'm comfortable making so I'm submitting the pull as-is for yall to look at and maybe make your own changes so it can work properly.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/transformers.py
+++ b/discord/app_commands/transformers.py
@@ -387,6 +387,46 @@ class RangeTransformer(IdentityTransformer):
     @property
     def max_value(self) -> Optional[Union[int, float]]:
         return self._max
+    
+    async def convert(self, ctx, argument):
+        # compatability with classic converters for hybrid command usage
+        # TransformerError used because i cant import BadArgument
+        if self._opt_type == AppCommandOptionType.string:
+            if self._min and len(argument) < self._min:
+                raise TransformerError(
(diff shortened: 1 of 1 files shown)


discord/app_commands/transformers.py
```diff
@@ -387,6 +388,46 @@ def min_value(self) -> Optional[Union[int, float]]:
     @property
     def max_value(self) -> Optional[Union[int, float]]:
         return self._max
+    
+    async def convert(self, ctx, argument):
```
CaedenPH, line 392: What is the return type and the type of both `ctx` and `argument`?

HexCodeFFF: ctx is of discord.ext.commands.Context, but I can’t import that because it causes a circular import error. The return type is also dependent on self._opt_type but I suppose a typing.Union[int,float,str] would work

HexCodeFFF: Oh and argument should just be a str

Rapptz:
Hybrid commands use commands.Range not app_commands.Range

HexCodeFFF:
i am very stupid and did not realize that existed, my badddddddddd

PR #9076, opened 2022-11-17
Uppercase color codes

Guddi8:
## Summary
Capitalize all hex color codes in the docstring so they are the same for all colors

## Checklist

- [ ] This PR fixes an issue.


Guddi8:
It is not meant to be useful, it should look better and be the same for all the colors

olijeffers0n (approved):
I'll approve it as an "It won't break anything" because it does update the docs to reflect the Python Implementation, however, the hex codes are more often written in lowercase so it is up to a vote/the mergers opinion here

PR #9078, opened 2022-11-19
[commands] Raise CheckFailure when cog check fails

dim-witted:
## Summary

Raise a `CheckFailure` when `cog_check`s fail, with its own error message. The original error message implied an issue related to the checks on the command itself and not the entire cog, so this just to easier differentiate between cog check and command check.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
Hello, thanks for the PR.

This is very much something I would prefer library users would do rather than the library since this is a rather slight shift in behaviour. If you want to differentiate between these then ideally you would have separate exceptions for these in your cog check.

PR #9082, opened 2022-11-23
Fixed typo in 'discord.Status.invisible'

cibere:
## Summary

The current docs have:
```
The member is “invisible”. In reality, this is only used in sending a presence a la Client.change_presence(). When you receive a user’s presence this will be offline instead.
```
though there is a typo with `a la`. I changed `a la` to `update to` in this PR

## Checklist

- [x] This PR fixes an issue.


cibere:
apparently this isn't a typo, so I will close this

PR #9093, opened 2022-11-27
Fix async iterators with before/after limits requesting past their bounds

Eta0:
## Summary

This affects `Messageable.history`, `ScheduledEvent.users`, `Client.fetch_guilds`, and `Guild.audit_logs`.

To illustrate the problem, `Messageable.history` counted returned messages to tell when to stop iteration, but did so before filtering away those past the `before` or `after` boundaries. When both `oldest_first=False` and an `after` boundary were provided, this led to the history iterator continuing to retrieve messages older than the `after` boundary, which would then all be filtered away, continuing until the message limit or the beginning of the entire channel was reached. A similar situation would also occur with `oldest_first=True` and a `before` boundary provided.

This commit changes the logic in these methods to count items after filtering, so they stop requesting more as soon as the in-bounds items are exhausted.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -1774,23 +1774,25 @@ class Messageable:
         channel = await self._get_channel()
 
         while True:
-            retrieve = min(100 if limit is None else limit, 100)
+            retrieve = 100 if limit is None else min(limit, 100)
             if retrieve < 1:
                 return
 
             data, state, limit = await strategy(retrieve, state, limit)
-
-            # Terminate loop on next iteration; there's no data left after this
-            if len(data) < 100:
-                limit = 0
 
             if reverse:
                 data = reversed(data)
             if predicate:
                 data = filter(predicate, data)
 
-            for raw_message in data:
+            count = 0
+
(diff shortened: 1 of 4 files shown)


Rapptz (commented):
Thanks for the PR.

First of all, nice find.

I do have some comments:

1. You're changing the calls to `min` even though it's not relevant to the PR, so after rereading a few times I realised this was purely a stylistic change.
2. Using the shadowing trick for getting the count is a bit of an ugly pattern.

It won't impact the merge of this PR but I felt it was worth mentioning.

Eta0:
Fair points. Index variables not having their own scope is [a bit contentious](https://eli.thegreenplace.net/2015/the-scope-of-index-variables-in-pythons-for-loops/) (the article linked actually features that pattern). It sort of loses its only point for conciseness once you need to declare `count` in advance to prevent it potentially being undefined. After reviewing it again I would also replace the `enumerate` with just `count += 1` in the loop.

Thanks for the quick review & merge!

PR #9109, opened 2022-12-04
Created an emotionbot

2021H1030039G:
This bot is an example that analyses user emotions and responds accordingly. It is a repl.it based script that performs 'CRUD' operations for basic keywords related to emotion.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/emotionbot.py
+++ b/examples/emotionbot.py
@@ -1,0 +1,96 @@
+import discord
+import os
+import requests
+import json
+import random
+from replit import db
+
+client = discord.Client(intents=discord.Intents.default())
+sad_words = ["sad", "depressed", "unhappy", "angry", "miserable"]
+
+starter_encouragements = [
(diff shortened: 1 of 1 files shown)


cyrus01337:
I don't think this would work out as a valid example considering there are a number of problems with it (`requests` in an async app which causes blocking, not using the `commands` extension when clearly using commands to name a couple), even if this was a valid use case for an example, others exist that already showcase the features demonstrated far better.

Rapptz:
Thanks for the PR.

Unfortunately, I can't accept it as there are too many issues that would take time to correct with incorrect programming practices pertaining to asynchronous programming.

While I can appreciate the ease of use when it comes to deploying on repl.it, the service uses shared IPs which causes frequent issues on Discord when an IP gets rate limited causing problems for our support.

PR #9127, opened 2022-12-08
add type option for overwrites_for

Devkalion:
## Summary

Fix typechecking for overwrites_for Channel method

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -557,7 +557,7 @@ class GuildChannel:
 
         Parameters
         -----------
-        obj: Union[:class:`~discord.Role`, :class:`~discord.abc.User`, :class:`~discord.Discord`, :class:`~discord.Object`]
+        obj: Union[:class:`~discord.Role`, :class:`~discord.abc.User`, :class:`~discord.Member`, :class:`~discord.Object`]
             The role or user denoting whose overwrite to get.
 
         Returns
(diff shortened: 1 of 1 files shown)


discord/abc.py
```diff
@@ -552,12 +552,12 @@ def created_at(self) -> datetime:
...
         return utils.snowflake_time(self.id)
 
-    def overwrites_for(self, obj: Union[Role, User, Object]) -> PermissionOverwrite:
+    def overwrites_for(self, obj: Union[Role, User, Member, Object]) -> PermissionOverwrite:
         """Returns the channel-specific overwrites for a member or a role.
 
         Parameters
         -----------
-        obj: Union[:class:`~discord.Role`, :class:`~discord.abc.User`, :class:`~discord.Object`]
+        obj: Union[:class:`~discord.Role`, :class:`~discord.abc.User`, :class:`~discord.Discord`, :class:`~discord.Object`]
```
AbstractUmbra, line 560: I don't think `discord.Discord` is correct...?

Also `User` here infers `abc.User` which is the abstract base class that both `discord.User` and `discord.Member` adhere to.

Devkalion: fix typo `discord.Discord`

as I see member does not inherits `abc.User`

```
class Member(discord.abc.Messageable, _UserTag):
```

AbstractUmbra: It does not need direct inheritance as it is a Snowflake | Protocol:-

<https://github.com/Rapptz/discord.py/blob/799e3c5bf09cedcd9b7e2b83df733910fa11532a/discord/abc.py#L199-L207>

PR #9128, opened 2022-12-08
Fix embed edge cases

Soheab:
## Summary

This PR fixes 2 things about `Embed`:

1. `.copy()` not including the following keys if any were assigned to an empty value (`""` for example): `type, title, description, url`.
Discussed in the bikeshedding thread ["Potential edge case for Embed.copy()"](https://canary.discord.com/channels/336642139381301249/1048902326586445886) in the server
~~2. `.from_dict()` not supporting `"colour"` as a key in the dict.
Reported by `Jack#7382` in the server [here](https://canary.discord.com/channels/336642139381301249/336642776609456130/1049371481576058880) (#general)~~ Reverted per [this comment](https://github.com/Rapptz/discord.py/pull/9128#issuecomment-1343495059)

I'm unsure if this is a breaking change or needs a `versionadded` string. Discord does not care about empty fields, so no errors are thrown.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -228,9 +228,15 @@ class Embed:
         # try to fill in the more rich fields
 
         try:
-            self._colour = Colour(value=data['color'])
+            color = data['color']
         except KeyError:
-            pass
+            try:
+                color = data['colour']
+            except KeyError:
+                color = None
+
+        if color is not None:
+            self._colour = Colour(value=color)
 
         try:
             self._timestamp = utils.parse_time(data['timestamp'])
@@ -742,16 +748,16 @@ class Embed:
                     result['timestamp'] = timestamp.replace(tzinfo=datetime.timezone.utc).isoformat()
 
         # add in the non raw attribute ones
-        if self.type:
+        if self.type is not None:
             result['type'] = self.type
 
-        if self.description:
+        if self.description is not None:
             result['description'] = self.description
 
-        if self.url:
+        if self.url is not None:
             result['url'] = self.url
 
-        if self.title:
+        if self.title is not None:
             result['title'] = self.title
 
         return result  # type: ignore # This payload is equivalent to the EmbedData type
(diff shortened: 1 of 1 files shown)


discord/embeds.py
```diff
@@ -228,9 +228,15 @@ def from_dict(cls, data: Mapping[str, Any]) -> Self:
...
 
         try:
-            self._colour = Colour(value=data['color'])
+            color = data['color']
         except KeyError:
-            pass
+            try:
+                color = data['colour']
+            except KeyError:
+                color = None
```
CaedenPH, line 236:
```suggestion
            color = data.get('colour', None)
```

discord/embeds.py
```diff
@@ -228,9 +228,15 @@ def from_dict(cls, data: Mapping[str, Any]) -> Self:
...
+
+        if color is not None:
+            self._colour = Colour(value=color)
```
CaedenPH, line 239: What happens if color is `None`? Is `_colour` referenced anywhere else where it might break?

Zephyrkul: See the `colour` property starting on line 318, as well as the fact that `self._colour` would also not be set on this logical path in current master

discord/embeds.py
Soheab: Unsure about that since it's either a personal preference or better practice according to Danny to use `try-except`. So going to wait on Danny for that.

Zephyrkul: There's no extra side effects being introduced by using `.get` here (e.g. no object being over-eagerly created, no necessity for differentiating between the key being present or not), so I doubt using it here would be problematic

Rapptz:
There's no alias on purpose since it's meant to be an exact payload from Discord or the output of to_dict, anything else is out of scope.

Soheab:
> There's no alias on purpose since it's meant to be an exact payload from Discord or the output of to_dict, anything else is out of scope.

Ah I was thinking it might've been an oversight because it's supposed to construct an Embed from a dict but fair enough if that's not the purpose. I have reverted it.

PR #9129, opened 2022-12-08
Update tree.py

Jevimartinez:
## Summary
Make the `guilds` parameter a keyword argument in order to avoid an error if it's not provided.

## Checklist
- [ x] This PR fixes an issue.


Rapptz:
This is explicitly by design required. With no parameters passed it looks like you're clearing every command in the tree. Making it explicit makes the reader know the intent of the code better.

PR #9157, opened 2022-12-26
Improving examples | Cogs (Not extensions)

### LINKED ISSUE #2525: Improving examples

bleudev:
## Summary

Add example with cogs (not extensions) from this issue #2525
(fix #2525)

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/examples/cogs/basic_not_extension.py
+++ b/examples/cogs/basic_not_extension.py
@@ -46,12 +46,10 @@ class Bot(commands.Bot):
             **options
         )
 
-    async def on_ready(self):  # Install cogs
-        print('Start installing cogs!')
+    async def setup_hook(self):
+        print('The start of cogs installation')
         await self.add_cog(MyCog(self))
-        print('Finish installing cogs!')
-
-        print("I'm ready")
+        print('The end of cogs installation')
 
 
 intents = discord.Intents.default()
--- a/examples/cogs/basic_not_extension.py
+++ b/examples/cogs/basic_not_extension.py
@@ -10,8 +10,8 @@ from typing import Optional, Type, Any
 
 class MyCog(commands.Cog):
     def __init__(self, bot) -> None:
-        self.bot = bot # Assign bot who is setuping cog
-    
+        self.bot = bot  # Assign bot who is setuping cog
+
     @commands.Cog.listener()
     async def on_message(self, message: discord.Message):
         if message.author.bot:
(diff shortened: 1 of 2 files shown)


GoogleGenius:
Please run this through black.

examples/cogs/basic_not_extension.py
```diff
@@ -0,0 +1,58 @@
...
+            await message.reply('Pong!')
+
+    @commands.bot_has_permissions(administrator=True)
+    @commands.command()
+    async def admin_check(self, ctx: commands.Context):
+        await ctx.send('You are admin!')
+
+# Create bot with cogs
+class Bot(commands.Bot):
+    def __init__(
```
GoogleGenius, line 29: Unnecessary...

ArthurKun021: I personally like having `_init_` there so you I can use the type hinting for the when I initialized some variables

pixeldeee:
> Please run this through black.

Ok

examples/cogs/basic_not_extension.py
pixeldeee: > Unnecessary to override the base class `__init__`

Maybe remove this method?

examples/cogs/basic_not_extension.py
```diff
@@ -0,0 +1,62 @@
...
+            help_command=help_command,
+            tree_cls=tree_cls,
+            description=description,
+            intents=intents,
+            **options
+        )
+
+    async def on_ready(self):  # Install cogs
+        print('Start installing cogs!')
+        await self.add_cog(MyCog(self))
```
LucasCoderT, line 51: It's heavily recommended to use `setup_hook` to do this.

pixeldeee:
```suggestion
    async def setup_hook(self):
        await self.add_cog(MyCog(self))
```
This code?

GoogleGenius: yes

pixeldeee:
@Rapptz pls review this pull

PR #9166, opened 2023-01-08
Add str and float examples to app_commands.Range documentation

lightlybakedman:
## Summary

Add examples for str and float to the documentation for app_commands.Range transformer as it was not clear what types could be used before, other than int.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/transformers.py
+++ b/discord/app_commands/transformers.py
@@ -553,10 +553,11 @@ else:
 
         Some example ranges limiting numeric input:
 
-        - ``Range[int, 10]`` means the input type is `int`, and the minimum is 10 with no maximum.
-        - ``Range[int, None, 10]`` means the input type is `int`, and the maximum is 10 with no minimum.
-        - ``Range[int, 1, 10]`` means the input type is `int`, and the minimum is 1 and the maximum is 10.
-        - ``Range[float, 1.0, 5.0]`` means the input type is `float`, and the minimum is 1.0 and the maximum is 5.0.
+        - ``Range[int, 10]`` means the minimum is 10 with no maximum.
+        - ``Range[int, None, 10]`` means the maximum is 10 with no minimum.
+        - ``Range[int, 1, 10]`` means the minimum is 1 and the maximum is 10.
+        - ``Range[float, 1.0, 5.0]`` means the minimum is 1.0 and the maximum is 5.0.
+        - ``Range[str, 1, 10]`` means the minimum length is 1 and the maximum length is 10.
 
         Some example ranges limiting string input:
 
--- a/discord/app_commands/transformers.py
+++ b/discord/app_commands/transformers.py
@@ -559,12 +559,6 @@ else:
         - ``Range[float, 1.0, 5.0]`` means the minimum is 1.0 and the maximum is 5.0.
         - ``Range[str, 1, 10]`` means the minimum length is 1 and the maximum length is 10.
 
-        Some example ranges limiting string input:
-
-        - ``Range[str, 10]`` means the minimum length is 10 with no maximum length.
-        - ``Range[str, None, 10]`` means the maximum length is 10 with no minimum length.
-        - ``Range[str, 1, 10]`` means the minimum length is 1 and the maximum length is 10.
-
         .. versionadded:: 2.0
 
         Examples
--- a/discord/app_commands/transformers.py
+++ b/discord/app_commands/transformers.py
@@ -567,6 +567,7 @@ else:
         .. code-block:: python3
 
             @app_commands.command()
+            async def range(interaction: discord.Interaction, value: app_commands.Range[int, 10, 12]):
                 await interaction.response.send_message(f'Your value is {value}', ephemeral=True)
         """
 
--- a/discord/app_commands/transformers.py
+++ b/discord/app_commands/transformers.py
@@ -551,11 +551,18 @@ else:
         During type checking time this is equivalent to :obj:`typing.Annotated` so type checkers understand
         the intent of the code.
 
-        Some example ranges:
+        Some example ranges limiting numeric input:
 
-        - ``Range[int, 10]`` means the minimum is 10 with no maximum.
-        - ``Range[int, None, 10]`` means the maximum is 10 with no minimum.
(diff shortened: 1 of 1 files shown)


discord/app_commands/transformers.py
```diff
@@ -565,7 +572,11 @@ class Range:
         .. code-block:: python3
 
             @app_commands.command()
-            async def range(interaction: discord.Interaction, value: app_commands.Range[int, 10, 12]):
+            async def numeric_range(interaction: discord.Interaction, value: app_commands.Range[int, 10, 12]):
+                await interaction.response.send_message(f'Your value is {value}', ephemeral=True)
+                
+            @app_commands.command()
+            async def string_range(interaction: discord.Interaction, value: app_commands.Range[str, 5, 10]):
```
CaedenPH, line 579: Is this example necessary?

CaedenPH (approved):
Nice work! Very useful examples

discord/app_commands/transformers.py
lightlybakedman: Definitely not necessary, but I'd say it's just useful as an additional example.

Rapptz:
Hello, thanks for the PR.

However, I'm not sure I think this is necessary? The documentation already mentions this is possible and the example syntax is easily adaptable so I'm not sure what adding more examples accomplishes since it just makes the documentation longer. Was there some sort of confusion that led to this addition or something?

lightlybakedman:
@Rapptz I think there could be confusion for users, myself included, since the existing docs don't explicitly mention what types can be used, and what "require a ... string type to fit within the range provided" means (length)

The short examples also only showcased the use of `int` and I just thought it would be more clear if examples for str and float were added, even if only 1 or 2. I do agree however that the additional code block example is unnecessary.

Soheab:
Want to add to that, everyone seems to be surprised that it's possible when it's mentioned in the help channels or outside so I don't think many know that this feature exists on discord/dpy.

discord/app_commands/transformers.py
```diff
@@ -551,11 +551,18 @@ class Range:
...
-        Some example ranges:
+        Some example ranges limiting numeric input:
 
-        - ``Range[int, 10]`` means the minimum is 10 with no maximum.
-        - ``Range[int, None, 10]`` means the maximum is 10 with no minimum.
-        - ``Range[int, 1, 10]`` means the minimum is 1 and the maximum is 10.
+        - ``Range[int, 10]`` means the input type is `int`, and the minimum is 10 with no maximum.
+        - ``Range[int, None, 10]`` means the input type is `int`, and the maximum is 10 with no minimum.
+        - ``Range[int, 1, 10]`` means the input type is `int`, and the minimum is 1 and the maximum is 10.
+        - ``Range[float, 1.0, 5.0]`` means the input type is `float`, and the minimum is 1.0 and the maximum is 5.0.
```
Rapptz, line 559:
```suggestion
        - ``Range[int, 10]`` means the minimum is 10 with no maximum.
        - ``Range[int, None, 10]`` means the maximum is 10 with no minimum.
        - ``Range[int, 1, 10]`` means the minimum is 1 and the maximum is 10.
        - ``Range[float, 1.0, 5.0]`` means the minimum is 1.0 and the maximum is 5.0.
        - ``Range[str, 1, 10]`` means the minimum length is 1 and the maximum length is 10.
```

The input type is incredibly explicit, repeating it doesn't clarify anything. I've added the string type in here to be more explicit that it's possible, but the section below this should just be removed.

discord/app_commands/transformers.py
```diff
@@ -551,11 +551,18 @@ class Range:
         During type checking time this is equivalent to :obj:`typing.Annotated` so type checkers understand
         the intent of the code.
 
-        Some example ranges:
+        Some example ranges limiting numeric input:
```
Rapptz, line 554:
```suggestion
        Some example ranges:
```

discord/app_commands/transformers.py
```diff
@@ -551,11 +551,18 @@ class Range:
...
+        - ``Range[int, 10]`` means the input type is `int`, and the minimum is 10 with no maximum.
+        - ``Range[int, None, 10]`` means the input type is `int`, and the maximum is 10 with no minimum.
+        - ``Range[int, 1, 10]`` means the input type is `int`, and the minimum is 1 and the maximum is 10.
+        - ``Range[float, 1.0, 5.0]`` means the input type is `float`, and the minimum is 1.0 and the maximum is 5.0.
+
+        Some example ranges limiting string input:
+
+        - ``Range[str, 10]`` means the minimum length is 10 with no maximum length.
+        - ``Range[str, None, 10]`` means the maximum length is 10 with no minimum length.
+        - ``Range[str, 1, 10]`` means the minimum length is 1 and the maximum length is 10.
```
Rapptz, line 565:
```suggestion
```

discord/app_commands/transformers.py
Rapptz, line 579:
```suggestion
```

Rapptz (changes requested):
I think this would have been better by being more explicit what's considered a string or numeric type. Which is, in some ways, surprising and unfortunate since these are just fundamental types when working with Python.

For now, as-is this PR is way too hand-holdy to the point of mostly stretching the docs or making the user seem dumber than they are. The target audience for the documentation is someone who already knows Python to an intermediate degree, which is (probably surprisingly) a higher bar than the one in the help channels or the support server.

discord/app_commands/transformers.py
```diff
@@ -565,7 +567,6 @@ class Range:
         .. code-block:: python3
 
             @app_commands.command()
-            async def range(interaction: discord.Interaction, value: app_commands.Range[int, 10, 12]):
                 await interaction.response.send_message(f'Your value is {value}', ephemeral=True)
```
Rapptz, line 570:
```suggestion
            async def range(interaction: discord.Interaction, value: app_commands.Range[int, 10, 12]):
                await interaction.response.send_message(f'Your value is {value}', ephemeral=True)
```

Rapptz (approved):
Thanks.

PR #9171, opened 2023-01-13
Fall back to Object for AuditLogEntry.target

Zephyrkul:
## Summary

Fixes `AuditLogEntry.target` sometimes being `None` despite there being a valid `_target_id` due to cache misses. Instead it'll return `Object` under this PR.

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/discord/audit_logs.py
+++ b/discord/audit_logs.py
@@ -719,7 +719,10 @@ class AuditLogEntry(Hashable):
                 return None
             return Object(id=self._target_id)
         else:
-            return converter(self._target_id)
+            converted = converter(self._target_id)
+            if self._target_id is not None and converted is None:
+                return Object(id=self._target_id)
+            return converted
 
     @utils.cached_property
     def category(self) -> Optional[enums.AuditLogActionCategory]:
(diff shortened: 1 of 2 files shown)


z03h:
`_convert_target_member` could be changed to return an Object with type Member.

Rapptz (changes requested):
Most transformers do this already, it's probably better to change the ones that don't so they do since it makes it possible to attach type information to the `Object`.

There might be a documentation change needed for that change too.

Zephyrkul:
> Most transformers do this already, it's probably better to change the ones that don't so they do since it makes it possible to attach type information to the `Object`.
Noticed that after looking into z03h's comment, will do

> There might be a documentation change needed for that change too.
Fair deal, on it

Zephyrkul:
> `_convert_target_member` could be changed to return an Object with type Member.

There's no such thing as this.

Rapptz:
It's `_convert_target_user` and `_convert_target_message`.

Zephyrkul:
Should be all set

Rapptz (approved):
Thanks!

PR #9175, opened 2023-01-14
update documentation of discord.user.avatar

AF-Source:
## Summary

currently, discord.user.avatar has an ambiguous description leaving people confused as to why avatar is not returned. This PR makes it more clear that in case of default avatars, user.avatar returns None.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/user.py
+++ b/discord/user.py
@@ -153,7 +153,7 @@ class BaseUser(_UserTag):
     def avatar(self) -> Optional[Asset]:
         """Optional[:class:`Asset`]: Returns an :class:`Asset` for the avatar the user has.
 
-        If the user does not have a traditional avatar, ``None`` is returned.
+        If the user has a non-traditional avatar or a global default discord avatar, ``None`` is returned.
         If you want the avatar that a user has displayed, consider :attr:`display_avatar`.
         """
         if self._avatar is not None:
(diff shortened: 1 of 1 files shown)


AF-Source:
Furthermore, the current documentation has "traditional avatar" which is not explained anywhere. I assume it means default avatars so maybe if that's the only case of "non traditional avatar", the wording can be removed completely

discord/user.py
```diff
@@ -153,7 +153,7 @@ def public_flags(self) -> PublicUserFlags:
     def avatar(self) -> Optional[Asset]:
         """Optional[:class:`Asset`]: Returns an :class:`Asset` for the avatar the user has.
 
-        If the user does not have a traditional avatar, ``None`` is returned.
+        If the user has a non-traditional avatar or a global default discord avatar, ``None`` is returned.
```
Rapptz, line 156:
```suggestion
        If the user has not uploaded an avatar, ``None`` is returned.
```

AF-Source: This is ambiguous since a user can upload a guild avatar but not his global one. I think you have to say 'global avatar'.

AF-Source:
```suggestion
        If the user has not uploaded a global avatar, ``None`` is returned.
```

Rapptz (approved):
Thanks.

PR #9186, opened 2023-01-18
update content parameter to not optional

stonymf:
Server returns `discord.errors.HTTPException: 400 Bad Request (error code: 50006): Cannot send an empty message` if content parameter not supplied

## Summary

We found an issue when trying to implement forum capabilities in our bot. It turns out that you have to supply a string for the content parameter in order for the Discord API to accept the request. Otherwise a 400 Bad Request is returned. Once we began supplying the content parameter, the forum threads were created without issue.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -2515,9 +2515,9 @@ class ForumChannel(discord.abc.GuildChannel, Hashable):
         self,
         *,
         name: str,
+        content: str,
         auto_archive_duration: ThreadArchiveDuration = MISSING,
         slowmode_delay: Optional[int] = None,
-        content: Optional[str] = None,
         tts: bool = False,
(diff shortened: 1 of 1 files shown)


Rapptz:
You can send embed only messages.

stonymf:
@Rapptz not sure I understand.

Rapptz:
The parameter is optional because you don't have to provide it. You *can* send a message without content, provided that you either upload an attachment or provide an embed. This change breaks that behaviour and is an incorrect breaking change.

stonymf:
@Rapptz roger that. Might gently suggest that the docs mention which parameters are necessary in order for the request to succeed.

PR #9189, opened 2023-01-18
Document (Group)Cog interaction_check

Soheab:
## Summary

A couple of lines that were left over from when `app_commands.Group` and `commands.Cog` could be subclassed call a method named `interaction_check` on the binding but this was never documented. Since there is an intense desire for a cog-level app commands check, I have defined such special method in `commands.Cog` and documented it on `GroupCog`.

This was discussed in following bikeshedding thread: ["Cog.cog_check for app commands"](https://canary.discord.com/channels/336642139381301249/1060563548138381333) in the server.

## Checklist

- [ ] This PR fixes an issue.


discord/ext/commands/cog.py
```diff
@@ -585,6 +585,16 @@ def cog_check(self, ctx: Context[BotT]) -> bool:
         """
         return True
 
+    @_cog_special_method
+    def interaction_check(self, interaction: discord.Interaction[BotT], /) -> bool:
+        """A special method that registers as a :func:`discord.app_commands.check`
+        for every app command and subcommand in this cog.
+
+        This function **can** be a coroutine and must take a sole parameter,
+        ``interaction``, to represent the :class:`~discord.Interaction`.
```
Rapptz, line 594: Needs a versionadded, which is actually 2.0 and not 2.2.

discord/ext/commands/cog.py
```diff
@@ -585,6 +585,16 @@ def cog_check(self, ctx: Context[BotT]) -> bool:
         """
         return True
 
+    @_cog_special_method
+    def interaction_check(self, interaction: discord.Interaction[BotT], /) -> bool:
```
Rapptz, line 589: This has to be async.

discord/ext/commands/cog.py
```diff
@@ -585,6 +585,16 @@ def cog_check(self, ctx: Context[BotT]) -> bool:
...
+        """A special method that registers as a :func:`discord.app_commands.check`
+        for every app command and subcommand in this cog.
+
+        This function **can** be a coroutine and must take a sole parameter,
```
Rapptz, line 593:
```suggestion
        This function **must** be a coroutine and must take a sole parameter,
```

Rapptz (changes requested):

discord/ext/commands/cog.py
Soheab: But the library uses utils.maybe_coroutine? 
https://github.com/Rapptz/discord.py/blob/f7e97954950ffb0e34238d70813454caa6f1a3ae/discord/app_commands/commands.py#L991-L995

discord/ext/commands/cog.py
```diff
@@ -585,6 +585,18 @@ def cog_check(self, ctx: Context[BotT]) -> bool:
         """
         return True
 
+    @_cog_special_method
+    def interaction_check(self, interaction: discord.Interaction[BotT], /) -> bool:
```
Rapptz, line 589: This PR is failing because `BotT` doesn't quite match `ClientT` so you'll have to use the latter.

Rapptz (approved):

PR #9196, opened 2023-01-20
Add info to docs on how `py -3` ignores virtualenv

marcospgp:
## Summary

Add info to docs on how `py -3` ignores virtualenv. Tried to be concise, could include more info such as suggesting use of `py` or `python` to take current virtualenv into account.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -88,6 +88,11 @@ However, for the quick and dirty:
 
 Congratulations. You now have a virtual environment all set up.
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


CaedenPH (approved):
Approving in terms of content but unsure of when someone would use py -3

Rapptz (approved):
CI failures are unrelated.

I still don't know if this callout is worth having but there's no harm in having it I guess. Thanks.

PR #9219, opened 2023-02-04
Add dark_embed and light_embed to Colour

TheMaster3558:
## Summary

This adds a new color. The color is the same color as the background of an embed and gives an embed rounded corners.
image

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/colour.py
+++ b/discord/colour.py
@@ -398,14 +398,20 @@ class Colour:
         return cls(0xFEE75C)
 
     @classmethod
-    def embed_grey(cls) -> Self:
+    def dark_embed(cls) -> Self:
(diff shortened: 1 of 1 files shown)


LeoCx1000:
I feel it should be named `.dark_embed` to be inline with `.dark_theme`, and also add `.light_embed`.

CaedenPH (approved):
Seems like a nice edition to me. I had to hardcode those values into my bot subclass, bit of an annoyance

Rapptz (approved):
Historically I've been against these because I considered these colours kind of static and prone to changing. I think the documentation for these don't need to document the actual colour value so it can be possible to change these whenever the client updates its colour schemes. However, this is outside the scope of this PR so it'll be done later.

PR #9223, opened 2023-02-07
Add `warn_message_content` kwarg to `Bot`

Auto-5:
## Summary

There is no way to stop the missing message_content warning for the `when_mentioned_or` prefix. The warning is correct because non-mention prefixes do not work in servers but can be incorrect when the bot owner only intends to use the commands in DMs. This kwarg option will help silence the warning for the second use case.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
If you want to stop a warning or something involving the logging module then use a logging filter. Adding these to the public API doesn't make much sense to me.

PR #9225, opened 2023-02-08
Add `tree` attribute to `app_commands.Command`

Auto-5:
## Summary

This adds the attribute `.tree` which contains the `CommandTree` to `app_command.Command`. This also makes it possible to access the `Cog` with `app_command.Command` only, but the command must be added to the tree first.
```py
try:
    cog = app_command.tree.client.get_cog(app_command._callback.__qualname__.split('.')[-2])
except IndexError:
    cog = None
```

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -58,6 +58,7 @@ from ..user import User
 from ..member import Member
 from ..permissions import Permissions
 from ..utils import resolve_annotation, MISSING, is_inside_class, maybe_coroutine, async_all
+from .tree import CommandTree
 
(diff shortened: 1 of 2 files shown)


Rapptz:
What is the use case for this?

Auto-5:
The use case is `.cog` attribute to `app_command.Command`, the same as https://discordpy.readthedocs.io/en/stable/ext/commands/api.html?highlight=command#discord.ext.commands.Command.cog

Rapptz:
The tree isn't the cog so it's not really the same.

PR #9245, opened 2023-02-18
Update create_stage_channel to support all features

NarmyOnDiscord:
## Summary

Adds support while creating stage channels for
- bitrate
- user_limit
- rtc_region
- video_quality_mode

Also removes topic, since thats not needed and your code would fail when you try to create a stage channel without a topic.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -1423,8 +1423,6 @@ class Guild(Hashable):
         rtc_region: Optional[str] = MISSING,
         video_quality_mode: VideoQualityMode = MISSING,
         overwrites: Mapping[Union[Role, Member], PermissionOverwrite] = MISSING,
-        
-        
     ) -> StageChannel:
         """|coro|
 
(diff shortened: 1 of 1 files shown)


Rapptz (changes requested):
Thanks. These need `versionadded` strings.

discord/guild.py
```diff
@@ -1466,12 +1476,22 @@ async def create_stage_channel(
...
+            options['bitrate'] = bitrate
+
+        if user_limit is not MISSING:
+            options['user_limit'] = user_limit
+
+        if rtc_region is not MISSING:
+            options['rtc_region'] = None if rtc_region is None else rtc_region
+
+        if video_quality_mode is not MISSING:
+            options['video_quality_mode'] = video_quality_mode.value
```
Rapptz, line 1493: This also needs a type check like the other methods

NarmyOnDiscord:
> Thanks. These need `versionadded` strings.

2.2, right?

NarmyOnDiscord (commented):
Added in the last commit

Rapptz (approved):
Thanks!

PR #9247, opened 2023-02-19
Changes to make command work in Python 3.11

la-png:
## Summary

In Python 3.11, if I specify MISSING directly, I will get a ValueError at startup, so I have changed to use default_factory.

## Checklist

- [x] This PR fixes an issue.


Rapptz:
This was fixed a long time ago in https://github.com/Rapptz/discord.py/commit/aa18e573c3b12950bc573e29217d8c47cdd71902 so this change feels unnecessary.

PR #9249, opened 2023-02-19
fix: Remove useless overload

CaedenPH:
## Summary

Removes a useless overload

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -268,12 +268,7 @@ def parse_time(timestamp: None) -> None:
 @overload
 def parse_time(timestamp: str) -> datetime.datetime:
     ...
-
-
-@overload
-def parse_time(timestamp: Optional[str]) -> Optional[datetime.datetime]:
-    ...
-
+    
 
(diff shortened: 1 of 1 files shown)


Rapptz:
Not useless. It's the base fallback case.

CaedenPH:
@Rapptz How does that differ from the actual `parse_time` function which accepts `Optional[str]`? The overload is the same as the function types - 
What does 
```py
@overload
def parse_time(timestamp: Optional[str]) -> Optional[datetime.datetime]:
    ...

def parse_time(timestamp: Optional[str]) -> Optional[datetime.datetime]:
```
do opposed to just 
```py
def parse_time(timestamp: Optional[str]) -> Optional[datetime.datetime]:
```?

Rapptz:
The way overloads work is that they ignore the "implementation" type hints.

CaedenPH:
> The way overloads work is that they ignore the "implementation" type hints.

Oh, I wasn't aware

bijij:
Mainly so the body of the implimentation itself can be type-checked during development.

PR #9258, opened 2023-02-24
Add property 'row' to Button.

Puncher1:
## Summary

Adds the attribute `row` of `ui.Button` to a property.  This also adds missing documentation of the `row` attribute to the `Button` component class.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/components.py
+++ b/discord/components.py
@@ -167,6 +167,12 @@ class Button(Component):
         The label of the button, if any.
     emoji: Optional[:class:`PartialEmoji`]
         The emoji of the button, if available.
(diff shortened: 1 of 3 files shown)


Rapptz:
Thanks for the PR. However, there are two things that aren't quite right.

1. `Item.row` exists.
2. This is a library construct, so it doesn't belong in the `discord.types` package since Discord doesn't send it.

Puncher1:
Ah okay 👍🏻. Well I saw it but I wondered why it's just `row` which is "missing".

Issue #9261, opened 2023-02-25
ScheduledEvent edit method requires location for external events


Baviaan:
### Summary

The edit method requires a location and end_time to edit the name of external events

### Reproduction Steps

1. Create an external scheduled event
2. Try to update the name of the event

### Minimal Reproducible Code

```python
import discord
from discord.ext import commands
from discord.utils import utcnow
import json
from datetime import timedelta

with open('config.json', 'r') as f:
    config = json.load(f)
token = config['BOT_TOKEN']
intents = discord.Intents.none()
intents.guilds = True

bot = commands.Bot(command_prefix="!", case_insensitive=True, intents=intents)

@bot.event
async def on_ready():
    print("Logged in.")
    guild = bot.guilds[0]
    start_time = utcnow() + timedelta(600)
    end_time = utcnow() + timedelta(1200)
    external = discord.EntityType.external
    privacy = discord.PrivacyLevel.guild_only
    location = "https://example.com"
    event = await guild.create_scheduled_event(name="MWE", start_time=start_time, end_time=end_time, entity_type=external, privacy_level=privacy, location=location)
    await event.edit(name="Edit")
    #await event.edit(name="Edit", location=location, end_time=end_time)

bot.run(token)
```

### Expected Results

I expected the name of my event to be updated.

Per https://discord.com/developers/docs/resources/guild-scheduled-event#modify-guild-scheduled-event-json-params
the location and end time are only required if *changing* to an external event. This works correctly when calling the raw discord API directly.

### Actual Results

I got a TypeError location must be set for external events. (And if I set location, same thing for end_time.)

2023-02-25 12:40:31 WARNING  discord.ext.commands.bot Privileged message content intent is missing, commands may not work as expected.
2023-02-25 12:40:31 INFO     discord.client logging in using static token
2023-02-25 12:40:32 INFO     discord.gateway Shard ID None has connected to Gateway (Session ID: f9802665bab0c999b126bbd8ab516c4f).
Logged in.
2023-02-25 12:40:34 ERROR    discord.client Ignoring exception in on_ready
Traceback (most recent call last):
  File "\~/.local/lib/python3.10/site-packages/discord/client.py", line 409, in _run_event
    await coro(*args, **kwargs)
  File "[..]/source/mwe.py", line 25, in on_ready
    await event.edit(name="Edit")
  File "\~/.local/lib/python3.10/site-packages/discord/scheduled_event.py", line 443, in edit
[...]

PR #9264, opened 2023-02-25
Fix uneccessary arguments needed for edit scheduled event

Puncher1:
## Summary

This fixes errors which occurre when the event type is `external` and you use [`ScheduledEvent.edit`](https://discordpy.readthedocs.io/en/latest/api.html?highlight=create_scheduled_event#discord.ScheduledEvent.edit) because no `location` or `end_time` is given, even you provided those already in `create_scheduled_event`.
The updated code will check if `location` and `end_time` are already set in the event object. Only if it's not set there, it raises an error.

This fix will also automatically set `payload['channel_id']` to `None` if the `EntityType` is `external`.

This error fixes an existing issue: https://github.com/Rapptz/discord.py/issues/9261

### Testing

**Code**
```python
@bot.command()
async def edit_event(ctx: commands.Context):
    guild = ctx.guild
    channel = bot.get_channel(1007623668433682492)

    start_time = discord.utils.utcnow() + datetime.timedelta(600)
    end_time = discord.utils.utcnow() + datetime.timedelta(1200)
    type_external = discord.EntityType.external
    privacy = discord.PrivacyLevel.guild_only
    location = "https://example.com"

    for i in range(0, 5):
        try:
            if i == 0:
                # Edit name only; PASS
                event = await guild.create_scheduled_event(name="Before 0", start_time=start_time, end_time=end_time,
                                                           entity_type=type_external, privacy_level=privacy,
                                                           location=location)
                await event.edit(name="After 0")
            elif i == 1:
                # Edit to external and all required arguments; PASS
                event = await guild.create_scheduled_event(name="Before 1", channel=channel, start_time=start_time,
                                                           privacy_level=privacy)
                await event.edit(name="After 1", entity_type=type_external, end_time=end_time, location=location)

            elif i == 2:
                # Edit to external and end_time; FAIL
                event = await guild.create_scheduled_event(name="Before 2", channel=channel, start_time=start_time,
                                                           privacy_level=privacy)
                await event.edit(name="After 2", entity_type=type_external, end_time=end_time)

            elif i == 3:
                # Edit to external and location; FAIL
                event = await guild.create_scheduled_event(name="Before 3", channel=channel, start_time=start_time,
                                                           privacy_level=privacy)
                await event.edit(name="After 3", entity_type=type_external, location=location)

            elif i == 4:
                # Edit to external and only end_time is already set; FAIL
                event = await guild.create_scheduled_event(name="Before 4", channel=channel, end_time=end_time,
                                                           start_time=start_time, privacy_level=privacy)
                await event.edit(name="After 4", entity_type=type_external)

        except Exception as e:
            tb = e.__traceback__
            _, _, funcname, _ = traceback.extract_tb(tb)[-1]

            print(f"{i}  Failed in {funcname}: {e}")
        else:
            print(f"{i}  Pass")
```
**Output**
```
0  Pass
1  Pass
2  Failed in edit: location must be set when entity_type is external
3  Failed in edit: end_time must be set when entity_type is external
4  Failed in edit: location must be set when entity_type is external
```
The output looks as expected.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/scheduled_event.py
+++ b/discord/scheduled_event.py
@@ -326,9 +326,7 @@ class ScheduledEvent(Hashable):
         description: :class:`str`
             The description of the scheduled event.
         channel: Optional[:class:`~discord.abc.Snowflake`]
-            The channel to put the scheduled event in. If the channel is
-            a :class:`StageInstance` or :class:`VoiceChannel` then
-            it automatically sets the entity type.
+            The channel to put the scheduled event in. This will automatically set the entity type.
 
             Required if the entity type is either :attr:`EntityType.voice` or
             :attr:`EntityType.stage_instance`.
--- a/discord/scheduled_event.py
+++ b/discord/scheduled_event.py
@@ -426,38 +426,37 @@ class ScheduledEvent(Hashable):
             payload['entity_type'] = entity_type.value
 
         _entity_type = entity_type or self.entity_type
+        _channel = channel or self.channel
+        _location = location or self.location
+        _end_time = end_time or self.end_time
 
         if _entity_type in (EntityType.stage_instance, EntityType.voice):
-            if channel is MISSING or channel is None:
+            if _channel in (MISSING, None):
                 raise TypeError('channel must be set when entity_type is voice or stage_instance')
+            payload['channel_id'] = _channel.id
 
-            payload['channel_id'] = channel.id
-
-            if location not in (MISSING, None):
+            if _location not in (MISSING, None):
                 raise TypeError('location cannot be set when entity_type is voice or stage_instance')
             payload['entity_metadata'] = None
         else:
-            if channel not in (MISSING, None):
-                raise TypeError('channel cannot be set when entity_type is external')
             payload['channel_id'] = None
 
-            if location is MISSING or location is None:
+            if _location in (MISSING, None):
                 raise TypeError('location must be set when entity_type is external')
+            metadata['location'] = _location
 
-            metadata['location'] = location
-
-            if end_time is MISSING or end_time is None:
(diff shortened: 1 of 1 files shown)


Rapptz:
This fix is too excessive. Not to mention you changed `channel` and `end_time` to both discard explicit `None` values. Truth is fixing this is a bit complicated because it needs differentiation between explicitly set vs not and changing things depending on whether the `entity_type` is changed.

Puncher1:
@Rapptz Why is it too excessive? At the moment, you can't edit the `name` of the event without providing `location` and `end_time` if you created an event wiht `EntyType.external`. I agree, it's discarding `None` values, but isn't it only relevant that either the current values ​​or the already set values ​​of the object are correct. Because otherwise you would have to enter the already set values ​​again for each edit, which I think is unnecessary.

Rapptz:
It's excessive because it causes more bugs while fixing one of them.

I think this function as a whole needs to be rewritten or changed significantly in order to work well. There's a difference between modifying with explicit `entity_type` vs not. Like changing from type A to type B requires different required parameters and keeping them the same doesn't require certain parameters. The function needs to be rewritten to account for all those cases so it's more complicated than what you have here. I also feel like there need to be overloads that handle the different keyword parameter interactions.

Puncher1:
Okay, understand. Well, how would you make parameters required or not based on the value of another certain parameter? Is something like that already in the library so I have an example?

Puncher1:
Or how about doing it with a bool as a parameter e.g. `is_external`?

Rapptz:
Superseded by #9268

PR #9275, opened 2023-02-28
Fix create_scheduled_event param handling

Puncher1:
## Summary

**This fix includes following:**
* overloads for `Guild.create_scheduled_event` + not requiring some args
* `end_time` is required now when `entity_type` is `external` in `Guild.create_scheduled_event` 
* `privacy_level` isn't static anymore in `Guild.create_scheduled_event`
* better readability when `entity_type` is not missing for both `Guild.create_scheduled_event` and `ScheduledEvent.edit`

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -2974,14 +2974,14 @@ class Guild(Hashable):
         else:
             if not isinstance(entity_type, EntityType):
                 raise TypeError('entity_type must be of type EntityType')
+                
+            payload['entity_type'] = entity_type.value
 
         if entity_type is None:
             raise TypeError(
                 'invalid GuildChannel type passed, must be VoiceChannel or StageChannel '
                 f'not {channel.__class__.__name__}'
             )
-
-        payload['entity_type'] = entity_type.value
 
         if privacy_level is not MISSING:
             if not isinstance(privacy_level, PrivacyLevel):
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -2807,6 +2807,68 @@ class Guild(Hashable):
         data = await self._state.http.get_scheduled_event(self.id, scheduled_event_id, with_counts)
         return ScheduledEvent(state=self._state, data=data)
 
+    @overload
+    async def create_scheduled_event(
+        self,
+        *,
+        name: str,
+        start_time: datetime.datetime,
+        entity_type: Literal[EntityType.external] = ...,
+        privacy_level: PrivacyLevel = ...,
+        location: str = ...,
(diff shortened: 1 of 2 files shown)


Puncher1:
@Rapptz For the API call `channel_id` is optional and not forbidden when `entity_type` is `external` but it doesn't have an effect, because `location` must be passed anyway. Should it still be possible to pass a channel in combination with `EntityType.external` or would it be confusing?

Puncher1:
Can't find where the issue is of the lint check.

bijij:
> Can't find where the issue is of the lint check. 

You might be using 2023 black, dpy is pinned to 2022 black

~black devs could have never foreseen this issue~

Rapptz:
This PR is failing because there's *a lot* of trailing whitespace. Please install black and run it locally.

Puncher1:
@Rapptz Ready to merge. In case you didn't see my question:
For the API call `channel_id` is optional even when `entity_type` is `external` but it doesn't have an effect, because `location` must be set anyway. At the moment it throws an error when you provide `channel` in combination with `EntityType.external` even when it technically would work with the API call.

Rapptz:
> Should it still be possible to pass a channel in combination with EntityType.external or would it be confusing?

The current error is fine since it corresponds to the modal in the client.

discord/guild.py
```diff
@@ -2899,27 +2961,32 @@ async def create_scheduled_event(
...
             payload['scheduled_start_time'] = start_time.isoformat()
 
+        entity_type = entity_type or getattr(channel, '_scheduled_event_entity_type', MISSING)
         if entity_type is MISSING:
-            if channel is MISSING:
+            if channel and isinstance(channel, Object):
+                if channel.type == VoiceChannel:
+                    entity_type = EntityType.voice
+                elif channel.type == StageChannel:
+                    entity_type = EntityType.stage_instance
```
Rapptz, line 2970:
```suggestion
                if channel.type is VoiceChannel:
                    entity_type = EntityType.voice
                elif channel.type is StageChannel:
                    entity_type = EntityType.stage_instance
```

Rapptz (changes requested):

z03h:
Setting end_time should be outside the entity_type if checks to allow setting an end time for voice/stage events

Puncher1:
> Setting end_time should be outside the entity_type if checks to allow setting an end time for voice/stage events

Thanks

Rapptz (approved):

PR #9278, opened 2023-03-01
Fix udp discovery not using 74 byte layout

### LINKED ISSUE #9277: Can't connect to a VC. `discord.gateway.DiscordVoiceWebSocket.initial_connection` call never returns.
### Summary

`initial_connection` method coroutine never returns resulting in failing to connect to a VC

### Reproduction Steps

Call and await `connect()` on an instance of `VoiceChannel`.
Details on getting an instance of `VoiceChannel` (all checks are skipped for clarity):
```py
member = message.author
voice_state = member.voice
voice_channel = voice_state.channel
voice_client = await voice_channel.connect()  # this breaks
```

### Minimal Reproducible Code

_No response_

### Expected Results

Call to `connect` returns or raises an error.

### Actual Results

Call to `connect` never returns. It appears to be stuck on `discord.gateway.DiscordVoiceWebSocket.initial_connection`, specifically `recv = await self.loop.sock_recv(state.socket, 70)` line.
Seems to happen to other bots too (so, not only "my machine" situation).
No traceback (no errors; the method just doesn't return).

### Intents

discord.Intents.all()

### System Information

```
- Python v3.10.6-final
- discord.py v2.2.0-final
- aiohttp v3.8.4
- system info: Linux 5.15.0-56-generic #62-Ubuntu SMP Tue Nov 22 19:54:14 UTC 2022
```
Running in docker (`FROM python:3.10`).
Happens both with `discord.py==2.2.0` and `discord.py==2.1.0`

### Checklist

### Additional Context

Error started being encountered between 6:38 and 13:45 UTC, March 1st.
Tested with three separate bot accounts and two separate machines.
I traced all `discord.gateway.DiscordVoiceWebSocket.initial_connection` calls. Appears that it only enters the coroutine.
`await self.loop.sock_recv(state.socket, 70)` is assumed to be the issue because the only other await (`await self.select_protocol(state.ip, state.port, mode)`) never enters the coroutine.

MinnDevelopment:
## Summary

This should fix #9277 

The layout is documented here: https://discord.com/developers/docs/topics/voice-connections#ip-discovery

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -948,16 +948,16 @@ class DiscordVoiceWebSocket:
         state.voice_port = data['port']
         state.endpoint_ip = data['ip']
 
-        packet = bytearray(70)
+        packet = bytearray(74)
         struct.pack_into('>H', packet, 0, 1)  # 1 = Send
         struct.pack_into('>H', packet, 2, 70)  # 70 = Length
         struct.pack_into('>I', packet, 4, state.ssrc)
         state.socket.sendto(packet, (state.endpoint_ip, state.voice_port))
-        recv = await self.loop.sock_recv(state.socket, 70)
+        recv = await self.loop.sock_recv(state.socket, 74)
         _log.debug('received packet in initial_connection: %s', recv)
 
-        # the ip is ascii starting at the 4th byte and ending at the first null
-        ip_start = 4
+        # the ip is ascii starting at the 8th byte and ending at the first null
+        ip_start = 8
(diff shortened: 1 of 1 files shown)


AbstractUmbra:
Can confirm this fix works for me.

joaoperalta (approved):

Jourdelune:
I have also seen that voice listen (not supported by discord.py I know) does not work now, discord doesn't send the audio flux.

PR #9283, opened 2023-03-02
[commands] Pass failed argument's value to BadLiteralArgument

Soheab:
## Summary

This PR adds an `argument` attribute and parameter to `BadLiteralArgument`. This way, users can include the failed argument's value in their custom errors.

The parameter defaults to an empty string to make this a non-breaking change. Defaulting to `None` and typing as `Optional[]` would mark it as Optional for everything when that's never the case when the library raises it.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/errors.py
+++ b/discord/ext/commands/errors.py
@@ -912,9 +912,6 @@ class BadLiteralArgument(UserInputError):
 
     .. versionadded:: 2.0
 
-    .. versionchanged:: 2.3
-        Added the ``argument`` parameter and matching attribute.
-
     Attributes
     -----------
     param: :class:`inspect.Parameter`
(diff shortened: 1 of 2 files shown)


discord/ext/commands/errors.py
```diff
@@ -920,12 +923,17 @@ class BadLiteralArgument(UserInputError):
         A tuple of values compared against in conversion, in order of failure.
     errors: List[:class:`CommandError`]
         A list of errors that were caught from failing the conversion.
+    argument: :class:`str`
+        The argument's value that failed to be converted. Defaults to an empty string for a non-breaking change.
```
Rapptz, line 927:
```suggestion
        The argument's value that failed to be converted. Defaults to an empty string.
```

discord/ext/commands/errors.py
```diff
@@ -912,6 +912,9 @@ class BadLiteralArgument(UserInputError):
 
     .. versionadded:: 2.0
 
+    .. versionchanged:: 2.3
+        Added the ``argument`` parameter and matching attribute.
+
```
Rapptz, line 917: No need to document this.
```suggestion
```

Rapptz (changes requested):
Realistically, very few exceptions are allowed to be raised by users but it's good to keep in mind when making changes.

Rapptz (approved):

Issue #680, opened 2017-07-31
Who even uses Python.


Googie2149:
Discord.net is way better.

PR #9288, opened 2023-03-05
Support for receiving audio from voice channels

Sheppsu:
This PR is based off #6507. I made a new branch and PR for reasons outlined in [this comment](https://github.com/Rapptz/discord.py/pull/6507#issuecomment-1455003093) on my old PR.

## Summary

This pull request provides an implementation of receiving and processing audio packets from discord.

**User interaction**
There are 6 user-facing methods: 
 - `listen`: tells an already initialized `AudioReceiver` (threading.Thread subclass) object to start listening. The object is initialized and started when the socket connection is first formed. The `AudioReceiver` object upon starting listening is passed some arguments the user specifies, most notably an `AudioSink` object. This sink object will be responsible for handling processed audio.
 - `init_audio_processing_pool`: calls `ConnectionState.init_audio_processing_pool`. The listening functionality of this PR uses sub processes to handle audio processing and this function allows the user to decide certain settings. 
 - `stop_listening`: tells `_receiver` to stop listening (it will still receive audio packets, but just discard them).
 - `pause_listening`: tells the `_receiver` to not process any audio received until listening has been resumed. The audio data is simply thrown away right after being received.
 - `resume_listening`: tells the `_receiver` to begin processing received audio again.
 - `wait_for_listen_ready`: a function that waits until `listen` can be safely executed. It's important when `listen` is being called multiple times to ensure there is not conflict between separate listening sessions.
The user gains control over audio frames with sink objects. More on sink objects below.

**Audio Receiver**
The `AudioReceiver` thread is continuously receiving audio data using `VoiceClient.recv_audio`. If currently we want the audio, then it will be processed, but otherwise it's discarded. The raw audio frames are sent to `ConnectionState.process_audio`, which will handle distributing audio frames across sub processes. The function returns a future and a callback is added which will send processed audio frames to `AudioReceiver._audio_processing_callback`. The processed audio frames received at the callback are sent to the audio sink object. Finally, when listening is stopped, `AudioReceiver` will execute the callback specified by the user, similar to `AudioPlayer`.

**Connection State**
I wasn't entirely sure where the process pool object should lie, but I decided on the `ConnectionState` at least for now. The connection state is only responsible for initializing `AudioProcessPool` and submitting raw audio frames to it for processing. The pool is initialized with a maximum number of process workers and when audio is submitted, it's submitted to a specific process by index using the guild id modulus the maximum number of workers. This method of spreading the audio across different processes will ensure a fairly even distribution. 

**Audio Processing Pool**
The design of `AudioProcessPool` is based upon `ProcessPoolExecutor`, which I could not use due to its limitations. When audio is submitted to the pool, it first spawns the specified process if it hasn't already been spawned. The spawning function creates a duplex pipe and an `AudioUnpacker` object (multiprocessing.Process subclass), starts the process, and saves it to a dictionary. The dictionary is only accessed under a lock since it's used by two different threads (the current one and another one I will explain soon). Now that the process is running, the audio data can be sent to it. Finally, the process number and a future object are put into a `Queue` together and it checks that a separate receiving thread is running, starting a new one otherwise. The receiving thread will continuously fetch items from the queue, which tell it which processes will be sending back processed audio and where to resend it (the future object).

**Audio Unpacker**
`AudioUnpacker` is a subprocess that receives audio frames through a pipe, processes them, and sends them back. Audio frames are decrypted, decoded, and finally sent back as a `AudioFrame` object or one of the RTCP objects.

**Audio Sinks**
`AudioSink` is an abstract class that defines the base methods of a sink: `on_audio`, `on_rtcp`, `cleanup`. `on_audio` will receive processed audio frames, `on_rtcp` will receive RTCP packets, and `cleanup` is called when listening stops. This class can be subclassed by the user to carry out specific tasks with audio frames, however, there are also higher level sink objects: `AudioHandlingSink` and `AudioFileSink`. `AudioHandlingSink` implements functionality that takes audio frames and validates them by making sure they are received in order, account for delays and out-of-order packets. `AudioFileSink` subclasses `AudioHandlingSink`, receiving valid audio packets and then writing them to file using `AudioFile` objects. I'll go into depth on all of these objects in separate sections.

**Audio Handling Sink**
When an audio frame is received, it's immediately put into a queue and a separate validation thread will fetch it. The steps of the validation process are as follows:
1. If the audio frame is silent, empty the buffer (a list containing audio frames). Emptying the buffer involves sorting them by frame sequence, setting `_last_sequence`, and putting each audio frame into the queue that the validation thread pulls from. `_last_sequence` is a variable keeping track of the last validated audio frame. The emptying process is performed within a lock to make sure all the frames go in the queue before any frames from `on_audio`.
2. If the sequence of the frame is less than that of the last validated frame then it's dropped. 
3. 
   a. If the sequence of the frame matches up with the sequence of the last validated frame, then the frame is valid. `_last_sequence` is updated, the frame is passed to `on_valid_audio` (an abstract method in this class), and the buffer is emptied.
   b. If the sequence of the frame is not valid, then it is added to the buffer. If a certain amount of time has passed since the first audio frame was placed into the buffer, than it's forcefully emptied. This forceful empty of the buffer signifies that it will no longer wait for the lost frame. It's possible that it still arrives shortly after, but it would be dropped in step 2.

**Audio File Sink**
`AudioFileSink` creates an `AudioFile` type object for each user's audio data and manages them.  When `on_valid_audio` is called, the frame's ssrc is linked to an already existing `AudioFile` object or a new one is created. The audio frame is then passed to that object to handle. `AudioFileSink` implements the cleanup method which calls cleanup on all `AudioFile` objects. Finally, once cleanup is done, the user can call `convert_files`, which will call `convert` on all `AudioFile` objects.

**Audio File**
`AudioFile` receives audio frames and writes them to a pcm file. It saves the timestamp of the last received frame to calculate how much silence should be inserted in between some frames. When `cleanup` is called the file is closed and its set as being done. The `convert` method is abstract and should create a new file of a specific type using the raw audio data. There are two already made classes that implement `AudioFile`: `WaveAudioFile` and `MP3AudioFile`. 

## Possible problems

The biggest problem with this is that discord does not officially support receiving audio. It works, however, it's not guaranteed to be bug-free and production-safe. An example of this is a part of the code that I had to remove due to discord sending invalid RTP packets. As such, there's no telling when discord could push breaking changes without any warning for audio receive features. 

In addition, the only RTCP packet that I was actually able to test was the `RTCPReceiverReportPacket` since I don't think discord sends any other RTCP packets. I'm also fairly certain that the RTCP Receiver packet discord sends is invalid as well, in that it indicated there was 1 Receiver Report Block, while the sent data did not resemble any report blocks. Maybe I messed something up, but I examined the data quite closely and referred to reliable sources on the structure of the RTCP Receiver Packet.

Even if this pull request is held off on due to reasons related to discord not supporting audio receive, I'll still keep this code up to date until discord does actually support audio receive (if ever) so that's usable by those who wish.

## Testing the feature

I've created a new file in the examples directory named "basic_voice_listening.py"

If you wanna contact me then just dm me on discord: sheppsu

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -943,6 +943,9 @@ class DiscordVoiceWebSocket:
             self._keep_alive = VoiceKeepAliveHandler(ws=self, interval=min(interval, 5.0))
             self._keep_alive.start()
         elif op == self.SPEAKING:
+            # Imported here to avoid circular import
+            from .member import Member
+
             ssrc = data["ssrc"]
             if ssrc in self._speaking_map:
                 self._speaking_map[ssrc]["speaking"] = data["speaking"]
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -1022,7 +1022,7 @@ class DiscordVoiceWebSocket:
         self._close_code = code
         await self.ws.close(code=code)
 
-    def get_member_from_ssrc(self, ssrc) -> Optional[Union['Member', int]]:
+    def get_member_from_ssrc(self, ssrc: int) -> Optional[Union['Member', int]]:
         if ssrc in self._speaking_map:
             user = self._speaking_map[ssrc]["user"]
             if isinstance(user, int) and (member := self._connection.guild.get_member(user)) is not None:
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -58,9 +58,9 @@ if TYPE_CHECKING:
     from typing_extensions import Self
 
     from .client import Client
+    from .member import Member
     from .state import ConnectionState
     from .voice_client import VoiceClient
-    from .member import Member
 
 
 class ReconnectWebSocket(Exception):
@@ -826,7 +826,7 @@ class DiscordVoiceWebSocket:
         self.loop: asyncio.AbstractEventLoop = loop
         self._keep_alive: Optional[VoiceKeepAliveHandler] = None
         self._close_code: Optional[int] = None
-        self._speaking_map = {}
+        self._speaking_map: Dict[int, Dict[str, Union[Member, int]]] = {}
         self.secret_key: Optional[str] = None
         if hook:
             self._hook = hook
@@ -948,10 +948,7 @@ class DiscordVoiceWebSocket:
             else:
                 user_id = int(data["user_id"])
                 user = self._connection.guild.get_member(user_id)
-                self._speaking_map[ssrc] = {
-                    "user": user if user is not None else user_id,
-                    "speaking": data["speaking"]
-                }
+                self._speaking_map[ssrc] = {"user": user if user is not None else user_id, "speaking": data["speaking"]}
 
         await self._hook(self, msg)
 
@@ -1028,7 +1025,7 @@ class DiscordVoiceWebSocket:
     def get_member_from_ssrc(self, ssrc) -> Optional[Union['Member', int]]:
         if ssrc in self._speaking_map:
             user = self._speaking_map[ssrc]["user"]
-            if type(user) == int and (member := self._connection.guild.get_member(user)) is not None:
+            if isinstance(user, int) and (member := self._connection.guild.get_member(user)) is not None:
                 self._speaking_map[ssrc]["user"] = member
                 return member
             return user
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -34,7 +34,7 @@ import threading
 import traceback
 import zlib
 
-from typing import Any, Callable, Coroutine, Deque, Dict, List, TYPE_CHECKING, NamedTuple, Optional, TypeVar
+from typing import Any, Callable, Coroutine, Deque, Dict, List, TYPE_CHECKING, NamedTuple, Optional, TypeVar, Union
 
 import aiohttp
 import yarl
@@ -60,6 +60,7 @@ if TYPE_CHECKING:
     from .client import Client
     from .state import ConnectionState
     from .voice_client import VoiceClient
+    from .member import Member
 
 
 class ReconnectWebSocket(Exception):
@@ -825,6 +826,7 @@ class DiscordVoiceWebSocket:
         self.loop: asyncio.AbstractEventLoop = loop
         self._keep_alive: Optional[VoiceKeepAliveHandler] = None
         self._close_code: Optional[int] = None
+        self._speaking_map = {}
         self.secret_key: Optional[str] = None
         if hook:
             self._hook = hook
@@ -939,6 +941,17 @@ class DiscordVoiceWebSocket:
             interval = data['heartbeat_interval'] / 1000.0
             self._keep_alive = VoiceKeepAliveHandler(ws=self, interval=min(interval, 5.0))
             self._keep_alive.start()
+        elif op == self.SPEAKING:
+            ssrc = data["ssrc"]
+            if ssrc in self._speaking_map:
+                self._speaking_map[ssrc]["speaking"] = data["speaking"]
+            else:
+                user_id = int(data["user_id"])
+                user = self._connection.guild.get_member(user_id)
+                self._speaking_map[ssrc] = {
+                    "user": user if user is not None else user_id,
+                    "speaking": data["speaking"]
+                }
 
         await self._hook(self, msg)
 
@@ -1011,3 +1024,11 @@ class DiscordVoiceWebSocket:
 
         self._close_code = code
         await self.ws.close(code=code)
+
+    def get_member_from_ssrc(self, ssrc) -> Optional[Union['Member', int]]:
+        if ssrc in self._speaking_map:
+            user = self._speaking_map[ssrc]["user"]
+            if type(user) == int and (member := self._connection.guild.get_member(user)) is not None:
+                self._speaking_map[ssrc]["user"] = member
+                return member
+            return user
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -43,6 +43,7 @@ from . import utils
 from .activity import BaseActivity
 from .enums import SpeakingState
 from .errors import ConnectionClosed
+from .object import Object
 
 _log = logging.getLogger(__name__)
 
@@ -826,7 +827,7 @@ class DiscordVoiceWebSocket:
         self.loop: asyncio.AbstractEventLoop = loop
         self._keep_alive: Optional[VoiceKeepAliveHandler] = None
         self._close_code: Optional[int] = None
-        self._speaking_map: Dict[int, Dict[str, Union[Member, int]]] = {}
+        self._speaking_map: Dict[int, Dict[str, Union[Member, Object]]] = {}
         self.secret_key: Optional[str] = None
         if hook:
             self._hook = hook
@@ -948,7 +949,10 @@ class DiscordVoiceWebSocket:
             else:
                 user_id = int(data["user_id"])
                 user = self._connection.guild.get_member(user_id)
-                self._speaking_map[ssrc] = {"user": user if user is not None else user_id, "speaking": data["speaking"]}
+                self._speaking_map[ssrc] = {
+                    "user": user if user is not None else Object(id=user_id, type=Member),
+                    "speaking": data["speaking"],
+                }
 
         await self._hook(self, msg)
 
@@ -1022,10 +1026,10 @@ class DiscordVoiceWebSocket:
         self._close_code = code
         await self.ws.close(code=code)
 
-    def get_member_from_ssrc(self, ssrc: int) -> Optional[Union['Member', int]]:
+    def get_member_from_ssrc(self, ssrc: int) -> Optional[Union['Member', Object]]:
         if ssrc in self._speaking_map:
             user = self._speaking_map[ssrc]["user"]
-            if isinstance(user, int) and (member := self._connection.guild.get_member(user)) is not None:
+            if isinstance(user, Object) and (member := self._connection.guild.get_member(user.id)) is not None:
                 self._speaking_map[ssrc]["user"] = member
                 return member
             return user
--- a/discord/sink.py
+++ b/discord/sink.py
@@ -612,7 +612,16 @@ class AudioHandlingSink(AudioSink):
     dealing with out-of-order packets and delays.
     """
 
-    __slots__ = ("_last_sequence", "_buffer", "_buffer_wait", "_frame_queue", "_is_validating", "_buffer_till", "_lock")
+    __slots__ = (
+        "_last_sequence",
+        "_buffer",
+        "_buffer_wait",
+        "_frame_queue",
+        "_is_validating",
+        "_buffer_till",
+        "_lock",
+        "_done_validating",
+    )
     # how long to wait for missing a packet
     PACKET_WAIT_TIME = 2
     # how long to wait for a new packet before closing the _validation_loop thread
@@ -628,6 +637,7 @@ class AudioHandlingSink(AudioSink):
         self._is_validating: threading.Event = threading.Event()
         self._buffer_till: Dict[int, Optional[float]] = defaultdict(lambda: None)
         self._lock: threading.Lock = threading.Lock()
+        self._done_validating: threading.Event = threading.Event()
 
     def on_audio(self, frame: AudioFrame) -> None:
         """Puts frame in a queue and lets a processing loop thread deal with it."""
@@ -640,12 +650,13 @@ class AudioHandlingSink(AudioSink):
     def _start_validation_loop(self) -> None:
         if not self._is_validating.is_set():
             threading.Thread(target=self._validation_loop).start()
-            # prevent multiple threads spawning
+            # prevent multiple threads spawning and make sure it spawns, otherwise giving a warning
             if not self._is_validating.wait(timeout=self.VALIDATION_LOOP_START_TIMEOUT):
                 _log.warning("Timeout reached waiting for _validation_loop thread to start")
 
     def _validation_loop(self) -> None:
         self._is_validating.set()
+        self._done_validating.clear()
         while True:
             try:
                 frame = self._frame_queue.get(timeout=self.VALIDATION_LOOP_TIMEOUT)
@@ -654,6 +665,8 @@ class AudioHandlingSink(AudioSink):
 
             self._validate_audio_frame(frame)
         self._is_validating.clear()
+        if not self._empty_entire_buffer():
+            self._done_validating.set()
 
     def _validate_audio_frame(self, frame: AudioFrame) -> None:
         # 1. If audio is a silent frame, empty buffer and reset last sequence
@@ -663,10 +676,6 @@ class AudioHandlingSink(AudioSink):
         # 3b. Else, put in a buffer
         # Buffer should be forcefully emptied (same as in 3a) when a missing frame is not received
         # after a specific amount of time
-
-        if frame.audio == SILENT_FRAME:
-            self._empty_buffer(frame.ssrc)
-            return
 
         last_sequence = self._last_sequence[frame.ssrc]
         if frame.sequence <= last_sequence:
@@ -679,7 +688,7 @@ class AudioHandlingSink(AudioSink):
         else:
             self._append_to_buffer(frame)
 
-    def _append_to_buffer(self, frame):
+    def _append_to_buffer(self, frame) -> None:
         self._buffers[frame.ssrc].append(frame)
         buffer_till = self._buffer_till[frame.ssrc]
         if buffer_till is None:
@@ -688,19 +697,28 @@ class AudioHandlingSink(AudioSink):
             self._buffer_till[frame.ssrc] = None
             self._empty_buffer(frame.ssrc)
 
-    def _empty_buffer(self, ssrc) -> None:
+    def _empty_entire_buffer(self) -> bool:
+        result = False
+        for ssrc in self._buffers.keys():
+            result = result or self._empty_buffer(ssrc)
+        return result
+
+    def _empty_buffer(self, ssrc) -> bool:
         buffer = self._buffers[ssrc]
         if len(buffer) == 0:
-            return
+            return False
 
         sorted_buffer = sorted(buffer, key=lambda f: f.sequence)
         self._last_sequence[ssrc] = sorted_buffer[0].sequence - 1
         # prevent on_audio from putting frames in queue before these frames
+        # and no conflicts on starting validation loop
         self._lock.acquire()
         for frame in sorted_buffer:
-            self._frame_queue.put(frame)
+            self._frame_queue.put_nowait(frame)
+        self._start_validation_loop()
         self._lock.release()
         self._buffers[ssrc] = []
+        return True
 
     def on_valid_audio(self, frame: AudioFrame) -> Any:
         """When an audio packet is declared valid, it'll be passed to this function.
@@ -744,7 +762,7 @@ class AudioFileSink(AudioHandlingSink):
         Indicates whether cleanup has been called.
     """
 
-    __slots__ = ("file_type", "output_dir", "output_files", "done")
+    __slots__ = ("file_type", "output_dir", "output_files", "done", "_clean_lock")
 
     def __init__(self, file_type: Callable[[str, int], 'AudioFile'], output_dir: str = "."):
         super().__init__()
@@ -754,6 +772,7 @@ class AudioFileSink(AudioHandlingSink):
         self.output_dir: str = output_dir
         self.output_files: Dict[int, AudioFile] = {}
         self.done: bool = False
+        self._clean_lock: threading.Lock = threading.Lock()
 
     def on_valid_audio(self, frame: AudioFrame) -> None:
         """Takes an audio frame and passes it to a :class:`AudioFile` object. If
@@ -764,6 +783,7 @@ class AudioFileSink(AudioHandlingSink):
         frame: :class:`AudioFrame`
             The frame which will be added to the buffer.
         """
+        self._clean_lock.acquire()
         if self.done:
             return
 
@@ -773,6 +793,7 @@ class AudioFileSink(AudioHandlingSink):
             )
 
         self.output_files[frame.ssrc].on_audio(frame)
+        self._clean_lock.release()
 
     def on_rtcp(self, packet: RTCPPacket) -> None:
         """This function receives RTCP Packets, but does nothing with them since
@@ -790,11 +811,14 @@ class AudioFileSink(AudioHandlingSink):
 
     def cleanup(self) -> None:
         """Calls cleanup on all :class:`AudioFile` objects."""
+        self._clean_lock.acquire()
         if self.done:
             return
+        self._done_validating.wait()
+        self.done = True
         for file in self.output_files.values():
             file.cleanup()
-        self.done = True
+        self._clean_lock.release()
 
     def convert_files(self) -> None:
         """Calls cleanup if it hasn't already been called and then calls convert on all :class:`AudioFile` objects."""
@@ -846,7 +870,10 @@ class AudioFile:
         "converted",
         "path",
         "_last_timestamp",
+        "_last_sequence",
+        "_packet_count",
         "user",
+        "_clean_lock",
     )
 
     FRAME_BUFFER_LIMIT = 10
@@ -858,8 +885,11 @@ class AudioFile:
         self.converted: bool = False
         self.user: Optional[Union[Member, Object]] = None
         self.path: str = self.file.name
+        self._clean_lock: threading.Lock = threading.Lock()
 
         self._last_timestamp: Optional[int] = None
+        self._last_sequence: Optional[int] = None
+        self._packet_count = 0
 
     def on_audio(self, frame: AudioFrame) -> None:
         """Takes an audio frame and adds it to a buffer. Once the buffer
@@ -872,18 +902,27 @@ class AudioFile:
         frame: :class:`AudioFrame`
             The frame which will be added to the buffer.
         """
+        self._clean_lock.acquire()
         if self.done:
             return
+        if self._packet_count < 7:
+            self._packet_count += 1
         self._write_frame(frame)
+        self._clean_lock.release()
 
     def _write_frame(self, frame: AudioFrame) -> None:
-        if self._last_timestamp is not None:
-            # write silence
+        # When the bot joins a vc and starts listening and a user speaks for the first time,
+        # the timestamp encompasses all that silence, including silence before the bot even
+        # joined the vc. It goes in a pattern that the 6th packet has a 11 sequence skip, so
+        # this last part of the if statement gets rid of that silence.
+        if self._last_timestamp is not None and not (self._packet_count == 6 and frame.sequence - self._last_sequence == 11):
             silence = frame.timestamp - self._last_timestamp - OpusDecoder.SAMPLES_PER_FRAME
             if silence > 0:
                 self.file.write(b"\x00" * silence * OpusDecoder.SAMPLE_SIZE)
-        self.file.write(frame.audio)
+        if frame.audio != SILENT_FRAME:
+            self.file.write(frame.audio)
         self._last_timestamp = frame.timestamp
+        self._last_sequence = frame.sequence
         self._cache_user(frame.user)
 
     def _cache_user(self, user: Optional[Union['Member', 'Object']]) -> None:
@@ -902,10 +941,12 @@ class AudioFile:
 
     def cleanup(self) -> None:
         """Writes remaining frames in buffer to file and then closes it."""
+        self._clean_lock.acquire()
         if self.done:
             return
         self.file.close()
         self.done = True
+        self._clean_lock.release()
 
     def convert(self, new_name: Optional[str] = None) -> None:
         """Converts the file to its formatted file type.
@@ -1050,32 +1091,27 @@ class AsyncEventWrapper:
 class AudioReceiver(threading.Thread):
     def __init__(
         self,
-        sink: AudioSink,
         client: 'VoiceClient',
-        *,
-        decode: bool = True,
-        after: Optional[Callable[..., Awaitable[Any]]] = None,
-        after_kwargs: Optional[dict] = None,
     ) -> None:
         if not has_nacl:
             raise RuntimeError("PyNaCl library is required to use audio receiving")
 
         super().__init__()
-        self.sink: AudioSink = sink
+        self.sink: Optional[AudioSink] = None
         self.client: VoiceClient = client
         self._state: ConnectionState = self.client._state
         self.loop = self.client.client.loop
 
-        self.decode: bool = decode
-        self.mode = self.client.mode
-        self.secret_key = self.client.secret_key
-        self.after: Optional[Callable[..., Awaitable[Any]]] = after
-        self.after_kwargs: Optional[dict] = after_kwargs
+        self.decode: bool = True
+        self.after: Optional[Callable[..., Awaitable[Any]]] = None
+        self.after_kwargs: Optional[dict] = None
 
         self._end: AsyncEventWrapper = AsyncEventWrapper()
+        self._on_standby: AsyncEventWrapper = AsyncEventWrapper()
+        self._on_standby.set()
         self._resumed: AsyncEventWrapper = AsyncEventWrapper()
-        self._resumed.set()
         self._clean: AsyncEventWrapper = AsyncEventWrapper()
+        self._clean.set()
         self._connected: threading.Event = client._connected
 
     def _do_run(self) -> None:
@@ -1087,7 +1123,9 @@ class AudioReceiver(threading.Thread):
             if data is None:
                 continue
 
-            future = self._state.process_audio(data, self.decode, self.mode, self.secret_key, self.client.guild.id)
+            future = self._state.process_audio(
+                data, self.decode, self.client.mode, self.client.secret_key, self.client.guild.id
+            )
             future.add_done_callback(self._audio_processing_callback)
 
     def _audio_processing_callback(self, future: Future) -> None:
@@ -1110,10 +1148,6 @@ class AudioReceiver(threading.Thread):
         except Exception as exc:
             self.stop()
             _log.exception("Exception occurred in voice receiver", exc_info=exc)
-        finally:
-            self.sink.cleanup()
-            self._call_after()
-            self._clean.set()
 
     def _call_after(self) -> None:
         if self.after is not None:
@@ -1123,9 +1157,35 @@ class AudioReceiver(threading.Thread):
             except Exception as exc:
                 _log.exception('Calling the after function failed.', exc_info=exc)
 
+    def _cleanup_listen(self) -> None:
+        self.sink.cleanup()
+        self._call_after()
+        self.sink = None
+        self._clean.set()
+
+    def start_listening(
+        self,
+        sink: AudioSink,
+        *,
+        decode: bool = True,
+        after: Optional[Callable[..., Awaitable[Any]]] = None,
+        after_kwargs: Optional[dict] = None,
+    ) -> None:
+        self.sink = sink
+        self.decode = decode
+        self.after = after
+        self.after_kwargs = after_kwargs
+        self._on_standby.clear()
+        self._clean.clear()
+        self._resumed.set()
+
     def stop(self) -> None:
         self._end.set()
-        self._resumed.set()
+
+    def stop_listening(self) -> None:
+        self._resumed.clear()
+        self._on_standby.set()
+        self._cleanup_listen()
 
     def pause(self) -> None:
         self._resumed.clear()
@@ -1137,19 +1197,22 @@ class AudioReceiver(threading.Thread):
         return self._end.is_set()
 
     def is_listening(self) -> bool:
-        return self._resumed.is_set() and not self._end.is_set()
+        return self._resumed.is_set() and not self._on_standby.is_set()
 
     def is_paused(self) -> bool:
-        return not self._end.is_set() and not self._resumed.is_set()
+        return not self._resumed.is_set() and not self._on_standby.is_set()
+
+    def is_on_standby(self) -> bool:
+        return self._on_standby.is_set()
 
     def is_cleaning(self) -> bool:
-        return self._end.is_set() and not self._clean.is_set()
+        return self._on_standby.is_set() and not self._clean.is_set()
 
     async def wait_for_resumed(self, *, loop=None) -> None:
         await self._resumed.async_wait(self.loop if loop is None else loop)
 
-    async def wait_for_done(self, *, loop=None) -> None:
-        await self._end.async_wait(self.loop if loop is None else loop)
+    async def wait_for_standby(self, *, loop=None) -> None:
+        await self._on_standby.async_wait(self.loop if loop is None else loop)
 
     async def wait_for_clean(self, *, loop=None) -> None:
         await self._clean.async_wait(self.loop if loop is None else loop)
--- a/discord/sink.py
+++ b/discord/sink.py
@@ -727,7 +727,7 @@ class AudioFile:
 
     Attributes
     ----------
-    file: BinaryIO
+    file: :term:`py:file object`
         File object of the audio file this object refers to.
     ssrc: :class:`int`
         ssrc of the user associated with this audio file
--- a/discord/sink.py
+++ b/discord/sink.py
@@ -4,7 +4,7 @@ import struct
 import subprocess
 import threading
 import wave
-from typing import IO, TYPE_CHECKING, Any, Callable, NamedTuple, Optional, Sequence, Tuple, Union
+from typing import TYPE_CHECKING, Any, BinaryIO, Callable, NamedTuple, Optional, Sequence, Tuple, Union
 
 from .enums import RTCPMessageType
 from .errors import ClientException
@@ -685,7 +685,7 @@ class AudioFileSink(AudioSink):
             os.remove(filepath)
         self.output_files = {}
 
-    def convert_file(self, file: IO, new_name: Optional[str] = None) -> Any:
+    def convert_file(self, file: BinaryIO, new_name: Optional[str] = None) -> Any:
         """Takes a file object with raw audio data and creates
         another file object with formatted audio data
 
@@ -693,7 +693,7 @@ class AudioFileSink(AudioSink):
 
         Parameters
         ----------
-        file: :class:`IO`
+        file: :term:`py:file object`
         new_name: Optional[:class:`str`]
         """
         raise NotImplementedError()
@@ -710,14 +710,14 @@ class WaveAudioFileSink(AudioFileSink):
 
     CHUNK_WRITE_SIZE = 64
 
-    def convert_file(self, file: IO, new_name: Optional[str] = None) -> Any:
+    def convert_file(self, file: BinaryIO, new_name: Optional[str] = None) -> Any:
         """Takes a file object that contains raw audio data and writes
         it to a wave audio file.
 
         Parameters
         ----------
-        file: :class:`IO`
-            File object that contains raw audio data
+        file: :term:`py:file object`
+            File object that contains raw audio data.
         new_name: Optional[:class:`str`]
             Name for the wave file excluding ".wav". Defaults to current name if None.
 
@@ -739,14 +739,14 @@ class WaveAudioFileSink(AudioFileSink):
 class MP3AudioFileSink(AudioFileSink):
     """Extends :class:`AudioFileSink` and defines a convert_file method."""
 
-    def convert_file(self, file: IO, new_name: Optional[str] = None) -> Any:
+    def convert_file(self, file: BinaryIO, new_name: Optional[str] = None) -> Any:
         """Takes a file object that contains raw audio data and writes
         it to a mp3 audio file.
 
         Parameters
         ----------
-        file: :class:`IO`
-            File object that contains raw audio data
+        file: :term:`py:file object`
+            File object that contains raw audio data.
         new_name: Optional[:class:`str`]
             Name for the wave file excluding ".mp3". Defaults to current name if None.
 
--- a/discord/sink.py
+++ b/discord/sink.py
@@ -1,11 +1,13 @@
 import asyncio
-from dataclasses import dataclass
 import logging
+import multiprocessing
 import os
+import queue
 import struct
 import subprocess
 import threading
 import wave
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Awaitable, BinaryIO, Callable, Dict, List, Optional, Sequence, Tuple, Union
 
 from .enums import RTCPMessageType
@@ -13,6 +15,13 @@ from .errors import ClientException
 from .object import Object
 from .opus import Decoder as OpusDecoder
 from .player import CREATE_NO_WINDOW
+
+try:
+    import nacl.secret
+    import nacl.utils
+except ImportError:
+    # Warning is given in VoiceClient
+    pass
 
 if TYPE_CHECKING:
     from .member import Member
@@ -39,6 +48,7 @@ __all__ = (
 
 
 _log = logging.getLogger(__name__)
+_mp_ctx = multiprocessing.get_context("spawn")
 
 
 @dataclass
@@ -317,7 +327,7 @@ class RTCPSourceDescriptionPacket(RTCPPacket):
 
     __slots__ = ("chunks",)
 
-    chunks: List
+    chunks: List[RTCPSourceDescriptionChunk]
 
     def __init__(self, version_flag: int, rtcp_type: RTCPMessageType, length: int, data: bytes):
         super().__init__(version_flag, rtcp_type, length)
@@ -363,8 +373,8 @@ class RTCPGoodbyePacket(RTCPPacket):
 
     Attributes
     ----------
-    ssrc_byes: Sequence[:class:`int`]
-        List of SSRCs that are disconnecting.
+    ssrc_byes: Tuple[:class:`int`]
+        List of SSRCs that are disconnecting. Not guaranteed to contain any values.
     reason: :class:`bytes`
         Reason for disconnect.
     """
@@ -374,7 +384,7 @@ class RTCPGoodbyePacket(RTCPPacket):
         "reason",
     )
 
-    ssrc_byes: Tuple
+    ssrc_byes: Union[Tuple[int], Tuple]
     reason: bytes
 
     def __init__(self, version_flag: int, rtcp_type: RTCPMessageType, length: int, data: bytes):
@@ -649,6 +659,7 @@ class AudioFileSink(AudioSink):
         "file_type",
         "output_dir",
         "output_files",
+        "_lock",
     )
 
     def __init__(self, file_type: Callable[[str, int], 'AudioFile'], output_dir: str = "."):
@@ -658,6 +669,8 @@ class AudioFileSink(AudioSink):
         self.output_dir: str = output_dir
         self.output_files: Dict[int, AudioFile] = {}
         self.done: bool = False
+
+        self._lock = threading.Lock()
 
     def __del__(self):
         if hasattr(self, "done") and not self.done:
@@ -675,10 +688,13 @@ class AudioFileSink(AudioSink):
         if self.done:
             return
 
+        self._lock.acquire()
         if frame.ssrc not in self.output_files:
             self.output_files[frame.ssrc] = self.file_type(
                 os.path.join(self.output_dir, f"audio-{frame.ssrc}.pcm"), frame.ssrc
             )
+        self._lock.release()
+
         self.output_files[frame.ssrc].on_audio(frame)
 
     def on_rtcp(self, packet: RTCPPacket) -> None:
@@ -693,7 +709,7 @@ class AudioFileSink(AudioSink):
             :class:`RTCPSourceDescriptionPacket`, :class:`RTCPGoodbyePacket`,
             :class:`RTCPApplicationDefinedPacket`
         """
-        pass
+        return
 
     def cleanup(self) -> None:
         """Calls cleanup on all :class:`AudioFile` objects."""
@@ -755,6 +771,7 @@ class AudioFile:
         "_last_timestamp",
         "_frame_buffer",
         "user",
+        "_lock",
     )
 
     FRAME_BUFFER_LIMIT = 10
@@ -770,6 +787,7 @@ class AudioFile:
         # This gives leeway for frames sent out of order
         self._frame_buffer: List[AudioFrame] = []
         self.user: Optional[Union[Member, Object]] = None
+        self._lock = threading.Lock()
 
     def on_audio(self, frame: AudioFrame) -> None:
         """Takes an audio frame and adds it to a buffer. Once the buffer
@@ -782,9 +800,11 @@ class AudioFile:
         frame: :class:`AudioFrame`
             The frame which will be added to the buffer.
         """
+        self._lock.acquire()
         self._frame_buffer.append(frame)
         if len(self._frame_buffer) >= self.FRAME_BUFFER_LIMIT:
             self._write_buffer()
+        self._lock.release()
 
     def _write_buffer(self, empty: bool = False) -> None:
         self._frame_buffer = sorted(self._frame_buffer, key=lambda frame: frame.sequence)
@@ -938,6 +958,32 @@ class MP3AudioFile(AudioFile):
         self._convert_cleanup(path)
 
 
+class AsyncEventWrapper:
+    def __init__(self, event: Optional[threading.Event] = None):
+        self.event: threading.Event = event or threading.Event()
+        self._waiters: queue.Queue = queue.Queue()
+
+    def __getattr__(self, item):
+        return getattr(self.event, item)
+
+    def set(self) -> None:
+        self.event.set()
+        # Queue.empty() is not reliable, so instead we just catch when the queue throws an Empty error
+        try:
+            while True:
+                future = self._waiters.get_nowait()
+                future._loop.call_soon_threadsafe(future.set_result, True)
+        except queue.Empty:
+            pass
+
+    async def async_wait(self, loop) -> None:
+        if self.is_set():
+            return
+        future = loop.create_future()
+        self._waiters.put(future)
+        await future
+
+
 class AudioReceiver(threading.Thread):
     def __init__(
         self,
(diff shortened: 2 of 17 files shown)


discord/sink.py
```diff
@@ -0,0 +1,775 @@
...
+    "RTCPReceiverReportBlock",
+    "RTCPSourceDescriptionChunk",
+    "RTCPSourceDescriptionItem"
+)
+
+
+_log = logging.getLogger(__name__)
+
+
+RTCPReceiverReportBlock = namedtuple("RTCPReceiverReportBlock", ("ssrc", "f", "c", "ehsn", "j", "lsr", "dlsr"))
```
AbstractUmbra, line 41: I believe it would be better to use `typing.NamedTuple` for clearer types?

discord/sink.py
```diff
@@ -0,0 +1,775 @@
...
+        up until the time this SR packet was generated. The count is reset if
+        the sender changes its SSRC identifier.
+    report_blocks: Sequence[:class:`RTCPReceiverReport`]
+        Sequence of :class:`RTCPReceiverReport` objects that tell statistics.
+        Receivers do not carry over statistics when a source changes its SSRC
+        identifier due to a collision.
+    extension: :class:`bytes`
+        Profile-specific extension that may or may not contain a value.
+    """
+    __slots__ = RTCPPacket.__slots__ + (
```
AbstractUmbra, line 190: if you add `black`'s "magic comma" at the end of a container type, it formats them for you:-

```diff
- __slots__ = ("foo", "bar")
+ __slots__ = ("foo", "bar",)
```
Which `black` then formats to:-
```py
__slots__ = (
  "foo",
  "bar",
)
```

This is the standard across the library, except in cases where `__slots__` only contains one item, in which case we disable `black` on those lines and make this format ourselves.

discord/sink.py
```diff
@@ -0,0 +1,775 @@
...
+    def __init__(self, version_flag: int, rtcp_type: RTCPMessageType, length: int, data: bytes):
+        super().__init__(version_flag, rtcp_type, length)
+
+        self.chunks = []
+        for _ in range(self.rc):
+            chunk, offset = self._parse_chunk(data)
+            data = data[offset:]
+            self.chunks.append(chunk)
+
+    def _parse_chunk(self, data):
```
AbstractUmbra, line 260: I understand it's a private method so feel free to ignore, but shouldn't this still have a TypedDict data payload type?

discord/sink.py
```diff
@@ -0,0 +1,775 @@
...
+        after cleanup is called.
+    """
+    __slots__ = (
+        "output_dir", "output_files", "done", "_timestamps",
+        "_frame_buffer", "_ssrc_to_user"
+    )
+
+    FRAME_BUFFER_LIMIT = 10
+
+    def __init__(self, output_dir="/"):
```
AbstractUmbra, line 527: Missing type annotations.

discord/sink.py
```diff
@@ -0,0 +1,775 @@
...
+        directory, name = os.path.split(path)
+        name = new_name + ext if new_name is not None else ".".join(name.split(".")[:-1]) + ext
+        return os.path.join(directory, name)
+
+
+class WaveAudioFileSink(AudioFileSink):
+    """Extends :class:`AudioFileSink` and defines a convert_file method."""
+    CHUNK_WRITE_SIZE = 64
+
+    def convert_file(self, file, new_name=None):
```
AbstractUmbra, line 642: Missing type annotations.

discord/sink.py
```diff
@@ -0,0 +1,775 @@
...
+            wavf.setframerate(OpusDecoder.SAMPLING_RATE)
+            while frames := file.read(OpusDecoder.FRAME_SIZE*self.CHUNK_WRITE_SIZE):
+                wavf.writeframes(frames)
+        return path
+
+
+class MP3AudioFileSink(AudioFileSink):
+    """Extends :class:`AudioFileSink` and defines a convert_file method."""
+
+    def convert_file(self, file, new_name=None):
```
AbstractUmbra, line 671: Missing type annotations.

discord/sink.py
```diff
@@ -0,0 +1,775 @@
...
+class AudioReceiver(threading.Thread):
+    def __init__(
+        self,
+        sink: AudioSink,
+        client: 'VoiceClient',
+        *,
+        decode: bool = True,
+        after: Optional[Callable[[AudioSink, Optional[Exception]], Any]] = None,
+    ) -> None:
+        threading.Thread.__init__(self)
```
AbstractUmbra, line 708: This is an old backport style call to `super().__init__()`, you're free to switch to this.

discord/sink.py
```diff
@@ -0,0 +1,775 @@
...
+        self.daemon = False
+        self.sink = sink
+        self.client = client
+        self.decode = decode
```
AbstractUmbra, line 712: These should have type annotations.

discord/utils.py
```diff
@@ -70,6 +70,7 @@
 import types
 import warnings
 import logging
+from collections import namedtuple
```
AbstractUmbra, line 73: Ideally we should be using `typing.NamedTuple` wherever possible.

discord/voice_client.py
```diff
@@ -44,21 +44,23 @@
 import logging
 import struct
 import threading
-from typing import Any, Callable, List, Optional, TYPE_CHECKING, Tuple, Union
+import select
+from typing import Any, Callable, List, Optional, TYPE_CHECKING, Tuple, Union, Dict
```
AbstractUmbra, line 48: Please run `black` and `isort` on the codebase as is the standard.

discord/voice_client.py
```diff
@@ -564,6 +586,39 @@ def _encrypt_xsalsa20_poly1305_lite(self, header: bytes, data) -> bytes:
...
+        box = nacl.secret.SecretBox(bytes(self.secret_key))
+
+        nonce = bytearray(24)
+        nonce[:4] = data[-4:]
+        data = data[:-4]
+
+        return self.strip_header_ext(box.decrypt(bytes(data), bytes(nonce)))
+
+    @staticmethod
+    def strip_header_ext(data):
```
AbstractUmbra, line 615: MIssing type annotations.

discord/voice_client.py
```diff
@@ -635,6 +690,78 @@ def resume(self) -> None:
         if self._player:
             self._player.resume()
 
+    def listen(self, sink: AudioSink, *, decode: bool = True,
+               after: Optional[Callable[[AudioSink, Optional[Exception]], Any]] = None) -> None:
```
AbstractUmbra, line 694: Please run `black`.

examples/basic_voice_listening.py
```diff
@@ -0,0 +1,141 @@
+import discord
+import argparse
+
+
+def vc_required(func):
+    async def get_vc(self, msg):
```
AbstractUmbra, line 6: Examples are heavily scrutinized, and as such you do need type annotations and brief comments.

This decorated check should have a small piece of text explaining what it does.

examples/basic_voice_listening.py
```diff
@@ -0,0 +1,141 @@
...
+    return get_vc
+
+
+class ArgumentParser(argparse.ArgumentParser):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.error_msg = None
+
+    def convert_arg_line_to_args(self, arg_line: str):
+        return arg_line.split()  # this is good enough for our arguments
```
AbstractUmbra, line 20: In an ideal world this would use `shlex.split()`, no?

examples/basic_voice_listening.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+start_arg_parser = ArgumentParser()
+start_arg_parser.add_argument('-o', '--out', choices=['wav', 'mp3'], default='mp3')
+
+
+class Client(discord.Client):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.connections = {voice.guild.id: voice for voice in self.voice_clients}
```
AbstractUmbra, line 37: This will never be populated on `__init__`?

discord/gateway.py
```diff
@@ -825,6 +826,7 @@ def __init__(
         self.loop: asyncio.AbstractEventLoop = loop
         self._keep_alive: Optional[VoiceKeepAliveHandler] = None
         self._close_code: Optional[int] = None
+        self._speaking_map = {}
```
AbstractUmbra, line 829: Missing type annotations.

discord/gateway.py
```diff
@@ -1011,3 +1024,11 @@ async def close(self, code: int = 1000) -> None:
 
         self._close_code = code
         await self.ws.close(code=code)
+
+    def get_member_from_ssrc(self, ssrc) -> Optional[Union['Member', int]]:
+        if ssrc in self._speaking_map:
+            user = self._speaking_map[ssrc]["user"]
+            if type(user) == int and (member := self._connection.guild.get_member(user)) is not None:
```
AbstractUmbra, line 1031: `isinstance` is more preferable, no?

AbstractUmbra (changes requested):
Thanks a bunch for the PR.

I had a look and this seems to be missing a lot of type information which the library now uses fully.

It also seems like the styling rules (`black` has rules in our `pyproject.toml` file) have not been used, and this would fail checks pre-merge.

I will advise that I personally am not fond of the example you introduced.
It uses commands and command-like situations but still uses `discord.Client`.
I understand the want for minimalist but at this point I feel `commands.Bot` should be used, as it allows for checks and the like.
It is also missing comments to explain how, what and why for some things, and type information (like the rest of the library).

examples/basic_voice_listening.py
```diff
@@ -0,0 +1,141 @@
...
+        self.out = None
+
+        self.commands = {
+            'start': self.start_listening,
+            'stop': self.stop_listening,
+            'pause': self.pause_listening,
+            'resume': self.resume_listening,
+        }
```
LucasCoderT, line 45: Why not use `ext.commands` here?

discord/sink.py
```diff
@@ -0,0 +1,775 @@
...
+        "ssrc", "report_blocks", "extension"
+    )
+
+    def __init__(self, version_flag: int, rtcp_type: RTCPMessageType, length: int, data: bytes):
+        super().__init__(version_flag, rtcp_type, length)
+
+        fmt = ">I"
+        buf_size = struct.calcsize(fmt)
+        self.ssrc = struct.unpack(fmt, data[:buf_size])[0]
+        self._parse_report_and_extension(data[buf_size:])
```
mikeshardmind, line 234: - Should probably prefer `!` to `>` (same byte order, but `!` has a specific meaning as to being network order)
- Shouldn't need to call calcsize on a literal
- Shouldn't slice a buffer passed to unpack, use unpack_from
- this last function should probably be removed rather than just exist with code commented out.


```suggestion
        self.ssrc = struct.unpack_from("!I", data, 0)[0]
        self.report_blocks = []
        self.data = data[4:]
```

discord/sink.py
```diff
@@ -0,0 +1,775 @@
...
+
+    def __init__(self, version_flag: int, rtcp_type: RTCPMessageType, length: int, data: bytes):
+        super().__init__(version_flag, rtcp_type, length)
+
+        fmt = ">IQ3I"
+        buf_size = struct.calcsize(fmt)
+        self.ssrc, self.nts, self.rts, self.spc, self.soc = \
+            struct.unpack(fmt, data[:buf_size])
+
+        self._parse_report_and_extension(data[buf_size:])
```
mikeshardmind, line 203: Same deal as other suggestion...

```suggestion
        self.ssrc, self.nts, self.rts, self.spc, self.soc = \
            struct.unpack_from("!IQ3I, data, 0)
        self.report_blocks = []
        self.extension = data[:24]
```

discord/sink.py
```diff
@@ -0,0 +1,775 @@
...
+        super().__init__(version_flag, rtcp_type, length)
+
+        if self.rc == 0:
+            buf_size = 0
+            self.ssrc_byes = []
+        else:
+            buf_size = self.rc*struct.calcsize("I")
+            self.ssrc_byes = struct.unpack(f">{self.rc}I", data[:buf_size])
+        reason_length = struct.unpack(">B", data[buf_size:buf_size+1])
+        self.reason = b"" if reason_length == 0 else struct.unpack(f">{reason_length}s", data[buf_size+1:])
```
mikeshardmind, line 311: similar to previous things on struct performance and convention + a fix 

```suggestion
            buf_size = self.rc * 4
            self.ssrc_byes = struct.unpack_from(f"!{self.rc}I", data, 0)
        reason_length = struct.unpack_from("!B", data, buf_size)
        self.reason = b"" if reason_length == 0 else struct.unpack_from(f"!{reason_length}s", data, buf_size+1)
```

discord/sink.py
```diff
@@ -0,0 +1,775 @@
...
+    __slots__ = RTCPPacket.__slots__ + ("ssrc", "name", "app_data")
+
+    def __init__(self, version_flag: int, rtcp_type: RTCPMessageType, length: int, data: bytes):
+        super().__init__(version_flag, rtcp_type, length)
+
+        fmt = ">I4s"
+        buf_size = struct.calcsize(fmt)
+        self.ssrc, self.name = struct.unpack(fmt, data[:buf_size])
+        self.name = self.name.decode("ascii")
+        self.app_data = data[buf_size:]
```
mikeshardmind, line 342: More of the same
```suggestion
        self.ssrc, self.name = struct.unpack_from("!I4s, data, 0)
        self.name = self.name.decode("ascii")
        self.app_data = data[8:]
```

mikeshardmind (commented):
I've stopped going through it at this point. There are a lot of small issues with struct all over the place here, including a situation noted where things were unpacked from the tuples returned from struct which should have noticeable caused errors had this been tested and the code path been hit.

Jourdelune:
Yes, There seems to be some issue in the code, I wrote this basic implementation:
``py
import discord


DEV_GUILD = discord.Object(id=913766363791757353)

class Client(discord.Client):
    def __init__(self, *, intents: discord.Intents):
        super().__init__(intents=intents)

    async def setup_hook(self):
        self.tree.copy_global_to(guild=DEV_GUILD)
        await self.tree.sync(guild=DEV_GUILD)

bot = Client(intents=intents)


@bot.event
async def on_ready():
    print(f'Logged in as {bot.user} (ID: {bot.user.id})')
    print('------')


@bot.tree.command()
async def start(interaction: discord.Interaction):
    """Start listening."""

    vc = interaction.user.voice
        
    if not vc:
        return await interaction.response.send_message('You\'re not in a vc right now')

    voice_client = discord.utils.get(bot.voice_clients, guild=interaction.guild)
    if voice_client and voice_client.channel.id != vc.channel.id:
        await voice_client.move_to(vc.channel)
    else:
        voice_client = await vc.channel.connect()

    voice_client.listen(discord.MP3AudioFileSink(output_dir='/tmp'), after=on_listening_stopped)
    await interaction.response.send_message(f'Started listening')


@bot.tree.command()
async def stop(interaction: discord.Interaction):
    """Stop listening."""

    voice_client = discord.utils.get(bot.voice_clients, guild=interaction.guild)
    if not voice_client:
        return await interaction.response.send_message(f'I am not connected to a voice channel.')

    if not voice_client.is_listening():
        return await interaction.response.send_message("Not currently listening")
    
    voice_client.stop_listening()
    await voice_client.disconnect()
 
    await interaction.response.send_message(f'No longer listening.')


@bot.tree.command()
async def pause_listening(interaction: discord.Interaction):
    """Pause listening."""

    voice_client = discord.utils.get(bot.voice_clients, guild=interaction.guild)
    if not voice_client:
        return await interaction.response.send_message(f'I am not connected to a voice channel.')

    if not voice_client.is_listening():
        return await interaction.response.send_message("Not currently listening")
    
    if voice_client.is_listening_paused():
        return await interaction.response.send_message("Listening already paused")
    
    voice_client.pause_listening()
    await interaction.response.send_message("Listening has been paused")


@bot.tree.command()
async def resume_listening(interaction: discord.Interaction):
    """Resume listening."""

    voice_client = discord.utils.get(bot.voice_clients, guild=interaction.guild)

    if not voice_client:
        return await interaction.response.send_message(f'I am not connected to a voice channel.')

    if not voice_client.is_listening():
        return await interaction.response.send_message("Not currently listening")
    
    if not voice_client.is_listening_paused():
        return await interaction.response.channel.send("Already resumed")
    
    voice_client.resume_listening()
    await interaction.response.send_message("Listening has been resumed")


def on_listening_stopped(sink, exc=None):
    sink.convert_files()


bot.run("TOKEN")
```

When you try to record with `/start` and after close the record with `/stop`, you have an error:
```
  File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/sink.py", line 741, in run
    self._do_run()
  File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/sink.py", line 731, in _do_run
    packet = self.client.recv_audio_packet(dump=not self._resumed.is_set())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/voice_client.py", line 842, in recv_audio_packet
    raise err[0]
TypeError: exceptions must derive from BaseException

During handling of the above exception, another exception occurred:

  File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/sink.py", line 754, in _call_after
    self.after(self.sink, error)
  File "/home/jourdelune/Bureau/Interaction/Bot/test.py", line 100, in on_listening_stopped
    sink.convert_files()
  File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/sink.py", line 613, in convert_files
    self.output_files[ssrc] = self.convert_file(f, new_name)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/sink.py", line 691, in convert_file
    process = subprocess.Popen(args, creationflags=subprocess.CREATE_NO_WINDOW)
                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'subprocess' has no attribute 'CREATE_NO_WINDOW'
```.

You can delete the args 'CREATE_NO_WINDOW'.
I share this basic implementation for someone that want test the code^^.
Also that can be useful if the callback `on_listening_stopped` in voice_client.listen(discord.WaveAudioFileSink(output_dir='/tmp'), after=on_listening_stopped) is awaited because you want maybe send the audio file in a channel.

**Comment by AbstractUmbra** [comment]
> First, thank you a lot for your integration, voice recording is highly requested by discord.py users (including me) and I really hope that this commit will be merged.
> There seems to be some issue in the code, I write this basic implementation:
> ```py
> import discord
> from discord import app_commands
> 
> 
> DEV_GUILD = discord.Object(id=913766363791757353)
> 
> class Client(discord.Client):
>     def __init__(self, *, intents: discord.Intents):
>         super().__init__(intents=intents)
>         self.tree = app_commands.CommandTree(self)
> 
>     async def setup_hook(self):
>         self.tree.copy_global_to(guild=DEV_GUILD)
>         await self.tree.sync(guild=DEV_GUILD)
> 
> intents = discord.Intents.default()
> bot = Client(intents=intents)
> 
> 
> @bot.event
> async def on_ready():
>     print(f'Logged in as {bot.user} (ID: {bot.user.id})')
>     print('------')
> 
> 
> @bot.tree.command()
> async def start(interaction: discord.Interaction):
>     """Start listening."""
> 
>     vc = interaction.user.voice
>         
>     if not vc:
>         return await interaction.response.send_message('You\'re not in a vc right now')
> 
>     voice_client = discord.utils.get(bot.voice_clients, guild=interaction.guild)
>     if voice_client and voice_client.channel.id != vc.channel.id:
>         await voice_client.move_to(vc.channel)
>     else:
>         voice_client = await vc.channel.connect()
> 
>     voice_client.listen(discord.MP3AudioFileSink(output_dir='/tmp'), after=on_listening_stopped)
>     await interaction.response.send_message(f'Started listening')
> 
> 
> @bot.tree.command()
> async def stop(interaction: discord.Interaction):
>     """Stop listening."""
> 
>     voice_client = discord.utils.get(bot.voice_clients, guild=interaction.guild)
>     if not voice_client:
>         return await interaction.response.send_message(f'I am not connected to a voice channel.')
> 
>     if not voice_client.is_listening():
>         return await interaction.response.send_message("Not currently listening")
>     
>     voice_client.stop_listening()
>     await voice_client.disconnect()
>  
>     await interaction.response.send_message(f'No longer listening.')
> 
> 
> @bot.tree.command()
> async def pause_listening(interaction: discord.Interaction):
>     """Pause listening."""
> 
>     voice_client = discord.utils.get(bot.voice_clients, guild=interaction.guild)
>     if not voice_client:
>         return await interaction.response.send_message(f'I am not connected to a voice channel.')
> 
>     if not voice_client.is_listening():
>         return await interaction.response.send_message("Not currently listening")
>     
>     if voice_client.is_listening_paused():
>         return await interaction.response.send_message("Listening already paused")
>     
>     voice_client.pause_listening()
>     await interaction.response.send_message("Listening has been paused")
> 
> 
> @bot.tree.command()
> async def resume_listening(interaction: discord.Interaction):
>     """Resume listening."""
> 
>     voice_client = discord.utils.get(bot.voice_clients, guild=interaction.guild)
> 
>     if not voice_client:
>         return await interaction.response.send_message(f'I am not connected to a voice channel.')
> 
>     if not voice_client.is_listening():
>         return await interaction.response.send_message("Not currently listening")
>     
>     if not voice_client.is_listening_paused():
>         return await interaction.response.channel.send("Already resumed")
>     
>     voice_client.resume_listening()
>     await interaction.response.send_message("Listening has been resumed")
> 
> 
> def on_listening_stopped(sink, exc=None):
>     sink.convert_files()
> 
> 
> bot.run("TOKEN")
> ```
> 
> When you try to record with `/start` and after close the record with `/stop`, you have an error:
> ```
> Traceback (most recent call last):
>   File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/sink.py", line 741, in run
>     self._do_run()
>   File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/sink.py", line 731, in _do_run
>     packet = self.client.recv_audio_packet(dump=not self._resumed.is_set())
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/voice_client.py", line 842, in recv_audio_packet
>     raise err[0]
> TypeError: exceptions must derive from BaseException
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/sink.py", line 754, in _call_after
>     self.after(self.sink, error)
>   File "/home/jourdelune/Bureau/Interaction/Bot/test.py", line 100, in on_listening_stopped
>     sink.convert_files()
>   File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/sink.py", line 613, in convert_files
>     self.output_files[ssrc] = self.convert_file(f, new_name)
>                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/sink.py", line 691, in convert_file
>     process = subprocess.Popen(args, creationflags=subprocess.CREATE_NO_WINDOW)
>                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: module 'subprocess' has no attribute 'CREATE_NO_WINDOW'
> ```
> 
> You can delete the args 'CREATE_NO_WINDOW'.
> I share this basic implementation for someone that want test the code^^.
> Also that can be useful if the callback `on_listening_stopped` in `voice_client.listen(discord.WaveAudioFileSink(output_dir='/tmp'), after=on_listening_stopped)` is awaited and return the file because you want maybe send the audio in a channel.

The issue here is that this argument only exists and works on Windows platforms.
The PR should include a system check before executing.

The python docs do not mention this strangely.

**Comment by mikeshardmind** [comment]
> The python docs do not mention this strangely.

They actually do. [CREATE_NO_WINDOW](https://docs.python.org/3/library/subprocess.html#subprocess.CREATE_NO_WINDOW) is specifically under the windows constants section, and where mentioned prior to this, it's in reference to the also windows only [STARTUPINFO](https://docs.python.org/3/library/subprocess.html#subprocess.STARTUPINFO)

File: discord/sink.py (hunk 1 above)
Comment (Sheepposu) [reply discord/sink.py:41]: Yeah that's true. I'll switch to using `typing.NamedTuple`.

File: discord/sink.py (hunk 2 above)
Comment (Sheepposu) [reply discord/sink.py:190]: Thanks for the tip

File: discord/sink.py (hunk 3 above)
Comment (Sheepposu) [reply discord/sink.py:260]: Not sure I understand what you mean. If this function were typed, would it not be `_parse_chunk(self, data: bytes) -> Tuple[RTCPSourceDescriptionChunk, int]`? I can add the typing since private methods elsewhere in the project also have typing.

Comment (AbstractUmbra) [reply discord/sink.py:260]: Perfect.

File: discord/sink.py (hunk 7 above)
Comment (Sheepposu) [reply discord/sink.py:708]: I used `thread.Thread` to mimic the `AudioPlayer` class, but I suppose that doesn't matter at all.

File: discord/utils.py (hunk 9 above)
Comment (Sheepposu) [reply discord/utils.py:73]: Will do.

File: discord/voice_client.py (hunk 10 above)
Comment (Sheepposu) [reply discord/voice_client.py:48]: Will do.

File: discord/gateway.py (hunk 17 above)
Comment (Sheepposu) [reply discord/gateway.py:1031]: yep it is

File: discord/sink.py (hunk 19 above)
Comment (Sheepposu) [reply discord/sink.py:234]: > this last function should probably be removed rather than just exist with code commented out.

I've kept it there since it would be used if discord sent valid rtcp packets. 
All the other points are fair. I've checked your other reviews as well, so I'll go through and make the changes.

**Comment by Sheepposu** [comment]
I'll work on redoing the current example file with the critique I've gotten from everyone.

**Comment by Sheepposu** [comment]
Is it preferable that the example be based on the basic_voice.py example or instead use a slash command implementation? Could also simply scrap the example all together if it's not really needed.

**Comment by Jourdelune** [comment]
The basic voice example has been write over a year ago so maybe you can implement slash command. Also I have found an another issue:
```
ERROR:discord.sink:Calling the after function failed.
  File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/sink.py", line 756, in _call_after
    self.after(self.sink, error)
  File "/home/jourdelune/Bureau/Interaction/Bot/commands/transcribe.py", line 41, in on_listening_stopped
    sink.convert_files()
  File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/sink.py", line 613, in convert_files
    new_name = f"audio-{user.name}#{user.discriminator}-{ssrc}" if user is not None else None
                        ^^^^^^^^^
AttributeError: 'int' object has no attribute 'name'
```
user is the id of the user, not a python object. Here is the code:
```py
from typing import Literal

import discord
import asyncio


class Transcribe(commands.Cog):
    def __init__(self, bot):
        self.bot = bot


    @app_commands.command(name="transcribe", description="transcribe.describe")
    @app_commands.checks.cooldown(1, 5.0, key=lambda i: (i.user.id))
    @app_commands.describe(style='transcribe.describe.style')
    async def transcribe(self, interaction: discord.Interaction, style: Literal['message', 'embed', 'webhook']):
        """Transcribe voice channel."""
        voice_channel = interaction.user.voice
        
        if voice_channel is None:
            return await interaction.response.send_message(await interaction.translate('transcribe.not_in_voice'))

        voice_channel = interaction.user.voice.channel
        voice_client = discord.utils.get(self.bot.voice_clients, guild=interaction.guild)

        if voice_client is None:
            voice_client = await voice_channel.connect()
        else:
            if voice_channel != voice_client.channel:
                return await interaction.response.send_message(await interaction.translate('transcribe.already'))

        voice_client.play(discord.FFmpegPCMAudio('./sounds/start.mp3'))

        await interaction.response.send_message(await interaction.translate('transcribe.description'))
        voice_client.listen(discord.WaveAudioFileSink(output_dir='/tmp/'), after=self.on_listening_stopped)
        await asyncio.sleep(3)
        voice_client.stop_listening()
        
    def on_listening_stopped(self, sink, exc=None, *, args = None): 
        sink.convert_files() 
        print(sink, exc, args)
    

async def setup(bot):
    await bot.add_cog(Transcribe(bot))
```
and I don't know if the library already does it but that can be interesting if you add the guild id to the file name, a bot can be in more that one channel at the same time (on different guild).

discord/sink.py
```diff
@@ -0,0 +1,858 @@
...
+        """Goes through all files, opens them and calls convert_file, then deletes them.
+
+        convert_file must be defined, else NotImplementedError will be raised
+        """
+        if not self.done:
+            self.cleanup()
+        for ssrc, file in self.output_files.items():
+            filepath = file.name
+            user = self._ssrc_to_user[ssrc]
+            new_name = f"audio-{user.name}#{user.discriminator}-{ssrc}" if user is not None else None
```
Jourdelune, line 681: new_name = f"audio-{user}-{ssrc}" if user else None

discord/sink.py
```diff
@@ -0,0 +1,858 @@
...
+            with open(filepath, "rb") as f:
+                self.output_files[ssrc] = self.convert_file(f, new_name)
+
+            os.remove(filepath)
+        self.output_files = {}
```
Jourdelune, line 686: that can be interesting to return the converted file and also allows the user to add arguments to the callback in listen:after parameters

discord/sink.py
Jourdelune: finally it seems that use is:
- the identifier
- or the object discord.py
I search why.

AbstractUmbra: Line #680 is the reason. Since AudioFrame.user can be int or Member, it doesn't check first.

Had this PR been reviewed with an active TypeChecker this would have been caught.

discord/utils.py
```diff
@@ -1374,3 +1374,8 @@ def _shorten(
 
 def _to_kebab_case(text: str) -> str:
     return CAMEL_CASE_REGEX.sub('-', text).lower()
+
+
+def strip_namedtuple_docs(cls: NamedTuple) -> None:
```
Gobot1234, line 1379: This also shouldn't pass type checking I don't think? I'd really advise running Pyright on this PR

Sheepposu: I've switched to using dataclasses so this is irrelevant now.

discord/sink.py
Sheepposu: Implemented a function that decides the file name format based on the type of `user`

discord/sink.py
```diff
@@ -0,0 +1,914 @@
...
+        The synchronization source identifier for the originator of this SR packet.
+    report_blocks: Sequence[:class:`RTCPReceiverReportPacket`]
+        Sequence of :class:`RTCPReceiverReportPacket` objects that tell statistics.
+        Receivers do not carry over statistics when a source changes its SSRC
+        identifier due to a collision.
+    extension: :class:`bytes`
+        Profile-specific extension that may or may not contain a value.
+    """
+
+    __slots__ = RTCPPacket.__slots__ + (
```
Gobot1234, line 264: This isn't how slots work, you only need to define the slots you use in this subclass

discord/sink.py
```diff
@@ -0,0 +1,914 @@
...
+    "RTCPSourceDescriptionChunk",
+    "RTCPSourceDescriptionItem",
+)
+
+
+_log = logging.getLogger(__name__)
+
+
+@dataclass
+class RTCPReceiverReportBlock:
```
Gobot1234, line 42: Considering how many of these are made, I feel like these would benefit from having slots

discord/sink.py
Sheepposu: Ah, I was unaware of this

Jourdelune:
When I look at the code, I don't understand how it can be implemented.
In fact, the callback contained in after, called by the listen function, only gets 2 arguments, the sink and a potential exception, I don't think that's enough. For example, if it records a bot discord in a voice channel on two different servers, how do you know when a recording ends from which server it came from...? 

Without this information it is difficult to make a useful use of the recording, I think it would be good to implement adding arguments to the callback, this way by adding the server id in the callback arguments it would be possible to know where the recording is coming from and for example send the file to the room. It might also be useful to specify which user each audio recording is from, although this can be done via the filename in sink.output_files.

Sheepposu:
> When I look at the code, I don't understand how it can be implemented. In fact, the callback contained in after, called by the listen function, only gets 2 arguments, the sink and a potential exception, I don't think that's enough. For example, if it records a bot discord in a voice channel on two different servers, how do you know when a recording ends from which server it came from...?
> 
> Without this information it is difficult to make a useful use of the recording, I think it would be good to implement adding arguments to the callback, this way by adding the server id in the callback arguments it would be possible to know where the recording is coming from and for example send the file to the channel. It might also be useful to specify which user each audio recording is from, although this can be done via the filename in sink.output_files.

You make good points on user experience with the functionality. I'll make it so that args and kwargs can be passed, and also create a new class specifically for the audio file that can carry information on the user and such.

Vexs:
The callback doesn't need to receive that information that way- you could for example, use a sink factory that took that information in from the scope.

Sheepposu:
> The callback doesn't need to receive that information that way- you could for example, use a sink factory that took that information in from the scope.

Could you expand on the idea of a sink factory?

Vexs:
Sorry, mixed up some terms and the actual problem itself on my part. Don't need a sink factory- what he's looking for would just be a closure.

Consider the following:
```
	def listen_stopped_closure(self, guild, event)
		def on_listening_stopped(sink, exc=None, *, args = None): 
        	    sink.convert_files() 
                    event.set()
        	    whatever etc etc 
           return on_listening_stopped

    ....
        event = asyncio.Event()
        voice_client.listen(discord.WaveAudioFileSink(output_dir='/tmp/'), after=self.listen_stopped_closure(voice_channel.guild, event))
        await event.wait()
        voice_client.stop_listening()
```

Jourdelune:
I have see an another issue:
```py
from typing import Literal

import discord
import asyncio

class Transcribe(commands.Cog):
    def __init__(self, bot):
        self.bot = bot
        self.listen_guild = {}


    @app_commands.command(name="transcribe", description="transcribe.describe")
    @app_commands.checks.cooldown(1, 5.0, key=lambda i: (i.user.id))
    @app_commands.describe(style='transcribe.describe.style')
    async def transcribe(self, interaction: discord.Interaction, style: Literal['message', 'embed', 'webhook']):
        """Transcribe voice channel."""
        voice_channel = interaction.user.voice
        
        if voice_channel is None:
            return await interaction.response.send_message(await interaction.translate('transcribe.not_in_voice'))

        voice_channel = interaction.user.voice.channel
        voice_client = discord.utils.get(self.bot.voice_clients, guild=interaction.guild)

        if voice_client is None:
            voice_client = await voice_channel.connect()
        else:
            if voice_channel != voice_client.channel:
                return await interaction.response.send_message(await interaction.translate('transcribe.already'))

        self.listen_guild[interaction.guild.id] = Listener()
        
        voice_client.play(discord.FFmpegPCMAudio('./sounds/start.mp3'))
      
        await interaction.response.send_message(await interaction.translate('transcribe.description'))

        voice_client.listen(discord.AudioFileSink(discord.WaveAudioFile, '/tmp/'), after=self.on_listening_stopped, guild=interaction.guild)
        await asyncio.sleep(3)
        voice_client.stop_listening()
    

    async def on_listening_stopped(self, sink, exc, guild): 
        sink.convert_files() # block here
        for file in sink.output_files.values(): 
            print(file)
                
    

async def setup(bot):   
    await bot.add_cog(Transcribe(bot))
```

In on_listening_stopped, the function sink.convert_files() block everything, I have found that the function convert in WaveAudioFile block the code when it try to read the file:
```py
class WaveAudioFile(AudioFile):
    CHUNK_WRITE_SIZE = 64

    def convert(self, new_name: Optional[str] = None) -> None:
        """Write the raw audio data to a wave file.

        Extends :class:`AudioFile`

        Parameters
        ----------
        new_name: Optional[:class:`str`]
            Name for the wave file excluding ".wav". Defaults to current name if None.
        """

        path = self._get_new_path(self.path, "wav", new_name)
        with wave.open(path, "wb") as wavf:
            wavf.setnchannels(OpusDecoder.CHANNELS)
            wavf.setsampwidth(OpusDecoder.SAMPLE_SIZE // OpusDecoder.CHANNELS)
            wavf.setframerate(OpusDecoder.SAMPLING_RATE)

            print(self.file.read(OpusDecoder.FRAME_SIZE * self.CHUNK_WRITE_SIZE)) # block here

            while frames := self.file.read(OpusDecoder.FRAME_SIZE * self.CHUNK_WRITE_SIZE):
                wavf.writeframes(frames)

        os.remove(self.path)
        file = open(path, "rb")
        file.close()
        self.file = file
```
I can't suggest fix for the moment because I don't understand the cause.

Jourdelune:
```py
import discord


DEV_GUILD = discord.Object(id=913766363791757353)

class Client(discord.Client):
    def __init__(self, *, intents: discord.Intents):
        super().__init__(intents=intents)

    async def setup_hook(self):
        self.tree.copy_global_to(guild=DEV_GUILD)
        await self.tree.sync(guild=DEV_GUILD)

bot = Client(intents=intents)


@bot.event
async def on_ready():
    print(f'Logged in as {bot.user} (ID: {bot.user.id})')
    print('------')


@bot.tree.command()
async def start(interaction: discord.Interaction):
    """Start listening."""

    vc = interaction.user.voice
        
    if not vc:
        return await interaction.response.send_message('You\'re not in a vc right now')

    voice_client = discord.utils.get(bot.voice_clients, guild=interaction.guild)
    if voice_client and voice_client.channel.id != vc.channel.id:
        await voice_client.move_to(vc.channel)
    else:
        voice_client = await vc.channel.connect()

    voice_client.listen(discord.AudioFileSink(discord.MP3AudioFile, '/tmp/'), after=on_listening_stopped)
    await interaction.response.send_message(f'Started listening')


@bot.tree.command()
async def stop(interaction: discord.Interaction):
    """Stop listening."""

    voice_client = discord.utils.get(bot.voice_clients, guild=interaction.guild)
    if not voice_client:
        return await interaction.response.send_message(f'I am not connected to a voice channel.')

    if not voice_client.is_listening():
        return await interaction.response.send_message("Not currently listening")
    
    voice_client.stop_listening()
    await voice_client.disconnect()
 
    await interaction.response.send_message(f'No longer listening.')


@bot.tree.command()
async def pause_listening(interaction: discord.Interaction):
    """Pause listening."""

    voice_client = discord.utils.get(bot.voice_clients, guild=interaction.guild)
    if not voice_client:
        return await interaction.response.send_message(f'I am not connected to a voice channel.')

    if not voice_client.is_listening():
        return await interaction.response.send_message("Not currently listening")
    
    if voice_client.is_listening_paused():
        return await interaction.response.send_message("Listening already paused")
    
    voice_client.pause_listening()
    await interaction.response.send_message("Listening has been paused")


@bot.tree.command()
async def resume_listening(interaction: discord.Interaction):
    """Resume listening."""

    voice_client = discord.utils.get(bot.voice_clients, guild=interaction.guild)

    if not voice_client:
        return await interaction.response.send_message(f'I am not connected to a voice channel.')

    if not voice_client.is_listening():
        return await interaction.response.send_message("Not currently listening")
    
    if not voice_client.is_listening_paused():
        return await interaction.response.channel.send("Already resumed")
    
    voice_client.resume_listening()
    await interaction.response.send_message("Listening has been resumed")


async def on_listening_stopped(sink, exc=None):
    sink.convert_files()
    for file in sink.output_files.values():
        file.file.read() # block here
        print('ok')

bot.run("Token")
```
Here is an example of the issue with MP3

Jourdelune:
It is not totally fixed, indeed `file.file.read()` also blocks^^.
```py
async def on_listening_stopped(sink, exc=None):
    sink.convert_files()   
    for file in sink.output_files.values():
        print(file.user.id)
        print(file.file.read()) # block
```

If the code running in a cogs, the `file.user` will be a int but outside a cogs, it will be a user object, shouldn't this be constant? Otherwise it is confusing.

Sheepposu:
> It is not totally fixed, indeed `file.file.read()` also blocks^^.

I find this strange because the file attribute in this case should be None after `convert` finishes running. 

> Also if the code running in a cogs, the `file.user` will be a int but outside a cogs, it will be a user object, shouldn't this be constant? Otherwise it is confusing.

I'm not sure I understand what you mean. The value of `file.user` depends on the "best" value that can be returned in the audio frames. `VoiceClient` asks the gateway for a user object associated with the ssrc and gets back either nothing, a user id, or a `Member` object. It passes that value to the Sink inside the `AudioFrame` object. `AudioFile` caches that value depending on its currently cached value (e.g. if the current value is a user id it and its given a `Member` object it will replace the user id). There's not a way for the value to be constant, though it will almost always return either `None` or `Member`.

Jourdelune:
> Also if the code running in a cogs, the file.user will be a int but outside a cogs, it will be a user object, shouldn't this be
> constant? Otherwise it is confusing.
ah right I thought user was something you could control.

> I find this strange because the file attribute in this case should be None after convert finishes running.
Indeed, you can test the code by replacing the function `on_listening_stopped` if you wish.

Sheepposu:
I believe it's not blocking the code, but actually an exception is occurring and it's not being shown in the console due to the error handling, which calls _log.exception (is this supposed to show in the console by default?). Either way, the user should use `file.path` to open the file themselves or do whatever.

Jourdelune:
I agree, I think it would be better to trigger an error because otherwise it may cause the user to misunderstand.

dolfies:
> I'm not sure I understand what you mean. The value of file.user depends on the "best" value that can be returned in the audio frames. VoiceClient asks the gateway for a user object associated with the ssrc and gets back either nothing, a user id, or a Member object. It passes that value in the AudioFrame object to the Sink. AudioFile caches that value depending on its currently cached value (e.g. if the current value is a user id and it's given a Member object, it will replace the user id). There's not a way for the value to be constant, though it will almost always return either None or Member.

Wouldn't it be better to use `discord.Object`?

Sheepposu:
> Wouldn't it be better to use `discord.Object`?

I assume you mean replacing `int` with `discord.Object`? That works well for the use of `discord.Object` so I'll consider implementing that.

Jourdelune:
also it's usefull to raise error in the after callback, All the errors are otherwise glossed over, which does not facilitate development.

Jourdelune:
There is also a very common silent error, sometimes the converter does not find the file to convert.
```py
import traceback
async def on_listening_stopped(sink, exc=None):
    try:
        sink.convert_files()
        print('ok')
    except Exception:
        traceback.print_exc()
```
```
  File "/home/jourdelune/Bureau/Interaction/Bot/test.py", line 61, in on_listening_stopped
    sink.convert_files()
  File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/sink.py", line 707, in convert_files
    file.convert(self._create_name(file))
  File "/home/jourdelune/.local/lib/python3.11/site-packages/discord/sink.py", line 866, in convert
    with open(self.path, "rb") as file:
         ^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/audio-201351.pcm'
```

Sheepposu:
> There is also a very common silent error, sometimes the converter does not find the file to convert.

Do you have any steps for reproducing this? I've never come across this problem, and I'm not sure how this exception could come about in the first place.

Jourdelune:
yes it's due to me putting the files in /tmp/, so it's not code related :p

Sheepposu:
Should be fixed, just put something like this inside the loop:
```Python
while voice_client.is_listening() or voice_client.is_listen_cleaning():
    await asyncio.sleep(0.1)
```

Sheepposu:
Ah, thanks for catching that.

Jourdelune:
Thank you for the fix, but maybe indeed of use
```py
while voice_client.is_listening() or voice_client.is_listen_cleaning():
```
the `voice_client.stop_listening` can only end once all this has happened, right? (i.e. only when it has stopped listening and deleted the file).

Sheepposu:
`voice_client.stop_listening` calls `AudioReceiver.stop`, which simply sets some `threading.Event` objects telling the thread that it can end the main loop and cleanup. Requiring `voice_client.stop_listening` to wait for all that cleanup (which also includes calling the after function) before returning feels like it could create some annoying bottlenecks, so I prefer this method.

mikeshardmind:
> Should be fixed, just put something like this inside the loop:
> 
> ```python
> while voice_client.is_listening() or voice_client.is_listen_cleaning():
>     await asyncio.sleep(0.1)
> ```

This is a really... not great user API. Waiting in a loop like this is significantly worse than deferring this to asyncio and the event loop. If this is meant to be a typical pattern of waiting until done listening, there should be a single user `await` to... wait for it to be done.

This can be accomplished with your choice of tools in asyncio, there are several. The most appropriate here if voice_clients are not meant to be reused would be an event, otherwise, a future. Neither of these presents a situation where someone might get stuck in a loop like this.

Sheepposu:
Alright I made a more better functionality with one async function that waits for listen to be ready to execute again. I made this class below to replace a couple threading.Event objects so that wait can be used asynchronously, and it works, though I wasn't entirely sure if 1. this is a good solution 2. this class is actually thread-safe (I'm pretty sure it is).

```Python
class _AsynchronousThreadSafeEvent(threading.Event):
    def __init__(self):
        super().__init__()
        self._waiters: Queue = Queue()

    def set(self) -> None:
        super().set()
        while not self._waiters.empty():
            future = self._waiters.get()
            future._loop.call_soon_threadsafe(future.set_result, True)

    async def async_wait(self) -> None:
        if self.is_set():
            return
        future = asyncio.get_running_loop().create_future()
        self._waiters.put(future)
        await future
```

mikeshardmind:
Issues mentioned in the [Prior PR](https://github.com/Rapptz/discord.py/pull/6507#issuecomment-844401202)  remain relevent here. I'll follow up in more detail with a way to empirically show the effects of this later

mikeshardmind:
I'll take the time to write something that models realistically normal delays later, but the sledgehammer version that demonstrates that this fails to isolate the real-time audio processing:

```py
# don't use this in a production bot
import random
import threading
import time


def sledgehammer():
    while True:
        # random to prevent any implementation details from getting smart, intentionally slow operation
        _a = pow(random.getrandbits(4096), random.getrandbits(4096), random.getrandbits(4096) + 1)
        time.sleep(0.9)

x = threading.Thread(target=sledgehammer, daemon=True)
x.start()

# put your minimal test bot for the feature below
```

on a modern CPU, this introduces ~128ms without a threading context switch about once a second.
This shouldn't be enough of a delay to get forcibly disconnected from discord, but it is unacceptable large an artificial delay in a production bot.

I'll write something that models real-world "acceptable" blocking later, but attempting to record using this PR and this bit of a sledgehammer easily demonstrates the lack of isolating the real-time processing the library needs to do from any disturbances caused by user code.

It's worth noting that 5x 15ms gaps in a 1-minute window, as well as a single 50ms gap *ever* is a noticeable aberration when it comes to audio and human listening, so while the above is an extreme example to get to the point quickly and make it easy to show the issue, there are realistic thresholds that would cause an issue for audio recording, but otherwise be acceptable in a bot outside of this concern. (i.e. it isn't possible to handwave this away as only an issue if someone has other problems with their bot)

Sheepposu:
Ah I see. Thanks for explaining in detail.

Sheepposu:
I have a couple questions:
1. Is this problem not also present in the functionality that plays audio?
2. As I understand it, this seems like a problem that lies with the limitations of threading, that being it all performs within one process. So is this problem even solvable without using a separate process to handle the audio? A solution like that is certainly doable, but are there any simpler solutions?

mikeshardmind:
> I have a couple questions:
> 
> 1. Is this problem not also present in the functionality that plays audio?
> 2. As I understand it, this seems like a problem that lies with the limitations of threading, that being it all performs within one process. So is this problem even solvable without using a separate process to handle the audio? A solution like that is certainly doable, but are there any simpler solutions?

1. It is, but the interfaces provided for voice playing are extensible enough that people aware of the limitations and who care about them end up using things like wavelink + lavalink, which does end up putting all the playing into another process that's actually designed with realtime audio concerns in mind. The interfaces here aren't extensible enough to easily have someone do similar in the future. If audio sending was currently not yet supported, and the only option proposed had such an issue (or if I was active at the time it was originally implemented) I'd have advocated for the base implementation to be more resilient to these issues as well. The current ecosystem has developed enough to make that moot though.

2. Pretty much a python limitation rather than threading specifically. There's very little that can be done to isolate pure python code from the effects of other code running in the same process. A dedicated subprocess designed to handle real time-sensitive concerns is what most people would consider the correct option here. There are a handful of other things that can be done with native code as well, but that would vastly increase the complexity of this PR, and it isn't necessary for performance or design reasons.

Sheepposu:
Thanks for answering. I'll replace `threading` with `multiprocessing` to get rid of this issue.

Sheepposu:
> * 1. This creates a thread and spawns a process per channel being listened to.
> * 2. This doesn't handle the unavailability of pynacl gracefully
> * 3. This misuses `__del__` and `__new__`
> * 4. This discards silent voice frames.
> * 5. The logic for reordering out-of-order frames is slightly broken by this implementation.

1. That's a good point. It would be much better to create one or more process(es) for handling the audio processing of all channels. Not sure how many channels one process could handle, not to mention it's also dependent on the number of sources it's listening to per channel. Maybe a good option then would be to use a ProcessPoolExecutor object across all the voice clients.
2. Could you explain?
3. For the __del__ method, is it because it possibly calls blocking code? I can see why that's a problem, and in that case I can pass an argument to only close the file and not write to it if that's better. For __new__ could you explain?
4. In hindsight the silent frame can be used to know that a user stopped talking, so I'll actually pass it back to the receiver. 
5. Could you explain?

I'll go through and fix all the smaller issues I spot, including the ones you mentioned. I'll also just delete the commented out code since it could be easily restored if necessary.

mikeshardmind:
> > * 1. This creates a thread and spawns a process per channel being listened to.
> > * 2. This doesn't handle the unavailability of pynacl gracefully
> > * 3. This misuses `__del__` and `__new__`
> > * 4. This discards silent voice frames.
> > * 5. The logic for reordering out-of-order frames is slightly broken by this implementation.
> 
> 1. That's a good point. It would be much better to create one or more process(es) for handling the audio processing of all channels. Not sure how many channels one process could handle, not to mention it's also dependent on the number of sources it's listening to per channel. Maybe a good option then would be to use a ProcessPoolExecutor object across all the voice clients.
> 2. Could you explain?
> 3. For the **del** method, is it because it possibly calls blocking code? I can see why that's a problem, and in that case I can pass an argument to only close the file and not write to it if that's better. For **new** could you explain?
> 4. In hindsight the silent frame can be used to know that a user stopped talking, so I'll actually pass it back to the receiver.
> 5. Could you explain?
> 
> I'll go through and fix all the smaller issues I spot, including the ones you mentioned. I'll also just delete the commented out code since it could be easily restored if necessary.

1. probably best to use a pool that users can set the size of. Can distribute across the pool by guild_id % pool_size to get a relatively even distribution without needing to track active listening in each process (snowflakes are evenly distributed, it won't end up perfect, but it should be good enough, sharding works on a similar principle)
2. See `has_nacl` use in voice_client.py for a model of handling this for sink.py
3a.  (RE: `__del__`), 99% of the time defining `__del__` on an object is a mistake, and should instead be handled with either a context manager or use of `atexit` `__del__` doesn't carry the same guarantees each of the others does. 
3b. (RE: `__new__` ) AudioPacket having a `__new__` which returns non-instances of AudioPacket breaks typing and isn't neccessary. There's nothing preventing this from just being a function returning one of the types it an return in this case rather than a class with `__new__` that does this.
4. thanks
5. The buffer is small and you are writing 5 frames of 10 at a time while discarding any local knowledge of sequence info from those frames. there are subtle ordering issues with this approach that can occur on even small timescales. (see below for logic beyond just expanding the size of the buffer) For writing to a file, the buffer can be significantly larger without the primary drawbacks associated with realtime communication and large jitter buffers (delay, underrun) as there isn't someone listening in realtime to this. An adaptive buffer using system time and sequence info could reasonably tolerate issues with this better. There won't ever be a perfect solution to this, memory use tradeoffs at a minimum will always exist, but considering the writing is to a file, not to another realtime socket for processing here, a larger adaptive buffer is appropriate.

\* The logical improvements here would be to keep track of the last uninterrupted sequence number and always be willing to write the next in sequence immediately, the buffer doesn't need to contain it as there *can't* be a misordering in that case. This approach should be paired with an internal time for the packet being received (system, not from the packet: time.monotonic) to have a maximum time to wait for the next in sequence, then jumping to the earliest sequence number in heap (O(1)) and using the time it was received, dropping anything sequenced prior to it if received after it. Pairing this approach with utilizing the 5 frames of silence from discord as a signal would improve the expected tolerance to networking delays significantly, as well as not allow out-of-order writes resulting from discarding info about what has already been written (ie [1,2,4,5,6,7,8,9,10,11], then receiving 3 after [1,2,4,5,6] are written out)

Sheepposu:
Thanks for expanding on your points. I'll work on implementing all the feedback wherever I find time.

Veylkh:
> Thanks for expanding on your points. I'll work on implementing all the feedback wherever I find time.

Thank you for your work.

Is there anything I could help with?

Sheepposu:
> Is there anything I could help with?

Thank you for the offer, but I think it's easier for me to implement the changes myself. I've just had too much schoolwork recently to work on it at all.

Veylkh:
> > Is there anything I could help with?
> 
> Thank you for the offer, but I think it's easier for me to implement the changes myself. I've just had too much schoolwork recently to work on it at all.

Any news on the matter?

Sheppsu:
> Any news on the matter?
For a while I've had most of the code put together, but I got really busy with other projects. I recently started working on finalizing this next commit, so I'll push it today or tomorrow. Only problem is it's hard to test it to its full capacity, which involves the bot being in multiple voice channels and listening in them all at the same time. I'll most likely worry about that later, and first make sure the general structure will be able to function well.

Sheppsu:
I've updated the summary section to include a detailed summary of how it now functions.

Sheppsu:
I'm now noticing an inaccuracy with the audio saved to file that isn't really reproducible, or at least I don't know how to reproduce it. It has to do with too much silence being added in between audio frames. I'll try finding out what the cause is.

Veylkh:
> > Any news on the matter?
> 
> For a while I've had most of the code put together, but I got really busy with other projects. I recently started working on finalizing this next commit, so I'll push it today or tomorrow. Only problem is it's hard to test it to its full capacity, which involves the bot being in multiple voice channels and listening in them all at the same time. I'll most likely worry about that later, and first make sure the general structure will be able to function well.

Great news. Let me know if you need someone to help for the testing.

Sheppsu:
> > > Any news on the matter?
> > 
> > 
> > For a while I've had most of the code put together, but I got really busy with other projects. I recently started working on finalizing this next commit, so I'll push it today or tomorrow. Only problem is it's hard to test it to its full capacity, which involves the bot being in multiple voice channels and listening in them all at the same time. I'll most likely worry about that later, and first make sure the general structure will be able to function well.
> 
> Great news. Let me know if you need someone to help with the testing.

Yes, I appreciate any help finding problems with the code.

Sheppsu:
I figured out the issue to that silence problem: when the bot joined the vc and began listening, I talked and discord sent frames which indicated the amount of time I had been silent, including before the bot joined the vc. However, there's a distinct pattern that it can be identified by. I noticed that discord always sends silent frames out of order, for example: it sends sequence 65, but then sends 10 silent frames with sequences 70-79, and frames 66-69 appear afterward, which creates a gap after the silent frames (when not sorted) that looks like this: 66, 67, 68, 69, 80. When that silence problem occurs I noticed the same pattern, except the silent frames are missing. So instead, `AudioFileSink` just checks for a 11 sequence skip on the 6th frame of the listening session, which has worked every time in my testing.

nickdarnell:
Looked at integrating this PR into my own fork, and I had a question for you and @Rapptz.  It looks like the voice channels are limited to a single connection, re: `interaction.guild.voice_client`.  Is this an API issue or something the discord API actually prevents a bot from being in more than one voice channel?

nickdarnell:
Also, instead of AudioFile, it would preferrable (at least to me) that your base class be an AudioBuffer that doesn't presume to need to be a file.  Ideally i would pass in the buffer you would use rather than it needing to be a specific path, and instead could be an memory buffer, and I could pass in something like io.BytesIO() instead of a path.

Sheppsu:
> Looked at integrating this PR into my own fork, and I had a question for you and @Rapptz. It looks like the voice channels are limited to a single connection, re: `interaction.guild.voice_client`. Is this an API issue or something the discord API actually prevents a bot from being in more than one voice channel?

I have no clue and taking a quick look at discord's api docs, I see no mention of it either.

> Also, instead of AudioFile, it would preferrable (at least to me) that your base class be an AudioBuffer that doesn't presume to need to be a file. Ideally i would pass in the buffer you would use rather than it needing to be a specific path, and instead could be an memory buffer, and I could pass in something like io.BytesIO() instead of a path.

I'll consider making a sink for byte buffers, but for now the closest thing to that is subclassing `AudioHandlingSink` and implementing the `on_valid_audio` method to put frames into buffers.

Rapptz:
> It looks like the voice channels are limited to a single connection, re: interaction.guild.voice_client. Is this an API issue or something the discord API actually prevents a bot from being in more than one voice channel?

That's how the API works. One connection per guild_id.

nickdarnell:
@Sheppsu Took me a bit to track it down, but haven't made a fix for it yet in my own code.  Just wanted to let you know that sequence does not go up forever, it rolls over at 65535, so after you listen to audio for a long enough time for one user, it rolls over and suddenly the validation code stops sending any valid audio because now all the sequences have started over at 0.

Sheppsu:
> @Sheppsu Took me a bit to track it down, but haven't made a fix for it yet in my own code. Just wanted to let you know that sequence does not go up forever, it rolls over at 65535, so after you listen to audio for a long enough time for one user, it rolls over and suddenly the validation code stops sending any valid audio because now all the sequences have started over at 0.

Ah ok, I'll have to patch that then

nickdarnell:
AudioProcessPool needs a bit of love.  If you create any temporary bots, that do any recording, their subprocesses from the pool don't get cleaned up when the bot closes.  Those subprocesses should go away when the bot does for sure at the very least.

Sheppsu:
> AudioProcessPool needs a bit of love. If you create any temporary bots, that do any recording, their subprocesses from the pool don't get cleaned up when the bot closes. Those subprocesses should go away when the bot does for sure at the very least.

That's a good point. I'll add some cleanup functionality to the process pool.

monkeyplayz1:
alr

nickdarnell:
@Rapptz I'd say this PR is in a pretty good state, I've had a lot of success with it, especially with the latest non-blocking cleanup changes.  Think it can be merged soon?  Looks like there are conflicts right now @Sheppsu

Rapptz:
As good time as any I guess, but I've been thinking about the state of the various voice receive implementations and decided that I'm probably not going to merge any of them into the mainline repository going forward. If you want a voice receive implementation they're most likely better off as extension modules (e.g. `discord.ext.my_voice_recv`). One of the contributors for this library has already [taken the initiative to do this](https://github.com/imayhaveborkedit/discord-ext-voice-recv) and is probably a better way forward since it detaches the maintenance burden from me to the person who wrote the PR.

Fundamentally the reason for all this is because voice receive is not documented by Discord in any capacity nor do they show signs of wanting to. Only a few libraries have voice receive implementations and they're all pretty much unmaintained or in a very frozen state. Having more code like that in the library would be mostly a headache for me.

To that end, thanks @Sheppsu for the PR even if I decided against merging it. It should be relatively easy to move this over to an extension module and put it on PyPI. If you need help in doing this you're free to go to the [Discord server](https://discord.gg/dpy) and ask for help there.

Sheppsu:
That's understandable and I respect your decision. I'll consider creating an extension module when I find the time (maybe sometime this week).

Sheppsu:
Finally made an [extension](https://github.com/Sheppsu/discord-ext-listening) of this PR.

PR #9296, opened 2023-03-11
Add and document `Guild.get_emoji`

rdrescher909:
## Summary

Adds a `Guild.get_emoji` method for getting emojis from a Guild by id.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -737,6 +737,26 @@ class Guild(Hashable):
         """
         return self._threads.get(thread_id)
 
+    def get_emoji(self, emoji_id: int, /) -> Optional[Emoji]:
+        """Returns an emoji with the given ID.
+
+        .. versionadded:: 2.3
+
+        Parameters
+        ----------
+        emoji_id: int
+            The ID to search for.
+
+        Returns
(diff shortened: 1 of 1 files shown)


discord/guild.py
```diff
@@ -737,6 +737,26 @@ def get_thread(self, thread_id: int, /) -> Optional[Thread]:
...
+
+        Returns
+        --------
+        Optional[:class:`Emoji`]
+            The returned Emoji or ``None`` if not found.
+        """
+        emoji = self._state.get_emoji(emoji_id)
+        if emoji and emoji.guild == self:
+            return emoji
+        return None
```
Puncher1, line 758: Check with `if x is not None` rather than with `if x` if you want to check if emoji is not `None`.
Also return `emoji` at the end for better readability. 

```suggestion
        if emoji is not None and emoji.guild != self:
            return None
        return emoji
```

Rapptz: I partially agree with this but I think the current code is also fine.

Rapptz (approved):

PR #9300, opened 2023-03-12
Add missing params `default_reaction_emoji` and `default_forum_layout` to `create_forum`

Puncher1:
## Summary

This PR adds 2 missing parameters `default_reaction_emoji` and `default_forum_layout` to `Guild.create_forum`

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -74,6 +74,7 @@ from .enums import (
     Locale,
     AutoModRuleEventType,
     ForumOrderType,
+    ForumLayoutType,
 )
 from .mixins import Hashable
 from .user import User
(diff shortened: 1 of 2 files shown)


Rapptz (approved):
This PR (and the previous one) is misusing `MISSING` and `None` but I'll fix it myself later.

Puncher1:
Ah yeah, sorry I missed that. So, for my understanding, is the following correct?: 

- `var: x = MISSING` means it's not required but that you can "catch" it you use `MISSING`
- `var: Optional[x] = None` means it's not required too, but if not set it's sent as a `null` value to the endpoint. 
- `var: Optional[x] = MISSING` means the same as the first except of being able to be `None`. E.g. this is used in an `edit` method.

Rapptz:
Yes. Though using `Optional[X] = None` is also there for compatibility in some methods.

Puncher1:
Alright! Thanks

PR #9306, opened 2023-03-17
Add support for widget image

Puncher1:
## Summary

This PR adds a new method `widget_image_url` to `Guild`, which returns an URL to the widget image.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -3948,11 +3948,6 @@ class Guild(Hashable):
             The style which should be applied for the widget.
             Default to ``shield``. Returns ``None`` if widget is not enabled.
 
-        Raises
-        -------
-        HTTPException
-            Retrieving the widget image url failed.
-
         Returns
         --------
         Optional[:class:`str`]
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -75,6 +75,7 @@ from .enums import (
     AutoModRuleEventType,
     ForumOrderType,
     ForumLayoutType,
+    WidgetStyle,
 )
 from .mixins import Hashable
 from .user import User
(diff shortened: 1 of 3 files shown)


discord/guild.py
```diff
@@ -3931,6 +3932,38 @@ async def widget(self) -> Widget:
 
         return Widget(state=self._state, data=data)
 
+    WidgetStyle = Literal['shield', 'banner1', 'banner2', 'banner3', 'banner4']
```
Gobot1234, line 3935: I wouldn't advise putting this in the namespace as then you can do Guild.WidgetStyle which sits a little weird with me

discord/guild.py
```diff
@@ -3931,6 +3933,36 @@ async def widget(self) -> Widget:
...
+        Parameters
+        -----------
+        style: :class:`str`
+            The style which should be applied for the widget.
+            Default to ``shield``. Returns ``None`` if widget is not enabled.
+
+        Raises
+        -------
+        HTTPException
+            Retrieving the widget image url failed.
```
Rapptz, line 3954: This doesn't raise anything.

Rapptz (commented):
I'm still unsure what the proper name for this is, since it's inconsistent with other methods that are similar, i.e. `Asset` but we talked about this in the discord server.

discord/guild.py
Puncher1: Ah okay. Well it does when the user specifies a wrong `style` but that's not in the bot's scope anymore, the link just returns an error in the browser.

Puncher1:
Yes. I didn't found a good name for it. I first thought about `widget_image_link` but idk if this is the way to go.

Puncher1:
How about turning this into an helper function in `utils.py` like `oauth_url`?

Puncher1:
Closing this because it doesn't seem to be useful. Can be reopened again if this changes.

PR #9318, opened 2023-03-29
Add and change `Guild` attributes and `Guild.edit` params

Puncher1:
### Addition

**`Guild`**
* `max_stage_video_channel_users` attribute
* `safety_alerts_channel` property

**`Guild.edit` parameters**
* `raid_alerts_disabled` which adds/removes `RAID_ALERTS_DISABLED` feature
* `safety_alerts_channel` -> `None` has no effect but doesn't raise an HTTP exception

### Changes

`Guild.edit`
* Better error handling if wrong type is passed for `public_updates_channel` and `rules_channel`. It's not possible to set them `None` either, when community is enabled. 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -267,6 +267,10 @@ class Guild(Hashable):
         Indicates if the guild has widget enabled.
 
         .. versionadded:: 2.0
+    max_stage_video_channel_users: Optional[:class:`int`]
+        The maximum amount of users in a stage video channel.
+
+        .. versionadded:: 2.3
     """
 
     __slots__ = (
@@ -315,6 +319,8 @@ class Guild(Hashable):
         'approximate_member_count',
         'approximate_presence_count',
         'premium_progress_bar_enabled',
+        '_safety_alerts_channel_id',
+        'max_stage_video_channel_users',
     )
 
     _PREMIUM_GUILD_LIMITS: ClassVar[Dict[Optional[int], _GuildLimit]] = {
@@ -478,6 +484,7 @@ class Guild(Hashable):
         self.max_presences: Optional[int] = guild.get('max_presences')
         self.max_members: Optional[int] = guild.get('max_members')
         self.max_video_channel_users: Optional[int] = guild.get('max_video_channel_users')
+        self.max_stage_video_channel_users: Optional[int] = guild.get('max_stage_video_channel_users')
         self.premium_tier: int = guild.get('premium_tier', 0)
         self.premium_subscription_count: int = guild.get('premium_subscription_count') or 0
         self.vanity_url_code: Optional[str] = guild.get('vanity_url_code')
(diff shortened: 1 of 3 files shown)


discord/guild.py
```diff
@@ -267,6 +267,10 @@ class Guild(Hashable):
         Indicates if the guild has widget enabled.
 
         .. versionadded:: 2.0
+    max_stage_video_channel_users: Optional[:class:`int`]
```
bijij, line 270: This name to me feels like it's an instance where we should deviate from discord.

Puncher1: What about `max_stage_video_users`?

Puncher1: Changed it to `max_stage_video_users`

discord/guild.py
```diff
@@ -801,6 +809,18 @@ def public_updates_channel(self) -> Optional[TextChannel]:
...
 
+    @property
+    def safety_alerts_channel(self) -> Optional[TextChannel]:
+        """Optional[:class:`TextChannel`]: Return's the guild's channel used for safety alerts. E.g. this is used in raid protection.
+        The guild must be a Community guild.
+
+        If no channel is set, then this returns ``None``.
+
+        .. versionadded:: 2.3
+        """
```
Rapptz, line 820:
```suggestion
        """Optional[:class:`TextChannel`]: Return's the guild's channel used for safety alerts, if set. 

        For example, this is used for the raid protection setting. The guild must have the ``COMMUNITY`` feature.

        .. versionadded:: 2.3
        """
```

discord/guild.py
```diff
@@ -1894,16 +1916,14 @@ async def edit(
             .. versionchanged:: 2.0
 
                 Now accepts an enum instead of :class:`str`.
-        rules_channel: Optional[:class:`TextChannel`]
+        rules_channel: :class:`TextChannel`
```
Rapptz, line 1919: Why are you removing the `Optional` aspect of this?

discord/guild.py
```diff
@@ -1894,16 +1916,14 @@ async def edit(
...
             The new channel that is used for rules. This is only available to
-            guilds that contain ``COMMUNITY`` in :attr:`Guild.features`. Could be ``None`` for no rules
-            channel.
+            guilds that contain ``COMMUNITY`` in :attr:`Guild.features`.
 
             .. versionadded:: 1.4
-        public_updates_channel: Optional[:class:`TextChannel`]
+        public_updates_channel: :class:`TextChannel`
```
Rapptz, line 1924: Why are you removing the `Optional` aspect of this?

discord/guild.py
```diff
@@ -1934,6 +1954,17 @@ async def edit(
         reason: Optional[:class:`str`]
             The reason for editing this guild. Shows up on the audit log.
 
+        raid_alerts_disabled: :class:`bool`
+            Whether the alerts for raid protection should be disabled for the guild.
+
+            .. versionadded:: 2.3
+
+        safety_alerts_channel: :class:`TextChannel`
```
Rapptz, line 1962: This should be `Optional`.

Rapptz (changes requested):
Note that these are `Optional` because there needs to be a way to *remove* and disable these.

discord/guild.py
Puncher1: It cannot be `None` if community is enabled, but I noticed it does when it's not a community, my fault I'll revert this and this for `public_updates_channel`.

discord/guild.py
Puncher1: It does not have an effect if `safety_alerts_channel` is set to `None`. I thought would be better to prevent confusion since it's not possible on Discord's site either.

Rapptz (approved):
LGTM.

For future changes please restrict things to one PR per change.

PR #9326, opened 2023-04-01
[docs] Fix on_socket_raw_send and on_raw_thread_member_remove param types

Puncher1:
## Summary

This PR adds missing param type for `on_socket_raw_send` and fixes wrong param type for `on_raw_thread_member_remove`. 

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -491,8 +491,9 @@ Debug
         WebSocket. The voice WebSocket will not trigger this event.
 
     :param payload: The message that is about to be passed on to the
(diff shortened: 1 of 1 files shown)


Rapptz (changes requested):
The type is there for implementation lenience even if it's not strictly true right now. It used to be true and will probably be true again at some point in the future without causing a change in guarantee. The other change of `member` -> `payload` is fine.

Rapptz (approved):

PR #9328, opened 2023-04-02
Support for AutoMod Members

Puncher1:
## Summary

This PR adds support for AutoMod Members. This also adds the new param `mention_raid_protection` in `AutoModTrigger`.
Related PR: https://github.com/discord/discord-api-docs/pull/6040

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/automod.py
+++ b/discord/automod.py
@@ -67,6 +67,7 @@ class AutoModRuleAction:
     -----------
     type: :class:`AutoModRuleActionType`
         The type of action to take.
+        Defaults to :attr:`~AutoModRuleActionType.block_message`.
 
         .. versionchanged:: 2.3
             This is an optional parameter.
@@ -139,16 +140,13 @@ class AutoModRuleAction:
         if sum(v is None for v in (channel_id, duration, custom_message)) < 2:
             raise ValueError('Only one of channel_id, duration, or custom_message can be passed.')
 
+        self.type: AutoModRuleActionType = AutoModRuleActionType.block_message
         if type is not None:
             self.type = type
         elif channel_id is not None:
             self.type = AutoModRuleActionType.send_alert_message
         elif duration is not None:
             self.type = AutoModRuleActionType.timeout
-        elif custom_message is not None:
-            self.type = AutoModRuleActionType.block_message
-        else:
-            raise ValueError('Please pass the action type explicitly if not using channel_id, custom_message, or duration.')
 
     def __repr__(self) -> str:
         return f'<AutoModRuleAction type={self.type} channel={self.channel_id} duration={self.duration}>'
--- a/discord/automod.py
+++ b/discord/automod.py
@@ -112,9 +112,7 @@ class AutoModRuleAction:
         ...
 
     @overload
-    def __init__(
-        self, *, type: Literal[AutoModRuleActionType.block_message], custom_message: Optional[str] = ...
-    ) -> None:
+    def __init__(self, *, type: Literal[AutoModRuleActionType.block_message], custom_message: Optional[str] = ...) -> None:
         ...
 
     @overload
@@ -154,7 +152,7 @@ class AutoModRuleAction:
             self.type = AutoModRuleActionType.block_message
 
     def __repr__(self) -> str:
-        return f'<AutoModRuleAction type={self.type} channel={self.channel_id} duration={self.duration}>'
+        return f'<AutoModRuleAction type={self.type.value} channel={self.channel_id} duration={self.duration}>'
 
     @classmethod
     def from_data(cls, data: AutoModerationActionPayload) -> Self:
--- a/discord/automod.py
+++ b/discord/automod.py
@@ -68,9 +68,6 @@ class AutoModRuleAction:
     type: :class:`AutoModRuleActionType`
         The type of action to take.
         Defaults to :attr:`~AutoModRuleActionType.block_message`.
-
-        .. versionchanged:: 2.3
-            This is an optional parameter.
     channel_id: Optional[:class:`int`]
         The ID of the channel or thread to send the alert message to, if any.
         Passing this sets :attr:`type` to :attr:`~AutoModRuleActionType.send_alert_message`.
--- a/discord/automod.py
+++ b/discord/automod.py
@@ -92,7 +92,7 @@ class AutoModRuleAction:
         ...
 
     @overload
-    def __init__(self, *, type: Literal[AutoModRuleActionType.send_alert_message] = ..., channel_id: int = ...) -> None:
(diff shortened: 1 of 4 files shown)


bijij (commented):
I'm very against making a breaking change here, not sure what harm there is in still having `block_message` be the default.

At most I'd say not explicitly passing type should spit back a deprecation warning.

docs/api.rst
```diff
@@ -3290,6 +3302,12 @@ of :class:`enum.Enum`.
 
         The rule will timeout a user.
 
+    .. attribute:: block_member_interactions
+
+        Similar to :attr:`timeout` the rule will timeout a user except of having a duration.
```
kaylynn234, line 3307: The wording here needs to be updated. It's very unclear what you're trying to say.

docs/api.rst
```diff
@@ -3290,6 +3302,12 @@ of :class:`enum.Enum`.
...
+        This will request the user to edit it's profile.
```
kaylynn234, line 3308: "it" is usually used to refer to objects rather than people, so I don't think it's the right fit here. It would probably be better to go with "their" instead?

(and just for the record, "it's" is the incorrect word to use here. "it's" is a shortening of "it is", whereas "its" is the possessive term, and refers to ownership of something)

docs/api.rst
```diff
@@ -3262,6 +3262,12 @@ of :class:`enum.Enum`.
         The rule will trigger when combined number of role and user mentions
         is greater than the set limit.
 
+    .. attribute:: member_profile
+
+        The rule will trigger when a user's profile contains a keyword.
```
kaylynn234, line 3267: It might be worth clarifying *which* keywords trigger the rule, since this could seem ambiguous otherwise.

Imagine you're somebody reading the documentation - which keywords trigger the rule? Are these keywords explicitly configured? If so, where are they configured?

kaylynn234 (commented):
Implementation aside, I think the documentation could be improved a bit. Nothing major!

Puncher1:
> I'm very against making a breaking change here, not sure what harm there is in still having `block_message` be the default.
> 
> At most I'd say not explicitly passing type should spit back a deprecation warning.

There's just no reason why `block_message` should be the default, but since it would be a breaking change, I agree.

discord/automod.py
```diff
@@ -89,16 +91,43 @@ def __init__(self, *, channel_id: Optional[int] = ...) -> None:
...
+    def __init__(
+        self, *, type: Literal[AutoModRuleActionType.block_message] = ..., custom_message: Optional[str] = ...
+    ) -> None:
+        ...
+
+    @overload
+    def __init__(self, *, type: Literal[AutoModRuleActionType.block_member_interactions] = ...) -> None:
+        ...
+
     def __init__(
```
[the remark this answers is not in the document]
Puncher1: I would change the first error from
```py
raise ValueError(f'{attribute} can only be passed if type is {action_type.name}.')
```
to
```py
raise ValueError(f'{attribute} can only be passed if type is {action_type.name} or None.')
```
Otherwise I think this is good, thanks!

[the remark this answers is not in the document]
Puncher1: One thing I noticed in 
```py
if type is None and sum(v is None for v in (channel_id, duration, custom_message)) < 2:
   ^^^^^^^^^^^^
```
`type` is irrelevant since it also applies when `type` is not `None`.

discord/automod.py
```diff
@@ -89,16 +91,43 @@ def __init__(self, *, channel_id: Optional[int] = ...) -> None:
         ...
 
     @overload
```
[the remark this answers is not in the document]
Puncher1: 1. Yeah true, neither `custom_message` should. 
2. Why? `type` is optional.

docs/api.rst
Puncher1: What about this?
```
Similar to :attr:`timeout` the rule will timeout a user except of having a limit.
```

docs/api.rst
Puncher1: Ah okay, thanks for clarifying. As you may have already noticed, English is not my native language. 😄

docs/api.rst
Puncher1: Yeah, the description is based on the description of `.. attribute:: keyword` which is basically based on the same functionality. Also it's just the enum `AutoModTriggerType` which is needed to create a `AutoModTrigger`. There it's better documented inside a table anyway: 
```py
+-----------------------------------------------+------------------------------------------------+
| :attr:`AutoModRuleTriggerType.member_profile` | :attr:`keyword_filter`, :attr:`regex_patterns`,|
|                                               | :attr:`allow_list`                             |
+-----------------------------------------------+------------------------------------------------+
 ```
 So I think it's fine.

File: docs/api.rst (hunk 1 above)
Comment (kaylynn234) [reply docs/api.rst:3307]: I think the second is a good improvement. It might be good to reorder the sentence a bit for clarity:
```rst
Similar to :attr:`timeout`, except the user will be timed out indefinitely.
```
This also leads nicely into the next sentence, which is a bonus!

discord/automod.py
```diff
@@ -85,36 +88,76 @@ class AutoModRuleAction:
...
+
+        if self.type is AutoModRuleActionType.send_alert_message and channel_id is None:
+            raise ValueError('channel_id must be set if AutoModRuleActionType is send_alert_message')
+        else:
+            self.channel_id: Optional[int] = channel_id
+
+        if self.type is AutoModRuleActionType.timeout and duration is None:
+            raise ValueError('duration must be set if AutoModRuleActionType is timeout')
+        else:
+            self.duration: Optional[datetime.timedelta] = duration
```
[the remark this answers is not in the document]
Puncher1: Initially I wanted it at the end of the constructor, so I forgot to replace `self.type` with `type`. But it's better to have it at the end anyway, so I put it back now.

Rapptz (changes requested):

0xallie:
Is there anything still holding this PR back?

Duued:
> Is there anything still holding this PR back?

You may have forgotten to re-request review from Rapptz

0xallie:
Not my PR, just curious about its state.

Puncher1:
It's still not documented on Discord's side, see here: https://github.com/discord/discord-api-docs/pull/6040

Puncher1:
This PR can be merged already. The original PR cannot isn't merged yet due to merge conflicts but the "not released" label has been removed, see this comment: https://github.com/discord/discord-api-docs/pull/6040#issuecomment-1634744229

Rapptz (changes requested):
This PR just needs to change the 2.3 references to 2.4.

Rapptz (approved):
Thanks!

Issue #9329, opened 2023-04-03
Have `DefaultHelpCommand` use `get_bot_mapping`


DanielBaulig:
### Summary

DefaultHelpCommand does not use and ignores the mapping provided to send_bot_help.

### Reproduction Steps

It's impossible to customize the DefaultHelpCommand implementation by altering it's `get_bot_mapping` implementation.

```
from discord.ext import commands

class CustomizedDefaultHelpCommand(commands.DefaultHelpCommand):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
    def get_bot_mapping(self):
        mapping = super().get_bot_mapping()
        # make some arbitrary changes to mapping
        return mapping
```

It [seems](https://github.com/Rapptz/discord.py/blob/master/discord/ext/commands/help.py#L1234-L1262) DefaultHelpCommand just ignores the mapping argument passed to it and "scours" the bot manually itself. There doesn't seem to be a good reason for that, but maybe I am missing something?

Would it be possible to rewrite DefaultHelpCommand's `send_bot_help` implementation to use the provided mapping instead of gathering one itself? That way it would be easily possible to adjust which commands to display and how without having to reimplement `send_bot_help` itself. 

### Minimal Reproducible Code

```python
See above.
```

### Expected Results

Overriding and changing `get_bot_mapping` should result in the actual output of `send_bot_help` to change.

### Actual Results

Output of `send_bot_help` remains unphased.

### Intents

message_content = True, but seems irrelevant

### System Information

- Python v3.10.10-final
- discord.py v2.2.2-final
- aiohttp v3.8.4
- system info: Linux 6.1.12-arch1-1 #1 SMP PREEMPT_DYNAMIC Tue, 14 Feb 2023 22:08:08 +0000

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.

### Additional Context

I've been trying to change the structure of the !help output to better suit my cog layout/structure. Also see the following SO question for more background: https://stackoverflow.com/questions/75908271/how-to-make-a-cog-a-prefix-command-group

DanielBaulig:
I'd be happy to provide a pull request to make the necessary changes if I can expect it to be accepted.

PR #9330, opened 2023-04-03
Use get_bot_mapping to generate bot help (Resolves issue #9329)

DanielBaulig:
##Summary 

Resolves issue: https://github.com/Rapptz/discord.py/issues/9329

This change makes it so that `DefaultHelpCommand` uses it's `mapping` argument (generated by `get_bot_mapping()`), to generate it's help output. This change does not directly change DefaultHelpCommand behavior and will continue to generate the same output, but allows subclasses to override get_bot_mapping to make alterations to the return value of get_bot_mapping to, for example, add or remove elements.

## Checklist

    - [ ] I have updated the documentation to reflect the changes. (Not applicable, actually behavior now matches documentation)
- [x] This PR fixes an issue (https://github.com/Rapptz/discord.py/issues/9329)

### DIFF
--- a/discord/ext/commands/help.py
+++ b/discord/ext/commands/help.py
@@ -1241,18 +1241,17 @@ class DefaultHelpCommand(HelpCommand):
 
         no_category = f'\u200b{self.no_category}:'
 
-        def get_category(command, *, no_category=no_category):
-            cog = command.cog
+        def get_cog_name(cog, *, no_category=no_category):
             return cog.qualified_name + ':' if cog is not None else no_category
+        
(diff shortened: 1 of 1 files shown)


Rapptz (changes requested):
This PR seems to change a lot of the code for what, at a glance, would appear to be a minimal change. Likewise, there are trailing spaces which would fail CI. I recommend running `black` on the files.

For example, this PR seems to change the grouping logic that was set, and seems to remove comments and change function names which doesn't seem to be in the spirit of the change.

PR #9333, opened 2023-04-04
Use get_role where appropriate in role decorators

Gobot1234:
## Summary

Small micro-optimisation to use the new method.

## Checklist

- [ ] This PR fixes an issue.


diceroll123:
Should `item` be a `Union[int, str]` for `bot_has_role` as well as `bot_has_any_role`? Seems like an oversight since `has_role` has this annotation

PR #9334, opened 2023-04-04
Improve handling of ArrayFlags

mikeshardmind:
## Summary
This is a micro-optimization. Figured since it was easy to improve on, and these need to be created for potentially a lot of events, it didn't hurt to do slightly better.

Why Discord did it this way instead of as an integer bitfield like everything reasonable is unknowable, but this does less work to get to the same result.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/flags.py
+++ b/discord/flags.py
@@ -25,6 +25,7 @@ DEALINGS IN THE SOFTWARE.
 from __future__ import annotations
 
 from functools import reduce
+from operator import or_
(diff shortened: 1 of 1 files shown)


Rapptz (changes requested):
I'd merge this but it's obscenely unreadable compared to the original code. It could use a small comment to clarify it for future readers.

mikeshardmind:
I've added an in-depth comment on this. I don't know why Discord went this route with it, but better to make sure anyone reading it later has an explanation, some of what's done there would be obscure to future readers.

Rapptz (approved):
Thanks.

PR #9339, opened 2023-04-05
Add channel attribute in Interaction

Puncher1:
## Summary

This PR adds a new attribute `channel` in Interaction.
Related PR: https://github.com/discord/discord-api-docs/pull/6051

### Discord Limitation
Because Discord doesn't send `recipients` in the payload when an interaction was sent from a DM channel, a new type `InteractionDMChannel` is added in which `recipients` is a `NotRequired` field.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -3061,11 +3061,7 @@ class GroupChannel(discord.abc.Messageable, discord.abc.PrivateChannel, Hashable
         self.owner_id: Optional[int] = utils._get_as_snowflake(data, 'owner_id')
         self._icon: Optional[str] = data.get('icon')
         self.name: Optional[str] = data.get('name')
-        self.recipients: List[User] = []
-
-        recipients = data.get('recipients')
-        if recipients is not None:
-            self.recipients = [self._state.store_user(u) for u in recipients]
+        self.recipients: List[User] = [self._state.store_user(u) for u in data.get('recipients', [])]
 
         self.owner: Optional[BaseUser]
         if self.owner_id == self.me.id:
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -2887,7 +2887,12 @@ class DMChannel(discord.abc.Messageable, discord.abc.PrivateChannel, Hashable):
 
     def __init__(self, *, me: ClientUser, state: ConnectionState, data: DMChannelPayload):
         self._state: ConnectionState = state
-        self.recipient: Optional[User] = state.store_user(data['recipients'][0])
+        self.recipient: Optional[User] = None
+
+        recipients = data.get('recipients')
+        if recipients is not None:
+            self.recipient = state.store_user(recipients[0])
+
         self.me: ClientUser = me
         self.id: int = int(data['id'])
 
@@ -3056,7 +3061,11 @@ class GroupChannel(discord.abc.Messageable, discord.abc.PrivateChannel, Hashable
         self.owner_id: Optional[int] = utils._get_as_snowflake(data, 'owner_id')
         self._icon: Optional[str] = data.get('icon')
         self.name: Optional[str] = data.get('name')
-        self.recipients: List[User] = [self._state.store_user(u) for u in data.get('recipients', [])]
+        self.recipients: List[User] = []
+
+        recipients = data.get('recipients')
+        if recipients is not None:
+            self.recipients = [self._state.store_user(u) for u in recipients]
 
         self.owner: Optional[BaseUser]
         if self.owner_id == self.me.id:
--- a/discord/interactions.py
+++ b/discord/interactions.py
@@ -194,13 +194,7 @@ class Interaction(Generic[ClientT]):
 
             channel = data['channel']
             if ch_type in (ChannelType.group, ChannelType.private):
-                try:
-                    channel = factory(me=self._client.user, data=channel, state=self._state)  # type: ignore
-                except KeyError:
-                    # Discord limitation
-                    channel = PartialMessageable(
-                        state=self._state, guild_id=self.guild_id, id=int(channel['id']), type=ch_type
-                    )
+                channel = factory(me=self._client.user, data=channel, state=self._state)  # type: ignore
             else:
                 guild = self._state._get_guild(self.guild_id)
                 channel = factory(guild=guild, state=self._state, data=channel)  # type: ignore
--- a/discord/interactions.py
+++ b/discord/interactions.py
@@ -186,7 +186,7 @@ class Interaction(Generic[ClientT]):
         self.channel: Optional[InteractionChannel] = None
 
         raw_channel = data.get('channel', {})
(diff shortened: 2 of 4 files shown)


discord/channel.py
```diff
@@ -3056,7 +3061,11 @@ def _update_group(self, data: GroupChannelPayload) -> None:
         self.owner_id: Optional[int] = utils._get_as_snowflake(data, 'owner_id')
         self._icon: Optional[str] = data.get('icon')
         self.name: Optional[str] = data.get('name')
-        self.recipients: List[User] = [self._state.store_user(u) for u in data.get('recipients', [])]
+        self.recipients: List[User] = []
+
+        recipients = data.get('recipients')
+        if recipients is not None:
+            self.recipients = [self._state.store_user(u) for u in recipients]
```
bijij, line 3068: Undo this, the old impl achieved the same result.

```suggestion
        self.recipients: List[User] = [self._state.store_user(u) for u in data.get('recipients', [])]
```

discord/interactions.py
```diff
@@ -173,6 +184,23 @@ def _from_data(self, data: InteractionPayload):
         self.version: int = data['version']
         self.channel_id: Optional[int] = utils._get_as_snowflake(data, 'channel_id')
```
bijij, line 185: the docs mention we should stop using this, and instead use the id in the `channel` field. Maybe do that too.

bijij (changes requested):
Overall, not sure why we've made `channel` an attribute instead of a property
I'd have just added a new `_channel_type` attrib and reworked the property to rely on that.

Aside from that I've left some nitpicks as review comments.

Puncher1:
> Overall, not sure why we've made `channel` an attribute instead of a property I'd have just added a new `_channel_type` attrib and reworked the property to rely on that.
> 
> Aside from that I've left some nitpicks as review comments.

Yeah, maybe better. Then, if `InvalidData` is raised, it does only effect the channel and not the whole object.

discord/interactions.py
```diff
@@ -98,6 +106,9 @@ class Interaction(Generic[ClientT]):
         The guild ID the interaction was sent from.
     channel_id: Optional[:class:`int`]
         The channel ID the interaction was sent from.
+
+        .. deprecated:: 2.3
+            Use :attr:`channel` instead.
```
Rapptz, line 111:
```suggestion
```

Just because Discord deprecates something doesn't mean it's deprecated for us.

discord/types/channel.py
```diff
@@ -153,13 +153,14 @@ class ForumChannel(_BaseTextChannel):
 class DMChannel(_BaseChannel):
     type: Literal[1]
     last_message_id: Optional[Snowflake]
-    recipients: List[PartialUser]
+    recipients: NotRequired[List[PartialUser]]
```
Rapptz, line 156: This and the change below it seem motivated purely by the interaction channel type changes, instead of changing the overall type to fit this it might be better to tentatively create a new DM channel type for interaction DM channel.

discord/interactions.py
```diff
@@ -213,6 +225,30 @@ def _from_data(self, data: InteractionPayload):
...
+            if ch_type in (ChannelType.group, ChannelType.private):
+                channel = factory(me=self._client.user, data=self._channel, state=self._state)  # type: ignore
+            else:
+                guild = self._state._get_guild(self.guild_id)
+                channel = factory(guild=guild, state=self._state, data=self._channel)  # type: ignore
+
+            return channel
+
+        return None
+
```
Rapptz, line 251: None of this code belongs inside a `@property`. Removing the `_cs_channel` cached property was also a bizarre decision considering it would have made this more efficient on secondary access.

discord/interactions.py
```diff
@@ -213,6 +225,30 @@ def _from_data(self, data: InteractionPayload):
             except KeyError:
                 pass
 
+    @property
+    def channel(self) -> Optional[InteractionChannel]:
+        """Optional[Union[:class:`abc.GuildChannel`, :class:`abc.PrivateChannel`, :class:`Thread`]]: The channel
+        the interaction was sent from.
+
+        .. versionadded:: 2.3
```
Rapptz, line 233:
```suggestion
```

This wasn't added in 2.3.

discord/interactions.py
```diff
@@ -172,6 +183,7 @@ def _from_data(self, data: InteractionPayload):
         self.token: str = data['token']
         self.version: int = data['version']
         self.channel_id: Optional[int] = utils._get_as_snowflake(data, 'channel_id')
+        self._channel = data.get('channel', {})
```
Rapptz, line 186: Storing a random dictionary in an attribute is very rarely a good idea.

Rapptz (changes requested):
Thanks for the PR but I think this implementation needs a lot of work.

discord/interactions.py
Puncher1: > None of this code belongs inside a `@property`.

What about the `InvalidData` error I mentioned above?:

> Yeah, maybe better. Then, if `InvalidData` is raised, it does only effect the channel and not the whole object.

Yeah, I agree with the cached property.

discord/interactions.py
```diff
@@ -151,6 +156,7 @@ class Interaction(Generic[ClientT]):
         '_cs_channel',
         '_cs_namespace',
         '_cs_command',
+        '_channel',
```
Rapptz, line 159:
```suggestion
```

discord/interactions.py
```diff
@@ -229,19 +251,17 @@ def guild(self) -> Optional[Guild]:
...
         """
-        guild = self.guild
-        channel = guild and guild._resolve_channel(self.channel_id)
-        if channel is None:
-            if self.channel_id is not None:
-                type = ChannelType.text if self.guild_id is not None else ChannelType.private
-                return PartialMessageable(state=self._state, guild_id=self.guild_id, id=self.channel_id, type=type)
-            return None
-        return channel
+        return self._channel
```
Rapptz, line 259:
```suggestion
        data = self.data.get('channel', {})
        raw_type = data.get('type')

        # Shouldn't be possible
        if raw_type is None:
            return None
    
        factory, channel_type = _threaded_channel_factory(raw_type)
        if factory is None:
            raise InvalidData('Unknown channel type {type} for channel ID {id}.'.format_map(data))

        if channel_type in (ChannelType.group, ChannelType.private):
            return factory(me=self._client.user, data=data, state=self._state)  # type: ignore
        else:
            guild = self._state._get_or_create_unavailable_guild(self.guild_id)
            return factory(guild=guild, state=self._state, data=data)  # type: ignore

        return None
```

Rapptz (changes requested):

Puncher1:
Had to add the check if `raw_ch_type` is not `None` again to bypass pytest. I think `channel` has to be added to the mock data in the tests.

Rapptz (approved):
Thanks. This looks alright.

PR #9349, opened 2023-04-09
Support for Soundboard and VC effects

Puncher1:
###### Last updated: Sep 21, 2024

* Bots can receive voice channel effects over gateway events
* Bots can receive sounds over gateway events and over API requests
* Bots can create, edit and delete sounds
* Bots can send sounds

### This PR includes
* **Support for voice channel effects**
  * `VoiceChannelEffect`, `VoiceChannelSoundEffect`, `VoiceChannelEffectAnimation` models
  * `VoiceChannelEffectAnimationType` enum
  * `on_voice_channel_effect` event
* **Support for soundboard sounds**
  * `SoundboardSound`, `SoundboardDefaultSound` models
  * `create_`, `edit_`, `delete_` sound
  * `fetch_` methods (default sounds, normal sounds, specific sound)
  * ~Request sounds over the gateway (*may be superfluous since `fetch_` methods exist*)~ 
  * `VoiceChannel.send_sound` method
  * `soundboard_sounds` property and `get_` methods (both for `Guild` and `Client`)
  * `on_soundboard_sound_` events (create, update, delete) + `GUILD_SOUNDBOARD_SOUNDS_UPDATE` event (dispatches `update` event for every sound in the list)
* Related Audit Log events
* `Intents.expressions` -> `emojis` and `emojis_and_stickers` are now both aliases

### Current Problems [solved]

All problems on Discord's side, which were stated in the related [soundboard PR](https://github.com/discord/discord-api-docs/pull/6260) are **fixed** ✅ 
>* There is no GET route for individual sounds or a guild's list of sounds. Guild sound lists can currently only be retrieved via the Gateway. GET routes should be added * 
>* PATCHing a sound currently clears the sound's volume and emoji if those aren't included in the request. The PR documents this current behavior, let me know once it is fixed **
>* Looks like id and sound_id are always the same, but id is optional ***

\* Now available
\*\* The PR doesn't document this behaviour anymore
\*\*\* `id` has been removed

### Others

Default sounds are identified with a `sound_id`, `override_path` has been removed

Related PRs: https://github.com/discord/discord-api-docs/pull/6025 (voice channel effects), https://github.com/discord/discord-api-docs/pull/6260 (soundboard)

<br/>

## Checklist

**Testing pending**

- [ ] This PR fixes an issue.


SvenLie:
When can we expect this to be undrafted and merged? @Puncher1

Puncher1:
@SvenLie The GET method is still not supported which, at least for me, doesn't make sense to merge it yet.

Puncher1:
### Update
* For the current status, please refer to the PR description
* The 3.x checks are failing. Looks like there's an error in the installation, @Rapptz 
* Testing is pending, after that this PR is ready for review

Rapptz:
3.x is failing due to aiohttp not building on 3.12 and 3.x is latest Python, so CI is blocked on dependencies and not much I can do there.

Puncher1:
This has been tested and is ready for review now.

PaulMarisOUMary:
I believe the issues with 3.x have been resolved now. 🎉

discord/channel.py
```diff
@@ -111,6 +125,126 @@ class ThreadWithMessage(NamedTuple):
...
+    emoji: Optional[:class:`PartialEmoji`]
+        The emoji of the effect.
+    sound: Optional[:class:`VoiceChannelSoundEffect`]
+        The sound of the effect. Returns ``None`` if it's an emoji effect.
+    """
+
+    __slots__ = ('channel', 'user', 'animation', 'emoji', 'sound')
+
+    def __init__(self, *, state: ConnectionState, data: VoiceChannelEffectPayload, guild: Guild):
+        self.channel: VoiceChannel = guild.get_channel(int(data['channel_id']))  # type: ignore # will always be a VoiceChannel
```
Rapptz, line 214: This assertion seems rather strong at the moment since it might change so that soundboards play in stage channels though that feels unlikely. This has no bearing on the PR however.

discord/channel.py
```diff
@@ -111,6 +125,126 @@ class ThreadWithMessage(NamedTuple):
...
+        self.user: Member = guild.get_member(int(data['user_id']))  # type: ignore # will always be a Member
```
Rapptz, line 215: This assertion is definitely wrong due to the member intent and cache requirements.

discord/channel.py
```diff
@@ -111,6 +125,126 @@ class ThreadWithMessage(NamedTuple):
...
+            animation_type = try_enum(VoiceChannelEffectAnimationType, data['animation_type'])  # type: ignore # cannot be None here
+            self.animation = VoiceChannelEffectAnimation(id=animation_id, type=animation_type)
+
+        emoji = data.get('emoji')
+        self.emoji: Optional[PartialEmoji] = PartialEmoji.from_dict(emoji) if emoji is not None else None
+        self.sound: Optional[VoiceChannelSoundEffect] = None
+
+        sound_id: Optional[int] = utils._get_as_snowflake(data, 'sound_id')
+        if sound_id is not None:
+            sound_volume = data['sound_volume']  # type: ignore # sound_volume cannot be None here
```
Rapptz, line 229:
```suggestion
            sound_volume = data['sound_volume'] or 0.0
```

Sometimes it's simpler to just coerce data in case Discord starts sending garbage.

discord/utils.py
```diff
@@ -609,7 +609,7 @@ def _get_as_snowflake(data: Any, key: str) -> Optional[int]:
         return value and int(value)
 
 
-def _get_mime_type_for_image(data: bytes):
+def _get_mime_type_for_file(data: bytes):
```
Rapptz, line 612: Please make this a new helper instead.

This change makes it so people can pass e.g. audio files for avatars and have it silently work when it would have failed prior.

discord/channel.py
```diff
@@ -111,6 +125,126 @@ class ThreadWithMessage(NamedTuple):
...
+        }
+        super().__init__(state=state, data=data)
+
+    def __repr__(self) -> str:
+        attrs = [
+            ('id', self.id),
+            ('volume', self.volume),
+        ]
+        inner = ' '.join('%s=%r' % t for t in attrs)
+        return f"<{self.__class__.__name__} {inner}>"
```
Rapptz, line 175: For a simple one it's better to do it in a simpler manner.

```suggestion
        return f"<{self.__class__.__name__} id={self.id} volume={self.volume}>"
```

discord/soundboard.py
```diff
@@ -0,0 +1,324 @@
...
+    def __ne__(self, other: object) -> bool:
+        return not self.__eq__(other)
+
+    def _update(self, data: BaseSoundboardSoundPayload):
+        self.volume: float = data['volume']
+
+    @property
+    def file(self) -> Asset:
+        """:class:`Asset`: Returns the sound file asset."""
+        return Asset._from_soundboard_sound(self._state, self.id)
```
Rapptz, line 100: This is pretty odd, you might be better off inheriting from `_AssetMixin` similar to how `Emoji` classes do it.

discord/state.py
```diff
@@ -454,6 +513,14 @@ def emojis(self) -> Sequence[Emoji]:
     def stickers(self) -> Sequence[GuildSticker]:
         return utils.SequenceProxy(self._stickers.values())
 
+    @property
+    def soundboard_sounds(self) -> Sequence[SoundboardSound]:
+        all_sounds = []
+        for guild in self.guilds:
+            all_sounds.extend(guild.soundboard_sounds)
+
+        return utils.SequenceProxy(all_sounds)
```
Rapptz, line 522: `SequenceProxy` is there to facilitate copy-on-write. Since you're creating a list flat out there's no reason to use it.

discord/asset.py
```diff
@@ -326,6 +326,14 @@ def _from_user_banner(cls, state: _State, user_id: int, banner_hash: str) -> Sel
             animated=animated,
         )
 
+    @classmethod
+    def _from_soundboard_sound(cls, state: _State, sound_id: int) -> Self:
+        return cls(
+            state,
+            url=f'{cls.BASE}/soundboard-sounds/{sound_id}',
+            key=str(sound_id),
```
Rapptz, line 334: If the key of the asset is the ID then it should probably be an asset mixin. Asset keys are more like hashes, at least so far.

Rapptz (changes requested):
I've only done a cursory pass at this. It has merge conflicts currently as well, though this feature is pretty complicated and definitely needs documentation on Discord's end for me to merge it.

There are a few things I'm kind of iffy about, some more important than others:

1. The name `SoundboardSound` is pretty silly, I'm unsure if going `Soudboard` is any better but it doesn't read as oddly. 
2. The machinery to handle opcode 31 in the gateway seems very complicated for what it is. The chunking mechanism is the way it is because it's expected that multiple events flow in and it needs to be synced and waited. I have no reason to believe soundboard sounds operate the same way but I also have no idea if they do. I tried asking to get clarification on this though I haven't gotten a response yet. I'll update you when I do.

discord/utils.py
Puncher1: Okay, but then it needs two `_bytes_to_base64_...` functions to differ between image and audio files or a parameter for `_bytes_to_base64_data` to pass the type of data, e.g. `'image'` or `'audio'`. Otherwise I would have to call both functions in `_bytes_to_base64_data` and use try except to somehow get the mime type which doesn't solve the problem, e.g.:

```py
def _bytes_to_base64_data(data: bytes) -> str:
    # ...
    try:
        mime = _get_mime_type_for_image(data)
    except ValueError:
        try:
            mime = _get_mime_type_for_audio(data)
        except ValueError:
            raise ValueError('Unsupported file type given')
    # ...
```
Still allows passing audio files even image files are expected.

Rapptz: I don't think it needs all that complication. You have `_get_mime_type_for_image` and `_get_mime_type_for_audio` then you inside `_bytes_to_base64_data` you can just do `audio=True` or `audio=False` (default) and appropriately branch it. I don't expect an endpoint to allow both forms of media so the try except is just noisy.

Puncher1: Yeah exactly, I meant something like this. Of course the try except wouldn't solve the problem as I mentioned. 👍

Puncher1:
Alright I'm done with the changes. Because of the name, how about just `Sound`? And should it be also applied for properties and attributes (e.g. `Client.soundboard_sounds` -> `Client.sounds`)?

DA-344:
This will need changes because of the latests commits on the DDevs Docs PR: https://github.com/discord/discord-api-docs/pull/6260/commits/ec8656239280b89c512e92921b1e43717fcaa6c6

Puncher1:
> ...
> 2. The machinery to handle opcode 31 in the gateway seems very complicated for what it is. The chunking mechanism is the way it is because it's expected that multiple events flow in and it needs to be synced and waited. I have no reason to believe soundboard sounds operate the same way but I also have no idea if they do. I tried asking to get clarification on this though I haven't gotten a response yet. I'll update you when I do.

@Rapptz You can now get soundboard sounds via `GET /guilds/{guild_id}/soundboard-sounds`, opcode 31 is now quite superfluous.

Rapptz:
> You can now get soundboard sounds via GET /guilds/{guild_id}/soundboard-sounds, opcode 31 is now quite superfluous.

If it's superfluous outside of multiple guilds -> sound then it should probably be removed. There isn't an immediate benefit to the complicated machinery being used for soundboards.

Puncher1:
Messed up a bit (almost lost all changes lol), so I had to force push.

Puncher1:
The soundboard PR on Discord's side is merged now as well.

Rapptz (approved):

PR #9350, opened 2023-04-09
Change _get_as_snowflake to handle empty strings

Puncher1:
## Summary

This PR changes the behavior of `utils._get_as_snowflake` to handle empty strings.

Normally, the value of a snowflake field shouldn't be an empty string, but due of a Discord bug this can still be the case which the following issue shows: https://github.com/discord/discord-api-docs/issues/6069

This change should prevent such a case which results in such an error:
```
  File "c:\users\andri\documents\github\discord.py-1\discord\ext\commands\core.py", line 229, in wrapped
  File "C:\Users\andri\Documents\GitHub\Punchax\cogs\!api.py", line 90, in audit_log
    async for entry in ctx.guild.audit_logs(limit=1):
  File "c:\users\andri\documents\github\discord.py-1\discord\guild.py", line 3896, in audit_logs
    yield AuditLogEntry(
  File "c:\users\andri\documents\github\discord.py-1\discord\audit_logs.py", line 592, in __init__
    self._from_data(data)
  File "c:\users\andri\documents\github\discord.py-1\discord\audit_logs.py", line 651, in _from_data
    channel = self.guild.get_channel_or_thread(channel_id) or Object(id=channel_id)
  File "c:\users\andri\documents\github\discord.py-1\discord\object.py", line 97, in __init__
    raise TypeError(f'id parameter must be convertible to int not {id.__class__.__name__}') from None
TypeError: id parameter must be convertible to int not str

The above exception was the direct cause of the following exception:

  File "c:\users\andri\documents\github\discord.py-1\discord\ext\commands\bot.py", line 1350, in invoke
    await ctx.command.invoke(ctx)
  File "c:\users\andri\documents\github\discord.py-1\discord\ext\commands\core.py", line 1023, in invoke
    await injected(*ctx.args, **ctx.kwargs)  # type: ignore
  File "c:\users\andri\documents\github\discord.py-1\discord\ext\commands\core.py", line 238, in wrapped
    raise CommandInvokeError(exc) from exc
discord.ext.commands.errors.CommandInvokeError: Command raised an exception: TypeError: id parameter must be convertible to int not str
```

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -605,6 +605,8 @@ def _get_as_snowflake(data: Any, key: str) -> Optional[int]:
     except KeyError:
         return None
     else:
+        if value == '':
+            return None
         return value and int(value)
 
 


discord/utils.py
```diff
@@ -605,6 +605,8 @@ def _get_as_snowflake(data: Any, key: str) -> Optional[int]:
     except KeyError:
         return None
     else:
+        if value == '':
+            return None
         return value and int(value)
```
mikeshardmind, line 610: There's no reason to do a separate check here, just modify the already existing one that covered None when discord actually sent what they were supposed to 🙄 
```suggestion
        return int(value) if value else None
```

mikeshardmind: this could also alternatively be fixed by changing line 604 to 

```py
        value = data[key] or None
```
but I think this would be worse for the normal case (haven't checked)

Rapptz:
Discord should fix this. This code is a hot path and changes are not accepted trivially.

PR #9361, opened 2023-04-15
Support for improved markdown

Puncher1:
## Summary

Just a small change to add support for headers `#` and unordered lists `-` in the markdown regex.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -900,7 +900,7 @@ _MARKDOWN_ESCAPE_REGEX = re.compile(fr'(?P<markdown>{_MARKDOWN_ESCAPE_SUBREGEX}|
 
 _URL_REGEX = r'(?P<url><[^: >]+:\/[^ >]+>|(?:https?|steam):\/\/[^\s<]+[^<.,:;\"\'\]\s])'
 
-_MARKDOWN_STOCK_REGEX = fr'(?P<markdown>[_\\~|\*`]|{_MARKDOWN_ESCAPE_COMMON})'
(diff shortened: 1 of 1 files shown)


Rapptz (commented):
This technically works, but it's not particularly correct since `#` and `-` (and, by extension numbers such as `1.`, `2.`, etc.) are required to be in the beginning of the line (or optionally with spaces) which the regex doesn't deal with. 

I'll just deal with it in a future commit.

Puncher1:
> This technically works, but it's not particularly correct since `#` and `-` (and, by extension numbers such as `1.`, `2.`, etc.) are required to be in the beginning of the line (or optionally with spaces) which the regex doesn't deal with.

Yeah, but at the moment things like bold texts `**Example**` are also not handled correctly because this would also be escaped: `Example *` (see image below), so I just adapted to this for my changes.

image

 Accordingly I thought `1.`, `2.` etc. would be excessive because this would be a mess if you use it in the middle of a text, which is quite often the case.

Rapptz:
This is what the `as_needed` keyword argument is for.

Puncher1:
> This is what the `as_needed` keyword argument is for. It's just hard to consistently do this without accidentally breaking something. For `#` and `-` they're actually useless unless they're in the beginning so it's not the same.
> 
> For example, consider the very common text you'll encounter such as `Danny#0007`, this would now be escaped when it wouldn't be necessary.

Yeah, fair.

PR #9362, opened 2023-04-15
[docs] Add missing attributes to AuditLogDiff

Puncher1:
## Summary

This PR adds missing attributes for `AuditLogDiff`.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -4016,11 +4016,11 @@ AuditLogDiff
 
         See also :attr:`Thread.applied_tags`
 
-        :type: List[:class:`ForumTag`]
+        :type: List[Union[:class:`ForumTag`, :class:`Object`]]
 
     .. attribute:: available_tags
 
-        The available flags of a forum.
+        The available tags of a forum.
 
         See also :attr:`ForumChannel.available_tags`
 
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -3422,6 +3422,12 @@ AuditLogDiff
 
         :type: :class:`str`
 
+    .. attribute:: guild
+
+        The guild of something.
+
+        :type: :class:`Guild`
+
```
Rapptz, line 3430: I can't imagine a scenario where this comes up right now, but this string leaves much to be desired.

Rapptz (changes requested):

docs/api.rst
Puncher1: It's in the `TRANSFORMERS` in `AuditLogChanges`:
https://github.com/Rapptz/discord.py/blob/cb6170a7247958294b85a3bd0ab29b71088e86a2/discord/audit_logs.py#L353

But I don't know right now where that would be needed either. So I simply took "[...] of something", as it is also the case with `name`. Should I remove it from the docs accordingly?

Rapptz: I guess it's fine, though I'll think of something for the replacement string.

Rapptz (approved):

PR #9371, opened 2023-04-21
add logging to examples

ImVaskel:
## Summary

With the changes to logging (i.e having the library create a logger for you by default), I thought it would be a good idea to migrate all the examples (and newbot) to use logging rather than printing to be more in-line with the rest of the library.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/app_commands/transformers.py
+++ b/examples/app_commands/transformers.py
@@ -28,7 +28,7 @@ client = MyClient()
 @client.event
 async def on_ready():
     logging.info(f'Logged in as {client.user} (ID: {client.user.id})')
-    logging.infot('------')
+    logging.info('------')
 
 
 # A transformer is a class that specifies how a parameter in your code
--- a/examples/app_commands/transformers.py
+++ b/examples/app_commands/transformers.py
@@ -1,6 +1,7 @@
 # This example builds on the concepts of the app_commands/basic.py example
 # It's suggested to look at that one to understand certain concepts first.
 
+import logging
 from typing import Literal, Union, NamedTuple
 from enum import Enum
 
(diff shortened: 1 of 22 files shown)


examples/app_commands/transformers.py
```diff
@@ -26,8 +27,8 @@ async def setup_hook(self):
 
 @client.event
 async def on_ready():
-    print(f'Logged in as {client.user} (ID: {client.user.id})')
-    print('------')
+    logging.info(f'Logged in as {client.user} (ID: {client.user.id})')
+    logging.infot('------')
```
Rapptz, line 31:
```suggestion
    logging.info('------')
```

examples/modals/basic.py
```diff
@@ -61,7 +60,7 @@ async def on_error(self, interaction: discord.Interaction, error: Exception) ->
         await interaction.response.send_message('Oops! Something went wrong.', ephemeral=True)
 
         # Make sure we know what the error actually is
-        traceback.print_exception(type(error), error, error.__traceback__)
+        logging.error('', exc_info=(type(error), error, error.__traceback__))
```
Rapptz, line 63: This is pretty ugly and potentially confusing. Likewise, just an exception instance works here.
```suggestion
        logging.error('', exc_info=error)
```

Rapptz (changes requested):
I'm not really sure on the merits of this PR. I understand that `logging` is used internally within the library and setting it up is a good idea, however when it comes to examples and actual simple applications I think their complication is not worth showcasing for a vast majority of people. 

My disposition for this PR is to not merge it as a result but I guess I can sit on it for a bit.

PR #9375, opened 2023-04-23
Support for interactions-only bots

Auto-5:
## Summary
This creates the `App` class to run a webserver so you can receive interactions by the [other of the two ways](https://discord.com/developers/docs/interactions/receiving-and-responding#receiving-an-interaction), via HTTP `INTERACTIONS_ENDPOINT_URL` instead of gateway (websocket). It uses `discord.py` models like `discord.Interaction` so responding to interactions is the same, just receiving is different.

The `FastAPI` dependency is optional from `discord[web]`. The file `examples/app_commands/app.py` is a working example. Only `Interaction.response.send_message` has been tested.

Some bots have no reason to be stateful, so it saves resources for heartbeats, sharding and stateful things.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/__init__.py
+++ b/discord/app_commands/__init__.py
@@ -18,3 +18,4 @@ from .transformers import *
 from .translator import *
 from . import checks as checks
 from .checks import Cooldown as Cooldown
+from .app import *
--- a/discord/app_commands/app.py
+++ b/discord/app_commands/app.py
@@ -1,0 +1,74 @@
+import aiohttp
+from fastapi import FastAPI
+from fastapi.requests import Request
+from fastapi.responses import Response, JSONResponse, PlainTextResponse
+from nacl.signing import VerifyKey
+from nacl.exceptions import BadSignatureError
+from .. import utils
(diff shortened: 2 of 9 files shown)


Rapptz:
Thanks for the PR.

I'm not inclined to merge this as it is currently because it's not exactly how we're planning on implementing this feature. For example, it's pretty opinionated to require a specific web framework (e.g. FastAPI) when there are other web frameworks that people would want to use. The appropriate layer to implement this to be agnostic in discord.py would be ASGI middleware that plays nicely with every web framework that implements ASGI but this does not do this. We were discussing this method in the [discord.py server](https://canary.discord.com/channels/336642139381301249/1013275205872918618/1013275205872918618).

Secondly, it's not a good idea to remove the ConnectionState handling as is done here since it causes some issues. For example, removing *all* the resolved items as it was done here just makes most of the discord.py useless for slash commands.

If you want to help contribute this, please join the discord.py server and discuss it in the thread linked.

PR #9188, opened 2023-01-18
Fix Piped Audio Input Ends Prematurely

thedtvn:
## Summary

check out [#9001](https://github.com/Rapptz/discord.py/issues/9001)

## Checklist

- [x] If code changes were made then they have been tested.
    - [ ] I have updated the documentation to reflect the changes.
- [x] This PR fixes an issue.
- [ ] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

PR #9380, opened 2023-04-30
Fix Piped Audio Input Ends Prematurely

### LINKED ISSUE #9001: Piped Audio Input Ends Prematurely
### Summary

Piped video/audio input from sources like `youtube-dl` does not terminate the pipe stream correctly, often cutting off the last bit of the stream.

### Reproduction Steps

- Stream audio from another process using `subprocess.Popen`; in my case I was using `youtube-dl`
- Wait for the audio to play until it nears the end of the stream

### Minimal Reproducible Code

```python
client = await ctx.author.voice.channel.connect()
url = "https://www.youtube.com/watch?v=KAwyWkksXuo"
ytdl = subprocess.Popen(["youtube-dl", "-f", "bestaudio/worst", "-i", url, "-o", "-"], stdout=subprocess.PIPE)
audsrc = discord.FFmpegPCMAudio(ytdl.stdout, pipe=True)
client.play(audsrc)
```

### Expected Results

Discord.py plays the stream until the very end, then closes FFMPEG and stops playback.

### Actual Results

The stream is cut off slightly before the track actually finishes. For the sample video (cbat), it terminates roughly 6 seconds before the stream actually finishes. In addition, FFMPEG terminates with code 255, indicating a forced termination of the program.

### Intents

members, message_content, messages

### System Information

- Python v3.10.7-final
- discord.py v2.0.1-final
- aiohttp v3.8.3
- system info: Darwin 21.6.0 Darwin Kernel Version 21.6.0: Mon Aug 22 20:17:10 PDT 2022; root:xnu-8020.140.49~2/RELEASE_X86_64

### Checklist

### Additional Context

My current solution for this involves a modification of the `FFmpegAudio` class in `player.py`.
```python
class FFmpegAudio(AudioSource):
    # ...
    def _pipe_writer(self, source: io.BufferedIOBase) -> None:
        while self._process:
            # arbitrarily large read size
            data = source.read(8192)
            if not data:
                # self._process.terminate() <--- Removed this line, replaced with following
                self._stdin.close()
                return
            try:
                if self._stdin is not None:
                    self._stdin.write(data)
            except Exception:
                _log.debug('Write error for %s, this is probably not a problem', self, exc_info=True)
                # at this point the source data is either exhausted or the process is fubar
                self._process.terminate()
                return
```

kairi003:
## Summary

This commit fix #9001 by ~~replacing `self._process.terminate()` with `self._stdin.close()` to fix the issue where the process was being interrupted after reading input.
I also add an `assert self._stdin is not None` statement to resolve the type of `self._stdin` and remove the redundant if statement.~~
adding `self._stdin.close()` ~~and `self._process.wait()`~~ only.

This PR is a duplicate of the PR #9188 closed for no reason.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/player.py
+++ b/discord/player.py
@@ -206,15 +206,18 @@ class FFmpegAudio(AudioSource):
             _log.info('ffmpeg process %s successfully terminated with return code of %s.', proc.pid, proc.returncode)
 
     def _pipe_writer(self, source: io.BufferedIOBase) -> None:
-        assert self._stdin is not None
         while self._process:
             # arbitrarily large read size
             data = source.read(8192)
             if not data:
-                self._stdin.close()
+                if self._stdin is not None:
+                    self._stdin.close()
+                self._process.wait()
+                self._process.terminate()
                 return
             try:
-                self._stdin.write(data)
+                if self._stdin is not None:
+                    self._stdin.write(data)
             except Exception:
                 _log.debug('Write error for %s, this is probably not a problem', self, exc_info=True)
                 # at this point the source data is either exhausted or the process is fubar
--- a/discord/player.py
+++ b/discord/player.py
@@ -206,15 +206,15 @@ class FFmpegAudio(AudioSource):
             _log.info('ffmpeg process %s successfully terminated with return code of %s.', proc.pid, proc.returncode)
 
     def _pipe_writer(self, source: io.BufferedIOBase) -> None:
+        assert self._stdin is not None
         while self._process:
             # arbitrarily large read size
             data = source.read(8192)
             if not data:
-                self._process.terminate()
+                self._stdin.close()
                 return
             try:
-                if self._stdin is not None:
-                    self._stdin.write(data)
+                self._stdin.write(data)
             except Exception:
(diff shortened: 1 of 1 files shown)


discord/player.py
```diff
@@ -210,7 +210,12 @@ def _pipe_writer(self, source: io.BufferedIOBase) -> None:
             # arbitrarily large read size
             data = source.read(8192)
             if not data:
-                self._process.terminate()
+                if self._stdin is not None:
+                    self._stdin.close()
+                try:
+                    self._process.wait(timeout=60.0)
+                except subprocess.TimeoutExpired:
+                    self._process.terminate()
```
Rapptz, line 218:
```suggestion
```

Let's just do this instead since the process is cleaned outside of this function.

Rapptz (changes requested):

discord/player.py
kairi003: Of course, that sounds good.
And sorry, I left a comment to your suggestion in the commit message by mistake.

Rapptz (approved):

PR #9406, opened 2023-05-14
Support app command contexts

RedKinda:
## Summary

Adds `@dm_only` and `@private_only` decorators to allow specifying the new `contexts` field in app commands, as specified [here](https://gist.github.com/devsnek/a47893004bc0ee577eb8e9fb76b4d57f), as well as the relevant init args.

The docs mention that if this field is specified, the permission calculation ignores the `dm_permission` field. Not sure how this should be reflected in the library for backwards compatibility purposes.

## Checklist

- [ ] If code changes were made then they have been tested. (needs a bit more testing for groups and context commands)
- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -623,7 +623,7 @@ class Command(Generic[GroupT, P, T]):
 
         Due to a Discord limitation, this does not work on subcommands.
     allowed_contexts: Optional[:class:`~discord.flags.AppCommandContext`]
-        A list of contexts that the command is allowed to be used in.
+        The contexts that the command is allowed to be used in.
         Overrides ``guild_only`` if this is set.
     nsfw: :class:`bool`
         Whether the command is NSFW and should only work in NSFW channels.
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -622,7 +622,7 @@ class Command(Generic[GroupT, P, T]):
         Whether the command should only be usable in guild contexts.
 
         Due to a Discord limitation, this does not work on subcommands.
-    allowed_contexts: Optional[:class:`~discord.AppCommandContext`]
+    allowed_contexts: Optional[:class:`~discord.flags.AppCommandContext`]
         A list of contexts that the command is allowed to be used in.
         Overrides ``guild_only`` if this is set.
     nsfw: :class:`bool`
@@ -1179,7 +1179,7 @@ class ContextMenu:
     guild_only: :class:`bool`
         Whether the command should only be usable in guild contexts.
         Defaults to ``False``.
-    allowed_contexts: Optional[:class:`.AppCommandContext`]
+    allowed_contexts: Optional[:class:`~discord.flags.AppCommandContext`]
         The contexts that this context menu is allowed to be used in.
         Overrides ``guild_only`` if set.
     nsfw: :class:`bool`
@@ -1425,7 +1425,7 @@ class Group:
         Whether the group should only be usable in guild contexts.
 
         Due to a Discord limitation, this does not work on subcommands.
-    allowed_contexts: Optional[:class:`.AppCommandContext`]
+    allowed_contexts: Optional[:class:`~discord.flags.AppCommandContext`]
         The contexts that this group is allowed to be used in. Overrides
         guild_only if set.
     nsfw: :class:`bool`
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -48,9 +48,9 @@ from typing import (
 import re
 from copy import copy as shallow_copy
 
-from discord.flags import AppCommandContext
 
 from ..enums import AppCommandOptionType, AppCommandType, ChannelType, Locale
+from ..flags import AppCommandContext
 from .models import Choice
 from .transformers import annotation_to_parameter, CommandParameter, NoneType
 from .errors import AppCommandError, CheckFailure, CommandInvokeError, CommandSignatureMismatch, CommandAlreadyRegistered
@@ -622,7 +622,7 @@ class Command(Generic[GroupT, P, T]):
         Whether the command should only be usable in guild contexts.
 
         Due to a Discord limitation, this does not work on subcommands.
-    restrict_contexts: Optional[List[:class:`~discord.AppCommandContext`]]
+    restrict_contexts: Optional[:class:`~discord.AppCommandContext`]
         A list of contexts that the command is allowed to be used in.
         Overrides ``guild_only`` if this is set.
     nsfw: :class:`bool`
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -622,7 +622,7 @@ class Command(Generic[GroupT, P, T]):
         Due to a Discord limitation, this does not work on subcommands.
     allowed_contexts: Optional[List[:class:`~discord.AppCommandContext`]]
         A list of contexts that the command is allowed to be used in.
-        Overrides guild_only if tihs is set.
+        Overrides ``guild_only`` if this is set.
     nsfw: :class:`bool`
         Whether the command is NSFW and should only work in NSFW channels.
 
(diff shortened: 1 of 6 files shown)


discord/app_commands/commands.py
```diff
@@ -618,6 +620,9 @@ class Command(Generic[GroupT, P, T]):
         Whether the command should only be usable in guild contexts.
 
         Due to a Discord limitation, this does not work on subcommands.
+    allowed_contexts: Optional[List[:class:`~discord.AppCommandContext`]]
+        A list of contexts that the command is allowed to be used in.
+        Overrides guild_only if tihs is set.
```
izxxr, line 625:
```suggestion
        Overrides guild_only if this is set.
```

discord/app_commands/commands.py
```diff
@@ -1167,6 +1177,9 @@ class ContextMenu:
     guild_only: :class:`bool`
         Whether the command should only be usable in guild contexts.
         Defaults to ``False``.
+    allowed_contexts: Optional[List[:class:`.AppCommandContext`]]
+        The contexts that this context menu is allowed to be used in.
+        Overrides `guild_only` if set.
```
izxxr, line 1182:
```suggestion
        Overrides ``guild_only`` if set.
```

discord/app_commands/commands.py
```diff
@@ -2418,8 +2450,102 @@ async def my_guild_only_command(interaction: discord.Interaction) -> None:
...
+            allowed_contexts = getattr(f, '__discord_app_commands_contexts__', None) or []
+            f.__discord_app_commands_contexts__ = allowed_contexts  # type: ignore # Runtime attribute assignment
+
+        if AppCommandContext.GUILD not in allowed_contexts:
+            allowed_contexts.append(AppCommandContext.GUILD)
+
+        return f
+
+    # Check if called with parentheses or not
+    if func is None:
```
izxxr, line 2467: I don't know if this is really needed because in other decorators that don't need extra parameters, library doesn't allow them to be used with parentheses (for example `@client.event`) so I think this would be inconsistent. The same goes for other two decorators.

discord/app_commands/models.py
```diff
@@ -160,6 +168,8 @@ class AppCommand(Hashable):
         The default member permissions that can run this command.
     dm_permission: :class:`bool`
         A boolean that indicates whether this command can be run in direct messages.
+    allowed_contexts: Optional[List[:class:`AppCommandContext`]]
+        A list of contexts that this command can be run in. Overrides the `dm_permission` attribute.
```
izxxr, line 172:
```suggestion
        A list of contexts that this command can be run in. Overrides the ``dm_permission`` attribute.
```

discord/enums.py
```diff
@@ -728,6 +729,12 @@ class AppCommandPermissionType(Enum):
     channel = 3
 
 
+class AppCommandContext(Enum):
+    GUILD = 0
+    BOT_DM = 1  # this is DMs only
+    PRIVATE_CHANNEL = 2  # this is DMs + GDMs
```
izxxr, line 735: These should be lowercase as with other enums. Also, this class needs to be documented in api.rst manually.

```suggestion
    guild = 0
    bot_dm = 1  # this is DMs only
    private_channel = 2  # this is DMs + GDMs
```

izxxr (changes requested):

discord/app_commands/commands.py
RedKinda: `@guild_only` has had this check before so I added them to the new decorators as well. Happy to remove it

izxxr: Right, my bad sorry 👍

Rapptz (changes requested):
Thanks for the PR.

Just some minor bikeshedding/comments for the first pass:

1. `AppCommandContext` is *not* an enum, it's actually a flag type. Discord has recently been doing bitfields using an array because they falsely believe it leads to better developer experience since they are targeting developers that do not know how to bitshift. I suggest migrating this over to use the new `ArrayFlags` type and work with it. I have tried to get them to change this to a regular integer but they seem to want to not want to change it to the more space-efficient and sane format.
2. `allowed_contexts` and any external reference to `context` in this well *context* should probably be called `restrict` instead. The type should still be called `AppCommandContext` but the attribute (and maybe eventual decorator) should be called `restrict`.
3. `private_only` should probably be renamed to `private_channel_only` for consistency with the library (e.g. `abc.PrivateChannel`).
4. The flag names for `AppCommandContext` should not use the Discord naming scheme since they are bad, instead it should probably be `guild`, `dm_channel`, and `private_channel`.

discord/app_commands/commands.py
```diff
@@ -48,6 +48,8 @@
 import re
 from copy import copy as shallow_copy
 
+from discord.flags import AppCommandContext
```
Rapptz, line 51:
```suggestion
from .flags import AppCommandContext
```

and move it down to where the others are

discord/app_commands/commands.py
```diff
@@ -618,6 +622,9 @@ class Command(Generic[GroupT, P, T]):
         Whether the command should only be usable in guild contexts.
 
         Due to a Discord limitation, this does not work on subcommands.
+    restrict_contexts: Optional[List[:class:`~discord.AppCommandContext`]]
```
Rapptz, line 625: Originally I meant that the attribute should just be `restrict` or `restrictions` but I think now `restrict` would make more sense as the decorator that applies multiple if one would ever happen and the attribute name itself should just be reverted back to `allowed_contexts`.

Sorry for the churn. Bikeshedding tends to be hard :)

```suggestion
    allowed_contexts: Optional[:class:`~discord.AppCommandContext`]
```

Note that since this is a flag now the `List[...]` is no longer applicable.

discord/flags.py
```diff
@@ -1716,21 +1735,86 @@ def slurs(self):
...
 
-    @classmethod
-    def none(cls: Type[Self]) -> Self:
-        """A factory method that creates a :class:`AutoModPresets` with everything disabled."""
-        self = cls.__new__(cls)
-        self.value = self.DEFAULT_VALUE
-        return self
+@fill_with_flags()
+class AppCommandContext(ArrayFlags):
+    DEFAULT_VALUE = 3
```
Rapptz, line 1741: Please move this below the docstring

discord/flags.py
```diff
@@ -1716,21 +1735,86 @@ def slurs(self):
...
+    r"""Wraps up the Discord :class:`AppCommand` context.
+
+    .. versionadded:: 2.3
+
+
```
Rapptz, line 1746:
```suggestion

```

Rapptz (changes requested):
You're missing the documentation for the new decorators. They need to be added to the API reference. Other than that this PR looks good.

discord/app_commands/commands.py
Rapptz: Nah, those names are fine. I was originally thinking of a new decorator that does e.g. `@restrict(guilds=True, private_channel=False)` but due to the inconsistency in the way the thing was set up (no group DM only nonsense) this idea ended up being kind of mediocre so no point pursuing it.

discord/app_commands/commands.py
```diff
@@ -2430,6 +2473,135 @@ def inner(f: T) -> T:
         return inner(func)
 
 
+def private_channel_only(func: Optional[T] = None) -> Union[T, Callable[[T], T]]:
```
Gobot1234, line 2476: This should either drop the ability to optionally use parens or add the necessary overloads here otherwise you'll get type errors.

discord/app_commands/commands.py
```diff
@@ -2430,6 +2473,135 @@ def inner(f: T) -> T:
...
+
+    Due to a Discord limitation, this decorator does nothing in subcommands and is ignored.
+
+    Examples
+    ---------
+
+    .. code-block:: python3
+
+        @app_commands.command()
+        @app_commands.allow_contexts(guilds=False, dms=False, private_channels=True)
```
Zalk0, line 2577:
```suggestion
        @app_commands.allow_contexts(guilds=True, dms=False, private_channels=True)
```
Didn't you want to set guilds to True in the example as that's what you say in the response message just below

Rapptz:
Superseded by #9760

PR #9425, opened 2023-05-24
[docs] document exceptions being ignored in teardown and cog_unload

SebbyLaw:
## Summary

Documents the exception ignoring behavior of ``teardown`` and ``cog_unload``, to alleviate confusion EG https://github.com/Rapptz/discord.py/discussions/9421

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/ext/commands/extensions.rst
+++ b/docs/ext/commands/extensions.rst
@@ -54,6 +54,10 @@ Cleaning Up
 
 Although rare, sometimes an extension needs to clean-up or know when it's being unloaded. For cases like these, there is another entry point named ``teardown`` which is similar to ``setup`` except called when the extension is unloaded.
 
+.. warning::
+
+    Exceptions raised in the ``tearDown`` function are ignored, and the extension is still unloaded.
+
 .. code-block:: python3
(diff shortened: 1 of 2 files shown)


docs/ext/commands/extensions.rst
```diff
@@ -54,6 +54,10 @@ Cleaning Up
 
 Although rare, sometimes an extension needs to clean-up or know when it's being unloaded. For cases like these, there is another entry point named ``teardown`` which is similar to ``setup`` except called when the extension is unloaded.
 
+.. warning::
+
+    Exceptions raised in the ``tearDown`` function are ignored, and the extension is still unloaded.
```
bijij, line 59:
```suggestion
    Exceptions raised in the ``teardown`` function are ignored, and the extension is still unloaded.
```

Rapptz (changes requested):
I don't think this deserves an admonition and should just be regular text instead. Destructors in most languages are meant to be infallible, not particularly strange that the same is true here.

SebbyLaw:
Updated to not use admonitions

PR #9426, opened 2023-05-25
Make _Overwrites.is_member and is_role use _Ovewrites.ROLE and MEMBER

Sengolda:
## Summary
It just makes the checks use the ROLE and MEMBER vars

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -323,10 +323,10 @@ class _Overwrites:
         }
 
     def is_role(self) -> bool:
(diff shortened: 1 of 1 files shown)


Rapptz:
Thanks for the PR. I avoided the extraneous name lookup since it's clear from the function name what they are (not to mention the associated constants are right above). I know it looks like a silly micro-optimisation but the readability of the method isn't hampered by it.

PR #9427, opened 2023-05-25
[commands] Add `displayed_name` to commands.Parameter

Soheab:
## Summary

This PR adds a `(_)displayed_name` attribute and kwarg to `commands.((P)aram)eter` and uses it in places where relevant. This kwarg allows users to set a different name to parameters than defined in the function's signature.

This was discussed in a bikeshedding post called ["`name` kwarg for commands.(param)eter"](https://canary.discord.com/channels/336642139381301249/1110480296870158346) in the [discord.py server](https://discord.com/invite/dpy)

I'm not sure if this needs documenting anywhere else like the exceptions and `Command.signature`.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -151,6 +151,7 @@ def get_signature_parameters(
             parameter._default = default.default
             parameter._description = default._description
             parameter._displayed_default = default._displayed_default
+            parameter._displayed_name = default._displayed_name
 
         annotation = parameter.annotation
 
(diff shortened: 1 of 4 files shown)


Rapptz:
I'm not really sure how I feel about this feature as a whole. I get the point and the use case of it, at least a little. However I'm not sure if it's common enough to warrant it, not to mention this implementation seems to add a new `displayed_name` attribute instead of just using `name`. I assume this was done for a reason though currently I can't think of it.

Rapptz (approved):
After giving it some thought and hearing it out on Discord, this feature seems fine enough.

Thanks.

PR #9429, opened 2023-05-26
[tasks] add on_valid_exception hook

alkshmir:
## Summary
First of all, thank you very much for great library!

I want the bot to ignore exceptions due to a 3rd party library during the task, however, need to log the occurrence of the error for the observability as well.
Thus, I have implemented `on_valid_exception` to enable injecting additional process in the event of `valid_exception`.

Example Usage:
```python
from discord.ext import tasks

import discord
import logging

_log = logging.getLogger('discord')

class MyError(Exception):
    pass

    def __init__(self, *args, **kwargs):

        # an attribute we can access from our task
        self.counter = 0
        self.my_background_task.add_exception_type(MyError)

    async def setup_hook(self) -> None:
        # start the task to run in the background
        self.my_background_task.start()

    async def on_ready(self):
        print(f'Logged in as {self.user} (ID: {self.user.id})')
        print('------')

    @tasks.loop(seconds=3)  # task runs every 3 seconds
    async def my_background_task(self):
        channel = self.get_channel(12345)  # channel ID goes here
        self.counter += 1
        if 3 < self.counter < 7:
            raise MyError('test')
        await channel.send(self.counter)

    @my_background_task.on_valid_exception
    async def my_handler(self, *args):
        exc: Exception = args[-1]
        _log.warn('handled an exception on my_background_task()', exc_info=exc)

    @my_background_task.before_loop
    async def before_my_task(self):
        await self.wait_until_ready()  # wait until the bot logs in

client = MyClient(intents=discord.Intents.default())
client.run('token')
```

Thank you for your time and consideration.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/tasks/__init__.py
+++ b/discord/ext/tasks/__init__.py
@@ -162,6 +162,7 @@ class Loop(Generic[LF]):
 
         self._before_loop = None
         self._after_loop = None
+        self._on_valid_exception = None
         self._is_being_cancelled = False
         self._has_failed = False
         self._stop_next_iteration = False
@@ -238,10 +239,11 @@ class Loop(Generic[LF]):
                 try:
                     await self.coro(*args, **kwargs)
                     self._last_iteration_failed = False
-                except self._valid_exception:
+                except self._valid_exception as e:
                     self._last_iteration_failed = True
                     if not self.reconnect:
                         raise
+                    await self._call_loop_function('on_valid_exception', e)
                     await asyncio.sleep(backoff.delay())
                 else:
                     if self._stop_next_iteration:
@@ -286,6 +288,7 @@ class Loop(Generic[LF]):
         copy._injected = obj
         copy._before_loop = self._before_loop
         copy._after_loop = self._after_loop
+        copy._on_valid_exception = self._on_valid_exception
         copy._error = self._error
         setattr(obj, self.coro.__name__, copy)
         return copy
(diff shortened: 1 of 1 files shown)


SebbyLaw:
The valid exception list by default contains several exceptions (`OSError`, `discord.GatewayNotFound`, `discord.ConnectionClosed`, `aiohttp.ClientError`, `asyncio.TimeoutError`) it should probably be documented that these could trigger the event.

Is there any reason why this is preferential to wrapping your task code in a try: except?

alkshmir:
I was so dumb that I was assuming catching exceptions within task itself does not trigger retry and exponential backoff, however, by registering the exception type to `_valid_exception` and raising the same exception in the `except` phrase it is found to be able to trigger retry.

The following sample worked as exactly what I wanted to do.
```python
    def __init__(self, *args, **kwargs):

        # an attribute we can access from our task
        self.counter = 0
        self.my_background_task.add_exception_type(MyError)

    @tasks.loop(seconds=3)  # task runs every 3 seconds
    async def my_background_task(self):
        try:
            channel = self.get_channel(12345)  # channel ID goes here
            self.counter += 1
            if 3 < self.counter < 7:
                raise MyError('test')
            await channel.send(self.counter)
        except MyError as e:
            _log.warn("handled an exception on my_background_task()", exc_info=e)
            raise
```

PR #9432, opened 2023-05-28
[Context] New function: maybe_reply

Developer-Anony:
## Summary

Added new feature to `discord.ext.commands.Context`.
This function is called `maybe_reply`.
Basically this function tries to reply to the `Context.message` but if cannot it just sends a message in context's channel.

In a command it would be used like this:
```py
client = commands.Bot(...)

@client.command()
async def ping(ctx: commands.Context, delete_message: bool = False) -> None:

    if delete_message == True:
        await ctx.message.delete()
    await ctx.maybe_reply('Test executed!')
```
That code would be an example of this new function for Context class. Photo:
dpy new func
## Checklist
- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/context.py
+++ b/discord/ext/commands/context.py
@@ -745,7 +745,7 @@ class Context(discord.abc.Messageable, Generic[BotT]):
     async def maybe_reply(self, content: Optional[str] = None, **kwargs: Any):
         """|coro|
         
-        A shortcut method to both :meth:`reply` and :meht:`send` to either reply to a
+        A shortcut method to both :meth:`reply` and :meth:`send` to either reply to a
         message or send a message in the context's channel.
 
         For interaction based contexts, this is the same as :meth:`send`.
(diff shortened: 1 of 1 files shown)


discord/ext/commands/context.py
```diff
@@ -741,6 +741,42 @@ async def reply(self, content: Optional[str] = None, **kwargs: Any) -> Message:
             return await self.send(content, reference=self.message, **kwargs)
         else:
             return await self.send(content, **kwargs)
+        
+    async def maybe_reply(self, content: Optional[str] = None, **kwargs: Any):
+        """|coro|
+        
+        A shortcut method to both :meth:`reply` and :meht:`send` to either reply to a
```
AbstractUmbra, line 748:
```suggestion
        A shortcut method to both :meth:`reply` and :meth:`send` to either reply to a
```

AbstractUmbra (changes requested):
Minor typo.

Rapptz:
Thanks for the PR. However I don't think this is a feature I want to add to the library and this type of implicit failed request tends to be a bad idea unless it has a strong justification, which this feature lacks.

Note that replying to a message has this feature built-in using [`MessageReference.fail_if_not_exists`](https://discordpy.readthedocs.io/en/stable/api.html#discord.MessageReference.fail_if_not_exists).

PR #9445, opened 2023-06-07
Ensure tasks are not garbage collected early

mikeshardmind:
## Summary

Adds a task store to `Client` and utilizes it in `_schedule_event`.

This fixes an issue identified in Discord here: https://discord.com/channels/336642139381301249/1115863485692256286
While it is unclear if this is the specific cause of the issue they experienced, it is possible to confirm that this is an issue.
A minimal synthetic example of this as an issue has been shared in Discord, and I'll include it as a comment below as well.

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -38,6 +38,7 @@ from typing import (
     List,
     Literal,
     Optional,
+    Set,
     Sequence,
     Tuple,
     Type,
@@ -289,6 +290,7 @@ class Client:
         self._application: Optional[AppInfo] = None
         self._connection._get_websocket = self._get_websocket
         self._connection._get_client = lambda: self
+        self.__task_store: Set[asyncio.Task] = set()
 
         if VoiceClient.warn_nacl:
             VoiceClient.warn_nacl = False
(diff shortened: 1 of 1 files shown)


mikeshardmind:
Synthetic example of this as an issue:
```py

import gc
import asyncio
from typing import Coroutine, Any, Callable

class MinimalRepro:

    def __init__(self):
        self.loop = asyncio.get_running_loop()

    async def on_error(self, event_method: str, /, *args: Any, **kwargs: Any) -> None:
        # original implementation used `_log.exception`
    
        print('Ignoring exception in %s', event_method)

    # copied as-is
    async def _run_event(
        self,
        coro: Callable[..., Coroutine[Any, Any, Any]],
        event_name: str,
        *args: Any,
        **kwargs: Any,
    ) -> None:
        try:
            await coro(*args, **kwargs)
        except asyncio.CancelledError:
            pass
        except Exception:
            try:
                await self.on_error(event_name, *args, **kwargs)
            except asyncio.CancelledError:
                pass
    
    #copied as-is
    def _schedule_event(
        self,
        coro: Callable[..., Coroutine[Any, Any, Any]],
        event_name: str,
        *args: Any,
        **kwargs: Any,
    ) -> asyncio.Task:
        wrapped = self._run_event(coro, event_name, *args, **kwargs)
        # Schedules the task
        return self.loop.create_task(wrapped, name=f'discord.py: {event_name}')


async def useless():
    await asyncio.sleep(60)
    print("done being useless")

async def repro_case():
    m = MinimalRepro()
    m._schedule_event(useless, "useless")
    gc.collect()


if __name__ == '__main__':
    asyncio.run(repro_case(), debug=True)
```

discord/client.py
```diff
@@ -289,6 +289,7 @@ def __init__(self, *, intents: Intents, **options: Any) -> None:
         self._application: Optional[AppInfo] = None
         self._connection._get_websocket = self._get_websocket
         self._connection._get_client = lambda: self
+        self.__task_store: set[asyncio.Task] = set()
```
Gobot1234, line 292:
```suggestion
        self.__task_store: Set[asyncio.Task] = set()
```
Should use old classes for consistency right?

mikeshardmind: Yeah, it should. This has been changed and an import added for it, my fault for forgetting that's not valid in 3.8

mikeshardmind:
After further investigation (discussed in discord) I'm having trouble showing definitively that this can happen in an actual bot. There may be some implementation detail keeping tasks alive here, and I really don't want to dig into it, seems to not be the actual cause of issues here.

PR #9454, opened 2023-06-19
Convert input for UserConverter and MemberConverter to lower case during last lookup step

atomflunder:
## Summary

This PR converts the input for `UserConverter` and `MemberConverter` classes to lower case during last lookup step, for the new discord username system. The new username system does not allow for uppercase characters in your username and thus I think it would be a good idea to lower case the input automatically for that scenario.  
If the user still has a discriminator it will function as before.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -217,7 +217,7 @@ class MemberConverter(IDConverter[discord.Member]):
             predicate = lambda m: m.name == username and m.discriminator == discriminator
         else:
             lookup = argument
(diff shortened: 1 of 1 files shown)


Rapptz:
Thanks for the PR, however I don't think this change is a positive one in the current landscape. This will probably make more sense as the number of legacy users is basically zero, but as it stands right now this change is breaking for a vast majority of users.

atomflunder:
Okay, that does make sense to me. Thank you for your answer.

PR #9457, opened 2023-06-20
Fix false positives in `animated` for `PartialEmoji.from_str`

### LINKED ISSUE #9456: PartialEmoji.from_str incorrectly assumes emojis starting with `a` but not including `:` are animated
### Summary

PartialEmoji.from_str incorrectly assumes emojis starting with `a` but not including `:` are animated

### Reproduction Steps

- Create an emoji that starts with `a` such as `anemoji`
- Save the emoji information as `emoji_name:emoji_id`, a listed valid representation for `PartialEmoji.from_str`.
- Call `PartialEmoji.from_str("anemoji:12345678910111213")`
- The returned PartialEmoji assumes that the emoji is animated when it isn't.

### Minimal Reproducible Code

```python
>>> my_emoji = PartialEmoji.from_str("anemoji:12345678910111213")
>>> print(repr(my_emoji))
>>> <PartialEmoji animated=True name='nemoji' id=12345678910111213>
```

### Expected Results

I expected following a valid format for `PartialEmoji.from_str` to return a valid representation of the emoji.

### Actual Results

Emojis with names starting with `a` are incorrectly assumed to be animated despite missing the animated separator.

### Intents

all

### System Information

- Python v3.11.4-final
- discord.py v2.3.0-final
- aiohttp v3.8.4
- system info: Linux 5.15.90.2-microsoft-standard-WSL2 SMP Fri Jun 2 00:45:15 UTC 2023

### Checklist

### Additional Context

_No response_

AbstractUmbra:
## Summary

Fixes #9456 
Test cases can be found here: <https://regex101.com/r/d8WD6t/1>

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/partial_emoji.py
+++ b/discord/partial_emoji.py
@@ -94,7 +94,7 @@ class PartialEmoji(_EmojiTag, AssetMixin):
 
     __slots__ = ('animated', 'name', 'id', '_state')
 
-    _CUSTOM_EMOJI_RE = re.compile(r'<?(?P<animated>a)?:?(?P<name>[A-Za-z0-9\_]+):(?P<id>[0-9]{13,20})>?')
+    _CUSTOM_EMOJI_RE = re.compile(r'<?(?P<animated>a(?=:))?:?(?P<name>[A-Za-z0-9\_]+):(?P<id>[0-9]{13,20})>?')
 
     if TYPE_CHECKING:
         id: Optional[int]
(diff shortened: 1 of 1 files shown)


Rapptz:
I'm unsure why a positive lookahead is used, you can just put it as part of the `animated` group, e.g. `<?(?P<animated>a:)?:?(?P<name>[A-Za-z0-9\_]+):(?P<id>[0-9]{13,20})>?`. Alternatively you could do a conditional but that makes it confusing to read.

Rapptz:
Actually a better regex might be `<?(?:(?P<animated>a)?:)?(?P<name>[A-Za-z0-9\_]+):(?P<id>[0-9]{13,20})>?` since this one avoids the lookahead, embeds the conditional, and makes it so `::foo:1283908120938` is not a valid input

AbstractUmbra:
Good call! Updated that now.

PR #9466, opened 2023-06-25
Add support for asyncio debug mode.

hawk-tomy:
## Summary

Add an easy way to enable asyncio debug mode.
[asyncio debug mode](https://docs.python.org/3/library/asyncio-dev.html#asyncio-debug-mode)

Now
```py
import asyncio, discord

    async def setup_hooks(self):
        asyncio.get_running_loop().set_debug(True)

client = MyClient(...)
client.run(token)
```

After
```py
client = discord.Client(...)
client.run(token, asyncio_debug=True)
```

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -786,6 +786,7 @@ class Client:
         log_formatter: logging.Formatter = MISSING,
         log_level: int = MISSING,
         root_logger: bool = False,
+        asyncio_debug: Optional[bool] = None,
     ) -> None:
         """A blocking call that abstracts away the event loop
         initialisation from you.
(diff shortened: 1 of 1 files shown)


Rapptz:
Hi, thanks for the PR.

I've actually thought about this debug mode flag before even before v1.0 was released, though I had more in mind with it than just asyncio debug mode. I wanted to use it as a way to turn on a debug mode for the library (separate from the debug logging) that would give a bit more debug info that I usually would not provide due to performance considerations but the idea never really panned out.

The actual debug mode stuff that could be added would probably have to be specced out in a future date or in the Discord server.

I bring this up because that idea is partially in conflict with this PR. So I'm in a bit of a limbo on what I want to do. I suppose as a middle ground this could be renamed into `debug` instead of `asyncio_debug`. However since the debug mode idea I have is separate from asyncio debugging then it needs to be available somewhere else. Basically the scope of the feature I want to have and the scope that this PR provides are somewhat mismatched.

hawk-tomy:
I'm sorry. I missmaked comment (updated now). I want add `asyncio_debug` parameter, but I wrote `debug` in a sample code.

Rapptz:
I know I saw in the changeset.

hawk-tomy:
> Basically the scope of the feature I want to have and the scope that this PR provides are somewhat mismatched.
I have thought about it, but I cannot solve this mismatch.
I will close this PR.

PR #9467, opened 2023-06-25
Minor changes

balanagireddym:
## Summary

Minor documnentation

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,7 +18,7 @@ Key Features
 
 - Modern Pythonic API using ``async`` and ``await``.
 - Proper rate limit handling.
-- Optimised in both speed and memory.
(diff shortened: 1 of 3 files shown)


LeoCx1000:
"optimised" and "favourite" are both spelt correctly. There is no need to change them to it's American spelling counterpart as they're already properly spelt.

Also "predownload" seems like the proper word in that context, as adding a hyphen makes it sounds like it is referring to a step that occurs before downloading, not to the action of downloading the entire track before streaming it to discord. (Or well, that's how it seems to me).

Rapptz:
Thanks but all of these are fine as-is.

PR #9471, opened 2023-06-27
Fix `Intents.emoji`, `emojis_and_stickers` having swapped alias/non-alias decors

Aluerie:
## Summary

Extremely minor detail, but I noticed that `Intents.emojis` flag docs say "Changed to an alias in 2.0." however it still has `@flag_value` decorator, while vice-versa for `Intents.emojis_and_stickers`, when looks like, it's supposed to be a new main flag. Therefore, two lines PR. 😮 

## Checklist

    - [ ] I have updated the documentation to reflect the changes. (docs are fine)
- [X] This PR fixes an issue.

### DIFF
--- a/discord/flags.py
+++ b/discord/flags.py
@@ -850,7 +850,7 @@ class Intents(BaseFlags):
         """
         return 1 << 2
 
-    @flag_value
+    @alias_flag_value
     def emojis(self):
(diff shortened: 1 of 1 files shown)


Rapptz (approved):
Thanks.

PR #9472, opened 2023-06-28
save a reference to created tasks

anio:
> Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done.
>
> [python docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task)

## Summary

Please review the link: https://textual.textualize.io/blog/2023/02/11/the-heisenbug-lurking-in-your-async-code/

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/interactions.py
+++ b/discord/interactions.py
@@ -827,7 +827,7 @@ class InteractionResponse(Generic[ClientT]):
                 except HTTPException:
                     pass
 
-            asyncio.create_task(inner_call())
+            task = asyncio.create_task(inner_call())
 
     async def edit_message(
         self,
@@ -941,7 +941,7 @@ class InteractionResponse(Generic[ClientT]):
                 except HTTPException:
                     pass
 
-            asyncio.create_task(inner_call())
+            task = asyncio.create_task(inner_call())
 
     async def send_modal(self, modal: Modal, /) -> None:
         """|coro|
@@ -1227,6 +1227,6 @@ class InteractionMessage(Message):
                 except HTTPException:
                     pass
 
-            asyncio.create_task(inner_call())
+            task = asyncio.create_task(inner_call())
         else:
             await self._state._interaction.delete_original_response()
--- a/discord/message.py
+++ b/discord/message.py
@@ -854,7 +854,7 @@ class PartialMessage(Hashable):
                 except HTTPException:
                     pass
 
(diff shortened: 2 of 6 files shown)


Rapptz:
Thanks for the PR.

1. This doesn't do anything
2. The intention is a duplicate of https://github.com/Rapptz/discord.py/pull/9445 and [discussion in the d.py server](https://canary.discord.com/channels/336642139381301249/1115863485692256286/1115863485692256286)

Gobot1234:
This has already been discussed on [discord](https://discord.com/channels/336642139381301249/1116114063173361714) and the conclusion was that since no-one has ever reported this being an issue, something is keeping the task alive. (This also wouldn't fix the problem even if it was there because as soon as `task` goes out of scope it would be GC'ed, you'd need to add them to somewhere they can escape the execution of the function)

mikeshardmind:
I looked into this more, and asyncio itself keeps the tasks alive. The documentation there is a lie, asyncio internals keep a strong reference to tasks that have been scheduled, at least until the initial checking for done callbacks, on top of the weakset (which opens some other questions as to if this is unintentional, implementation detail, or outdated docs, but I don't want to get into that here) The article there from textualize is interesting, as it is calling something a heisenbug when you can verify that asyncio keeps the task alive and do so with confidence that this is deterministically so, but given my own initial mistake in checking this, it's hard to blame the author.

As for why I thought there might have been an issue, there's actually a different issue with how I originally went about checking this which happened to cause an *appearance* of the same issue. I put a level of trust in the asyncio docs, and when something happened a specific way, I trusted that what was documented must be the cause without fully investigating it (at first) However, as the discussion thread linked above should show, if what I thought was an issue was, there would be more issues than this. Additionally, you can check for yourself where and how asyncio keeps references for scheduled tasks.

mikeshardmind:
Someone poked me on discord about the prior comment, I've already responded to them, but for anyone else curious on specifics, [this comment in an asyncio issue details where it can happen](https://github.com/python/cpython/issues/91887#issuecomment-1434816045 ) (it shouldn't be possible in discord.py dispatch) and it's "planned" to be handled upstream for tasks in the future to remove the very slim edge case where this issue is actually possible.

PR #9483, opened 2023-07-04
Update some typing enums

A5rocks:
## Summary

I updated some types. (I was checking [my own types](https://github.com/fleuralice/discord-docs) against these to check and found these and a few more: I still have to confirm those others.)

## Checklist

- [ ] This PR fixes an issue.

I don't think types are tested? And I think none of these should really be checked...


Rapptz (approved):
Thanks.

PR #9484, opened 2023-07-05
Renamed some things of guilds, members and roles for the Guild Premiums Subscription system

Developer-Anony:
## Summary
This PR basically changes the name of some things in the next files:
>> Modified member.py: Changed the 'premium_since' to 'nitro_subsriber_since'
>> Modified guild.py: Changed 'premium_tier' to 'boost_tier', changed 'premium_subscription_count' to 'boost_count', changed 'premium_progress_bar_enabled' to 'boost_progress_bar_enabled, changed 'Guild._PREMIUM_GUILD_LIMITS' to 'Guild._BOOSTED_GUILD_LIMITS'
>> Modified role.py: Changed 'is_premium_subscriber' to 'is_booster'
I did this because as you may know Discord will launch a complete Guild Premium Subscription system, this change will affect to the Nitro and Boost things in discord.py, as they are referred as `premium`, so they are renamed as `nitro` or `boost` (depending of what it refers).
If this PR is merged I know the users that use discord.py will need to change everything related to nitro and boost that are referred as `premium` to the new names, but, as I said, it is for the best (ig).

Also as I kept the old names as aliases.

## Checklist
- [ ] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -882,7 +882,7 @@ class Guild(Hashable):
     @property
     def boosters(self) -> List[Member]:
         """List[:class:`Member`]: A list of members who have "boosted" this guild."""
-        return [member for member in self.members if member.nitro_subsciber_since is not None]
+        return [member for member in self.members if member.nitro_subscriber_since is not None]
 
     @property
     def roles(self) -> Sequence[Role]:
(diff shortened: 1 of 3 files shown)


LostLuma:
FWIW your PR *is* a breaking change since users of the library would have to change their code when updating. Keep in mind that discord.py follows semantic versioning, more info here: https://discordpy.readthedocs.io/en/latest/version_guarantees.html

If you want any chance of merging such a PR you'd have to keep an alias with the old name for each thing you've changed.

Developer-Anony:
@LostLuma Thanks for the advice! Going to change it now.

Rapptz:
These fields are named directly from the API. Subscribers use different attributes. It makes no sense to change these.

PR #9486, opened 2023-07-07
Add support for attachment flags

Soheab:
## Summary

This PR adds support for attachment flags.

Depends on https://github.com/discord/discord-api-docs/pull/6272

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/flags.py
+++ b/discord/flags.py
@@ -1824,3 +1824,76 @@ class MemberFlags(BaseFlags):
     def started_onboarding(self):
         """:class:`bool`: Returns ``True`` if the member has started onboarding."""
         return 1 << 3
+
+
+@fill_with_flags()
+class AttachmentFlags(BaseFlags):
+    r"""Wraps up the Discord Attachment flags
+
+    .. versionadded:: 2.4
+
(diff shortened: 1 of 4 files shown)


Rapptz (changes requested):
In the library `is_` methods are always methods, i.e. callable. Since these are properties/flags you need to remove the `is_` prefix.

discord/flags.py
```diff
@@ -1824,3 +1825,76 @@ def bypasses_verification(self):
...
+
+    @flag_value
+    def clip(self):
+        """:class:`bool`: Returns ``True`` if the attachment is a clip."""
+        return 1 << 0
+
+    @flag_value
+    def thumbnail(self):
+        """:class:`bool`: Returns ``True`` if the attachment is a thumbnail."""
+        return 1 << 1
```
Rapptz, line 1895: I know these are undocumented on Discord's side but it sure would be nice to know what these even are for better documentation and maybe better naming, but this won't block this PR.

Rapptz (approved):
Thanks.

PR #9499, opened 2023-07-16
Rework login-flow to retrieve user object from appinfo

bijij:
## Summary

This PR reworks login flow to use the new `bot` key in the application payload for the ClientUser object, rather than making a 2nd HTTP request.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -610,8 +610,8 @@ class Client:
         token = token.strip()
 
         data = await self.http.static_login(token)
-        self._connection.user = ClientUser(state=self._connection, data=data)
-        self._application = await self.application_info()
+        self._connection.user = ClientUser(state=self._connection, data=data['bot'])  # type: ignore
+        self._application = AppInfo(self._connection, data)
         if self._connection.application_id is None:
             self._connection.application_id = self._application.id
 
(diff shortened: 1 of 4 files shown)


discord/client.py
```diff
@@ -610,8 +610,8 @@ async def login(self, token: str) -> None:
         token = token.strip()
 
         data = await self.http.static_login(token)
-        self._connection.user = ClientUser(state=self._connection, data=data)
-        self._application = await self.application_info()
+        self._connection.user = ClientUser(state=self._connection, data=data['bot'])  # type: ignore
```
bijij, line 613: Not sure if it would be better to raise above here if `'bot' not in data`.
I don't think it can ever happen though.

bijij:
Do we want add support for the new PATCH endpoint for applications?

Rapptz:
Sure.

bijij:
Seems the branch in the API docs has been merged.
Without any response regarding our query.

Given last week I checked and it didn't have the fields, I'm guessing that it's not happening.

Rapptz:
Yeah, they basically said to make an issue if we want it but my experience with that means that it's never happening.

PR #9506, opened 2023-07-22
Update button.py

mister-proger:
Added the 'callback` parameter to the button initializer

## Summary

Instead of initializing a button with a `callback` in two lines, it is now possible to feed the `callback` function directly to the initializer

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
Thanks for the PR but this is unpythonic.

mister-proger:
> Thanks for the PR but this is unpythonic.

Yes, I know, but why not do it? Like I still use a subclass where I added the same line. Of course, we approve of the masochism, but it’s easier for everyone to switch to assembler...

PR #9527, opened 2023-08-19
Add encoder params to VoiceClient.play

### LINKED ISSUE #7130: Audio bitrate limiting
### Summary

Being able to limit the bitrate of outgoing audio streams

The core library

### The Problem

Hosting bots on connections with low bandwidth can make the played audio lag, this could be solved by giving the option to throttle the outgoing audio from the bot to a lower bandwidth to sacrifice audio quality for stability

### The Ideal Solution

Being able to simply give `voice_channel.play` an option for bitrate (in kilobytes, i guess) - perhaps like the PCMVolumeTransformer is implemented.

```
vc.source = discord.PCMVolumeTransformer(vc.source)
vc.source.volume = float(volume)
```
->

```
vc.source = discord.BitrateTransformer(vc.source)
vc.source.bitrate = 32
```

### The Current Solution

None

### Additional Context

_No response_

imayhaveborkedit:
## Summary
Adds kwarg parameters to allow access to setting `VoiceClient`'s underlying `Encoder` initial parameters.  In addition, this also changes the APPLICATION_ constants to be like the other CTL TypedDicts.  Also fixes an error message typo.  I'm not really sure about using stringly typed constants but its what the `Encoder` constructor uses. 

Closes #7130 

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/discord/opus.py
+++ b/discord/opus.py
@@ -39,8 +39,15 @@ from .errors import DiscordException
 
 if TYPE_CHECKING:
     T = TypeVar('T')
+    APPLICATION_CTL = Literal['audio', 'voip', 'lowdelay']
     BAND_CTL = Literal['narrow', 'medium', 'wide', 'superwide', 'full']
     SIGNAL_CTL = Literal['auto', 'voice', 'music']
+
+
+class ApplicationCtl(TypedDict):
+    audio: int
+    voip: int
+    lowdelay: int
 
 
 class BandCtl(TypedDict):
@@ -90,9 +97,10 @@ OK      = 0
 BAD_ARG = -1
 
 # Encoder CTLs
-APPLICATION_AUDIO    = 2049
-APPLICATION_VOIP     = 2048
-APPLICATION_LOWDELAY = 2051
+APPLICATION_AUDIO    = 'audio'
+APPLICATION_VOIP     = 'voip'
+APPLICATION_LOWDELAY = 'lowdelay'
+# These remain as strings for backwards compat
 
 CTL_SET_BITRATE      = 4002
 CTL_SET_BANDWIDTH    = 4008
@@ -104,6 +112,12 @@ CTL_SET_SIGNAL       = 4024
 CTL_SET_GAIN             = 4034
 CTL_LAST_PACKET_DURATION = 4039
 # fmt: on
+
+application_ctl: ApplicationCtl = {
+    'audio': 2049,
+    'voip': 2048,
+    'lowdelay': 2051,
+}
 
 band_ctl: BandCtl = {
     'narrow': 1101,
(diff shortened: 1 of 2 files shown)


discord/voice_client.py
```diff
@@ -45,6 +45,7 @@
 import struct
 import threading
 from typing import Any, Callable, List, Optional, TYPE_CHECKING, Tuple, Union
+from typing_extensions import Unpack
```
Rapptz, line 48: This is not a runtime requirement so it will fail for most people. We usually import from `typing_extensions` purely at type checking time inside `TYPE_CHECKING`. That being said, this style of unpacking kwargs is *not* used in the repository at the moment. Originally the reason was because it was not formally accepted into Python but it was added during [PEP-692](https://peps.python.org/pep-0692/). However, last time we attempted to incorporate this the ergonomics were not ideal since it had a lot of bugs in Pyright and did not have good auto complete experience.

I am unsure if this has changed since then.

discord/opus.py
```diff
@@ -57,6 +64,15 @@ class SignalCtl(TypedDict):
     music: int
 
 
+class EncoderKwargs(TypedDict, total=False):
+    application: APPLICATION_CTL
```
Rapptz, line 68: I don't really see much point in allowing this in particular to be set?

discord/opus.py
```diff
@@ -57,6 +64,15 @@ class SignalCtl(TypedDict):
...
+    bitrate: int
+    fec: bool
+    packet_loss_pct: float
```
Rapptz, line 71:
```suggestion
    expected_packet_loss: float
```

Shortcuts in typing are usually not done unless necessary and this one's better off longer.

discord/opus.py
```diff
@@ -57,6 +64,15 @@ class SignalCtl(TypedDict):
     music: int
 
 
+class EncoderKwargs(TypedDict, total=False):
+    application: APPLICATION_CTL
+    bitrate: int
+    fec: bool
```
Rapptz, line 70: This one's technically a shortcut too but I'm unsure if it's better off in long form or not. No strong opinion, just noting it.

discord/opus.py
```diff
@@ -319,16 +342,36 @@ def get_opus_version() -> str:
...
+        *,
+        application: APPLICATION_CTL = 'audio',
+        bitrate: int = 128,
+        fec: bool = True,
+        packet_loss_pct: float = 0.15,
+        bandwidth: BAND_CTL = 'full',
+        signal_type: SIGNAL_CTL = 'auto',
+    ):
+        if application not in application_ctl:
+            raise KeyError(f'{application} is not a valid application setting. Try one of: {"".join(application_ctl)}')
```
Rapptz, line 356:
```suggestion
            raise ValueError(f'{application} is not a valid application setting. Try one of: {"".join(application_ctl)}')
```

discord/voice_client.py
```diff
@@ -599,6 +633,10 @@ def play(self, source: AudioSource, *, after: Optional[Callable[[Optional[Except
             Source is not a :class:`AudioSource` or after is not a callable.
         OpusNotLoaded
             Source is not opus encoded and opus is not loaded.
+        ValueError
+            An improper value was passed as an encoder parameter.
+        KeyError
+            An improper value was passed as an encoder parameter.
```
Rapptz, line 639:
```suggestion
```

Rapptz (changes requested):
This API is a bit weird because the encoder instances are usually shared which means this API can behave unexpectedly if you play with different parameters unless I'm missing something.

Unsure how this could be reconciled though.

discord/opus.py
imayhaveborkedit: Without `total=False` pyright complains if you don't have all the parameters passed in the function, if that's what you were talking about.

Rapptz: No I was talking about the application parameter.

imayhaveborkedit: That was originally the only constructor parameter that Encoder had so it didn't occur to me to either remove it or arbitrarily restrict access to only that setting.

imayhaveborkedit:
> the encoder instances are usually shared

So unless *I'm* missing something the Encoder instances exist as one per VoiceClient that handles one AudioSource.  In fact they *shouldn't* be shared with anything.  I'm not exactly sure what the concern is but all this really does is expose the previously publicly inaccessible Encoder configuration settings as a set of initial parameters.  The encoder is recreated if any options are specified so I don't think this can conflict with anything.  One could even change these settings at any point by accessing `voice_client._encoder.set_*` and it would work fine (i've done this before with bitrate for fun) but I didn't consider this important enough to expose an api for.

discord/voice_client.py
```diff
@@ -569,7 +571,13 @@ def _encrypt_xsalsa20_poly1305_lite(self, header: bytes, data) -> bytes:
 
         return header + box.encrypt(bytes(data), bytes(nonce)).ciphertext + nonce[:4]
 
-    def play(self, source: AudioSource, *, after: Optional[Callable[[Optional[Exception]], Any]] = None) -> None:
+    def play(
+        self,
+        source: AudioSource,
+        *,
+        after: Optional[Callable[[Optional[Exception]], Any]] = None,
+        **encoder_kwargs: Unpack[EncoderKwargs],
```
Rapptz, line 579: Can you manually unpack these kwargs instead of using this form? It prevents people from typoing.

discord/voice_client.py
```diff
@@ -579,9 +587,15 @@ def play(self, source: AudioSource, *, after: Optional[Callable[[Optional[Except
...
         passed, any caught exception will be logged using the library logger.
 
+        Extra parameters may be passed to the internal opus encoder if a PCM based
+        source is used.  Otherwise, they are ignored.
+
         .. versionchanged:: 2.0
             Instead of writing to ``sys.stderr``, the library's logger is used.
 
+        .. versionchanged:: 2.4
+            Added encoder parameters as keyword arguments in ``encoder_kwargs``.
```
Rapptz, line 597:
```suggestion
            Added encoder parameters as keyword arguments.
```

Rapptz (changes requested):
I guess if the encoder instances aren't shared then there's no real issue.

Rapptz (approved):

PR #9528, opened 2023-08-20
Fixes/improvements to FFmpegAudio and subclasses

imayhaveborkedit:
## Summary
- Enables FEC on `FFmpegOpusAudio`
- Adds [-blocksize](https://ffmpeg.org/ffmpeg-all.html#pipe "ffmpeg docs: pipe") to the ffmpeg cli arguments
- Fixes 'copy' codec not working when passed directly to `FFmpegOpusAudio`
- Improves error message for Ogg parsing failure
- Adds a debug logging message for when an ffmpeg process is spawned
- Fixes the `stderr` parameter
  - Deprecates support for passing `subprocess.PIPE`, as it did nothing.  Now emits a `DeprecationWarning`.
  - Corrects the type hint
  - Properly support non-fs backed file objects (like `BytesIO`) by adding a reader thread

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/player.py
+++ b/discord/player.py
@@ -175,6 +175,7 @@ class FFmpegAudio(AudioSource):
             self._pipe_thread.start()
 
     def _spawn_process(self, args: Any, **subprocess_kwargs: Any) -> subprocess.Popen:
+        _log.debug('Spawning ffmpeg process with command: %s', args)
         process = None
         try:
             process = subprocess.Popen(args, creationflags=CREATE_NO_WINDOW, **subprocess_kwargs)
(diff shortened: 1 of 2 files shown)


discord/player.py
```diff
@@ -222,9 +246,25 @@ def _pipe_writer(self, source: io.BufferedIOBase) -> None:
...
+                data: bytes = self._stderr.read(self.BLOCKSIZE)
+            except Exception:
+                _log.debug('Read error for %s, this is probably not a problem', self, exc_info=True)
+                return
+            if data is None:
+                return
+            # TODO: Do I need to check if this explodes?  catch OSError or something?
+            #       What happens if dest can no longer be written to?  Keep reading without writing or exit?
+            #       Does it pose an issue if the stderr pipe doesn't get read from?
+            dest.write(data)
```
Rapptz, line 263: If an exception happens here then it should ideally properly clean up. It's probably as simple as just doing a `return` like you did elsewhere rather than propagating it to the thread exit.

PR #9529, opened 2023-08-20
Replace audioop with Python equivalent 

### LINKED ISSUE #9477: Use of deprecated module audioop
### Summary

discord.py uses the audioop module which will be removed in Python 3.13

### Reproduction Steps

On Python 3.11: `python -Walways -c "import discord"`

Output:
```
...\Python311\Lib\site-packages\discord\player.py:28: DeprecationWarning: 'audioop' is deprecated and slated for removal in Python 3.13
  import audioop
```

### System Information

```
- Python v3.11.2-final
- discord.py v2.3.1-final
- aiohttp v3.8.3
```

### Additional Context

The only usage is here: https://github.com/Rapptz/discord.py/blob/0871b34fc86af49deb6db2436d3e77f5516fdcba/discord/player.py#L633

There doesn't seem to be any standard library alternative, or external library that I can obviously see. It's been a while since it was deprecated so I'm not sure how much chance there is someone will turn it into an external library.

One option could be to try and reimplement it in Python, which seems like it could be doable as it doesn't seem too complicated: https://github.com/python/cpython/blob/036da3bd43aa2593d17d2fb73d4794f9965c577d/Modules/audioop.c#L944, not really sure though.

davidhozic:
## Summary

Replaces audioop with Python equivalent inside ``discord.PCMVolumeTransformer.read()``.
Resolves #9477 .

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/player.py
+++ b/discord/player.py
@@ -25,10 +25,10 @@ from __future__ import annotations
 
 import threading
 import subprocess
-import audioop
 import asyncio
 import logging
 import shlex
+import array
 import time
 import json
 import sys
@@ -36,6 +36,7 @@ import re
 import io
 
 from typing import Any, Callable, Generic, IO, Optional, TYPE_CHECKING, Tuple, TypeVar, Union
+from math import floor
 
 from .enums import SpeakingState
 from .errors import ClientException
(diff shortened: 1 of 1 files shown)


Rapptz:
Thanks for the PR but this isn't what we plan on doing.

PR #9540, opened 2023-08-27
Fix IndexError in Colour.from_str('')

ambdroid:
## Summary

I think this is more in line with the documented behavior.

## Checklist

    - [n/a] I have updated the documentation to reflect the changes.
- [x] This PR fixes an issue.

### DIFF
--- a/discord/colour.py
+++ b/discord/colour.py
@@ -196,6 +196,9 @@ class Colour:
             The string could not be converted into a colour.
         """
 
+        if value == '':
+            raise ValueError('cannot convert empty string to colour')
+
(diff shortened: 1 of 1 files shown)


discord/colour.py
```diff
@@ -196,6 +196,9 @@ def from_str(cls, value: str) -> Self:
             The string could not be converted into a colour.
         """
 
+        if value == '':
+            raise ValueError('cannot convert empty string to colour')
```
Rapptz, line 200:
```suggestion
        if not value:
            raise ValueError('unknown colour format given')
```

PR #9556, opened 2023-09-10
patching doc extensions to work with later sphinx versions

CrosswaveOmega:
fix for  `doc/extensions/builder.py` to allow for compatibility with sphinx 7.2.5 and up.

## Checklist

- [X ] If code changes were made then they have been tested.
    - [X ] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.

### DIFF
--- a/docs/extensions/builder.py
+++ b/docs/extensions/builder.py
@@ -1,5 +1,5 @@
 from sphinx.builders.html import StandaloneHTMLBuilder
-from sphinx.builders.gettext import MessageCatalogBuilder, I18nBuilder, timestamp, ltz, should_write, GettextRenderer
+from sphinx.builders.gettext import MessageCatalogBuilder, I18nBuilder, ctime,should_write, GettextRenderer
 from sphinx.locale import __
 from sphinx.util import status_iterator
 from sphinx.util.osutil import ensuredir
@@ -69,7 +69,7 @@ class DPYMessageCatalogBuilder(MessageCatalogBuilder):
             'project': self.config.project,
             'last_translator': self.config.gettext_last_translator,
             'language_team': self.config.gettext_language_team,
-            'ctime': datetime.datetime.fromtimestamp(timestamp, ltz).strftime('%Y-%m-%d %H:%M%z'),
+            'ctime': ctime,
             'display_location': self.config.gettext_location,
             'display_uuid': self.config.gettext_uuid,
         }


gribbg:
I'm running into this same issue.

Note that `.astimezone()` can provide access to the local timezone:
```python
>>> datetime.datetime.now().astimezone().tzinfo
datetime.timezone(datetime.timedelta(days=-1, seconds=72000), 'Eastern Daylight Time')
```

gribbg:
Actually, it's probably best to just use the old code in the `except` block:
```python
except Exception as exc:
    from sphinx.buildert.gettext import timestamp, ltz
    ctime = datetime.datetime.fromtimestamp(timestamp, ltz).strftime('%Y-%m-%d %H:%M%z')
```

CrosswaveOmega:
I was considering replacing the ctime calculation used in the except block with the ctime calculation used in the later sphinx build, but your way would work out better.

Rapptz:
Rather than doing the compatibility just bump the version along with the changes.

CrosswaveOmega:
Outside of a deprecation warning for status_iterator (which I patched), the extensions now work with the latest sphinx version (7.2.6).  
The only thing left is to resolve any cosmetic/front end issues caused by bumping the sphinx version.

No767:
Might be useful to run `docs/extensions/builder.py` through black and isort (useless it is already ran through)

CrosswaveOmega:
I didn't run black and isort originally, I did so in this commit.

setup.py
```diff
@@ -35,7 +35,7 @@
 extras_require = {
     'voice': ['PyNaCl>=1.3.0,<1.6'],
     'docs': [
-        'sphinx==4.4.0',
+        'sphinx==7.2.6',
```
No767, line 38: I want to note that Sphinx 7.2.6 does not support Python 3.8 ([changelog](https://www.sphinx-doc.org/en/master/changes.html#id32), [pr](https://github.com/sphinx-doc/sphinx/pull/11511)) since they dropped support for 3.8 in Sphinx 7.2.0. The only latest version that does support 3.8 is Sphinx 7.1.2

Other than that everything seems fine

No767 (changes requested):

setup.py
CrosswaveOmega: If that's the case, I can change the version to 7.1.2.  The patch will still work.

No767: Sounds good

PR #9559, opened 2023-09-12
Fix `View.from_message` not creating other Selects

z03h:
## Summary

`View.from_message` would ignore the component type and only create string Selects even if it was a Channel, Role, User, or Mentionable Select. 

`BaseSelect.from_component` did not copy the original custom id.

This broke [dynamic selects](https://github.com/Rapptz/discord.py/blob/3827671bf748c6f44f75f72fe1a84935c24c01b9/discord/ui/view.py#L619) since component type and custom id would not match.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
In hindsight, this `from_component` implementation is pretty bad since it's mixing the responsibility of `__repr__` with the construction when those should be separate. This can be fixed later though.

PR #9574, opened 2023-09-20
Pass exceptions from coro to on_error

LeoSko:
## Summary

This PR tries to resolve issue when `on_ready` event processing throws exception and it doesn't appear in any logs making it seem like bot is stuck at loading. I think this also affects other events, but at least this PR seems to fix for `on_ready` on my machine.

_Spent about an hour looking at logging and debugging after insignificant changes only to find out I'm try to access field which doesn't exist._

## Checklist

    - [ ] I have updated the documentation to reflect the changes. - if changes are OK, I can take a look to update them.
- [x] This PR fixes an issue. - _at least for me_
- [ ] ~~This PR adds something new (e.g. new method or parameters).~~
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed) - **please verify, but it seems to be non-breaking**
- [ ] ~~This PR is **not** a code change (e.g. documentation, README, ...)~~

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -443,9 +443,9 @@ class Client:
             await coro(*args, **kwargs)
         except asyncio.CancelledError:
             pass
-        except Exception:
+        except Exception as e:
             try:
-                await self.on_error(event_name, *args, **kwargs)
+                await self.on_error(event_name, exception=e, *args, **kwargs)
             except asyncio.CancelledError:
                 pass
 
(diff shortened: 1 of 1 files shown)


Rapptz:
Thanks for the PR.

Inside `on_error` the way to get the exception information is through [`sys.exc_info`](https://docs.python.org/3/library/sys.html#sys.exc_info). This is [documented in the event reference](https://discordpy.readthedocs.io/en/stable/api.html#discord.on_error).

PR #9586, opened 2023-09-30
[app_commands] fix typehints for __class_getitem__ in Transform and Range

owocado:
## Summary

`_TransformMetadata` was removed in commit https://github.com/Rapptz/discord.py/commit/11618cd1bac5b18f378f1dc31cf9d7bc2585dbe4#diff-f265b23b93514ed2b2e81c5a2d802297e0b72659aa10712c6bacf454e9940b77L290-L293

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/transformers.py
+++ b/discord/app_commands/transformers.py
@@ -525,7 +525,7 @@ else:
         .. versionadded:: 2.0
         """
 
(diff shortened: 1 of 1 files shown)


Rapptz (approved):
Thanks. These type hints don't end up mattering since it's not in a type checking block but good catch nonetheless.

PR #9588, opened 2023-09-30
Refactor ``from_components`` for select classes

Soheab:
## Summary

This PR "refactors" the `from_component` class method for selects to not be weird and use the `__item_repr_attributes__ ` class var.

Not sure if this is the best way to do this but it seems to work fine: https://mystb.in/MeasuresConspiracyDeaf

The classmethod is not considered private but also not documented, so I'm not sure if the `isinstance` and `else` is needed.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ui/select.py
+++ b/discord/ui/select.py
@@ -813,7 +813,10 @@ class ChannelSelect(BaseSelect[V]):
         ordering. The row number must be between 0 and 4 (i.e. zero indexed).
     """
 
-    __component_attributes__ = BaseSelect.__component_attributes__ + ('channel_types', 'default_values',)
+    __component_attributes__ = BaseSelect.__component_attributes__ + (
+        'channel_types',
+        'default_values',
+    )
 
     def __init__(
         self,
--- a/discord/ui/select.py
+++ b/discord/ui/select.py
@@ -58,7 +58,6 @@ from ..role import Role
 from ..user import User, ClientUser
 from ..abc import GuildChannel
 from ..threads import Thread
-from ..components import SelectMenu as SelectComponent
 
 __all__ = (
     'Select',
@@ -225,6 +224,13 @@ class BaseSelect(Item[V]):
         'max_values',
         'disabled',
     )
+    __component_attributes__: Tuple[str, ...] = (
+        'custom_id',
+        'placeholder',
+        'min_values',
+        'max_values',
+        'disabled',
+    )
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


discord/ui/select.py
```diff
@@ -343,12 +344,29 @@ def is_dispatchable(self) -> bool:
...
-    def from_component(cls, component: SelectMenu) -> Self:
-        return cls(
-            **{k: getattr(component, k) for k in cls.__item_repr_attributes__},
-            custom_id=component.custom_id,
-            row=None,
-        )
+    def from_component(cls, component: SelectComponent) -> Self:
+        if not isinstance(component, SelectComponent):
+            raise TypeError('Excepted an instance of SelectComponent not {component.__class__.__name__}')
+
```
Rapptz, line 350:
```suggestion
```

discord/ui/select.py
```diff
@@ -343,12 +344,29 @@ def is_dispatchable(self) -> bool:
...
+        kwrgs = {
+            'custom_id': component.custom_id,
+            'placeholder': component.placeholder,
+            'min_values': component.min_values,
+            'max_values': component.max_values,
+            'disabled': component.disabled,
+        }
```
Rapptz, line 357: This is probably better off being a separate attribute, such as `__from_component_attributes__` with a tuple of attributes or something in each one similar to the old design but decoupled from the `__repr__`.

Rapptz (changes requested):

discord/ui/select.py
Soheab: Again with the `BaseSelect.... += (..., ...)` ?

Rapptz: Not `+=` because that modifies the original but yes. The problem with the old one was that it was coupled with `__repr__` which made the output way too verbose.

discord/ui/select.py
```diff
@@ -344,11 +351,19 @@ def is_dispatchable(self) -> bool:
...
+            ComponentType.string_select: Select,
+            ComponentType.user_select: UserSelect,
+            ComponentType.role_select: RoleSelect,
+            ComponentType.channel_select: ChannelSelect,
+            ComponentType.mentionable_select: MentionableSelect,
+        }
+        if component.type not in type_to_cls:
+            raise TypeError(f'Unknown component type {component.type}')
+
+        constructor = type_to_cls[component.type]
```
Rapptz, line 364:
```suggestion
        constructor = type_to_cls.get(component.type, Select)
```

I want this to be mostly crash free.

Rapptz (approved):

PR #9590, opened 2023-09-30
Add support for guild incidents

Soheab:
## Summary

This PR adds support for the `incidents_data` field on guilds.

Relevant API Docs:
- https://github.com/discord/discord-api-docs/pull/6396

This was discussed in [ discord.py >](https://discord.gg/dpy)[💬Guild incident data](https://canary.discord.com/channels/336642139381301249/1158118289390702744)

## Checklist

- [ ] This PR fixes an issue.


Rapptz (changes requested):
With this API, there's no way to edit the incident data from a clean slate. What I mean is, if e.g. `Guild.incidents_data` is `None` because it's unset then there's no way to add new incident data.

I'm honestly unsure how I feel about the attribute as a whole, it does mirror the API as-is so it's not particularly controversial but it doesn't feel very ergonomic.

It would be nice if there was e.g. `Guild.invites_paused() -> bool` and `Guild.dms_paused() -> bool` helper functions. I don't know if the `are_` prefix is good for these, it feels awkward.

Soheab:
> With this API, there's no way to edit the incident data from a clean slate. What I mean is, if e.g. `Guild.incidents_data` is `None` because it's unset then there's no way to add new incident data.
> 
> I'm honestly unsure how I feel about the attribute as a whole, it does mirror the API as-is so it's not particularly controversial but it doesn't feel very ergonomic.
> 
> It would be nice if there was e.g. `Guild.invites_paused() -> bool` and `Guild.dms_paused() -> bool` helper functions. I don't know if the `are_` prefix is good for these, it feels awkward. 

> With this API, there's no way to edit the incident data from a clean slate. What I mean is, if e.g. `Guild.incidents_data` is `None` because it's unset then there's no way to add new incident data.

Looks like I forgot to change the type but it does always return an instance of `IncidentData` to use `.edit` on.

> I'm honestly unsure how I feel about the attribute as a whole, it does mirror the API as-is so it's not particularly controversial but it doesn't feel very ergonomic.

Not sure either. Maybe the two fields should just be on `Guild.edit` and have those helper methods mentioned above? Or a method called `edit_security_actions(...) -> Guild` or something like that (`edit_incidents`, ...).

discord/guild.py
```diff
@@ -4292,3 +4333,45 @@ async def create_automod_rule(
...
+    @property
+    def invites_paused_until(self) -> Optional[datetime.datetime]:
+        """:class:`datetime.datetime`: When invites get enabled again.
+
+        .. versionadded:: 2.4
+        """
+        if not self._incidents_data:
+            return None
+
+        return utils.parse_time(self._incidents_data.get("invites_disabled_until"))
```
Rapptz, line 4346:
```suggestion
        return utils.parse_time(self._incidents_data.get('invites_disabled_until'))
```

discord/guild.py
```diff
@@ -4292,3 +4333,45 @@ async def create_automod_rule(
...
+    @property
+    def dms_paused_until(self) -> Optional[datetime.datetime]:
+        """:class:`datetime.datetime`: When direct messages get enabled again.
+
+        .. versionadded:: 2.4
+        """
+        if not self._incidents_data:
+            return None
+
+        return utils.parse_time(self._incidents_data.get("dms_disabled_until"))
```
Rapptz, line 4357:
```suggestion
        return utils.parse_time(self._incidents_data.get('dms_disabled_until'))
```

discord/guild.py
```diff
@@ -4292,3 +4333,45 @@ async def create_automod_rule(
         )
 
         return AutoModRule(data=data, guild=self, state=self._state)
+
+    @property
+    def invites_paused_until(self) -> Optional[datetime.datetime]:
+        """:class:`datetime.datetime`: When invites get enabled again.
```
Rapptz, line 4339:
```suggestion
        """Optional[:class:`datetime.datetime`]: If invites are paused, returns when invites will get enabled in UTC,
        otherwise returns None.
```

discord/guild.py
```diff
@@ -4292,3 +4333,45 @@ async def create_automod_rule(
...
+        .. versionadded:: 2.4
+        """
+        if not self._incidents_data:
+            return None
+
+        return utils.parse_time(self._incidents_data.get("invites_disabled_until"))
+
+    @property
+    def dms_paused_until(self) -> Optional[datetime.datetime]:
+        """:class:`datetime.datetime`: When direct messages get enabled again.
```
Rapptz, line 4350:
```suggestion
        """Optional[:class:`datetime.datetime`]: If DMs are paused, returns when DMs will get enabled in UTC,
        otherwise returns None.
```

Rapptz (changes requested):
This implementation is much better.

discord/guild.py
```diff
@@ -509,6 +511,7 @@ def _from_data(self, guild: GuildPayload) -> None:
         self.owner_id: Optional[int] = utils._get_as_snowflake(guild, 'owner_id')
         self._large: Optional[bool] = None if self._member_count is None else self._member_count >= 250
         self._afk_channel_id: Optional[int] = utils._get_as_snowflake(guild, 'afk_channel_id')
+        self._incidents_data: Optional[IncidentData] = guild.get("incidents_data")
```
Rapptz, line 514:
```suggestion
        self._incidents_data: Optional[IncidentData] = guild.get('incidents_data')
```

Rapptz (approved):
This looks good, thanks.

PR #9594, opened 2023-10-01
Support 3.11

tuna2134:
## Summary

Add 3.11 ~~and 3.12~~ to setup.py

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -96,7 +96,6 @@ setup(
         'Programming Language :: Python :: 3.9',
         'Programming Language :: Python :: 3.10',
         'Programming Language :: Python :: 3.11',
-        'Programming Language :: Python :: 3.12',
         'Topic :: Internet',
         'Topic :: Software Development :: Libraries',
         'Topic :: Software Development :: Libraries :: Python Modules',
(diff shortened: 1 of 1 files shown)


No767:
Is 3.12 supported on master? I know 3.11 is, but unsure if 3.12 is supported or not

tuna2134:
> Is 3.12 supported on master? I know 3.11 is, but unsure if 3.12 is supported or not

3.12 will release almost, so I added

No767:
> > Is 3.12 supported on master? I know 3.11 is, but unsure if 3.12 is supported or not
> 
> 3.12 will release almost, so I added 

None of the workflows use 3.12 right now nor have I tested 3.12 on the master branch. Unless you can prove that 3.12 indeed works with no issues and everything works as intended, I would highly suggest not marking 3.12 support since it would be misleading

tuna2134:
If 3.12 will released, we can merge this
I am not recommend to merge this.

No767:
Fair enough

tuna2134:
Well, tomorrow, 3.12 is going to release.

No767:
Note that aiohttp doesn't build on 3.12, so we have to wait until aiohttp gets support for 3.12 before testing it

tuna2134:
I see

tuna2134:
I removed 3.12

PR #9599, opened 2023-10-03
add Class in enums , utils files

pondsan1412:
in discord.py==2.3.2 missing a lot class of used for app_commands due cannot import  'appcommandoptiontype' from 'discord.enums' and others

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/enums.py
+++ b/enums.py
@@ -1,0 +1,1025 @@
+"""
+The MIT License (MIT)
+
+Copyright (c) 2015-2021 Rapptz
+Copyright (c) 2021-present Pycord Development
+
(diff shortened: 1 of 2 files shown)


Rapptz:
These files already exist (the proper ones, not the ones you have given). See [here](https://github.com/Rapptz/discord.py/blob/master/discord/utils.py) and [here](https://github.com/Rapptz/discord.py/blob/master/discord/enums.py).

Your issue is that you have a fork installed alongside with the library which is causing conflicts. Uninstall other libraries if you want this to work.

PR #9600, opened 2023-10-05
[Message] New CDN Attachments parameters, new class and new method

Developer-Anony:
## Summary
This PR adds a new class for the CDN urls in message contents. This class just represents that link and its attributes (new one like the "ex", "is" and "hm"), channel id and attachment id.
This also adds a new method to the Message class: `get_url_attachments()`, uses `re` to search all `https://cdn.discordapp.com/attachments/` ocurrences.

If you feel this is unneccesary, it doesn't have to be added, need some fixes tell me, I will fix it as soon as I can.

## Checklist
- [ ] This PR fixes an issue.


Developer-Anony:
The is_premium, recorder things are from an old PR so i reseted it also to its master state lol

Developer-Anony:
Forgot to change member.py, going to do it as soon as I can

Rapptz:
Thanks for the PR but on top of having a bunch of unnecessary changes, this type of behaviour feels like something you should do yourself in your application rather than at the library level.

PR #9603, opened 2023-10-07
Add set voice channel status endpoint

Puncher1:
## Summary

This PR adds a new parameter `status` to `VoiceChannel.edit` to set the voice channel status via the endpoint `/channels/{channel_id}/voice-status`

**Note:** The implementation of the events and permission is seperated into a second PR (https://github.com/Rapptz/discord.py/pull/9604) because it can already be merged, which this PR cannot due to the experiment lock.

Related PR: https://github.com/discord/discord-api-docs/pull/6400

## Checklist

- [x] If code changes were made then they have been tested.
    - [x] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.
- [x] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

PR #9604, opened 2023-10-07
Add voice channel status events and permission

Puncher1:
## Summary
**Work in Progress**

This PR adds the events and permission for the voice channel status.

**Note:** The implementation of the endpoint to set the voice channel status is seperated into a second PR (https://github.com/Rapptz/discord.py/pull/9603) because it cannot be merged due to the experiment lock, which this PR already can.

Related PR: https://github.com/discord/discord-api-docs/pull/6398/

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -2839,6 +2839,40 @@ of :class:`enum.Enum`.
 
         .. versionadded:: 2.4
 
+    .. attribute:: voice_channel_status_update
+
+        The status of a voice channel was updated.
+
+        When this is the action, the type of :attr:`~AuditLogEntry.target` is
+        a :class:`VoiceChannel`.
+
+        When this is the action, the type of :attr:`~AuditLogEntry.extra` is
+        set to an unspecified proxy object with 2 attributes:
+
+        - ``status``: The status of the voice channel.
+        - ``channel``: The channel of which the status was updated.
+
+        When this is the action, :attr:`AuditLogEntry.changes` is empty.
+
+        .. versionadded:: 2.4
+
+    .. attribute:: voice_channel_status_delete
+
+        The status of a voice channel was deleted.
+
+        When this is the action, the type of :attr:`~AuditLogEntry.target` is
+        a :class:`VoiceChannel`.
+
+        When this is the action, the type of :attr:`~AuditLogEntry.extra` is
+        set to an unspecified proxy object with 2 attributes:
+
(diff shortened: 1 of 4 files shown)


Puncher1:
I used the constructor instead of an `_update` method to assign the status because a channel update would overwrite `self.status` to `None`. In the client however, if you edit the channel the status gets reset which I assume is a UI bug (see this comment: https://github.com/discord/discord-api-docs/pull/6398#issuecomment-1752029443)

docs/api.rst
```diff
@@ -370,6 +370,19 @@ Channels
...
+    Called whenever the status of a voice channel has changed.
+
+    .. versionadded:: 2.4
+
+    :param channel: The channel whose status has changed.
+    :type channel: :class:`VoiceChannel`
+    :param before: The status before the update.
+    :type before: Optional[:class:`str`]
+    :param after: The status after the update.
+    :type after: Optional[:class:`str`]
```
Rapptz, line 384: I think instead of an event like this we should probably make it a raw event instead with the past state potentially there, maybe as `MISSING` to differentiate it from `None` if it wasn't found at all.

Rapptz (changes requested):
Looks mostly okay, however not a big fan on the event at the moment. We're kind of moving away from this style of event but it was too late to commit to it in v2 due to churn.

docs/api.rst
Puncher1: So a raw model that holds e.g. `status` and `old_status` or or more like the raw thread update event with the data payload as an attribute? Why not using a regular event with before and after status + a raw event?

Rapptz: This event isn't worth having multiple. It's a waste.

I suspect you can do a raw model + `cached_status` for the old one, similar to `cached_message` in the raw message update event.

Rapptz (commented):
This PR is fine, but Discord officially says this functionality isn't stable. I don't really know why, considering it's been over a year. That aside, the version strings need to move to the new version target.

Puncher1:
> This PR is fine, but Discord officially says this functionality isn't stable. I don't really know why, considering it's been over a year. That aside, the version strings need to move to the new version target.

It's also not clear what happens regarding the hang status. But I make it merge-ready anyway.

codeofandrin:
Original PR has been closed
> Closing this as abandoned.

PR #9610, opened 2023-10-15
Implement editing application info

Puncher1:
## Summary

This PR implements the ability to edit the application info. This also adds 2 parameters to `AppInfo` which are needed to not interfere with the `edit` method.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/appinfo.py
+++ b/discord/appinfo.py
@@ -30,8 +30,11 @@ from . import utils
 from .asset import Asset
 from .flags import ApplicationFlags
 from .permissions import Permissions
+from .utils import MISSING
 
 if TYPE_CHECKING:
+    from typing import Dict, Any
+
     from .guild import Guild
     from .types.appinfo import (
         AppInfo as AppInfoPayload,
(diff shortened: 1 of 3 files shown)


discord/appinfo.py
```diff
@@ -232,6 +248,142 @@ def flags(self) -> ApplicationFlags:
...
+                payload['cover_image'] = cover_image
+            else:
+                payload['cover_image'] = utils._bytes_to_base64_data(cover_image)
+
+        if interactions_endpoint_url is not MISSING:
+            payload['interactions_endpoint_url'] = interactions_endpoint_url
+
+        if tags is not MISSING:
+            payload['tags'] = tags
+        print(payload)
```
Rapptz, line 383:
```suggestion

```

discord/appinfo.py
```diff
@@ -232,6 +248,142 @@ def flags(self) -> ApplicationFlags:
...
+    ) -> AppInfo:
+        r"""|coro|
+
+        Edits the application info.
+
+        .. versionadded:: 2.4
+
+        .. note::
+
+            All parameters can be ``None`` to remove the respective one.
```
Rapptz, line 274:
```suggestion
        Edits the application info.

        .. versionadded:: 2.4
```

It's better to say this in every parameter's documentation. I know it's repetitive but it's harder to miss, plus doesn't break when a new parameter is added that isn't optional.

Rapptz (changes requested):

discord/appinfo.py
Puncher1: Debug leftover...

Rapptz (approved):
Thanks.

PR #9611, opened 2023-10-16
Implement Guild Shop

Developer-Anony:
## Summary

This PR adds things for the guild shop function, that hasn't been out to all guilds but already in some.
The main route for this is `/channels/:guildId/shop`, this route also adds the guild subscriptions page in the route of `/channels/:guildId/role-subscriptions/guild-groups`

This PR adds methods to the HTTPClient class, and a new method in `~.Guild`. Also new classes such as GuildShop, also adds new types and enums

This is a draft PR as Discord hasn't fully finished this, and I am still waiting for a valid JSON response so I can create the needed TypedDicts lol

## Checklist

- [ ] This PR fixes an issue.


Developer-Anony:
Also, if there are some files that I didn't mention that are changed it is because of my IDE, that formats the code I like, copy-paste or git clone to make it more "beatuiful" ¯\_(ツ)_/¯

AbstractUmbra:
> Also, if there are some files that I didn't mention that are changed it is because of my IDE, that formats the code I like, copy-paste or git clone to make it more "beatuiful" ¯_(ツ)_/¯

Typically you should disable these tools on repositories that don't adhere to them already. As it stands this PR won't be accepted due to the huge and unnecessary changes it creates.
For instance, the entire `member.py` file has been changed entirely and this will change the blame for no benefit. I'd suggest resetting the files to their original state and only making your necessary changes.

Developer-Anony:
> > Also, if there are some files that I didn't mention that are changed it is because of my IDE, that formats the code I like, copy-paste or git clone to make it more "beatuiful" ¯_(ツ)_/¯
> 
> Typically you should disable these tools on repositories that don't adhere to them already. As it stands this PR won't be accepted due to the huge and unnecessary changes it creates. For instance, the entire `member.py` file has been changed entirely and this will change the blame for no benefit. I'd suggest resetting the files to their original state and only making your necessary changes.

Ye, I was planning to "reset" the file to its original state, but I am a bit busy so I cannot right now

Developer-Anony:
Also, this is not another comment of the text formatter in my IDE...
The products in the guild shops follow this routes: `/channels/:guildId/shop/:productId`, but none of the servers that have enabled the Guild Shop feature have any products (because it won't load until the experiment has been fully rolled).
So I won't create its class / object until I have the Guild Shop returning JSON and the products JSON

Developer-Anony:
> On top of what is said, where is this documented on the discord docs? Or the pr for the docs?

At the moment there is no documentation because, as I said, it is still a rolling experiment.
I know the routes are correct because I am in some guilds with this experiment, that is why I say that I don't know the returning object / payload.

discord/role.py
```diff
@@ -66,41 +66,48 @@ class RoleTags:
...
-        self.subscription_listing_id: Optional[int] = _get_as_snowflake(data, 'subscription_listing_id')
+        self.boost_listing_id: Optional[int] = _get_as_snowflake(data, 'subscription_listing_id')
+        self.subscription_listing_id: Optional[int] = self.boost_listing_id
 
         # NOTE: The API returns "null" for this if it's valid, which corresponds to None.
         # This is different from other fields where "null" means "not there".
         # So in this case, a value of None is the same as True.
         # Which means we would need a different sentinel.
-        self._premium_subscriber: bool = data.get('premium_subscriber', MISSING) is None
+        self._nitro_subscriber: bool = data.get('premium_subscriber', MISSING) is None
```
AbstractUmbra, line 96: Why was this changed?

No767:
> > On top of what is said, where is this documented on the discord docs? Or the pr for the docs?
> 
> At the moment there is no documentation because, as I said, it is still a rolling experiment.
> I know the routes are correct because I am in some guilds with this experiment, and the shop page won't load until, I guess, the experiment is fully rolled out, that is why I say that I don't know the returning object / payload.

I wouldn't depend on undocumented experiments since they could change at any moment's notice

AbstractUmbra (changes requested):
Aside from the rather unfortunate formatting tweaks that made this PR hard to review, I did notice some attribute changes within classes that seemed irrelevant to the PR?

discord/role.py
Developer-Anony: Oops, forgot to change that from a old PR

Developer-Anony: It is difficult when editing files in mobile git lmao, going to change it asap

Developer-Anony:
> > > On top of what is said, where is this documented on the discord docs? Or the pr for the docs?
> > 
> > 
> > At the moment there is no documentation because, as I said, it is still a rolling experiment.
> > I know the routes are correct because I am in some guilds with this experiment, and the shop page won't load until, I guess, the experiment is fully rolled out, that is why I say that I don't know the returning object / payload.
> 
> I wouldn't depend on undocumented experiments since they could change at any moment's notice

Yep, I know, that is why I will check on it every couple days until the experiment is fully rolled out.
The route the Guild Shop follows right now it is in this attachment (i took it myself): https://cdn.discordapp.com/attachments/1097158421540773910/1163516257811845190/sever_shop_route.png?ex=653fdbf4&is=652d66f4&hm=ec5cb163b925e64863a887d4105d61e62d829bae7459d30bacf30edf90ed82e8&

No767:
> > > > On top of what is said, where is this documented on the discord docs? Or the pr for the docs?
> > > 
> > > 
> > > At the moment there is no documentation because, as I said, it is still a rolling experiment.
> > > I know the routes are correct because I am in some guilds with this experiment, and the shop page won't load until, I guess, the experiment is fully rolled out, that is why I say that I don't know the returning object / payload.
> > 
> > 
> > I wouldn't depend on undocumented experiments since they could change at any moment's notice
> 
> Yep, I know, that is why I will check on it every couple days until the experiment is fully rolled out. The route the Guild Shop follows right now it is in this attachment (i took it myself): https://cdn.discordapp.com/attachments/1097158421540773910/1163516257811845190/sever_shop_route.png?ex=653fdbf4&is=652d66f4&hm=ec5cb163b925e64863a887d4105d61e62d829bae7459d30bacf30edf90ed82e8&

As a side note, depending on the client path is extremely unreliable. I just tested the supposed path against my own server (which does not have this feat enabled) and it broke as expected. I'd just recommend waiting until Discord decides to document this feature first.

discord/guild.py
```diff
@@ -48,6 +48,8 @@
 )
 import warnings
```
bijij, line 50: discord.types references need to be inside `if TYPE_CHECKING` blocks. This should also probably be aliased as `GuildShopPayload` to avoid a clash with the (currently non-existent) discord.py representation.

discord/guild.py
```diff
@@ -764,6 +766,22 @@ def get_emoji(self, emoji_id: int, /) -> Optional[Emoji]:
...
+
+        Returns
+        -------
+        Optional[:class:`GuildShop`]
+            The returned Guild Shop or ``None`` if no shop is enabled.
+        """
+
+        shop = self._state.http.get_guild_shop(self.id)
+        if shop:
+            return shop # type: ignore
```
bijij, line 783: Couple things:
(Note this assumes the issues outlined above are fixed, as in reality at the moment `shop` is a coroutine)

- This shouldn't directly return the dictionary parsed from the response.
  - Instead it should be converting the response object into a class representation.
  - The return type annotation should be updated to reflect this.
 -  The `type: ignore` comment should be removed.

discord/http.py
```diff
@@ -942,6 +942,18 @@ def get_reaction_users(
         if after:
             params['after'] = after
         return self.request(r, params=params)
+    
+    def get_guild_shop(
+        self,
+        guild_id: Snowflake
+    ) -> Response[Optional[guild.GuildShop]]:
```
bijij, line 949:
```suggestion
    ) -> Response[guild.GuildShop]:
```
I find it very unlikely that the json payload from a GET request will be a literal `null` in the event that the guild shop does not exist.

Much more likely you're gonna get some kind of error response which should be raised as an exception.

discord/guild.py
```diff
@@ -764,6 +766,22 @@ def get_emoji(self, emoji_id: int, /) -> Optional[Emoji]:
         if emoji and emoji.guild == self:
             return emoji
         return None
+    
+    def get_guild_shop(self) -> Optional[GuildShop]:
```
bijij, line 770:
```suggestion
    async def guild_shop(self) -> GuildShop:
```


Methods which make a HTTP request for data shouldn't use the `get_` prefix.

Also, I'm assuming discord returns a 404 when there is no guild shop, you should propagate that rather than wrapping returning an `Optional` type.

Also, since this is fetching from the API, this needs to be an `async` function.

discord/guild.py
```diff
@@ -764,6 +766,22 @@ def get_emoji(self, emoji_id: int, /) -> Optional[Emoji]:
...
+
+        .. versionadded:: 2.4
+
+        Returns
+        -------
+        Optional[:class:`GuildShop`]
+            The returned Guild Shop or ``None`` if no shop is enabled.
+        """
+
+        shop = self._state.http.get_guild_shop(self.id)
```
bijij, line 781:
```suggestion
        shop = await self._state.http.get_guild_shop(self.id)
```
This is a HTTP request call, it needs to be awaited. Otherwise `shop`'s value will just be a coroutine.

bijij (changes requested):
I'm aware that this pull request is currently a draft.
But I've left some comments for your awareness.

I haven't reviewed any of the files which should have changed reverted. (member.py, message.py, role.py)

discord/guild.py
Developer-Anony: Yep, I was going to check that but in mobile when editing big files it is a bit buggy

discord/guild.py
Developer-Anony: The type ignore is just to prevent MyPy syntax highlighting

Developer-Anony:
Thanks everyone for all the file reviews 😀!
I will try to push the changes soon today

discord/guild.py
Developer-Anony: Changes as you said.

discord/guild.py
Developer-Anony: Ok, changed the name from get_guild_shop to guild_shop and made it asynchronous.

discord/guild.py
Developer-Anony: Awaited the http call (forgot it was a coroutine, that is why I didn't await it).

discord/guild.py
Developer-Anony: Fixed that also in the new commits

discord/http.py
Developer-Anony: Yep, I also changed that.

Developer-Anony:
Well everyone, I got the returning payload from discord when accessing both shop and products, going to push changes asap

Developer-Anony:
This is just for everyone awareness
Guild Subscriptions are now also inside Guild Shop, so I am going to add methods to them to return guild Subscriptions

Developer-Anony:
Oh, yeah, forgot to update __all__ in some files...

Developer-Anony:
Routes have changes and I have updated them in the code.
Also, these endpoints are disabled for bots until further notice, so as soon as they are reactivated for bots I will push more commits.

Developer-Anony:
I will make no more commits until Fiscord disables the experiment lock in the endpoints this commit uses.

Developer-Anony:
Discord has blocked the access to this endpoint for bots, this means that this PR is no longer useful.
I am closing it. But, it would be weird seeing a bot create shop products lol.
Anyways, I will try to find some endpoints that bots can use to __see__ guild products

PR #9651, opened 2023-11-18
Fix silent errors in tasks

AsyncFor:
## Summary
By default, the logger in discord.ext.tasks is not configured. This commit ensures that errors are not passed silently.

## Checklist

- [x] This PR fixes an issue.


Rapptz:
Setting up the logger is the responsibility of the end user, not the library. Usually this is done by either calling this routine explicitly (as it's meant to be called with customisable parameters) or using Client.run.

PR #9658, opened 2023-11-27
Add Client.get_user_named

Soheab:
## Summary

This PR adds a method called `get_user_named` to Client as an equivalent to `Guild.get_member_named`.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -1089,7 +1089,7 @@ class Client:
         if discriminator == '0' or (len(discriminator) == 4 and discriminator.isdigit()):
             pred = lambda u: u.name == username and u.discriminator == discriminator
         else:
-            pred = lambda u: u.global_name == name or u.name == name
+            pred = lambda u: u.name == name or u.global_name == name
 
         return utils.find(pred, users)
 


discord/client.py
```diff
@@ -1048,6 +1048,51 @@ def get_user(self, id: int, /) -> Optional[User]:
...
+
+
+        Parameters
+        -----------
+        name: :class:`str`
+            The name of the user to lookup.
+
+        Returns
+        --------
+        Optional[:class:`User`]
```
lmaotrigine, line 1076:
```suggestion
        Optional[:class:`~discord.User`]
```

lmaotrigine (changes requested):
Docs build is failing

Rapptz:
I'm not really sure on the usefulness of this method on whether it's worth adding to the `Client` class, especially since discriminators aren't really a thing anymore.

Soheab:
I kept the discriminator impl because of bots and "dead" users I guess. It can be removed all at once anyways.

But not sure where else it could go.. in utils maybe?

PR #9667, opened 2023-12-03
Updated typo for automod_rule_trigger which is actually automod_rule_trigger_type

CastelRune:
…trigger_type

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -2825,7 +2825,7 @@ of :class:`enum.Enum`.
         set to an unspecified proxy object with 3 attributes:
 
         - ``automod_rule_name``: The name of the automod rule that was triggered.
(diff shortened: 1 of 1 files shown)


AbstractUmbra:
As was already discussed [in the support server](https://discord.gg/r3sSKJJ) and the [help thread](https://canary.discord.com/channels/336642139381301249/1180911005765742662/1180992676850585701) the naming for this attribute is corrected.

The reasoning behind the name was discussed prior to implementation and was agreed that it being simply named `automod_rule_trigger` would imply that it would be a `AutoModRuleTrigger` type, hence the `_type` suffix as is named in the [official API documentation](https://discord.com/developers/docs/resources/auto-moderation#auto-moderation-rule-object-trigger-types) for this type.

PR #9673, opened 2023-12-09
[Typing] Change value being set to match `Future[None]` annotation for `tasks.SleepHandle.future`

Sachaa-Thanasius:
Switch `True` to `None` within in `self.loop.call_later(..., self.future.set_result, True)`.

## Summary

Randomly noticed that in this [typeshed issue](https://github.com/python/typeshed/pull/11015#issuecomment-1810496592) that's integrating `TypeVarTuple` in some spots in asyncio, discord.py popped up in mypy_primer due to having the `SleepHandle.future` expect to be set to `None`, but later setting it to `True` instead. This corrects the value being set to match the `Future[None]` annotation.

(This hasn't really been an issue before now, so . . . almost feels like a nothing change. Figured it wouldn't hurt though.)

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/tasks/__init__.py
+++ b/discord/ext/tasks/__init__.py
@@ -111,12 +111,12 @@ class SleepHandle:
         self.loop: asyncio.AbstractEventLoop = loop
         self.future: asyncio.Future[None] = loop.create_future()
         relative_delta = discord.utils.compute_timedelta(dt)
-        self.handle = loop.call_later(relative_delta, self.future.set_result, True)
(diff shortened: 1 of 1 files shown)


Sachaa-Thanasius:
Just did a small test and this works, but taking a second glance at that earlier mypy_primer result, I don't know whether the current change is right, or alternatively `Future[None]` should be change to `Future[bool]`.

Rapptz:
Thanks! This was a rather amusing oversight.

PR #9682, opened 2023-12-12
Add type attribute to Invite

Puncher1:
## Summary

This PR adds the `type` attribute to `Invite`.
As mentioned in the related PR this change got rejected multiple times. But I don't see why. Is there a specific reason it shouldn't be implemented in the library, that I don't know? 

Related PR: https://github.com/discord/discord-api-docs/pull/6568

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
The reason why I believe this wasn't documented is because the various types of invites were all experiments that didn't pan out. Friend invites died, and I don't think group invites are a thing anymore.

Puncher1:
Alright. Let's wait though what the ddevs say to it, unless you already know more than I do about whether something like this will be documented in the future.

Puncher1:
The related PR on Discord's side has been merged.

PR #9687, opened 2023-12-16
make bot.owner_ids optional

StarrFox:
## Summary

bot.owner_ids is typed as Optional[Collection[int]] but it defaults to an empty set when there is no team and not None

this pr makes it default to None

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/bot.py
+++ b/discord/ext/commands/bot.py
@@ -183,7 +183,7 @@ class BotBase(GroupMixin[None]):
         self._help_command: Optional[HelpCommand] = None
         self.description: str = inspect.cleandoc(description) if description else ''
         self.owner_id: Optional[int] = options.get('owner_id')
-        self.owner_ids: Optional[Collection[int]] = options.get('owner_ids', set())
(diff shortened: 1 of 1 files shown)


Rapptz:
I see no real reason for this change. Users can provide an explicit `None` themselves and the code works fine as-is, the default the library provides doesn't really seem to matter, especially since the way you're meant to check for `owner_ids` existence is the implicit bool conversion, i.e. `if bot.owner_ids:` to catch the empty set case that exists regardless of the library.

StarrFox:
should the type instead be changed to not be Optional?

Rapptz:
The type is fine, a user can explicitly pass `owner_ids=None` and it'll work just fine. The library provided default doesn't have to be `None` for the type hint to be correct.

StarrFox:
are you sure it isn't confusing in the case were they don't have a team and expect owner_ids to be None by default and use an identity check (owner_ids is None)? 

I don't think anything is loss by using None as the default and it saves people having to find out later that it isn't actually None in this case

Rapptz:
That code would be buggy because it doesn't account for the empty set case as mentioned earlier.

PR #9699, opened 2024-01-02
Fix move_to related voice state bugs

imayhaveborkedit:
## Summary
1. Skip trying to move the bot to its current channel.
2. Fix `voice_client.channel` not being set when the bot moves itself.
3. Fixes an unrelated incorrect type

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -827,7 +827,7 @@ class DiscordVoiceWebSocket:
         self.loop: asyncio.AbstractEventLoop = loop
         self._keep_alive: Optional[VoiceKeepAliveHandler] = None
         self._close_code: Optional[int] = None
-        self.secret_key: Optional[str] = None
+        self.secret_key: Optional[List[int]] = None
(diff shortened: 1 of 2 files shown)


discord/gateway.py
```diff
@@ -827,7 +827,7 @@ def __init__(
         self.loop: asyncio.AbstractEventLoop = loop
         self._keep_alive: Optional[VoiceKeepAliveHandler] = None
         self._close_code: Optional[int] = None
-        self.secret_key: Optional[str] = None
+        self.secret_key: Optional[List[int]] = None
```
Rapptz, line 830: This is most likely `bytes` no?

imayhaveborkedit: In the one place it's actually used (making the SecretBox) it gets cast to bytes on use.  I could refactor everything to change it to bytes on arrival but it was just this hint that was inconsistent with `List[int]`.

PR #9702, opened 2024-01-07
Add disabled property to `DynamicItem` class

Snawe:
## Summary

Add `disabled` property to `DynamicItem` item.

## Checklist

### DIFF
--- a/discord/ui/dynamic.py
+++ b/discord/ui/dynamic.py
@@ -76,6 +76,8 @@ class DynamicItem(Generic[BaseT], Item['View']):
         like to control the relative positioning of the row then passing an index is advised.
         For example, row=1 will show up before row=2. Defaults to ``None``, which is automatic
         ordering. The row number must be between 0 and 4 (i.e. zero indexed).
+    disabled: Optional[:class:`bool`]
+        Whether the button is disabled or not. Defaults to :class:`.Button` default value.
 
     Attributes
     -----------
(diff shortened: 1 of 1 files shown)


discord/ui/dynamic.py
```diff
@@ -76,6 +76,8 @@ class DynamicItem(Generic[BaseT], Item['View']):
         like to control the relative positioning of the row then passing an index is advised.
         For example, row=1 will show up before row=2. Defaults to ``None``, which is automatic
         ordering. The row number must be between 0 and 4 (i.e. zero indexed).
+    disabled: :class:`bool`
```
diceroll123, line 79:
```suggestion
    disabled: Optional[:class:`bool`]
```

discord/ui/dynamic.py
```diff
@@ -165,6 +170,14 @@ def row(self) -> Optional[int]:
     def row(self, value: Optional[int]) -> None:
         self.item.row = value
 
+    @property
+    def disabled(self) -> bool:
```
diceroll123, line 174:
```suggestion
    def disabled(self) -> Optional[bool]:
```

Snawe: This one is not `Optional` since it comes from the `item` itself. There `False` is default.

Rapptz:
Thanks for the PR, but `disabled` isn't there on purpose since it's not actually part of the `Item` interface. I'm unsure if I want to add it to it since in the future there might be items that aren't capable of being disabled (selects were like this originally, TextInput remains this way, etc.).

Snawe:
Ok, that makes sense.
Do you have a suggested way how to disable all buttons of a view without it? I came across it because I have a view with x dynamic buttons.
Using
```py
        for button in self.view.children:
            button.disabled = True
```

I saw that the button pressed is an instance of `DynamicButton` (lacking a `disable` property), while all other items in the list are an instance of `Button`.

I think I would go with
```py
        for button in self.view.children:
            button.disabled = True
         self.item.disabled = True
```
as an alternative, but it just feels a bit weired to me :/

Rapptz:
Generally a view with dynamic items has *all of them* as dynamic items so you can just do `child.item.disabled = True` in the loop. Otherwise you can check using `isinstance` in the loop.

Snawe:
👍 might have missed that. Will check it when I am back home and close this PR afterwards! Thx

PR #9710, opened 2024-01-19
Fix Member.get_role returning None for the @everyone role

Soheab:
## Summary

This was noticed by "snawe" in their [help post](https://discord.com/channels/336642139381301249/1197947439064297575) in the discord.py server.

The problem here is that the library currently does not store the guild (`@everyone` role) ID in `Member._roles` thus doing `.has(guild_id)` (used in `Member.get_role`) return `None`. This is fixed by adding an if-statement in `Member.get_role` that checks whether the input equals to the member's guild ID and returns `Guild.default_role` instead.

I'm not sure whether the `versionchanged` is necessary here.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/member.py
+++ b/discord/member.py
@@ -1099,6 +1099,8 @@ class Member(discord.abc.Messageable, _UserTag):
         """Returns a role with the given ID from roles which the member has.
 
         .. versionadded:: 2.0
+        .. versionchanged:: 2.4
+            This can now return :attr:`Guild.default_role` if ``role_id`` is the same as the member's guild ID.
 
         Parameters
         -----------
@@ -1110,6 +1112,9 @@ class Member(discord.abc.Messageable, _UserTag):
         Optional[:class:`Role`]
             The role or ``None`` if not found in the member's roles.
         """
+        if role_id == self.guild.id:
+            return self.guild.default_role
+
         return self.guild.get_role(role_id) if self._roles.has(role_id) else None
 
     def is_timed_out(self) -> bool:


Snawe:
@Rapptz why is that intended? I understand that the `@everyone` role id is the same as the guild id. But consider the following usecase:

* A user can set a given role to allow X and passes `@everyone` as the role
* I now take the id and save it in the database
* If now a user wants to do X, I query the role id from the database and do a `Member.get_role()`.
* If it returns `None`, I print "you do not have the required role", else the user can do X.

If now on `@everyone` `get_role()` returns None you need to add an additional if to achieve this. Like:
```py
    if not member.get_role(id):
        if not guild.default_role.id == id:
            return
```
I don't see why this is the intended way to do this. Or is there another obvious way that I miss?

Soheab:
> * A user can set a given role to allow X and passes `@everyone` as the role
> * I now take the id and save it in the database

You could check whether a user is trying to pass the default role before adding it to your db and error out because like *everyone* has the *everyone* role.
Note that [`Guild.get_role`](https://discordpy.readthedocs.io/en/stable/api.html#discord.Guild.get_role) *does* return the default_role so that can be one way to check it.

Snawe:
Sure, I could not allow setting `@everyone` as a role, but I actually want to distinguish between no role set and `@everyone` set.
Sure, I could also set an arbitrary value (example negative), but again. 2 more ifs. One when the user tries to set the value, one when I check the permission.

> Even if `Member.get_role` returned it, trying to add it doesn't make any sense and will result in an error. You need to check it either way.

I don't want to add the role to the user. I basically want to check
* if no role is set in DB => only administrators are allowed
* if role is set, all users with given role are allowed to perform an action.

Rapptz:
From an API perspective, the member doesn't actually have that role in the role list and that's what you're seeing exposed here. The high level `roles` property manually inserts it into the list as a means of ease of use. Fundamentally the `@everyone` role is special. Whether that `if` statement should be on your end or my end is mostly a matter of style, I don't like paying the extra branch for it.

PR #9720, opened 2024-02-04
Add ConversionCancelled command error

### LINKED ISSUE #9719: Provide a standard way to cancel a `Converter` conversion
### Summary

We need a way to "cancel", not "fail" the conversion.

discord.ext.commands

### The Problem

Some bots provide a selection menu of targets when they cannot find an exact match for the command argument.
e.g. A search result of guild members' nicknames when nobody has the exact same name as the given argument string.

It makes sense to put this feature in the custom `Converter` to make it reusable in many commands.
And since it requires user interaction, the conversion may time out or be cancelled by the user.

But currently, a `Converter` can only succeed or fail.
There is no way to "cancel" the conversion, to cleanly "give up" the conversion and the invocation of the command.

### The Ideal Solution

Introduce a new `CommandError` subclass, `ConversionCancelled`.
Custom converters can raise this exception to indicate intentional cancellation of the conversion.
Similar to `CommandNotFound`, it is mostly expected to be suppressed by the `on_command_error()` handler.

### The Current Solution

_No response_

### Additional Context

_No response_

WieeRd:
## Summary

This introduces a new `CommandError` subclass, `ConversionCancelled`.

Fixes #9719

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/errors.py
+++ b/discord/ext/commands/errors.py
@@ -146,6 +146,25 @@ class ConversionError(CommandError):
         self.original: Exception = original
 
 
+class ConversionCancelled(CommandError):
+    """Exception raised when a Converter class intentionally cancels the conversion.
+
+    This indicates that the conversion was *cancelled* under expected conditions,
+    rather than *failed* due to an invalid input or an error in the conversion logic.
+    This is useful for handling user interventions and timeouts in custom converters.
+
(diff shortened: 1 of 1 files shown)


WieeRd:
wontfix'ed

PR #9723, opened 2024-02-07
updates avatar warning docs to show that animated pfps now work.

JDJGInc:
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/user.py
+++ b/discord/user.py
@@ -425,7 +425,7 @@ class ClientUser(BaseUser):
         avatar: Optional[:class:`bytes`]
             A :term:`py:bytes-like object` representing the image to upload.
             Could be ``None`` to denote no avatar.
-            Only image formats supported for uploading is JPEG, PNG, GIF, WEBP.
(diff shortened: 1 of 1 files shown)


JDJGInc:
I am not sure about the warning but I think the docs updating may be important.

Rapptz:
You can just remove the line and move it down to the avatar parameter docs.

JDJGInc:
Alright.

Rapptz:
That's fine but it should be `are` instead of `is` (the original had the same grammar mistake)

JDJGInc:
Would you look to make the order like:

``Only image formats supported for uploading are JPEG, GIF, PNG, and WEBP.``

or keep the order as is?

and then make it

``Only image formats supported for uploading are JPEG, PNG, GIF, and WEBP.``

Rapptz:
The order doesn't matter

JDJGInc:
image
alright here's it now then:

PR #9739, opened 2024-02-22
Add proxy variable

jhqwqmc:
## Summary

Add the use of the proxy variable to the from_url function of the webhook class.

During use, I found that the from_url function did not provide a proxy parameter to pass in. My area requires a proxy to access discord, so I added this parameter.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/webhook/async_.py
+++ b/discord/webhook/async_.py
@@ -1231,6 +1231,7 @@ class Webhook(BaseWebhook):
         cls,
         url: str,
         *,
+        proxy: Optional[str] = None,
         session: aiohttp.ClientSession = MISSING,
         client: Client = MISSING,
         bot_token: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


No767:
If you were to ask me, this isn't needed. I would wonder *why* even support proxies for webhooks?

Rapptz:
The reason for this is because you can set the proxies on the `ClientSession` you create instead.

jhqwqmc:
`ClientSession` does not provide proxy variables. You need to pass proxy variables to Webhook.

jhqwqmc:
Needing a proxy is also a helpless move. Our country has blocked the website discord.com.

Rapptz:
Actually it seems you're right, I forgot this change happened to aiohttp a while back. This PR is insufficient though. It needs to support `proxy_auth` parameter as well *and* be properly documented.

jhqwqmc:
I just added `proxy_auth`

Rapptz:
This is still missing documentation, and most likely needs to run black.

JDJGInc:
Oh wow

PR #9745, opened 2024-03-04
Move most static metadata to pyproject.toml.

Sachaa-Thanasius:
## Summary

It moves most of the library's metadata to pyproject.toml from setup.py, excluding the library version due to its particular type of dynamism and the fact that code would seemingly be required for it regardless of possible alternatives like setuptools_scm.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,83 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools"]
 build-backend = "setuptools.build_meta"
+
+[project]
+name = "discord.py"
+description = "A Python wrapper for the Discord API"
+readme = { file = "README.rst", content-type = "text/x-rst" }
+license = { file = "LICENSE" }
+requires-python = ">=3.8"
+authors = [{ name = "Rapptz" }]
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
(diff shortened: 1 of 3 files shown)


Sachaa-Thanasius:
Hmm. I'm not sure what's causing the issue here, but I'd guess that moving the data somehow . . . changed the impact of manually setting `PYTHONPATH` to the path of the working directory? Running the same tests on my machine works just fine after an editable install, without the `PYTHONPATH="$(pwd)"` portion of the command.

Sachaa-Thanasius:
@Rapptz Any pointers on what's causing these tests to fail? I'm guessing it has to do with the explicit `PYTHONPATH` since `aiohttp` can't be found for importing, and the tests sans `PYTHONPATH` seem to work fine locally.

pyproject.toml
```diff
@@ -1,7 +1,84 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools"]
```
Sachaa-Thanasius, line 2: setuptools automatically installs wheel if needed. Further info is in the first note in this section: https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use

Rapptz (approved):
Thanks. Let's give this a try.

PR #9382, opened 2023-05-01
Make DMChannels and GroupChannels mentionable

PythonCoderAS:
## Summary

This makes DMChannels and GroupChannels mentionable. These do properly show up when mentioned.

SCR-20230501-levz

## Checklist

- [ ] If code changes were made then they have been tested.
    - [x] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.
- [x] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

PR #9756, opened 2024-03-18
Add mention property to DMChannel class

Vioshim:
## Summary

Implements DM Channel mentions as these have been properly reintroduced in the UI's update.

image

## Checklist

- [ ] This PR fixes an issue.


Sherazranjha:
Yy

LeoCx1000:
Duplicte of #9382 - but also (in my opinion) making these mentionable feels pointless since only the DM's recipients can see the mention. It's not a global "take me to this user's DMs" button and could potentially cause confusion.

PR #9759, opened 2024-03-19
Add support for Polls

DA-344:
## Summary

Adds support for Polls.

### Related PRs:
- https://github.com/discord/discord-api-docs/pull/6746

### Additions:
- `Poll`, `PollAnswer` and `PollMedia` objects.
- `PollLayoutType` enum
- `on_poll_vote_add`/`remove` events.
- `poll` attribute to `Message`
- `poll` kwarg in `Messageable.send`

## Checklist

- [ ] This PR fixes an issue.


discord/poll.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def __init__(self, text: str, *, emoji: Union[PartialEmoji, Emoji, str] = MISSING) -> None:
+        self.text: str = text
+        self.emoji: Optional[Union[PartialEmoji, Emoji, str]] = emoji if emoji is not MISSING else None
+
+    @classmethod
+    def _from_data(cls, data: PollAnswerPayload) -> Self:
+        media = data.get('poll_media')
+        text = media.get('text')
+        emoji = media.get('emoji', None)
```
Gobot1234, line 92:
```suggestion
        media = data['poll_media']
```
if media is None other stuff breaks anyway.

DA-344: Yeah, you are right, I committed this change (with some others that are pretty similar like this).

No767:
Is there documentation or an announcement surrounding this feature?

DA-344:
> Is there documentation or an announcement surrounding this feature?

Not yet, but it is getting a pop-up saying you to try it, though you cannot send them (you get a 403: Forbidden).

Rapptz:
Thanks for the PR.

Discord has asked that I do not implement this yet and will give me private documentation when the time comes.

discord/poll.py
```diff
@@ -0,0 +1,644 @@
...
+
+    @classmethod
+    def from_dict(cls, *, data: PollMediaPayload) -> PollMedia:
+        """Returns a new instance of this class from a payload."""
+
+        return cls(text=data['text'], emoji=PartialEmoji.from_dict(data.get('emoji', {})))
+
+    def __iter__(self) -> Union[str, Optional[PollMediaEmoji]]:
+        yield self.text
+        yield self.emoji
```
AbstractUmbra, line 123: Why do we define `__iter__` here?

discord/poll.py
```diff
@@ -0,0 +1,644 @@
...
+# PollDuration = Literal[1, 4, 8, 24, 72, 168]
+# So, I discovered that any time is valid, so there is no need for this to be a strict
+# literal.
+
+
+PollMediaEmoji = Union[PartialEmoji, Emoji, str]
+
+
+
+class PollMedia:
```
AbstractUmbra, line 81: I think this could be a NamedTuple subclass?

discord/poll.py
```diff
@@ -0,0 +1,644 @@
...
+    """Represents a poll answer count.
+
+    This is not meant to be user-constructed but instead obtained by the results in
+    :attr:`Poll.answer_counts`
+
+    Attributes
+    ----------
+    id: :class:`int`
+        The answer ID.
+    me_voted: :class:`bool`
```
AbstractUmbra, line 202: I know this parameter name comes from the API payload but I personally find it the naming a bit lacking. How about `self_vote` or something along those lines?

AbstractUmbra (commented):
Small items to check.

Also if this PR now "works as intended" as has been stated then please un-draft it so that CI can be run.

discord/poll.py
DA-344: Oh, I forgot to remove that! Sorry!

discord/poll.py
DA-344: Yeah, will consider it 🤔

discord/poll.py
DA-344: I thought of it when I first created it, but I didn't subclass it at the end.
Going to change though.

DA-344:
I found an error when adding an emoji to a poll answer. Going to fix it ASAP.

DA-344:
All errors I found with the PR have been fixed.

__Fixed errors__:
- `Message.poll` raising an error
- `PollAnswer.from_params` not allowing emojis
- `Poll._from_data` raising errors even if the duration exceeded some milliseconds

__Breaking changes__ *(for the ones that are using the branch)*:
*None*

__Changes__:
- Fixed type hinting errors
- Modified internal parameters (i.e.: PollAnswer)
- `Message.poll` default parameter value is now `None`, this has been done to prevent `AttributeError`s when using `poll.message` and when adding poll answers.

**If you see any errors please tell me**

discord/poll.py
DA-344: This was removed as I said.

discord/poll.py
DA-344: You are right, this is now a NamedTuple.

bijij:
Out of curiosity, since the official docs don't seem to note it, can a poll be sent in a message with additional data such as attchements, embeds, content etc?

or is it some kind of mutually exclisive thing?

DA-344:
> Out of curiosity, since the official docs don't seem to note it, can a poll be sent in a message with additional data such as attchements, embeds, content etc?

No, if you add any of them you get an error. It would be good if you could though.

discord/flags.py
```diff
@@ -1255,6 +1255,55 @@ def auto_moderation_execution(self):
...
+        - :func:`on_raw_poll_vote_add` (both guilds and DMs)
+        - :func:`on_raw_poll_vote_remove` (both guilds and DMs)
+
+        .. versionadded:: 2.4
+        """
+        return (1 << 24) | (1 << 25)
+
+    @flag_value
+    def guild_polls(self):
+        """:class:`bool`: Whether poll related events are enabled.
```
bijij, line 1275: This should note this is specific to guilds.

discord/flags.py
```diff
@@ -1255,6 +1255,55 @@ def auto_moderation_execution(self):
...
+        - :func:`on_raw_poll_vote_add` (only for guilds)
+        - :func:`on_raw_poll_vote_remove` (only for guilds)
+
+        .. versionadded:: 2.4
+        """
+        return 1 << 24
+
+    @flag_value
+    def dm_polls(self):
+        """:class:`bool`: Whether poll related events are enabled.
```
bijij, line 1292: This should note this is specific to DMs.

discord/flags.py
```diff
@@ -1255,6 +1255,55 @@ def auto_moderation_execution(self):
...
+
+        See also :attr:`guild_polls` for guilds or :attr:`polls` for both.
```
bijij, line 1294:
```suggestion
        See also :attr:`guild_polls` and :attr:`polls`.
```

discord/flags.py
```diff
@@ -1255,6 +1255,55 @@ def auto_moderation_execution(self):
...
+
+        .. versionadded:: 2.4
+        """
+        return (1 << 24) | (1 << 25)
+
+    @flag_value
+    def guild_polls(self):
+        """:class:`bool`: Whether poll related events are enabled.
+
+        See also :attr:`dm_polls` for DMs or :attr:`polls` for both.
```
bijij, line 1277:
```suggestion
        See also :attr:`guild_polls` and :attr:`polls`.
```

discord/flags.py
```diff
@@ -1255,6 +1255,55 @@ def auto_moderation_execution(self):
         """
         return 1 << 21
 
+    @flag_value
+    def polls(self):
+        """:class:`bool`: Whether poll related events are enabled.
```
bijij, line 1260: Probably worth noting this is a combination of `guild_polls` and `dm_polls`

discord/poll.py
```diff
@@ -0,0 +1,630 @@
...
+        cls, *, data: Union[PollWithExpiryPayload, FullPollPayload], message: Message, state: ConnectionState
+    ) -> Self:
+        answers = [
+            PollAnswer(data=answer, poll=message.poll, message=message) for answer in data.get('answers')
+        ]  # 'message' will always have the 'poll' attr
+        multiselect = data.get('allow_multiselect', False)
+        layout_type = try_enum(PollLayoutType, data.get('layout_type', 1))
+        question_data = data.get('question')
+        question = question_data.get('text')
+        expiry = datetime.datetime.fromisoformat(data['expiry'])  # If obtained via API, then expiry is set.
```
trevorflahardy, line 416: `utils.parse_time` should be used here

trevorflahardy (commented):
Just some things I noticed whilst browsing. There are more, but they're more concept-related rather than strictly code edits. The fashion in which a `Poll` is both stateful and stateless will need to be addressed differently, as well as how `PollAnswer` is constructed and handled. 👍

discord/http.py
```diff
@@ -154,6 +156,7 @@ def handle_message_parameters(
     thread_name: str = MISSING,
     channel_payload: Dict[str, Any] = MISSING,
     applied_tags: Optional[SnowflakeList] = MISSING,
+    poll: Optional[Poll] = MISSING
```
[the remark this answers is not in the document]
trevorflahardy: > i would disagree, since it makes pyright angry on `await ctx.send('Hello', poll=poll)` where `poll` is `Optional[Poll]`

Ah yes, you're right. Additionally, `abc.Messageable.send` should have `poll: Optional[Poll] = None` and not `poll: Poll = MISSING`, like it has now.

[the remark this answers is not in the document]
trevorflahardy: > i would disagree, since it makes pyright angry on `await ctx.send('Hello', poll=poll)` where `poll` is `Optional[Poll]`

That would go back to the overload for `Context.send`. It's the same reason that you get an error when you do
image

The revision suggestion only addressed `handle_message_parameters`, which denotes that `poll=None` cannot be given to either a `.send` or a `.edit` method.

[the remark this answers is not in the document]
trevorflahardy: You're right, I was thrown off by the type hint for the send coroutines.

`handle_message_parameters` should have `Optional[Poll] = None` and all the send methods should have `poll: Optional[Poll] = None` in their function definition rather than `poll: Poll = MISSING`. In the `handle_message_parameters` block it should have `if poll is not None: ...`

```suggestion
    poll: Optional[Poll] = None
```
```python
def handle_message_parameters(poll: Optional[Poll] = None) -> ...:
    if poll is not None: ...
```

discord/poll.py
```diff
@@ -0,0 +1,686 @@
...
+            for user in data.get('users'):
+                yield User(state=self._state, data=user)
+
+        else:
+            guild = self._message.guild
+
+            for user in data.get('users'):
+                member = guild.get_member(int(user['id']))
+
+                yield member or User(state=self._state, data=user)
```
bijij, line 181: This needs to be reworked entirely, if the limit is >100 we should instead keep hitting the endpoint until it is exhaused, reaction users does the same thing.

Additionally, I believe this should be called `voters`

bijij (changes requested):

discord/types/poll.py
```diff
@@ -0,0 +1,95 @@
...
+    text: str
+    emoji: NotRequired[Optional[PartialEmoji]]
+
+
+class PollAnswer(TypedDict):
+    poll_media: PollMedia
+
+
+class PollAnswerWithID(PollAnswer):
+    answer_id: Snowflake
```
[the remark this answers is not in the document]
DA-344: No, thanks for reminding me of it! Fixed in latest commits.

discord/webhook/async_.py
```diff
@@ -1732,6 +1749,15 @@ async def send(
 
             .. versionadded:: 2.4
 
+        poll: :class:`Poll`
```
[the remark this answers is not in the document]
DA-344: No, I genuinely forgot about webhooks, thanks for reminding me lol

discord/poll.py
```diff
@@ -0,0 +1,685 @@
...
+
+        super().__init__(id=int(data['id']), message=message, state=state)
+
+    def __repr__(self) -> str:
+        return f'<PollAnswerCount id={self.id} resolved={self.resolved!r}> self_voted={self.self_voted}'
+
+    @property
+    def original_message(self) -> Message:
+        """:class:`Message`: Returns the original message the poll of this answer is in."""
+        return self._message # type: ignore # Here will always be a value
```
bijij, line 219:
```suggestion
        return self._message  # type: ignore # Here will always be a value
```

discord/poll.py
```diff
@@ -0,0 +1,685 @@
...
+        id: int,
+        text: str,
+        emoji: Optional[PollMediaEmoji] = None,
+        *,
+        poll: Poll,
+    ) -> Self:
+        poll_media: PollMediaPayload = {'text': text}
+        if emoji:
+            if isinstance(emoji, Emoji):
+                poll_media['emoji'] = {'name': emoji.name} # type: ignore
```
bijij, line 285:
```suggestion
                poll_media['emoji'] = {'name': emoji.name}  # type: ignore
```

discord/poll.py
```diff
@@ -0,0 +1,685 @@
...
+
+                if emoji.id:
+                    poll_media['emoji']['id'] = emoji.id # type: ignore
```
bijij, line 288:
```suggestion
                    poll_media['emoji']['id'] = emoji.id  # type: ignore
```

discord/poll.py
```diff
@@ -0,0 +1,685 @@
...
+            elif isinstance(emoji, PartialEmoji):
+                poll_media['emoji'] = emoji.to_dict()
+            else:
+                poll_media['emoji'] = {'name': str(emoji)} # type: ignore
```
bijij, line 292:
```suggestion
                poll_media['emoji'] = {'name': str(emoji)}  # type: ignore
```

discord/poll.py
```diff
@@ -0,0 +1,685 @@
...
+        return self._poll.get_answer_count(id=self.id)
+
+    def _to_dict(self) -> PollMediaPayload:
+        data: PollMediaPayload = {'text': self.text}
+
+        if self.emoji is not None:
+            if isinstance(self.emoji, PartialEmoji):
+                data['emoji'] = self.emoji.to_dict()
+            else:
+                data['emoji'] = {'name': str(self.emoji)} # type: ignore
```
bijij, line 340:
```suggestion
                data['emoji'] = {'name': str(self.emoji)}  # type: ignore
```

discord/poll.py
```diff
@@ -0,0 +1,685 @@
...
+
+                if hasattr(self.emoji, 'id'):
+                    data['emoji']['id'] = int(self.emoji.id) # type: ignore
```
bijij, line 343:
```suggestion
                    data['emoji']['id'] = int(self.emoji.id)  # type: ignore
```

discord/poll.py
```diff
@@ -0,0 +1,685 @@
...
+
+        self.multiselect: bool = multiselect
+        self.layout_type: PollLayoutType = layout_type
+
+        # NOTE: These attributes are set manually when calling
+        # _from_data, so it should be ``None`` now.
+        self._message: Optional[Message] = None
+        self._state: Optional[ConnectionState] = None
+        self._finalized: bool = False
+        self._counts: Optional[Dict[int, PollAnswerCount]] = None # {answer_id: answer_count}
```
bijij, line 422:
```suggestion
        self._counts: Optional[Dict[int, PollAnswerCount]] = None  # {answer_id: answer_count}
```

discord/poll.py
```diff
@@ -0,0 +1,685 @@
...
+        self._finalized = message.poll._finalized
+        self._counts = message.poll._counts
+
+        for answer in self._answers:
+            answer._update(message)
+
+    @classmethod
+    def _from_data(
+        cls, *, data: PollPayload, message: Message, state: ConnectionState
+    ) -> Self:
```
bijij, line 444:
```suggestion
    def _from_data(cls, *, data: PollPayload, message: Message, state: ConnectionState) -> Self:
```

discord/poll.py
```diff
@@ -0,0 +1,685 @@
...
+        answers = [
+            PollAnswer(data=answer, poll=message.poll, message=message) for answer in data.get('answers')
+        ]  # 'message' will always have the 'poll' attr
+        multiselect = data.get('allow_multiselect', False)
+        layout_type = try_enum(PollLayoutType, data.get('layout_type', 1))
+        question_data = data.get('question')
+        question = question_data.get('text')
+        expiry = utils.parse_time(data['expiry'])  # If obtained via API, then expiry is set.
+        duration = expiry - message.created_at
```
bijij, line 453:
```suggestion
        duration = expiry - message.created_at
```

discord/poll.py
```diff
@@ -0,0 +1,685 @@
...
+    def _to_dict(self) -> PollCreatePayload:
+        data: PollCreatePayload = {
+            'allow_multiselect': self.multiselect,
+            'question': self._question_media.to_dict(),
+            'duration': self._hours_duration,
+            'layout_type': self.layout_type.value,
+            'answers': [
+                {'poll_media': answer._to_dict()}
+                for answer in self.answers
+            ]
```
bijij, line 490:
```suggestion
            'answers': [{'poll_media': answer._to_dict()} for answer in self.answers],
```

discord/poll.py
```diff
@@ -0,0 +1,685 @@
...
+        Returns
+        -------
+        :class:`Poll`
+            This poll with the new answer appended.
+        """
+
+        if self._message:
+            raise RuntimeError(
+                'Cannot append answers a poll recieved via a message'
+            )
```
bijij, line 595:
```suggestion
            raise RuntimeError('Cannot append answers a poll recieved via a message')
```

bijij (changes requested):
I've gone through and marked where type-checkers or the formatter will fail this code.

We adhere strictly to our black formatting and pyright configuration.
I suggest you ensure in future you before committing run black and pyright over the `discord` directory in the repo.

discord/state.py
```diff
@@ -1612,6 +1612,46 @@ def parse_entitlement_delete(self, data: gw.EntitlementDeleteEvent) -> None:
...
+        message = self._get_message(raw.message_id)
+        guild = self._get_guild(raw.guild_id)
+
+        if guild:
+            user = guild.get_member(raw.user_id)
+        else:
+            user = self.get_user(raw.user_id)
+
+        if message and user:
+            self.dispatch('poll_vote_add', user, message, message.poll.get_answer(raw.answer_id)) # type: ignore
```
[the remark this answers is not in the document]
bijij: Probably worth logging some kind of message in the event that we have the message object in cache but it doesn't have a poll. It's technically an error state.

discord/poll.py
```diff
@@ -0,0 +1,630 @@
...
+    @property
+    def question(self) -> str:
+        """:class:`str`: Returns this poll answer question string."""
+        return self._question_media.text
+
+    @property
+    def emoji(self) -> Optional[PartialEmoji]:
+        """Optional[:class:`PartialEmoji`]: Returns the emoji for this poll's question."""
+
+        return None  # As of now, polls questions don't support emojis
```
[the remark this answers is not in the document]
DA-344: When Discord adds emoji support for questions this will be needed again, so it is better to return ``None`` until then.

discord/raw_models.py
```diff
@@ -503,3 +505,32 @@ def __init__(self, *, data: GuildApplicationCommandPermissions, state: Connectio
         self.permissions: List[AppCommandPermissions] = [
             AppCommandPermissions(data=perm, guild=self.guild, state=state) for perm in data['permissions']
         ]
+
+class RawPollVoteActionEvent(_RawReprMixin):
+    """Represents the payload for a :func:`on_raw_poll_vote_add` or :func:`on_raw_poll_vote_remove`
+    event.
+
```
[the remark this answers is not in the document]
DA-344: Changes made. (Forgot to add newline between classes)

discord/state.py
```diff
@@ -1612,6 +1612,54 @@ def parse_entitlement_delete(self, data: gw.EntitlementDeleteEvent) -> None:
...
+        guild = self._get_guild(raw.guild_id)
+
+        if guild:
+            user = guild.get_member(raw.user_id)
+        else:
+            user = self.get_user(raw.user_id)
+
+        if message and user:
+            if not message.poll:
+                _log.warning('Cannot dispatch "poll_vote_add" event because the cached message does not contain a poll. This is a caching error.')
```
bijij, line 1634:
```suggestion
                _log.warning(
                    'POLL_VOTE_ADD referencing message with ID: %s without a poll. Discarding.',
                    raw.message_id,
                )
```

I'd probably reword this, same for the one for remove, at a mimimum including the message ID would be desirable.

discord/http.py
```diff
@@ -256,6 +259,13 @@ def handle_message_parameters(
         }
         payload.update(channel_payload)
 
+    if poll is not MISSING:
+        if len(poll) == 0 or len(poll) > 10:
+            raise ValueError('Poll must contain between 1 and 10 answers')
+        if poll._hours_duration < 1 or poll._hours_duration > 168:
+            raise ValueError('Polls duration must be between 1 hour and 7 days')
+        payload['poll'] = poll._to_dict()  # type: ignore
```
bijij, line 267: You need to check poll isn't None prior to this.

discord/poll.py
```diff
@@ -0,0 +1,720 @@
...
+                continue
+
+            for raw_member in reversed(users):
+                member_id = int(raw_member['id'])
+                member = guild.get_member(member_id)
+
+                yield member or User(state=self._state, data=raw_member)
+
+
+class PollAnswerCount(PollAnswerBase):
```
bijij, line 219: I feel like this functionality should be merged with PollAnswer and PollAnswer base shouldn't exist.

discord/state.py
```diff
@@ -1612,6 +1612,54 @@ def parse_entitlement_delete(self, data: gw.EntitlementDeleteEvent) -> None:
         entitlement = Entitlement(data=data, state=self)
         self.dispatch('entitlement_delete', entitlement)
 
+    def parse_message_poll_vote_add(self, data: gw.PollVoteActionEvent) -> None:
```
bijij, line 1615: You should update the answer count, as well as the self_voted property when you recieve these events.

discord/poll.py
DA-344: As both poll answers allow this functionality, I thought it would be convinient to create a class where both PollAnswer & PollAnswerCount inherit from which already implements this method.

discord/poll.py
```diff
@@ -0,0 +1,630 @@
...
+        """:class:`Poll`: Returns the poll that this answer belongs to."""
+        return self._message.poll  # type: ignore
+
+
+class PollAnswer(PollAnswerBase):
+    """Represents a poll's answer.
+
+    .. container:: operations
+
+        .. describe:: str(x)
```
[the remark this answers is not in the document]
DA-344: Fixed.

discord/poll.py
```diff
@@ -0,0 +1,630 @@
...
+
+        .. describe:: len(x)
+
+            Returns the Poll's answer amount.
+
+    .. versionadded:: 2.4
+
+    Parameters
+    ----------
+    question: Union[:class:`PollMedia`, :class:`str`]
```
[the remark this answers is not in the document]
DA-344: Using a ``str`` is a shortcut of ``PollMedia(text="...")``, and is internally handled.

discord/http.py
DA-344: Fixed in latest commits

DA-344:
The PR is ready to be merged. Here's a summary of the changes:
- Added `on_poll_vote_add`/`on_poll_vote_remove` events. (With their respective raw events)
- New `Poll` class to support, well... polls. Can be subclassed.
- Implemented `PollAnswer` and `PollAnswerCount`
- Added `poll` parameter to ``Messageable.send``, ``PartialMessage.reply`` & ``Context.send`` (w/ ``Context.reply``)
- New ``end_poll`` method for ``PartialMessage``. Returns a ``Message`` on success.
- Added ``Permissions.send_polls``
- New ``RawPollVoteActionEvent`` raw model
- New ``Client.wait_for`` overloads to type hint poll related events.
- ``PollLayoutType`` enum.
- New ``Intents.polls``, ``Intents.guild_polls`` and ``Intents.dm_polls`` intents.

discord/poll.py
DA-344: You're right, I thought of it previously, but left it like that. I was already planning in merging them into ``PollAnswer``

discord/message.py
```diff
@@ -1937,6 +1969,11 @@ def _handle_components(self, data: List[ComponentPayload]) -> None:
     def _handle_interaction(self, data: MessageInteractionPayload):
         self.interaction = MessageInteraction(state=self._state, guild=self.guild, data=data)
 
+    def _handle_poll_vote(self, answer_id: int, added: bool):
```
bijij, line 1972: Is there any reason for this function to exist?
Why can't the state just mess with the poll directly?

discord/poll.py
```diff
@@ -0,0 +1,640 @@
...
+    ) -> None:
+        self.media: PollMedia = PollMedia.from_dict(data=data['poll_media'])
+        self.id: int = int(data['answer_id'])
+        self._message: Optional[Message] = message
+        self._state: Optional[ConnectionState] = message._state if message else None
+        self._vote_count: int = 0
+        self.self_voted: bool = False
+        self._poll: Poll = poll
+
+    def _handle_vote_event(self, added: bool) -> None:
```
bijij, line 166: This should be updating self_voted or whatever we're calling it.

discord/abc.py
```diff
@@ -1350,6 +1351,7 @@ async def send(
         view: View = ...,
         suppress_embeds: bool = ...,
         silent: bool = ...,
+        poll: Poll = ...,
```
bijij, line 1354: This doesn't seem right?
Unless something changed, you can't send a poll with any other form of "content", so there should be no overloads which accept both Poll and any other kind (e.g. content, view, embeds, attachments)

discord/client.py
```diff
@@ -1834,6 +1836,28 @@ async def wait_for(
...
 
+    @overload
+    async def wait_for(
+        self: Union[Bot, AutoShardedBot],
+        event: Literal['poll_vote_add', 'poll_vote_remove'],
+        /,
+        *,
+        check: Optional[Callable[[Union[User, Member], Message, PollAnswer], bool]] = None,
+        timeout: Optional[float] = None,
+    ) -> Tuple[Union[User, Member], PollAnswer]:
```
bijij, line 1847: The check type and the return tuple don't match, which one is correct?

discord/ext/commands/context.py
```diff
@@ -641,6 +642,7 @@ async def reply(
         suppress_embeds: bool = ...,
         ephemeral: bool = ...,
         silent: bool = ...,
+        poll: Poll = ...,
```
bijij, line 645: See my comment in abc.py

discord/http.py
```diff
@@ -256,6 +259,13 @@ def handle_message_parameters(
         }
         payload.update(channel_payload)
 
+    if poll is not MISSING:
+        if len(poll) == 0 or len(poll) > 10:
+            raise ValueError('Poll must contain between 1 and 10 answers')
+        if poll._hours_duration < 1 or poll._hours_duration > 168:
+            raise ValueError('Polls duration must be between 1 hour and 7 days')
```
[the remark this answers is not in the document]
bijij: I'm going to reiterate this, I'm still not sure we should be doing this.

discord/http.py
```diff
@@ -2503,6 +2513,43 @@ def edit_application_info(self, *, reason: Optional[str], payload: Any) -> Respo
...
+        answer_id: Snowflake,
+        after: Optional[Snowflake] = None,
+        limit: Optional[int] = None,
+    ) -> Response[poll.PollAnswerVoters]:
+        params = {}
+
+        if after:
+            params['after'] = int(after)
+
+        if limit:
```
bijij, line 2529: This is going to make `0` return 25 results instead of erroring, I think this should be an explicit `None` check instead.
```suggestion
        if limit is not None:
```

discord/poll.py
```diff
@@ -0,0 +1,641 @@
...
+
+        if self.emoji:
+            if isinstance(self.emoji, Emoji):
+                payload['emoji'] = {'name': self.emoji.name, 'id': self.emoji.id}
+
+            elif isinstance(self.emoji, PartialEmoji):
+                payload['emoji'] = self.emoji.to_dict()
+
+            else:
+                payload['emoji'] = {'name': str(self.emoji)}  # type: ignore
```
bijij, line 109:
```suggestion
        if isinstance(self.emoji, str):
            payload['emoji'] = {"id": None, "name": self.emoji}
        elif self.emoji is not None:
            payload['emoji'] = {"id": self.emoji.id, "name": self.emoji.name}
```

This can be refworked to remove the type-ignore.

discord/poll.py
```diff
@@ -0,0 +1,641 @@
...
+        data: PollMediaPayload = {'text': self.text}
+
+        if self.emoji is not None:
+            if isinstance(self.emoji, PartialEmoji):
+                data['emoji'] = self.emoji.to_dict()
+            else:
+                data['emoji'] = {'name': str(self.emoji)}  # type: ignore
+
+                if hasattr(self.emoji, 'id'):
+                    data['emoji']['id'] = int(self.emoji.id)  # type: ignore
```
bijij, line 242:
```suggestion
    def _to_dict(self) -> PollAnswerPayload:
        return {
            'poll_media': self.media.to_dict(),
        }
```

Why isn't this returning the full poll answer payload?

discord/poll.py
```diff
@@ -0,0 +1,641 @@
...
+                message.channel.id, message.id, self.id, after=after_id, limit=retrieve
+            )
+            users = data['users']
+
+            if len(users) == 0:
+                # No more voters to fetch, terminate process
+                break
+
+            limit -= len(users)
+            after = Object(id=int((users[-1])['id']))
```
bijij, line 321:
```suggestion
            after = Object(id=int(users[-1]['id']))
```
stylistic choice, but I think these parens do nothing but make it harder to read.

discord/poll.py
```diff
@@ -0,0 +1,641 @@
...
+class Poll:
+    """Represents a message's Poll.
+
+    .. container:: operations
+
+        .. describe:: str(x)
+
+            Returns the Poll's question
+
+        .. describe:: len(x)
```
bijij, line 344: If you're supporting len, you should support iter IMO.

discord/poll.py
```diff
@@ -0,0 +1,641 @@
...
+        self._hours_duration: float = duration.total_seconds() / 3600
+
+        self.multiselect: bool = multiselect
+        self.layout_type: PollLayoutType = layout_type
+
+        # NOTE: These attributes are set manually when calling
+        # _from_data, so it should be ``None`` now.
+        self._message: Optional[Message] = None
+        self._state: Optional[ConnectionState] = None
+        self._results: Optional[PollResultPayload] = None
```
bijij, line 402: Why are you storing a payload?

discord/poll.py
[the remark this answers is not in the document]
bijij:
```suggestion
```
Disagree, when could be never.

discord/poll.py
```diff
@@ -0,0 +1,641 @@
...
+
+        return None  # As of now, polls questions don't support emojis
+
+    @property
+    def answers(self) -> List[PollAnswer]:
+        """List[:class:`PollAnswer`]: Returns a read-only copy of the answers"""
+        return self._answers.copy()
+
+    @property
+    def expiry(self) -> Optional[datetime.datetime]:
```
bijij, line 502:
```suggestion
    def expires_at(self) -> Optional[datetime.datetime]:
```

discord/poll.py
```diff
@@ -0,0 +1,641 @@
...
+    def is_finalized(self) -> bool:
+        """:class:`bool`: Returns whether the poll has finalized.
+
+        It always returns ``False`` if the poll is not part of a
+        fetched message. You should consider accessing this method
+        via :attr:`Message.poll`
+        """
+        if not self._results:
+            return False
+        return self._results['is_finalized']
```
bijij, line 538: I'm not sure why you're not just storing the state as a bool, storing the result payload seems like a waste.

discord/poll.py
```diff
@@ -0,0 +1,641 @@
...
+    def message(self) -> Optional[Message]:
+        """:class:`Message`: The message this poll is from."""
+        return self._message
+
+    @property
+    def total_votes(self) -> int:
+        """:class:`int`: Returns the sum of all the answer votes."""
+        return sum([answer.vote_count for answer in self.answers])
+
+    def is_finalized(self) -> bool:
```
bijij, line 529: This needs an EN-GB equivalent.

discord/poll.py
```diff
@@ -0,0 +1,641 @@
...
+        return self
+
+    def get_answer(
+        self,
+        /,
+        id: int,
+    ) -> Optional[PollAnswer]:
+        """Returns the answer with the provided ID or ``None`` if not found.
+
+        Note that the ID, as Discord says, it is the index / row where the answer
```
bijij, line 602: This is an implementation detail you can't rely on it.

discord/poll.py
```diff
@@ -0,0 +1,641 @@
...
+        except KeyError:
+            self._results = None
+
+        return self
+
+    def _to_dict(self) -> PollCreatePayload:
+        data: PollCreatePayload = {
+            'allow_multiselect': self.multiselect,
+            'question': self._question_media.to_dict(),
+            'duration': self._hours_duration,
```
bijij, line 470: This whole property seems dangerous, you allow people to add new answers to a copy, but if they edit the duration this won't be updated.

discord/poll.py
```diff
@@ -0,0 +1,641 @@
...
+            'layout_type': self.layout_type.value,
+            'answers': [{'poll_media': answer._to_dict()} for answer in self.answers],
```
bijij, line 472:
```suggestion
            'answers': [answer.to_dict() for answer in self.answers],
```

If you take my above suggestion, this can be simplified.

bijij (changes requested):
There's still some substantial issues with the Poll model itself, and some questional UX decisions which may lead to issues in the future.

discord/abc.py
DA-344: I've been testing polls and they do can be sent with embeds, views and contents.
thingy

discord/poll.py
DA-344: Thanks for the comment, this has been updated so the duration is calculated when calling _to_dict.

discord/poll.py
DA-344: Added its equivalent (``is_finalised``)

discord/poll.py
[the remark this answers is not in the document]
bijij: I'd rather this be left open for @Rapptz to chime in.

AbstractUmbra:
What's the current status of this PR? Are we progressing, as I can see the last few comments from @bijij are unresolved?

DA-344:
> What's the current status of this PR? Are we progressing, as I can see the last few comments from @bijij are unresolved?

I am applying the suggested changes, most of them are already implemented and I forgot to add a comment and resolve the conversations, lol sorry.

discord/poll.py
DA-344: Change applied in latest commits.

discord/poll.py
DA-344: MB, updated in latest commits.

discord/poll.py
DA-344: Yeah that was there from previous changes that I reverted and forgot to remove the parentheses. Should be removed by now.

discord/poll.py
DA-344: Yeah, this is now updated in the latest commits.

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+            else:
+                poll_media['emoji'] = {'name': str(emoji)}  # type: ignore
+
+        payload: PollAnswerWithIDPayload = {'answer_id': id, 'poll_media': poll_media}
+
+        return cls(data=payload, message=message, poll=poll)
+
+    @property
+    def text(self) -> str:
+        """:class:`str`: Returns this answer display text."""
```
Rapptz, line 206:
```suggestion
        """:class:`str`: Returns this answer's displayed text."""
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+        return self.media.text
+
+    @property
+    def emoji(self) -> Optional[Union[PartialEmoji, Emoji]]:
+        """Optional[:class:`PartialEmoji`]: Returns this answer display emoji, is any."""
```
Rapptz, line 211:
```suggestion
        """Optional[Union[:class:`Emoji`, :class:`PartialEmoji`]]: Returns this answer's displayed emoji, if any."""
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+        if isinstance(self.media.emoji, str):
+            return PartialEmoji.from_str(self.media.emoji)
+        return self.media.emoji
```
Rapptz, line 214: You're doing this here but this should have been done in the constructor for `PollMedia`.

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+        """
+
+        emoji = data.get('emoji')
+
+        if emoji:
+            return cls(text=data['text'], emoji=PartialEmoji.from_dict(emoji))
+        return cls(text=data['text'])
+
+
+class PollAnswer:
```
Rapptz, line 126: Missing `versionadded`.

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+    'Poll',
+    'PollAnswer',
+    'PollMedia',
+)
+
+MISSING = utils.MISSING
+PollMediaEmoji = Union[PartialEmoji, Emoji, str]
+
+
+class PollMedia:
```
Rapptz, line 67: This isn't a `NamedTuple`, nor should it be. Either way, this is missing `versionadded`.

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+        if emoji:
+            if isinstance(emoji, Emoji):
+                poll_media['emoji'] = {'name': emoji.name}  # type: ignore
+
+                if emoji.id:
+                    poll_media['emoji']['id'] = emoji.id  # type: ignore
+            elif isinstance(emoji, PartialEmoji):
+                poll_media['emoji'] = emoji.to_dict()
+            else:
+                poll_media['emoji'] = {'name': str(emoji)}  # type: ignore
```
Rapptz, line 198: A lot of this can be simplified by doing something similar to this:
```suggestion
        poll_media: PollMediaPayload = {'text': text}
        if emoji is not None:
            emoji = PartialEmoji.from_str(emoji) if isinstance(emoji, str) else emoji._as_partial()
            emoji_data = emoji.to_dict()
            # Remove animated key?
            # emoji_data.pop('animated', None)
            poll_media['emoji'] = emoji_data
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+        if isinstance(self.media.emoji, str):
+            return PartialEmoji.from_str(self.media.emoji)
+        return self.media.emoji
+
+    @property
+    def vote_count(self) -> int:
+        """:class:`int`: Returns an approximate votes of this answer. If the poll
+        has finalized, it returns the exact amount of votes.
+        """
+        return self._vote_count
```
Rapptz, line 221:
```suggestion
    @property
    def vote_count(self) -> int:
        """:class:`int`: Returns an approximate count of votes for this answer.
        
        If the poll is finished, the count is exact.
        """
        return self._vote_count
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+
+    def _to_dict(self) -> PollAnswerPayload:
+        return {
+            'poll_media': self.media.to_dict(),
+        }
+
+    async def voters(
+        self, *, limit: Optional[int] = None, after: Optional[Snowflake] = None
+    ) -> AsyncIterator[Union[User, Member]]:
+        """Returns an :term:`asynchronous iterator` representing the users that have voted to this answer.
```
Rapptz, line 236:
```suggestion
        """Returns an :term:`asynchronous iterator` representing the users that have voted on this answer.
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+
+        The ``after`` parameter must represent a user
+        and meet the :class:`abc.Snowflake` abc.
+
+        .. note::
+
+            This can only be called when the parent poll has an attached message.
```
Rapptz, line 243:
```suggestion
        This can only be called when the parent poll was sent to a message.
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+        HTTPException
+            Retrieving the users failed.
+
+        Yields
+        ------
+        Union[:class:`User`, :class:`Member`]
+            The member (if retrievable) or the user that has voted
+            to this poll answer. The case where it can be a :class:`Member`
+            is in a guild message context. Sometimes it can be a :class:`User`
+            if the member has left the guild.
```
Rapptz, line 278:
```suggestion
            The member (if retrievable) or the user that has voted
            on this poll answer. The case where it can be a :class:`Member`
            is in a guild message context. Sometimes it can be a :class:`User`
            if the member has left the guild or if the member is not cached.
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+        """
+
+        if not self._message or not self._state:  # Make type checker happy
+            raise RuntimeError('You cannot fetch users in a non-message-attached poll')
```
Rapptz, line 282:
```suggestion
            raise ClientException('You cannot fetch users to a poll not sent with a message')
```

Import `ClientException` and document this in `Raises`.

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+            state = self._state
+            after_id = after.id if after else None
+
+            data = await state.http.get_poll_answer_voters(
+                message.channel.id, message.id, self.id, after=after_id, limit=retrieve
+            )
+            users = data['users']
+
+            if len(users) == 0:
+                # No more voters to fetch, terminate process
```
Rapptz, line 304:
```suggestion
                # No more voters to fetch, terminate loop
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+    .. container:: operations
+
+        .. describe:: str(x)
+
+            Returns the Poll's question
+
+        .. describe:: len(x)
+
+            Returns the Poll's answer amount.
+
```
Rapptz, line 334:
```suggestion
```

None of these operations actually make sense.

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+    .. versionadded:: 2.4
+
+    Parameters
+    ----------
+    question: Union[:class:`PollMedia`, :class:`str`]
```
Rapptz, line 339: If the parameter type and the attribute types are different then you need a dedicated attribute section as well.

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+        The poll's question media. Text can be up to 300 characters.
```
Rapptz, line 340:
```suggestion
        The poll's displayed question. The text can be up to 300 characters.
```

discord/poll.py
```diff
@@ -0,0 +1,640 @@
...
+
+        .. warning::
+
+            At the moment, this *does not* support emojis.
+
+    duration: :class:`datetime.timedelta`
+        The duration of the poll.
```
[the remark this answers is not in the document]
Rapptz: Why was this ignored?

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+
+            voters = [voter async for voter in poll_answer.voters()]
+            # voters is now a list of User
+
+        Parameters
+        ----------
+        limit: Optional[:class:`int`]
+            The maximum number of results to return.
+            If not provided, returns all the users who
+            voted to this poll answer.
```
Rapptz, line 263:
```suggestion
            voted on this poll answer.
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+
+    .. versionadded:: 2.4
+
+    Parameters
+    ----------
+    question: Union[:class:`PollMedia`, :class:`str`]
+        The poll's question media. Text can be up to 300 characters.
+    duration: :class:`datetime.timedelta`
+        The duration of the poll.
+    multiselect: :class:`bool`
```
Rapptz, line 343: Rename this attribute to `multiple` a la HTML since it's a better name.

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+        Whether users are allowed to select more than one answer. Defaults
+        to ``True``
```
Rapptz, line 345:
```suggestion
        Whether users are allowed to select more than one answer. 
        Defaults to ``True``.
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+    layout_type: :class:`PollLayoutType`
+        The layout type of the poll. Defaults to :attr:`PollLayoutType.default`
```
Rapptz, line 347:
```suggestion
        The layout type of the poll. Defaults to :attr:`PollLayoutType.default`.
```

discord/poll.py
```diff
@@ -0,0 +1,630 @@
...
+        multiselect: bool = False,
+        layout_type: PollLayoutType = PollLayoutType.default,
+    ) -> None:
+        if isinstance(question, str):
+            self._question_media: PollMedia = PollMedia(text=question, emoji=None)
+        else:
+            self._question_media: PollMedia = (
+                question  # At the moment this only supports text, so no need to add emoji support
+            )
+        self._answers: List[PollAnswer] = []
```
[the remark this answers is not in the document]
Rapptz: Why was this ignored?

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+        self._state: Optional[ConnectionState] = None
+        self._finalized: bool = False
+        self._expiry: Optional[datetime.datetime] = None
+
+    def _update(self, message: Message) -> None:
+        self._state = message._state
+        self._message = message
+
+        if not message.poll:
+            # Something should go very wrong for this if block to be called
```
Rapptz, line 394:
```suggestion
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+            return
+        # These attributes are accessed via message.poll as there
+        # is where all the data is stored
```
Rapptz, line 397:
```suggestion
        # The message's poll contains the more up to date data.
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+            'allow_multiselect': self.multiselect,
+            'question': self._question_media.to_dict(),
+            'duration': self.duration.total_seconds() / 3600,
+            'layout_type': self.layout_type.value,
+            'answers': [answer._to_dict() for answer in self.answers],
+        }
+        return data
+
+    def __str__(self) -> str:
+        return self._question_media.text
```
Rapptz, line 462:
```suggestion
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+
+    def __repr__(self) -> str:
+        return f"<Poll duration={self.duration} question=\"{self.question}\" answers={self.answers}>"
+
+    def __len__(self) -> int:
+        return len(self.answers)
```
Rapptz, line 468:
```suggestion
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+
+
+class PollMedia:
+    """Represents the poll media for a poll item.
+
+    .. container:: operations
+
+        .. describe:: str(x)
+
+            Returns this Poll Media\'s text.
```
Rapptz, line 74:
```suggestion
```

I don't think this is future proof.

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+    """
+
+    __slots__ = ('text', 'emoji')
+
+    def __init__(self, /, text: str, emoji: Optional[PollMediaEmoji] = None) -> None:
+        self.text: str = text
+        self.emoji: Optional[PollMediaEmoji] = emoji
+
+    def __str__(self) -> str:
+        return self.text
```
Rapptz, line 92:
```suggestion
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+        *,
+        multiselect: bool = False,
+        layout_type: PollLayoutType = PollLayoutType.default,
+    ) -> None:
+        if isinstance(question, str):
+            self._question_media: PollMedia = PollMedia(text=question, emoji=None)
+        else:
+            self._question_media: PollMedia = (
+                question  # At the moment this only supports text, so no need to add emoji support
+            )
```
Rapptz, line 375:
```suggestion
            # At the moment this only supports text, so no need to add emoji support
            self._question_media: PollMedia = question
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+    def message(self) -> Optional[Message]:
+        """:class:`Message`: The message this poll is from."""
+        return self._message
+
+    @property
+    def total_votes(self) -> int:
+        """:class:`int`: Returns the sum of all the answer votes."""
+        return sum([answer.vote_count for answer in self.answers])
+
+    def is_finalized(self) -> bool:
```
Rapptz, line 514:
```suggestion
    def is_finalised(self) -> bool:
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+        """:class:`bool`: Returns whether the poll has finalized.
```
Rapptz, line 515:
```suggestion
        """:class:`bool`: Returns whether the poll has finalised.
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+
+        It always returns ``False`` for stateless polls.
+        """
+        return self._finalized
+
+    is_finalised = is_finalized
```
Rapptz, line 521:
```suggestion
    is_finalized = is_finalised
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+
+    @property
+    def total_votes(self) -> int:
+        """:class:`int`: Returns the sum of all the answer votes."""
+        return sum([answer.vote_count for answer in self.answers])
+
+    def is_finalized(self) -> bool:
+        """:class:`bool`: Returns whether the poll has finalized.
+
+        It always returns ``False`` for stateless polls.
```
Rapptz, line 517:
```suggestion
        This always returns ``False`` for stateless polls.
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+    def copy(self) -> Self:
+        """Returns a stateless copy of this poll.
+
+        This is meant to be used when you want to edit a stateful poll.
+
+        Returns
+        -------
+        :class:`Poll`
+            The copy of the poll.
+        """
```
Rapptz, line 532: This looks like a partial deep copy. I guess the name might be fine, but it's a bit weird for a method named `copy` to do a (partial) deep copy rather than a shallow copy.

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+
+        new = self.__class__(question=self.question, duration=self.duration)
+        # We want to return a stateless copy of the poll, so we should not
+        # override new._answers as our answers may contain a state
+
```
Rapptz, line 537:
```suggestion

        # We want to return a stateless copy of the poll, so we should not
        # override new._answers as our answers may contain a state
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+            Cannot append answers to a poll recieved via message.
+
+        Returns
+        -------
+        :class:`Poll`
+            This poll with the new answer appended.
+        """
+
+        if self._message:
+            raise RuntimeError('Cannot append answers to a poll recieved via message.')
```
Rapptz, line 571:
```suggestion
            raise ClientException('Cannot append answers to a poll that is active')
```

Update docstring too.

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+
+        Raises
+        ------
+        RuntimeError
+            Cannot append answers to a poll recieved via message.
+
+        Returns
+        -------
+        :class:`Poll`
+            This poll with the new answer appended.
```
Rapptz, line 567:
```suggestion
            This poll with the new answer appended. This is to allow fluent-style chaining.
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+
+    def get_answer(
+        self,
+        /,
+        id: int,
+    ) -> Optional[PollAnswer]:
+        """Returns the answer with the provided ID or ``None`` if not found.
+
+        Note that the ID, as Discord says, it is the index / row where the answer
+        is located in the poll.
```
Rapptz, line 586:
```suggestion
```

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+        id: :class:`int`
+            The ID of the answer to get.
+
+        Returns
+        -------
+        Optional[:class:`PollAnswer`]
+            The answer.
+        """
+
+        return utils.get(self.answers, id=id)
```
Rapptz, line 599: You ignored this earlier but `get_` methods **should not** be O(n). Either change it to a dictionary or remove the method entirely.

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+            Ending the poll failed.
+
+        Returns
+        -------
+        :class:`Poll`
+            The updated poll.
+        """
+
+        if not self._message or not self._state:  # Make type checker happy
+            raise RuntimeError('This poll has no attached message.')
```
Rapptz, line 620: Once again, `ClientException`.

discord/raw_models.py
```diff
@@ -503,3 +505,33 @@ def __init__(self, *, data: GuildApplicationCommandPermissions, state: Connectio
...
+    .. versionadded:: 2.4
+
+    Attributes
+    ----------
+    user_id: :class:`int`
+        The ID of the user that added their vote.
```
Rapptz, line 519:
```suggestion
        The ID of the user that added or removed a vote.
```

discord/raw_models.py
```diff
@@ -503,3 +505,33 @@ def __init__(self, *, data: GuildApplicationCommandPermissions, state: Connectio
...
+    channel_id: :class:`int`
+        The Channel ID this poll answer was added.
```
Rapptz, line 521:
```suggestion
        The channel ID where the poll vote action took place.
```

discord/raw_models.py
```diff
@@ -503,3 +505,33 @@ def __init__(self, *, data: GuildApplicationCommandPermissions, state: Connectio
...
+    message_id: :class:`int`
+        The Message ID that contains the poll the user added their vote to.
```
Rapptz, line 523:
```suggestion
        The message ID that contains the poll the user added or removed their vote on.
```

discord/raw_models.py
```diff
@@ -503,3 +505,33 @@ def __init__(self, *, data: GuildApplicationCommandPermissions, state: Connectio
...
+    guild_id: Optional[:class:`int`]
+        The Guild ID the poll is in, or ``None`` if in a GDM or DM channel.
```
Rapptz, line 525:
```suggestion
       The guild ID where the vote got added or removed, if applicable.
```

discord/raw_models.py
```diff
@@ -503,3 +505,33 @@ def __init__(self, *, data: GuildApplicationCommandPermissions, state: Connectio
...
+    answer_id: :class:`int`
+        The poll answer's ID the user voted to.
```
Rapptz, line 527:
```suggestion
        The poll answer's ID the user voted on.
```

discord/webhook/async_.py
```diff
@@ -1734,6 +1742,15 @@ async def send(
...
+            The attached poll for this message.
```
Rapptz, line 1746:
```suggestion
            The poll to send with this message.
```

discord/webhook/sync.py
```diff
@@ -979,6 +983,14 @@ def send(
             in the UI, but will not actually send a notification.
 
             .. versionadded:: 2.2
+        poll: :class:`Poll`
+            The attached poll for this message.
```
Rapptz, line 987:
```suggestion
            The poll to send with this message.
```

docs/api.rst
```diff
@@ -1442,6 +1442,71 @@ Voice
     :param after: The voice state after the changes.
     :type after: :class:`VoiceState`
 
+Polls
```
Rapptz, line 1445: These events are alphabetically sorted, P comes before V not after. The English alphabet is `ABCDEFGHIJLKMNOPQRSTUVWXYZ`

docs/api.rst
```diff
@@ -1442,6 +1442,71 @@ Voice
...
+~~~~~~
+
+.. function:: on_poll_vote_add(user, answer)
+
+    Called when a :class:`Message`\'s attached :class:`Poll` gets a new vote. If any of ``user`` or ``message``
+    are not cached this event will not be called.
```
Rapptz, line 1451:
```suggestion
    Called when a :class:`Message`\'s attached :class:`Poll` gets a new vote. If the ``user`` or ``message``
    are not cached then this event will not be called.
```

docs/api.rst
```diff
@@ -1442,6 +1442,71 @@ Voice
...
+
+    This requires :attr:`Intents.message_content` and :attr:`Intents.polls` to be enabled.
+
+    .. note::
+
+        If the poll allows multiselect and the user votes for more than one answer
```
Rapptz, line 1457:
```suggestion
        If the poll allows multiple answers and the user votes for more than one answer
```

docs/api.rst
```diff
@@ -1442,6 +1442,71 @@ Voice
...
+        this event will be called as many times as answers the user has voted to.
+
+    .. versionadded:: 2.4
+
+    :param user: The user that voted for the message's poll.
+    :type user: Union[:class:`User`, :class:`Member`]
+    :param answer: The answer the user voted to.
```
Rapptz, line 1464:
```suggestion
    :param answer: The answer the user voted on.
```

docs/api.rst
```diff
@@ -1442,6 +1442,71 @@ Voice
...
+
+        If the poll allows multiselect and the user removes more than one vote
+        this event will be called as many times as votes removed.
+
+    .. versionadded:: 2.4
+
+    :param user: The user that removed their vote from the message's poll.
+    :type user: Union[:class:`User`, :class:`Member`]
+    :param answer: The answer the user removed the vote from.
+    :type answer: :class:`PollAnswer`
```
Rapptz, line 1484: You should collapse this with the one above, similar to the other events.

Rapptz (changes requested):
Please fix the CI failures.

discord/permissions.py
```diff
@@ -850,6 +866,7 @@ class PermissionOverwrite:
         send_voice_messages: Optional[bool]
         create_expressions: Optional[bool]
         create_events: Optional[bool]
+        send_polls: Optional[bool]
```
Rapptz, line 869: The alias needs to be here too.

discord/message.py
```diff
@@ -1439,6 +1444,28 @@ async def reply(self, content: Optional[str] = None, **kwargs: Any) -> Message:
 
         return await self.channel.send(content, reference=self, **kwargs)
 
+    async def end_poll(self) -> Message:
```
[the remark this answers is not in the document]
DA-344: Okay, I have added this.

discord/poll.py
DA-344: Yeah I thought of it when I created this method, it does a partial deep copy because, as the docstring says, it is meant to return a stateless poll copy when the actual object has a state. I don't know if I explain myself...

discord/permissions.py
DA-344: Added the alias.

discord/abc.py
```diff
@@ -1516,6 +1522,10 @@ async def send(
             in the UI, but will not actually send a notification.
 
             .. versionadded:: 2.2
+        poll: :class:`~discord.Poll`
+            The attached poll for this message.
```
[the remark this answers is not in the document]
DA-344: Commited this change.

discord/message.py
```diff
@@ -1439,6 +1444,28 @@ async def reply(self, content: Optional[str] = None, **kwargs: Any) -> Message:
...
+        """|coro|
+
+        Tries to end this message\'s :class:`Poll`.
```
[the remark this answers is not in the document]
DA-344: Committed this change

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+    emoji: Optional[Union[:class:`PartialEmoji`, :class:`Emoji`, :class:`str`]]
+        The attached emoji for this media. This will always be ignored for a poll
+        question media.
+    """
+
+    __slots__ = ('text', 'emoji')
+
+    def __init__(self, /, text: str, emoji: Optional[PollMediaEmoji] = None) -> None:
+        self.text: str = text
+        self.emoji: Optional[PollMediaEmoji] = emoji
```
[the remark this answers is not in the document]
DA-344: Thank you for the clarification. I've updated the code to parse strings to ``PartialEmoji``s.

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+    .. container:: operations
+
+        .. describe:: str(x)
+
+            Returns this Poll Media\'s text.
+
+    Attributes
+    ----------
+    text: :class:`str`
+        The displayed text
```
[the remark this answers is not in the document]
DA-344: I've commented in the wrong review 🤦‍♂️. Anyways, apologies for the oversight. Added the dot.

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+
+    def to_dict(self) -> PollMediaPayload:
+        """Returns an API valid payload for this instance."""
+
+        payload: PollMediaPayload = {'text': self.text}
+
+        if isinstance(self.emoji, str):
+            payload['emoji'] = {"id": None, "name": self.emoji}
+        elif self.emoji is not None:
+            payload['emoji'] = {"id": self.emoji.id, "name": self.emoji.name}
```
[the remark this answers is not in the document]
DA-344: I've updated the code to use single quotes instead of double quotes, as requested.

discord/poll.py
```diff
@@ -0,0 +1,624 @@
...
+
+            Returns this Poll Media\'s text.
+
+    Attributes
+    ----------
+    text: :class:`str`
+        The displayed text
+    emoji: Optional[Union[:class:`PartialEmoji`, :class:`Emoji`, :class:`str`]]
+        The attached emoji for this media. This will always be ignored for a poll
+        question media.
```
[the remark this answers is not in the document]
DA-344: Thanks for the improvement, committed the change.

discord/poll.py
```diff
@@ -0,0 +1,577 @@
...
+    @property
+    def question(self) -> str:
+        """:class:`str`: Returns this poll answer question string."""
+        return self._question_media.text
+
+    @property
+    def emoji(self) -> Optional[PartialEmoji]:
+        """Optional[:class:`PartialEmoji`]: Returns the emoji for this poll's question."""
+
+        return None  # As of now, polls questions don't support emojis
```
Rapptz, line 436:
```suggestion
```

I don't see the point of this.

Rapptz (changes requested):
Just one final nit

discord/client.py
```diff
@@ -1839,6 +1841,28 @@ async def wait_for(
...
+        /,
+        *,
+        check: Optional[Callable[[Union[User, Member], PollAnswer], bool]] = None,
+        timeout: Optional[float] = None,
+    ) -> Tuple[Union[User, Member], PollAnswer]:
+        ...
+
+    @overload
+    async def wait_for(
+        self: Union[Bot, AutoShardedBot],
```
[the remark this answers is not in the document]
DA-344: Yeah you are right, thanks for letting me know.

discord/poll.py
DA-344: Yeah, I thought of it in the last day, but just couldn't find the words to say it. Anyways, it was removed in the latest commits.

Rapptz (approved):
Thanks!

mwpowellhtx:
How far along is this? Been in a state of limbo for what, over a year?

Soheab:
> How far along is this? Been in a state of limbo for what, over a year?

Do you not see the purple "Merged" pill near the name?

Or the text above saying "Rapptz (Danny) merged commit [e43bd86](https://github.com/Rapptz/discord.py/commit/e43bd8692cf3e293e5fd1bbe7ec481dd8aa09c4a) into Rapptz:master [on May 10, 2024](https://github.com/Rapptz/discord.py/pull/9759#event-12767975555)".

Or the note before commenting:
"This PR was closed a year ago. Please consider opening a [new issue](https://github.com/Rapptz/discord.py/issues/new/choose) instead of leaving a comment here.".

mwpowellhtx:
> > How far along is this? Been in a state of limbo for what, over a year?
> 
> Do you not see the purple "Merged" pill near the name?

I did not, thank you for mentioning. Are there use case examples?

PR #9760, opened 2024-03-19
First pass at supporting user apps

Rapptz:
This is based off of previous work by Red in #9406 and their subsequent branch. I've merely adapted and fixed some of it.

There are a few things I'm mostly nebulous about that I need to mull over some more:

1. The `authorized_integration_owner` type is pretty useless, I can't imagine a use case for it nor a good way to expose it since its information is duplicated elsewhere in the schema.
2. `AppIntegrationType` as a type feels wrong. It should probably be an enum or renamed to `AppInstallationType` since that's.. what it is. The original naming feels like a misstep on Discord's side.
3. `AppContextType` is an `ArrayFlags` currently but due to a losing battle discussing this with them I think this might be better off as an `Enum` instead.
4. The naming schemes are a bit inconsistent, e.g. the decorators are `install_types`/`user_install`/`guild_install` but the inner types and properties are called `integration_types`. Again, this feels like a misstep on Discord's side.
5. Permissions probably need more mocking for the various `channel.permissions_for` code that people have littered around.
6. The new `interaction_metadata` type needs to be added to `Message`
7. The new fields in `AppInfo` need to be added as well.

### DIFF
--- a/discord/app_commands/__init__.py
+++ b/discord/app_commands/__init__.py
@@ -16,5 +16,6 @@ from .tree import *
 from .namespace import *
 from .transformers import *
 from .translator import *
+from .installs import *
 from . import checks as checks
 from .checks import Cooldown as Cooldown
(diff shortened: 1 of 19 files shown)


aboumerhy (commented):
good

KylomaskGamer:
yeepeeee

PR #9769, opened 2024-03-27
Ensure Client.close() has finished in __aexit__

mikeshardmind:
## Summary

This wraps the closing behavior in a task. Subsequent callers of .close() now await that same close finishing rather than short circuiting. This prevents a user-called close outside of __aexit__ from not finishing before no longer having a running event loop.

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -284,7 +284,7 @@ class Client:
         self._enable_debug_events: bool = options.pop('enable_debug_events', False)
         self._connection: ConnectionState[Self] = self._get_state(intents=intents, **options)
         self._connection.shard_count = self.shard_count
-        self._closed: bool = False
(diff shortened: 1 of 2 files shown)


discord/client.py
```diff
@@ -304,8 +304,8 @@ async def __aexit__(
         exc_value: Optional[BaseException],
         traceback: Optional[TracebackType],
     ) -> None:
-        if not self.is_closed():
-            await self.close()
+        # Don't double-call user implementations
+        await (self.close() if self._closing_task is None else self._closing_task)
```
Rapptz, line 308: Can you expand this out instead of using the ternary?

mikeshardmind: sure

PR #9774, opened 2024-04-01
copy over original row position for dynamic items

ika2kki:
## Summary

this PR addresses a `ui.DynamicItem` bug where the internal row position isn't copied over when swapping the original item with the dynamic variant after a call to `ui.View.from_message` whilst dispatching~ this makes it so the view's item positions go to a broken state given another item takes up an entire row or similar, like a `Select`.

bikeshedding post: https://discord.com/channels/336642139381301249/1224072326103044096

i've made a stylistic change in the code as well but i can revert this! 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ui/view.py
+++ b/discord/ui/view.py
@@ -615,16 +615,15 @@ class ViewStore:
 
         view = View.from_message(interaction.message, timeout=None)
 
-        base_item_index: Optional[int] = None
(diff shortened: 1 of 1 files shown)


Rapptz:
Thanks for everything bud.

PR #9776, opened 2024-04-01
Fix File(...) not accepting bytes as a source

vvsotnikov:
## Summary

Type hints in `discord.File()` included `bytes` as a possible value for the `fp` parameter, but it actually wasn't supported. This PR handles `bytes` input and adds a test to ensure this bug won't happen again.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/file.py
+++ b/discord/file.py
@@ -93,6 +93,10 @@ class File:
             self.fp: io.BufferedIOBase = fp
             self._original_pos = fp.tell()
             self._owner = False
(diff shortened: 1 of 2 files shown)


Gobot1234:
This isn't correctly handling this, bytes aren't meant to be passed as the contents of the file object, open can take bytes paths

vvsotnikov:
Ah, I see, thank you. I misread the docstring :)

PR #9787, opened 2024-04-14
[types] Fix signature for `commands.hybrid_command()` decorator when decorating free functions.

Sachaa-Thanasius:
## Summary
This fixes an error from pyright that considers the resolved value of `CogT` to be `Unknown` when it should be None, specificallly for a the first type parameter of `HybridCommand` when created via `commands.hybrid_command()` decorating a free function.

Uses the same callback trick that `commands.command()` does:
https://github.com/Rapptz/discord.py/blob/425edd2e10b9be3d7799c0df0cd1d43a1a34654e/discord/ext/commands/core.py#L1700-L1713

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/hybrid.py
+++ b/discord/ext/commands/hybrid.py
@@ -859,6 +859,8 @@ class HybridGroup(Group[CogT, P, T]):
 if TYPE_CHECKING:
     # Using a class to emulate a function allows for overloading the inner function in the decorator.
 
+    from typing import overload
+
     class _HybridCommandDecorator:
         @overload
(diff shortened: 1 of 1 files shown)


Sachaa-Thanasius:
Huh, so this is a different `CogT` than the one that `commands.command()` is working with, and this one is bound to `Cog`, not `Optional[Cog]`. That explains the CI errors. Wonder how much would break if hybrid's CogT was either changed to match core's or if it was replaced with the same `CogT` by importing it from `commands._types`.

Sachaa-Thanasius:
Actually, wait. `discord.app_commands.command` has this issue too. Only `discord.ext.commands.command` doesn't. Am I missing something for why the hybrid and app versions don't account for this?

It's bigger than I thought, regardless. I'll close this and make an issue instead.

PR #9789, opened 2024-04-15
Fix multiple arguments can't be parsed to a flag with List annotation

Kuro-Rui:
## Summary

Currently, flags with List annotation will only parse flags if there's only one argument, else it will return a
`BadFlagArgument: Could not convert to 'List' for flag 'flag'`

## Checklist

- [ ] This PR fixes an issue.


Kuro-Rui:
I learned the hard way that List annotation is for multiple passing, as stated in the docs. My mistake, I should've done more reading and research. Sorry for the inconvenience 🙏

PR #9796, opened 2024-04-19
Fix channel deletion not removing associated threads

### LINKED ISSUE #9795: Threads in deleted channels not removed from cache
### Summary

When a channel is deleted, any threads in that channel will remain in `guild.threads`

### Reproduction Steps

1) Create a thread in a channel
2) See that the thread is in `guild.threads`
3) Delete the channel containing the thread
4) See that the thread is still in `guild.threads`

### Minimal Reproducible Code

```python
import asyncio
channel = await guild.create_text_channel("random-channel")
thread = await channel.create_thread(name="my-thread")
print(thread in guild.threads)
await channel.delete()
await asyncio.sleep(3)
print(thread in guild.threads)
```

### Expected Results

True
False

### Actual Results

True
True

### Intents

All apart from these: presences, invites, webhooks, integrations, and dm typing/reactions

### System Information

```python
- Python v3.12.1-final
- discord.py v2.3.2-final
- aiohttp v3.9.4
```

### Checklist

### Additional Context

_No response_

LostLuma:
## Summary

Resolves #9795.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/state.py
+++ b/discord/state.py
@@ -831,6 +831,12 @@ class ConnectionState(Generic[ClientT]):
                             guild._scheduled_events.pop(s.id)
                             self.dispatch('scheduled_event_delete', s)
 
+                threads = guild._remove_threads_by_channel(channel_id)
+
+                for thread in threads:
+                    self.dispatch('thread_delete', thread)
+                    self.dispatch('raw_thread_delete', RawThreadDeleteEvent._from_thread(thread))
+
     def parse_channel_update(self, data: gw.ChannelUpdateEvent) -> None:
         channel_type = try_enum(ChannelType, data.get('type'))
         channel_id = int(data['id'])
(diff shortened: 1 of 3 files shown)


discord/state.py
```diff
@@ -831,6 +831,12 @@ def parse_channel_delete(self, data: gw.ChannelDeleteEvent) -> None:
                             guild._scheduled_events.pop(s.id)
                             self.dispatch('scheduled_event_delete', s)
 
+                threads = guild._remove_threads_by_channel(channel_id)
+
+                for thread in threads:
+                    self.dispatch('thread_delete', thread)
+                    self.dispatch('raw_thread_delete', RawThreadDeleteEvent._from_thread(thread))
```
Rapptz, line 838: I get why this is here but this would make it the only time (to my memory?) that a raw event doesn't correspond to its own Discord event I think.

LostLuma: It is the only one yeah. But I think it'd be pretty unfortunate to not receive the raw thread delete event only in certain situations and needing to work around that.

PR #9805, opened 2024-04-25
Flags positional (consume rest)

Vioshim:
## Summary

Implemented positional for flags, this should allow to perform flags after the parameter we're intending to use while still remaining compatible with slash commands.
 
For example (Assuming `--message` is the positional flag, and defaults to `"Hi"` ): 

`?say Hello`
`?say Hello --ping True`
`?say --message Hello --ping True`
`?say --ping True`

## Checklist

- [X] This PR fixes an issue.


Rapptz (changes requested):
The PR seems fine as-is, but do you mind adding documentation to the `commands.rst` file explaining this feature?

PR #9815, opened 2024-05-02
[commands] Add support for type statement and NewType

### LINKED ISSUE #9813: support for 3.12 type aliases
### Summary

Discord 2.3.2 doesn't support python 3.12 type alises

discord.ext.commands

### The Problem

When creating aliases like this
`MemberOrRole = Role | Member`
The transformer has no issues

When creating aliases with the type statement, the transformer raises the error `unsupported type annotation`.
`type MemberOrRole = Role | Member`

The only difference in this example is the statement.

### The Ideal Solution

Being able to just use  `type MemberOrRole = Role | Member`
or more complex situations like `type NumberOrStringSequence [T: (int, float, str)] = Sequence[T]`

### The Current Solution

_No response_

### Additional Context

The type statement for type aliases allows for generic type annotation, without the need for the typing module, and this greatly improves the coding experience for those that use 3.12. Just like mentioned on the [PEP-695](https://docs.python.org/3/whatsnew/3.12.html#whatsnew312-pep695:~:text=the%20new%20syntax%20allows%20declaring%20typevartuple%20and%20paramspec%20parameters%2C%20as%20well%20as%20typevar%20parameters%20with%20bounds%20or%20constraints%3A) it allows for declaring types, with constraints. That leads to a more controlled and readable environment.

bijij:
## Summary

This PR adds support for `typing.NewType` and `type` statement type aliases.
Closes #9813 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -168,7 +168,6 @@ T_co = TypeVar('T_co', covariant=True)
 _Iter = Union[Iterable[T], AsyncIterable[T]]
 Coro = Coroutine[Any, Any, T]
 MaybeAwaitable = Union[T, Awaitable[T]]
-NewTypeType = type(NewType('Foo', int))
 
 
 class CachedSlotProperty(Generic[T, T_co]):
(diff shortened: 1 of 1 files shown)


Rapptz:
Might be worth doing this for app_commands as well?

bijij:
At least doing a quick  test this seems to work out of the box since app_commands also relies on the code in `utils`, unless I'm missing something.

Rapptz:
Oh my mistake. I noticed this is actually within `utils`. For some reason on first read I thought this was under `commands.py`.

PR #9818, opened 2024-05-05
Add support for getting and editing integration_types_config application field

Soheab:
## Summary

This PR adds support for getting and editing the `integration_types_config` field for applications.

I exposed it as `default_x_install_params` properties instead of an object that parses it because I find the inner `oauth2_install_params` object very unnecessary. 

Something to note, setting `default_guild_install_x` and not `default_user_install_x` will disable `default_user_install_x` and I'm not sure if discord.py should handle or tell the user that, if, where and how.

Relevant API Docs:

- https://discord.com/developers/docs/change-log#march-18-2024-api-changes
- https://discord.com/developers/docs/resources/application#application-object

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/appinfo.py
+++ b/discord/appinfo.py
@@ -478,7 +478,6 @@ class AppInfo:
         if integration_types_config:
             payload['integration_types_config'] = integration_types_config
 
-        print("payload: ", integration_types_config)
         data = await self._state.http.edit_application_info(reason=reason, payload=payload)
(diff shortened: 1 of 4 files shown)


Rapptz (commented):
Hm. This API feels.. kinda stinky to be blunt. It's overly verbose and repetitive but I'm unsure how to make it better as it stands. Maybe the separate objects would be cleaner, I don't know.

PR #9829, opened 2024-05-13
Fix merge methods for AppCommandContext and AppInstallationType

JeronimusII:
## Summary

Fixes an issue with the `merge()` method of the `AppCommandContext` and `AppInstallationType` classes which would cause the "other" to always override the flags with True or False even when unspecified (None value).

Its caused by the getter methods being used which always turn None values into False so the condition `other.guild is None` for example never evaluates to True.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/installs.py
+++ b/discord/app_commands/installs.py
@@ -78,8 +78,8 @@ class AppInstallationType:
     def merge(self, other: AppInstallationType) -> AppInstallationType:
         # Merging is similar to AllowedMentions where `self` is the base
         # and the `other` is the override preference
-        guild = self.guild if other.guild is None else other.guild
-        user = self.user if other.user is None else other.user
+        guild = self._guild if other._guild is None else other._guild
+        user = self._user if other._user is None else other._user
         return AppInstallationType(guild=guild, user=user)
 
     def _is_unset(self) -> bool:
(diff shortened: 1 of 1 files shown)


Kile:
Just out of interest, why are not just using `is False`?

JeronimusII:
@Kile
When using `is False` the expression turns into something like:
`guild = self.guild if other.guild is False else other.guild`
Which essentially does the same as:
`guild = self.guild or other.guild`
Or when taking the getter methods into account:
`guild = bool(self._guild) or bool(other._guild)`

Now you cant override a `True` value in `self._guild` with a `False` in `other._guild` since `True or False` evaluates to `True` again.

The (I assume) expected behaviour is that the values of `self` are used unless explicitly overridden by the `other` values (not `None`).
Where the problem currently is that the getter methods turn a `None` value into `False` which causes the values to be always overridden with the ones from `other` (with `False` instead of `None`).
Which makes omitting a parameter in a decorator like `@discord.app_commands.allowed_contexts()` to keep the default value set in the constructor of `discord.ext.commands.Bot` (`allowed_contexts` parameter) not possible.
https://github.com/Rapptz/discord.py/blob/b9dc85e6f72f26dcc17f5d0808cfe35d439b3104/discord/app_commands/installs.py#L79-L80
https://github.com/Rapptz/discord.py/blob/b9dc85e6f72f26dcc17f5d0808cfe35d439b3104/discord/mentions.py#L140-L142

PR #9831, opened 2024-05-15
Fix Webhook's raising AttributeError when ``bot`` is not passed in ``Webhook.from_url``

DA-344:
## Summary

When using ``discord.Webhook.from_url`` and passing ``session`` instead of ``bot``, and then sending a message it raises an ``AttributeError``.

This fixes that by adding a "false" ``_get_poll`` method in ``_WebhookState``.

## Checklist

- [x] This PR fixes an issue.


DA-344:
Moved to #9832

PR #9833, opened 2024-05-16
feat: Removed endpoints that are no longer usable by bots.

Isabe1le:
## Summary

There were a few API methods that are no longer usable by bots left in the library. As this library no longer supports self-bots, these should be removed.

## Checklist

- [x] This PR fixes an issue.


Rapptz:
Thanks for the PR but I'm not accepting breaking changes at this time.

PR #9836, opened 2024-05-17
Add reaction type to raw events and users iterator

Puncher1:
## Summary

This PR adds the reaction type to `RawReactionActionEvent` and `reaction.users` async iterator.

Related PRs:
- https://github.com/discord/discord-api-docs/pull/6377/
- https://github.com/discord/discord-api-docs/pull/6873/

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/reaction.py
+++ b/discord/reaction.py
@@ -223,7 +223,7 @@ class Reaction:
             For pagination, reactions are sorted by member.
         type: Optional[:class:`ReactionType`]
             The type of reaction to return users from.
-            If not provided, returns users of reactions with type ``normal``.
+            If not provided, Discord only returns users of reactions with type ``normal``.
 
             .. versionadded:: 2.4
 
@@ -249,9 +249,6 @@ class Reaction:
         if limit is None:
             limit = self.count
 
-        if type is None:
-            type = ReactionType.normal
-
         while limit > 0:
             retrieve = min(limit, 100)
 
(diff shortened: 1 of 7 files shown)


Rapptz:
This has conflicts now.

Puncher1:
> This has conflicts now.

Resolved

discord/reaction.py
```diff
@@ -243,6 +249,9 @@ async def users(
         if limit is None:
             limit = self.count
 
+        if type is None:
```
Vioshim, line 252: This part is breaking, it shouldn't be Normal reaction if not provided but rather None, the API returns either normal/boost reactions if the type isn't provided.

Rapptz: This can be fixed by not providing the `type` query parameter if this is `None`.

Puncher1: > This part is breaking, it shouldn't be Normal reaction if not provided but rather None, the API returns either normal/boost reactions if the type isn't provided.

Are you sure? In the docs it says `0` for the default value which is `normal`.

mikeshardmind: docs are wrong here, you can observe this pretty easily by just trying current library behavior (which doesn't pass type) on a reaction that exists both as a super reaction and a normal one. You'll get both back.

mikeshardmind: sorry, someone was messing with reactions on the same message I was testing with at the time. I should have double checked who the users were. default value appears to be correct

Vioshim: My bad, had a similar situation myself when testing, the default is correct, my bad for delaying the PR ^^;

Puncher1: Alright. Well maybe it still makes sense to not pass the value if it's `None` to let Discord decide and reflect it in the docs as so.

Puncher1: I changed it now like I described above. Thanks

Rapptz (approved):

PR #9837, opened 2024-05-18
Account for user installations in Channel.permissions_for

Kile:
## Summary
Currently `Channel.permissions_for` do not account for user installed apps. So in dms they will have `_dm_permissions()` permissions even though they cannot see the chat history and in guilds the bot is not on, the library will error because it attempts to access guild data it does not have access to. To address this, I am introducing `Permissions._user_installed_permissions()` which includes a default for what permissions a user installed app has. This method is used whenever permissions are checked for the Bot/App when it is not part of the server or dm the permissions are checked for.

I know Discord provides this permission information in [Interaction.app_permissions](https://discordpy.readthedocs.io/en/stable/interactions/api.html#discord.Interaction.app_permissions) and [Context.bot_permissions](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#discord.ext.commands.Context.bot_permissions), but I think this is still important for compatibility and "migrating" to user installed commands (the reason I encountered this in the first place because I changed one of my commands to support user installs.). Furthermore it is never great to get an internal dpy error without explanation.

Please let me know if my code style needs change or the permissions in my new method need changing.

## Checklist

- [x] This PR fixes an issue.

## Minimal reproducible example

Here is some code to reproduce the issues d.py currently has:
```py
import discord
import logging

from asyncio import run

logging.basicConfig(level=logging)

class SyncingBot(commands.Bot):
    async def setup_hook(self):
        synced = await self.tree.sync()
        print(f"Synced with data:\n{synced}")

class TestCog(commands.Cog):
    def __init__(self, bot):
        self.bot = bot

    @discord.app_commands.command()
    @discord.app_commands.allowed_installs(guilds=True, users=True)
    @discord.app_commands.allowed_contexts(dms=True, guilds=True, private_channels=True)
    async def test_slash(self, interaction: discord.Interaction):
        # Inaccurate perms in dms (except with the bot), errors in servers the bot is not in
        await interaction.response.send_message(interaction.channel.permissions_for(interaction.user), ephemeral=True)

    @commands.hybrid_command()
    @discord.app_commands.allowed_installs(guilds=True, users=True)
    @discord.app_commands.allowed_contexts(dms=True, guilds=True, private_channels=True)
    async def test_hybrid(self, ctx: commands.Context):
        # Inaccurate perms in dms (except with the bot), errors in servers the bot is not in
        await ctx.send(ctx.channel.permissions_for(ctx.me), ephemeral=True)

async def main():
    bot = SyncingBot(command_prefix="!", intents=discord.Intents.default(), chunk_guilds_at_startup=False)
    await bot.add_cog(TestCog(bot))
    await bot.start("TOKEN")

run(main())
```

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -749,10 +749,10 @@ class GuildChannel:
             return Permissions.all()
 
         default = self.guild.default_role
-        if default:
-            base = Permissions(default.permissions.value)
-        else:
-            base = Permissions._user_installed_permissions(in_guild=True)
+        if default is None:
+            return Permissions._user_installed_permissions(in_guild=True)
+
+        base = Permissions(default.permissions.value)
 
         # Handle the role case first
         if isinstance(obj, Role):
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -717,6 +717,9 @@ class GuildChannel:
 
         .. versionchanged:: 2.4
             User installed apps are now taken into account.
+            The permissions returned by a user installed app mirror the 
+            permissions Discord returns in :attr:`~Interaction.app_permissions`,
+            though it is recommended to use that instead.
 
         Parameters
         ----------
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -753,7 +753,11 @@ class GuildChannel:
 
         default = self.guild.default_role
         if default is None:
-            return Permissions._user_installed_permissions(in_guild=True)
+
+            if self._state.self_id == obj.id:
+                return Permissions._user_installed_permissions(in_guild=True)
+            else:
+                return Permissions.none()
 
         base = Permissions(default.permissions.value)
 
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -752,7 +752,7 @@ class GuildChannel:
         if default:
             base = Permissions(default.permissions.value)
         else:
-            base = Permissions._user_installed_permissions()
+            base = Permissions._user_installed_permissions(in_guild=True)
 
         # Handle the role case first
         if isinstance(obj, Role):
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -699,6 +699,7 @@ class GuildChannel:
         - Member overrides
         - Implicit permissions
         - Member timeout
+        - User installed app
 
         If a :class:`~discord.Role` is passed, then it checks the permissions
         someone with that role would have, which is essentially:
@@ -713,6 +714,9 @@ class GuildChannel:
 
         .. versionchanged:: 2.0
             ``obj`` parameter is now positional-only.
+
+        .. versionchanged:: 2.4
+            User installed apps are now taken into account.
 
         Parameters
         ----------
@@ -745,7 +749,10 @@ class GuildChannel:
             return Permissions.all()
 
         default = self.guild.default_role
-        base = Permissions(default.permissions.value)
+        if default:
+            base = Permissions(default.permissions.value)
+        else:
+            base = Permissions._user_installed_permissions()
 
         # Handle the role case first
         if isinstance(obj, Role):
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -717,9 +717,9 @@ class GuildChannel:
 
         .. versionchanged:: 2.4
             User installed apps are now taken into account.
-            The permissions returned by a user installed app mirror the
+            The permissions returned for a user installed app mirrors the
             permissions Discord returns in :attr:`~discord.Interaction.app_permissions`,
-            though it is recommended to use that instead.
+            though it is recommended to use that attribute instead.
 
         Parameters
         ----------
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -717,8 +717,8 @@ class GuildChannel:
 
         .. versionchanged:: 2.4
             User installed apps are now taken into account.
-            The permissions returned by a user installed app mirror the 
-            permissions Discord returns in :attr:`~Interaction.app_permissions`,
+            The permissions returned by a user installed app mirror the
+            permissions Discord returns in :attr:`~discord.Interaction.app_permissions`,
             though it is recommended to use that instead.
 
         Parameters
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -288,10 +288,12 @@ class TextChannel(discord.abc.Messageable, discord.abc.GuildChannel, Hashable):
         return self._state._get_message(self.last_message_id) if self.last_message_id else None
 
     @overload
-    async def edit(self) -> Optional[TextChannel]: ...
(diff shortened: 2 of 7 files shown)


Kile:
Ok no idea how I fix those two tests failing

LostLuma:
Run black on channel.py, then those should pass.

Kile:
Well that didn't work

z03h:
Discord provides the permissions for your app/bot in the current channel with [`Interaction.app_permissions`](https://discordpy.readthedocs.io/en/stable/interactions/api.html#discord.Interaction.app_permissions). Hybrid commands can use [`Context.bot_permissions`](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#discord.ext.commands.Context.bot_permissions).

Kile:
That's fair, but for migrating purposes and because it seems not great to have internal error I still think this has a place

LostLuma:
> Well that didn't work

Sorry, I forgot that discord.py runs a specific version (22.6) in CI, which formats stuff a bit different. Before your `Make tests pass` commit it seems the thing it doesn't like is erroneous whitespace in [this docstring](https://github.com/Rapptz/discord.py/blob/403de1624da1dc61bd072e040d4608dd84b2e930/discord/channel.py#L3012) on the line I linked.

Rapptz:
> That's fair, but for migrating purposes and because it seems not great to have internal error I still think this has a place

We debated this long before but I wasn't sure if it was worth going further, and I still don't. Code that is installed and used in these new contexts should just use the correct code.

Kile:
Just crossposting from bikeshedding here
> I would argue keeping `Channel.permissions_for` the same is, while not a breaking change of its own, very likely going to cause some issues since if someone just adds the user install decorator to their command no error will appear so they won't check the docs where it tells them not to use it (unlike for me where there was an error, otherwise I would have not figured this out) so unless the console prints a warning (which I dont think you want) this not being accurate will remain unnoticed.

> By O(n) I am assuming you mean `if not obj in self.recipients:`. Sure, in theory O(n) is not great. But this is only for gdms and dms, so 1 < n < 10 which is not a very big n at all. I think this would be much more applicable if n would be guild members or smth but since the maximum is 9 I think that is a worthy tradeoff for accurate permissions. 
This is definititely annoying from Discord to do but I don't see a big tradeoff for dpy to fix where Discord went wrong

> Also regardless of migration I think you also value consistency. It would not be very consistent if Channel.permissions_for was accurate for some cases but not for others

Kile:
As discussed I have removed this for dms and gems. I have kept in the `in_guild` parameter for the new class method just because I felt like it would be misleading to leave it out, acting as though bots had tts perms in anywhere if user installed. Now to check those test failures.

Kile:
I have also added a recommendation to use Interaction.app_permissions instead in the docs. I didn't feel like more explanation for user apps for dms was necessary since, as the docstring already mentions, dm permissions are not really a thing and the default is close enough.

discord/channel.py
```diff
@@ -3019,13 +3019,13 @@ def permissions_for(self, obj: Any = None, /) -> Permissions:
...
-            The user to check permissions for. This parameter is ignored
-            but kept for compatibility with other ``permissions_for`` methods.
+            The user to check permissions for.
 
         Returns
         --------
         :class:`Permissions`
-            The resolved permissions.
+            The user to check permissions for. This parameter is ignored
+            but kept for compatibility with other ``permissions_for`` methods.
```
Rapptz, line 3028: Why did you move this..?

Rapptz (changes requested):

discord/abc.py
```diff
@@ -745,6 +752,9 @@ def permissions_for(self, obj: Union[Member, Role], /) -> Permissions:
             return Permissions.all()
 
         default = self.guild.default_role
+        if default is None:
+            return Permissions._user_installed_permissions(in_guild=True)
```
z03h, line 756: Doesn't this need a check to ensure it's the bot user? This could return incorrect permissions if one uses `permissions_for` on a user other than the bot.

discord/channel.py
Kile: My bad, removed it when I implemented this for dms and re-added it at the wrong place

discord/abc.py
Kile: That is a good point.
What do you think would be the most sensible way of returning user permissions? Just 0? Also, how do I access the bot/bot ID in this method?

z03h: You can get the bot id from `channel._state.self_id`. I don't have any better ideas other than 0.

discord/channel.py
```diff
@@ -3019,13 +3019,13 @@ def permissions_for(self, obj: Any = None, /) -> Permissions:
...
-            The user to check permissions for.
+            The user to check permissions for. This parameter is ignored
+            but kept for compatibility with other ``permissions_for`` methods.
 
         Returns
         --------
         :class:`Permissions`
-            The user to check permissions for. This parameter is ignored
-            but kept for compatibility with other ``permissions_for`` methods.
+        The resolved permissions.
```
Rapptz, line 3028: Needs to be indented.

LeoCx1000:
Wouldn't this be returning wrong information? The following wouldn't work with the current permissions returned, even though it's a sane assumption to make:-

```py
if ctx.channel.permissions_for(ctx.me).read_message_history:
    async for message in ctx.channel.history(): ...
```

Same goes for `.send_messages`- actually also for every other one since the bot isn't in the guild. This would lead to wrong assumptions potentially being made.

Kile:
> Wouldn't this be returning wrong information? The following wouldn't work with the current permissions returned, even though it's a sane assumption to make:-
> 
> ```python
> if ctx.channel.permissions_for(ctx.me).read_message_history:
>     async for message in ctx.channel.history(): ...
> ```
> 
> Same goes for `.send_messages`- actually also for every other one since the bot isn't in the guild. This would lead to wrong assumptions potentially being made.

`read_message_history` specifically should be `False` unless it is `True` is `read_messages` if `True`. And `read_messages` *has* to be `True` otherwise the permissions just evaluate to `0`. But to be honest I now see a decent argument to be made for it to just return `0`, makes it easier since the lib does not have to keep up with however Discord decides to handle this or whether the user is the bot or not. 

I think that would be a little inconvenient for a number of things though such as
- In the error handler send a message when an error occurred if you have send message perms (otherwise no error message). Technically you *can* send messages as interaction replies and followups
- Not sending an embed/image with a command response because permissions for attach_image/embed are `False`

Though I suppose you can work around that with an extra if statement.

discord/abc.py
Kile: Cheers, that should be implemented now

Rapptz (approved):

PR #9842, opened 2024-05-22
Add VoiceClient.wait_until_done

imayhaveborkedit:
## Summary

Adds a new function: `VoiceClient.wait_until_done`.  As the name suggests, it waits until the player is done, similar to the other `wait_until_` functions.

There are probably some places in the docs this could be mentioned but i'll add another commit for that later.

## Checklist

- [ ] This PR fixes an issue.


GnomedDev:
Previously when this was asked for as a part of `VoiceClient.play`, it was turned down for performance reasons. Since this affects all code paths, not just when wait_until_done is called, how are those performance issues mitigated?

Source: https://discord.com/channels/336642139381301249/336642776609456130/958395836751020032

PR #9843, opened 2024-05-24
[commands] fix HelpCommand not carrying over checks

### LINKED ISSUE #9827: HelpCommand checks gets overwritten when set to bot.help_command
### Summary

When you use HelpCommand.add_check, and set to bot.help_command, it overwritten

### Reproduction Steps

1. initialize the help command without `checks` kwarg
2. use HelpCommand.add_check, i.e in init or outside
3. assign to bot.help_command

### Minimal Reproducible Code

```python
help_command = commands.DefaultHelpCommand()
help_command.add_check(lambda ctx: False)
print(help_command._command_impl.checks) # [<function <lambda> at 0x000001E67FEE50D0>]
bot.help_command = help_command
print(help_command._command_impl.checks) # []
help_command.add_check(lambda ctx: False)
print(help_command._command_impl.checks) # [<function <lambda> at 0x000001E67FEE50D0>]
```

### Expected Results

checks should be carried over after bot.help_command is set and not get overwritten.

### Actual Results

checks are overwritten when setting to bot.help_command. This results in the checks not being applied when invoking the help command.

### Intents

discord.Intents.all()

### System Information

- Python v3.9.13-final
- discord.py v2.4.0-alpha
    - discord.py metadata: v2.4.0a5025+g041abf8b
- aiohttp v3.9.5
- system info: Windows 10 10.0.22631

### Checklist

### Additional Context

checks do carried over if you set the `checks` kwarg on `command_attrs`.
```py
help_command = commands.DefaultHelpCommand(command_attrs={'checks': [lambda ctx: False]})
help_command.add_check(lambda ctx: False)
bot.help_command = help_command
print(help_command._command_impl.checks)  # [<function <lambda> at 0x0000027E06E35160>, <function <lambda> at 0x0000027E06F089D0>]
```
Happens since `HelpCommand._add_to_bot` only copies `command_attrs` but left out `checks` on the actual command impl [here](https://github.com/Rapptz/discord.py/blob/master/discord/ext/commands/help.py#L380). Since command impl there gets overwritten, the checks disappear.

A simple fix would just be `command.checks = self._command_impl.checks` after [this line](https://github.com/Rapptz/discord.py/blob/master/discord/ext/commands/help.py#L380) but I don't like my approach

z03h:
## Summary

Calling `HelpCommand.add_check` before setting the help command would not carry over checks since a new command_impl is created.

Fixes #9827 
## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/help.py
+++ b/discord/ext/commands/help.py
@@ -297,6 +297,11 @@ class _HelpCommandImpl(Command):
         # Revert `on_error` to use the original one in case of race conditions
         self.on_error = self._injected.on_help_command_error
 
+    def update(self, **kwargs: Any) -> None:
+        cog = self.cog
+        self.__init__(self._original, **dict(self.__original_kwargs__, **kwargs))
+        self.cog = cog
+
 
 class HelpCommand:
     r"""The base implementation for help command formatting.
@@ -377,9 +382,8 @@ class HelpCommand:
         return obj
 
     def _add_to_bot(self, bot: BotBase) -> None:
-        command = _HelpCommandImpl(self, **self.command_attrs)
-        bot.add_command(command)
-        self._command_impl = command
+        self._command_impl.update(**self.command_attrs)
+        bot.add_command(self._command_impl)
 
     def _remove_from_bot(self, bot: BotBase) -> None:
         bot.remove_command(self._command_impl.name)


mikeshardmind:
It's not clear to me that this should be the behavior here, if I replace the help command, I would find checks from the prior command still existing to be surprising behavior, as checks belong to command objects and I'm replacing the command object with a different one.

z03h:
A little bit. This fixes help command checks being enjoy if you call help.add_check before bot.help_command=help.

Currently a new command instance gets created right before being added and doesn't copy the checks from the previous instance so they're lost.

PR #9846, opened 2024-05-26
Add notes about contexts and installation_types only being available for global commands

Soheab:
## Summary

**While in preview,** only global commands can be made into user installable apps and I think the docs should note that on the guild(s) decorator and kwargs.

## Checklist

- [ ] This PR fixes an issue.


Rapptz (changes requested):
The wording is fine, but can you expand "kwarg" to "keyword argument" in the documentation.

Rapptz (approved):
Thanks

PR #9852, opened 2024-05-29
Add note about `archiver_id`

tailoric:
## Summary

Discord removed the `archiver_id` field from their thread metadata.
See following discussion thread and api docs commit:
https://github.com/Rapptz/discord.py/discussions/9850
https://github.com/discord/discord-api-docs/commit/82c98ba5699b69706cfb493a7c6e08b67ae1a8f3

This pull request adds a note explaining that the value will always be `None` and that the id can only be acquired via the audit log now
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/threads.py
+++ b/discord/threads.py
@@ -123,8 +123,7 @@ class Thread(Messageable, Hashable):
         The user's ID that archived this thread.
 
         .. note::
-            This field has been removed from the :ddocs:`Thread Metadata Object <resources/channel#thread-metadata-object>`.
-            The ``archiver_id``, therefore, will always be ``None`` and can only be obtained via the audit log now.
+            Due to an API change, the ``archiver_id`` will always be ``None`` and can only be obtained via the audit log now.
 
     auto_archive_duration: :class:`int`
         The duration in minutes until the thread is automatically hidden from the channel list.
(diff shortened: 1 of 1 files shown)


discord/threads.py
```diff
@@ -121,6 +121,11 @@ class Thread(Messageable, Hashable):
         This is always ``True`` for public threads.
     archiver_id: Optional[:class:`int`]
         The user's ID that archived this thread.
+
+        .. note::
+            This field has been removed from the :ddocs:`Thread Metadata Object <resources/channel#thread-metadata-object>`.
+            The ``archiver_id``, therefore, will always be ``None`` and can only be obtained via the audit log now.
```
Rapptz, line 127:
```suggestion
        .. note::
            Due to an API change, the ``archiver_id``, therefore, will always be ``None`` and can only be obtained via the audit log now.
```

Simplify the wording.

Rapptz (changes requested):

discord/threads.py
```diff
@@ -121,6 +121,10 @@ class Thread(Messageable, Hashable):
         This is always ``True`` for public threads.
     archiver_id: Optional[:class:`int`]
         The user's ID that archived this thread.
+
+        .. note::
+            Due to an API change, the ``archiver_id`` will always be ``None`` and can only be obtained via the audit log now.
```
Rapptz, line 126:
```suggestion
            Due to an API change, the ``archiver_id`` will always be ``None`` and can only be obtained via the audit log.
```

Good catch, but I forgot to remove this too.

Rapptz (approved):
Thanks

PR #9855, opened 2024-06-04
Attach command on Choice for translator

Soheab:
## Summary

This PR adds a `command` property to `app_commands.Choice` that can only be used to access the.. command inside the translator's translate method. 

I wasn't sure on always attaching the command on the Choice since it doesn't store any state and is pretty lightweight.

### Why?

Currently, there is no way to access the command if `context.location` is `TranslationContextLocation.choice_name` in `app_commands.Translator.translate` which is useful for implementations that store the strings per command name.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/models.py
+++ b/discord/app_commands/models.py
@@ -87,6 +87,7 @@ if TYPE_CHECKING:
     from ..channel import TextChannel
     from ..threads import Thread
     from ..user import User
+    from ..app_commands import Command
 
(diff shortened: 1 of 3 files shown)


Rapptz:
`locale_str` contains kwargs which allows you to specify as much extra information as possible when constructing it, so this should really not be necessary, especially since it kind of bloats up the model.

PR #9864, opened 2024-06-14
Update Premium Buttons

DA-344:
## Summary

So Discord updated the premium interacion buttons and the current implementation needs some changes (+ some fixes).

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ui/button.py
+++ b/discord/ui/button.py
@@ -266,7 +266,6 @@ def button(
     style: ButtonStyle = ButtonStyle.secondary,
     emoji: Optional[Union[str, Emoji, PartialEmoji]] = None,
     row: Optional[int] = None,
-    sku_id: Optional[int] = None,
 ) -> Callable[[ItemCallbackType[V, Button[V]]], Button[V]]:
     """A decorator that attaches a button to a component.
 
--- a/discord/ui/button.py
+++ b/discord/ui/button.py
@@ -24,7 +24,7 @@ DEALINGS IN THE SOFTWARE.
 
 from __future__ import annotations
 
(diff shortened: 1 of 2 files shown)


docs/interactions/api.rst
```diff
@@ -336,7 +336,7 @@ Enumerations
         Represents a link button.
     .. attribute:: premium
 
-        Represents a gradient button denoting that buying a SKU is
+        Represents a blurple button denoting that buying a SKU is
```
LeoCx1000, line 339: Why does this need to mention a colour at all? It could simply be omitted.

discord/ui/button.py
```diff
@@ -275,11 +276,11 @@ def button(
 
     .. note::
 
-        Buttons with a URL cannot be created with this function.
+        Buttons with a URL or SKU cannot be created with this function.
         Consider creating a :class:`Button` manually instead.
         This is because buttons with a URL do not have a callback
         associated with them since Discord does not do any processing
-        with it.
+        with it, and buttons with a SKU are handled by Discord.
```
Rapptz, line 283: This should probably be:

```
    .. note::

        Buttons with a URL or an SKU cannot be created with this function.
        Consider creating a :class:`Button` manually instead.
        This is because these buttons cannot have a callback
        associated with them since Discord does not do any processing
        with them.
```

docs/interactions/api.rst
Rapptz:
```suggestion
        Represents a button denoting that buying a SKU is
```

Rapptz (changes requested):

Rapptz (approved):

PR #9870, opened 2024-06-25
Revert "Set socket family of connector to AF_INET"

oliver-ni:
This change was made since Discord doesn't support IPv6, and there were concerns about clients with DNS64 enabled without NAT64.

However, this breaks hosts who don't have v4 connectivity and are _actually_ running NAT64 (me!)

Having DNS64 without NAT64 is really an issue on the client's end. It would break far more than just discord.py, so I don't think we should be concerned about those cases. I'm proposing this change be reverted.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -48,7 +48,6 @@ from typing import (
 from urllib.parse import quote as _uriquote
 from collections import deque
 import datetime
-import socket
 
 import aiohttp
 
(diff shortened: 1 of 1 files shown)


SebbyLaw (approved):
LGTM

Rapptz:
I'm not a fan of this solution. This should probably be a `Client` setting instead with the default being `False`.

Rapptz:
I've given this some more thought and I'm just going to go ahead and allow it.

PR #9871, opened 2024-06-27
fix: Polls limiting duration at 7 days

DA-344:
## Summary

So Discord recently allowed polls to be 32 days long so the check in ``Poll._from_data`` that "fixes" the poll duration at 7 days should be removed.

This also makes sure that if Discord changes the maximum duration for polls again there won't be any problems.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/poll.py
+++ b/discord/poll.py
@@ -388,9 +388,6 @@ class Poll:
         # self.created_at = message.created_at
         # duration = self.created_at - expiry
 
-        if (duration.total_seconds() / 3600) > 168:  # As the duration may exceed little milliseconds then we fix it
-            duration = datetime.timedelta(days=7)
-
         self = cls(
             duration=duration,
             multiple=multiselect,


discord/poll.py
```diff
@@ -388,9 +388,6 @@ def _from_data(cls, *, data: PollPayload, message: Message, state: ConnectionSta
         # self.created_at = message.created_at
         # duration = self.created_at - expiry
 
-        if (duration.total_seconds() / 3600) > 168:  # As the duration may exceed little milliseconds then we fix it
-            duration = datetime.timedelta(days=7)
-
```
AbstractUmbra, line 393: Why are we removing this clamp instead of altering it to meet the new criteria?

AbstractUmbra: ...and upon such time we would remove said clamp, we shouldn't be doing it for features we hope discord may release.

DA-344: The check is removed so polls aren't capped at 7 days when obtained via API or gateway, in my opinion updating this every time Discord updates the polls maximum duration would not be right.

AbstractUmbra: If that is the argument now then I wonder why it was implemented and accepted in the first place.

mikeshardmind: It's unlikely discord ever adds an uncapped poll. The feature doesn't make sense. The point of a poll is getting an answer from a group of people, if it never ends, there's no answer ever.

Rapptz: It's fine to remove the check and let the server handle it, that's what the library usually does anyway except within `discord.ui`

DA-344:
Damn this thing closed lol, sorry about that gonna reopen it

PR #9875, opened 2024-07-03
Add more tests for colour, embeds, files, buttons and modals

junah201:
## Summary

I added more test case for colour, embeds, files, buttons and modals

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/tests/test_colour.py
+++ b/tests/test_colour.py
@@ -44,6 +44,7 @@ import pytest
         ('rgb(20%, 24%, 56%)', 0x333D8F),
         ('rgb(20%, 23.9%, 56.1%)', 0x333D8F),
         ('rgb(51, 61, 143)', 0x333D8F),
+        ('0x#333D8F', 0x333D8F),
     ],
 )
 def test_from_str(value, expected):
@@ -53,6 +54,7 @@ def test_from_str(value, expected):
 @pytest.mark.parametrize(
     ('value'),
     [
+        None,
         'not valid',
         '0xYEAH',
         '#YEAH',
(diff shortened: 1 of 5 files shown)


tests/test_colour.py
```diff
@@ -62,8 +64,85 @@ def test_from_str(value, expected):
...
+        (1, 0xFFCD00),
+        (21, 0xFFFC00),
+        ("ju", 0xFF4600),
+        (-0.5, 0x1900FF),
+        (143.96, 0xDBFF00),
+        (b"Hello", 0xFF0E00),
+    ],
+)
+def test_random_colour(seed, expected):
+    assert discord.Colour.random(seed=seed).value == expected
```
Rapptz, line 133:
```suggestion
```

I don't think this should be a guaranteed thing to test for.

tests/test_colour.py
```diff
@@ -62,8 +64,85 @@ def test_from_str(value, expected):
...
+        (discord.Colour.greyple(), 0x99AAB5),
+        (discord.Colour.dark_theme(), 0x313338),
+        (discord.Colour.fuchsia(), 0xEB459E),
+        (discord.Colour.yellow(), 0xFEE75C),
+        (discord.Colour.dark_embed(), 0x2B2D31),
+        (discord.Colour.light_embed(), 0xEEEFF1),
+        (discord.Colour.pink(), 0xEB459F),
+    ],
+)
+def test_static_colours(value, expected):
```
Rapptz, line 117: I just want to make a note that some of these have changed before, e.g. `blurple`. I'm not sure what the API stability on these are but we have changed them where it makes sense historically. This has no bearing on the tests being here but I just wanted to comment on it.

Rapptz:
Sorry for the late response on this PR, huge backlog.

Issue #9878, opened 2024-07-05
Adding the 26th option did not raise a ValueError on Select


junah201:
### Summary

Adding the 26th option did not raise a ValueError on Select

### Reproduction Steps

Following PR #9875, I was writing test code for `discord.ui.Select`. At that time, I recognize that adding the 26th option didn't raise a ValueError. In fact, `discord.ui.Select` raises a ValueError when adding the 27th option.

### Minimal Reproducible Code

```python
import discord

bot = discord.Client(
    intents=discord.Intents.all()
)


@bot.event
async def on_message(message: discord.Message):
    if message.content == "!test":
        select = discord.ui.Select()
        for i in range(1, 25 + 1):
            select.add_option(label=f"Test {i}", value=str(i))

        # add 26th option
        select.add_option(label="26", value="26")

        view = discord.ui.View()
        view.add_item(select)

        await message.channel.send("Select Test", view=view)

bot.run("token")
```

### Expected Results

A valueError is expected to occur when the 26th option is added to the select, not when sending the view to discord.

```py
ValueError: maximum number of options already provided
```

### Actual Results

When I send Select with 26 options, `discord.errors.HTTPException` occured.

```bash
discord.errors.HTTPException: 400 Bad Request (error code: 50035): Invalid Form Body
In components.0.components.0.options: Must be between 1 and 25 in length.
```

### Intents

discord.Intents.all

### System Information

$ python -m discord -v
- Python v3.10.7-final
- discord.py v2.5.0-alpha
    - discord.py metadata: v2.5.0a5058+ge5324907
- aiohttp v3.7.4.post0
- system info: Windows 10 10.0.22631

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.

### Additional Context

https://github.com/Rapptz/discord.py/blob/3a8063602919813e91d19101f9b5a0e9e7438d4b/discord/ui/select.py#L502-L503

junah201:
@MCausc78 
I understand your point, but since adding the 27th option raises a ValueError, it seems more like an unintended behavior rather than a question of responsibility. Additionally, regardless of whose responsibility it is, discord.py does appear to be performing this validation (with some bug).


https://github.com/Rapptz/discord.py/blob/3a8063602919813e91d19101f9b5a0e9e7438d4b/discord/ui/view.py#L169-L175

[...]

PR #9879, opened 2024-07-05
Fix incorrect Select max option condition

junah201:
## Summary

Resolve : #9878 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ui/select.py
+++ b/discord/ui/select.py
@@ -499,7 +499,7 @@ class Select(BaseSelect[V]):
             The number of options exceeds 25.
         """
 
-        if len(self._underlying.options) > 25:
+        if len(self._underlying.options) >= 25:
             raise ValueError('maximum number of options already provided')
 
         self._underlying.options.append(option)
--- a/tests/test_ui_selects.py
+++ b/tests/test_ui_selects.py
@@ -1,0 +1,39 @@
+"""
+The MIT License (MIT)
+
+Copyright (c) 2015-present Rapptz
+
+Permission is hereby granted, free of charge, to any person obtaining a
+copy of this software and associated documentation files (the "Software"),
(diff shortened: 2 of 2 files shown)


discord/ui/select.py
```diff
@@ -499,7 +499,7 @@ def append_option(self, option: SelectOption) -> None:
             The number of options exceeds 25.
         """
 
-        if len(self._underlying.options) > 25:
+        if len(self._underlying.options) >= 25:
```
DA-344, line 502: This is not right..? Discord allows 25 options so why do we raise a ValueError when we have 25?

AbstractUmbra: Because the current code path is adding another item...?

junah201: @DA-344 @AbstractUmbra 

The `append_option` function executes before the option is added to `self._underlying.options`, so `>= 25` is correct. You can reproduce this issue using the bot code and test code provided below.

### Bot code

```py
import discord

bot = discord.Client(
    intents=discord.Intents.all()
)

@bot.event
async def on_message(message: discord.Message):
    if message.content == "!test":
        select = discord.ui.Select()
        for i in range(1, 25 + 1):
            select.add_option(label=f"Test {i}", value=str(i))

        # add 26th option
        select.add_option(label="26", value="26")

        view = discord.ui.View()
        view.add_item(select)

        await message.channel.send("Select Test", view=view)

bot.run("token")
```

### Test code

```py
@pytest.mark.asyncio
async def test_add_option():
    select = discord.ui.Select()

    for i in range(1, 25 + 1):
        select.add_option(label=str(i), value=str(i))

    with pytest.raises(ValueError):
        select.add_option(label="26", value="26")
  ```

PR #9892, opened 2024-07-22
Add support for message snapshots used in message forwarding

redmvc:
## Summary

[Message forwarding](https://discord.com/developers/docs/resources/channel#message-reference-content-attribution-forwards) was introduced on July 18, 2024. Messages now have a new optional parameter, `message_snapshots`, which is an array of [Message Snapshots](https://discord.com/developers/docs/resources/channel#message-snapshot-object) containing immutable snapshots of a message being forwarded. Such messages will, additionally, always have a reference to the message being forwarded, and that reference will have `type = 1`.

This PR adds a `MessageSnapshot` class and a `message_snapshots` parameter to the `Message` class. It also adds the `type` parameter to the `MessageReference` class, in order to track whether a reference is a reply or a forward.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -218,6 +218,11 @@ class ChannelType(Enum):
         return self.name
 
 
+class MessageReferenceType(Enum):
+    default = 0
+    forward = 1
+
+
 class MessageType(Enum):
     default = 0
     recipient_add = 1
--- a/discord/message.py
+++ b/discord/message.py
@@ -32,6 +32,7 @@ from os import PathLike
 from typing import (
     Dict,
     TYPE_CHECKING,
+    Literal,
     Sequence,
     Union,
     List,
@@ -72,6 +73,7 @@ if TYPE_CHECKING:
         Message as MessagePayload,
         Attachment as AttachmentPayload,
         MessageReference as MessageReferencePayload,
+        MessageSnapshot as MessageSnapshotPayload,
         MessageApplication as MessageApplicationPayload,
         MessageActivity as MessageActivityPayload,
         RoleSubscriptionData as RoleSubscriptionDataPayload,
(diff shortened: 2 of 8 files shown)


Soheab:
Can't this use the existing Message class? With a potential `_from_snapshot` classmethod

redmvc:
Message snapshots don't have all Message information; in particular, they don't have an author.

discord/message.py
```diff
@@ -486,10 +572,19 @@ class MessageReference:
...
     """
 
-    __slots__ = ('message_id', 'channel_id', 'guild_id', 'fail_if_not_exists', 'resolved', '_state')
+    __slots__ = ('type', 'message_id', 'channel_id', 'guild_id', 'fail_if_not_exists', 'resolved', '_state')
 
-    def __init__(self, *, message_id: int, channel_id: int, guild_id: Optional[int] = None, fail_if_not_exists: bool = True):
+    def __init__(
+        self,
+        *,
+        type: int,
```
AbstractUmbra, line 580: If on line 597 we're using the Enum directly with `try_enum` when why does this parameter take an `int` instead of the relevant Enum type?

discord/message.py
```diff
@@ -529,6 +625,7 @@ def from_message(cls, message: PartialMessage, *, fail_if_not_exists: bool = Tru
             A reference to the message.
         """
         self = cls(
+            type=0,
```
AbstractUmbra, line 628: See earlier comment about using Enum values rather than their raw int.

discord/enums.py
```diff
@@ -218,6 +218,11 @@ def __str__(self) -> str:
         return self.name
 
 
+class MessageReferenceType(Enum):
+    default = 0
```
AbstractUmbra, line 222: Is `default` the name for a reply reference in the API, or is this an <us> creation?
If the latter, I believe it can be named better, like `reply`.

discord/message.py
```diff
@@ -452,6 +455,87 @@ def guild_id(self) -> Optional[int]:
...
+    -----------
+    type: :class:`MessageType`
+        The type of the snapshotted message.
+    content: :class:`str`
+        The actual contents of the snapshotted message.
+    embeds: List[:class:`Embed`]
+        A list of embeds the snapshotted message has.
+    attachments: List[:class:`Attachment`]
+        A list of attachments given to the snapshotted message.
+    timestamp: :class:`datetime.datetime`
```
AbstractUmbra, line 471:
```suggestion
    created_at: :class:`datetime.datetime`
```

Discord.py changes these attributes to saner names, e.g. `created_at`/`edited_at`.

discord/message.py
```diff
@@ -452,6 +455,87 @@ def guild_id(self) -> Optional[int]:
...
+
+        return [MessageSnapshot(state, snapshot['message']) for snapshot in message_snapshots]
+
+    def __init__(self, state: ConnectionState, data: MessageSnapshotPayload):
+        self.type: MessageType = try_enum(MessageType, data['type'])
+        self.content: str = data['content']
+        self.embeds: List[Embed] = [Embed.from_dict(a) for a in data['embeds']]
+        self.attachments: List[Attachment] = [Attachment(data=a, state=state) for a in data['attachments']]
+        self.timestamp: datetime.datetime = utils.parse_time(data['timestamp'])
+        self._edited_timestamp: Optional[datetime.datetime] = utils.parse_time(data['edited_timestamp'])
```
AbstractUmbra, line 507:
```suggestion
        self._edited_at: Optional[datetime.datetime] = utils.parse_time(data['edited_timestamp'])
```

Same as above, but also.. we store this private attribute but don't seem to do anything with it, is this intended? Or would this be utilised as a property that's been missed?

discord/enums.py
redmvc: Default is the name that the Discord API itself uses to call them

discord/message.py
redmvc: No, you're right, it should be a property

discord/message.py
```diff
@@ -499,6 +594,7 @@ def __init__(self, *, message_id: int, channel_id: int, guild_id: Optional[int]
     @classmethod
     def with_state(cls, state: ConnectionState, data: MessageReferencePayload) -> Self:
         self = cls.__new__(cls)
+        self.type = try_enum(MessageReferenceType, data.get('type', MessageReferenceType.default))
```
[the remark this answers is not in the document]
redmvc: Huh, why?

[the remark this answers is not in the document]
AbstractUmbra: This is the opposite of what we're trying to do.

discord/message.py
redmvc: Made a property, but I kept its name as _edited_timestamp since that's the same name as the private property in the Message class

discord/message.py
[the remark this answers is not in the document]
redmvc: Oh huh gotcha

discord/message.py
```diff
@@ -452,6 +455,92 @@ def guild_id(self) -> Optional[int]:
         return self._parent.guild_id
 
 
+class MessageSnapshot:
```
DA-344, line 458: I think this could have a ``cached_message`` attribute that, as its name indicates, has the cached message of this snapshot, if found.

redmvc: Is it ever found? I tested it for a while and none of the forwards I saw had that attribute non-empty

DA-344: As said, if the message is found in the cache then this would be non-empty. If a user forwards a message the bot has cached then the attribute would be filled, else not.

redmvc: Wait sorry I misunderstood. Yeah that seems like not a bad idea, though I'd need to look into how the cache works exactly

discord/message.py
```diff
@@ -452,6 +455,92 @@ def guild_id(self) -> Optional[int]:
...
+    """Represents a message snapshot attached to a forwarded message.
```
Rapptz, line 459: This needs a `.. versionadded:: 2.5` into it.

discord/message.py
Rapptz: This is a good idea, it shouldn't be a complicated add either. You already pass the connection state as a parameter but you need to bind it to make it into a property.

discord/enums.py
Rapptz, line 222: This should be called `reply` with an alias to `default`. The API docs even call it `REPLY` and `DEFAULT` despite not mentioning the former. https://discord.com/developers/docs/resources/message#message-reference-types

discord/message.py
```diff
@@ -462,6 +551,8 @@ class MessageReference:
 
     Attributes
     -----------
+    type: :class:`MessageReferenceType`
+        The type of message reference.
```
Rapptz, line 555: Needs a `versionadded`.

discord/message.py
```diff
@@ -486,10 +577,19 @@ class MessageReference:
...
-    def __init__(self, *, message_id: int, channel_id: int, guild_id: Optional[int] = None, fail_if_not_exists: bool = True):
+    def __init__(
+        self,
+        *,
+        type: MessageReferenceType,
+        message_id: int,
+        channel_id: int,
+        guild_id: Optional[int] = None,
+        fail_if_not_exists: bool = True,
+    ):
```
Rapptz, line 590: `MessageReference` is a 'data class' in discord.py which means users can create it, making this a breaking change for those who did since they no longer have a `type` parameter. You should give it a default value.

docs/api.rst
```diff
@@ -5166,6 +5166,14 @@ AllowedMentions
...
     :members:
 
+MessageSnapshot
+~~~~~~~~~~~~~~~~~
+
+.. attributetable:: MessageSnapshot
+
+.. autoclass:: MessageSnapshot
+    :members:
+
```
Rapptz, line 5176: Since this is an API received object it needs to be in the model section with the constructor hidden.

Rapptz (changes requested):
Thanks for the PR, some minor nits.

Puncher1:
As [new PR](https://github.com/discord/discord-api-docs/pull/7116) opened on Discord's side, this is missing following fields in `MessageSnapshot`: `stickers`, `sticker_items`, and `components`.

redmvc:
Sorry, it's unlikely I'll be able to resume work on this PR anytime soon. If anybody wants to take the code, I encourage it.

Rapptz:
Superceded by #9950

PR #9898, opened 2024-08-02
Aiodns win fix

mikeshardmind:
## Summary

See bikeshedding here: https://discord.com/channels/336642139381301249/1268727114568564847

3 of the 4 possible options each have a fix here. 

2 of the 3 changes both require `import sys`, so that's been added as a separate commit in case one of these changes isn't wanted to allow quickly selectively dropping components of this.

If documentation is wanted, I can follow up with that as well

Notes on potential breakage:

Any other library relying on the parts of the windows proactor event loop that are unique to that event loop may become incompatible. with these changes, as the conditions for setting this event loop are constrained to when there was already something relying on the selector event loop's semantics that discord.py uses when available (aiodns via aiohttp) this is not newly breaking, just changing how users are notified that they specified a reliance on 2 incompatible things.

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,7 @@ docs = [
 ]
 speed = [
     "orjson>=3.5.4",
-    "aiodns>=1.1",
+    "aiodns>=1.1; sys_platform != 'win32'",
     "Brotli",
     "cchardet==2.1.7; python_version < '3.10'",
 ]


BabyBoySnow:
Has Danny gave any input on the discord server on this?

Rapptz:
I really don't like this fix unfortunately. I know it's mainly because of upstream but I wish there was a better way of doing this -- and to be honest I'm not sure `aiodns` is worth the trouble or "speedup". So I'm leaning towards solution 1 and just axe it as a dependency on win32.

mikeshardmind:
Sure, dropped the other changes from the branch. Might be worth considering dropping aiodns altogether from extras and letting people grab it themselves, though with #9900 being merged, people can also manually specify a resolver.

Rapptz:
Yeah `aiodns` could probably be removed wholesale but for now just removing it for Windows is fine by me.

PR #9903, opened 2024-08-05
Fix Poll.duration rounding errors

ambdroid:
## Summary

In `Poll._from_data`, `expiry - message.created_at` may be a non-integer number of hours. If a received poll has a duration of only one hour, then `_to_dict()` may yield something like `{ ... 'duration': 0.9999991738888889 ... }` and attempting to re-send the poll would raise an an `Invalid form body` error as `duration` cannot be less than 1.
I put the fix for this in `_from_data` to maintain the current API behavior of making it the developer's responsibility to ensure that new polls have a valid duration.
 
## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/poll.py
+++ b/discord/poll.py
@@ -385,7 +385,7 @@ class Poll:
         question = question_data.get('text')
         expiry = utils.parse_time(data['expiry'])  # If obtained via API, then expiry is set.
         # expiry - message.created_at may be a few nanos away from the actual duration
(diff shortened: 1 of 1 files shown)


DA-344 (approved):
Although this looks good, stateful polls are not meant to be re-used as their already existent data may have conflicts with the new received ones. 
`Poll.copy()` purpose is this, allow already constructed and stateful polls to be re-used by returning a stateless one.

PR #9911, opened 2024-08-08
Add support for message call

Puncher1:
## Summary

This PR adds support for the message call object.

Related PR:
* https://github.com/discord/discord-api-docs/pull/6864

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -819,17 +819,22 @@ class MessageCall:
         A list of users that participated in the call.
     """
 
-    __slots__ = ('ended_timestamp', 'participants')
+    __slots__ = ('_message', 'ended_timestamp', 'participants')
 
     def __repr__(self) -> str:
         return f'<MessageCall participants={self.participants!r}>'
 
-    def __init__(self, *, state: ConnectionState, data: MessageCallPayload):
+    def __init__(self, *, state: ConnectionState, message: Message, data: MessageCallPayload):
+        self._message: Message = message
         self.ended_timestamp: Optional[datetime.datetime] = utils.parse_time(data.get('ended_timestamp'))
         self.participants: List[Optional[User]] = []
 
         for user_id in data['participants']:
-            self.participants.append(state.get_user(int(user_id)))
+            user_id = int(user_id)
+            if user_id == self._message.author.id:
+                self.participants.append(self._message.author)  # type: ignore # can't be a Member here
+            else:
+                self.participants.append(state.get_user(user_id))
 
 
 class RoleSubscriptionInfo:
@@ -1869,13 +1874,8 @@ class Message(PartialMessage, Hashable):
         self.position: Optional[int] = data.get('position')
         self.application_id: Optional[int] = utils._get_as_snowflake(data, 'application_id')
         self.stickers: List[StickerItem] = [StickerItem(data=d, state=state) for d in data.get('sticker_items', [])]
-        self.call: Optional[MessageCall] = None
 
-        call = data.get('call')
-        if call is not None:
-            self.call = MessageCall(state=state, data=call)
-
-            # This updates the poll so it has the counts, if the message
+        # This updates the poll so it has the counts, if the message
         # was previously cached.
         self.poll: Optional[Poll] = None
         try:
@@ -1963,7 +1963,7 @@ class Message(PartialMessage, Hashable):
         else:
             self.role_subscription = RoleSubscriptionInfo(role_subscription)
 
-        for handler in ('author', 'member', 'mentions', 'mention_roles', 'components'):
+        for handler in ('author', 'member', 'mentions', 'mention_roles', 'components', 'call'):
             try:
                 getattr(self, f'_handle_{handler}')(data[handler])
             except KeyError:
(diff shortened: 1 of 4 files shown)


discord/message.py
```diff
@@ -804,6 +806,32 @@ def cover(self) -> Optional[Asset]:
...
+
+    def __repr__(self) -> str:
+        return f'<MessageCall participants={self.participants!r}>'
+
+    def __init__(self, *, state: ConnectionState, data: MessageCallPayload):
+        self.ended_timestamp: Optional[datetime.datetime] = utils.parse_time(data.get('ended_timestamp'))
+        self.participants: List[Optional[User]] = []
+
+        for user_id in data['participants']:
+            self.participants.append(state.get_user(int(user_id)))
```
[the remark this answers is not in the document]
Puncher1: `Message.mentions` is actually an empty list as you can see in the payload:
```
{
    ...
    "call": {
        "ended_timestamp": "2024-08-08T17:51:12.456000+00:00",
        "participants": [
            "305354423801217025",
            "670700188087615489"
        ]
    },
    ...
    "mentions": [],
    ...
}
```
Also `state.get_user` returns either a `User` object or `None`.

discord/message.py
```diff
@@ -804,6 +806,32 @@ def cover(self) -> Optional[Asset]:
         return None
 
 
+class MessageCall:
+    """Represents a message's call data in a private channel from a :class:`~discord.Message`.
+
+    .. versionadded:: 2.5
```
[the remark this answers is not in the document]
dolfies: Would be nice to keep the same name as the previous implementation though

discord/message.py
```diff
@@ -804,6 +806,49 @@ def cover(self) -> Optional[Asset]:
...
+
+    Attributes
+    -----------
+    ended_timestamp: Optional[:class:`datetime.datetime`]
+        The timestamp the call has ended.
+    participants: List[Optional[:class:`User`]]
```
Rapptz, line 818: `List[Optional[X]]` should never be done in the library, so properly filter these. [This is how it was done in the previous implementation](https://github.com/Rapptz/discord.py/blob/1863a1c6636f53592519320a173ec9573c090c0b/discord/message.py#L184-L194).

Rapptz (changes requested):
Mostly okay, I'd prefer consistency with the previous version.

The previous name was `CallMessage` and not `MessageCall`. I still think that name makes more sense than the current one because `Call` is the adjective for `Message`, similar to `PartialMessage` rather than `MessagePartial`.

discord/message.py
Puncher1: Alright, yeah idk why I thought this would be a good idea 😄 I guess I should also implement the mechanism with searching the users in `self.mentions` for efficiency reasons.

Puncher1:
> I think `Message.system_content` needs update to handle `MessageType.call`.

yes you're right, it currently returns an empty string.

Puncher1:
The problem about this is: I don't know e.g. when Discord uses the next hour. When is it `an hour` and when is it `2 hours`. I would assume that 91min is `2 hours` and 89min is `an hour` but it could also be that 60..115min is `an hour` and the rest is `2 hours`.

Rapptz:
You can see the previous implementation:

https://github.com/Rapptz/discord.py/blob/1863a1c6636f53592519320a173ec9573c090c0b/discord/message.py#L319-L330

However it's pretty out-of-date. From the client, the strings are as follows:

```
SYSTEM_MESSAGE_CALL_STARTED_HOOK: $[!!{username}!!](usernameHook) started a call.,
SYSTEM_MESSAGE_CALL_STARTED: [!!{username}!!](usernameOnClick) started a call.,
SYSTEM_MESSAGE_CALL_STARTED_WITH_DURATION_HOOK: $[!!{username}!!](usernameHook) started a call that lasted !!{callDuration}!!.,
SYSTEM_MESSAGE_CALL_MISSED_HOOK: You missed a call from $[!!{username}!!](usernameHook).,
SYSTEM_MESSAGE_CALL_MISSED_WITH_DURATION_HOOK: You missed a call from $[!!{username}!!](usernameHook) that lasted !!{callDuration}!!.,
SYSTEM_MESSAGE_JOIN_CALL: Join the call
```

The client formats the duration string like this...
```js
m = function(e) {
    let t = null != e.call ? e.call.duration : null;
    return null != t ? t.humanize() : null
}(t);
```

Which isn't very helpful. We can just take your liberty here and just round, so 2.5 hours => 3 hours but 2.3 hours => 2 hours.

The skeleton should be something like this:

```py
if self.type is MessageType.call:
    call_ended = self.call.ended_timestamp is not None
    missed = self._state.user not in self.call.participants

    if call_ended:
        duration = ...
        if missed:
            return 'You missed a call from {0.author.name} that lasted {1}'.format(self, duration)
        else:
            return '{0.author.name} started a call that lasted {1}.'.format(self, duration)
    else:
        if missed:
            return '{0.author.name} started a call \N{EM DASH} Join the call.'.format(self)
        else:
            return '{0.author.name} started a call.'.format(self)
```

Puncher1:
I've just noted that `mentions` doesn't contain the users anymore:
```
self._message.mentions=[]
```
So we have to retrieve the users directly from the cache.

Puncher1:
I've added the system call message. The "calculation" is based on how Discord determines it (approx.). I've manipulated the message object in the client to see what the output is. I also made some unit tests, which can be found here: 

https://github.com/Puncher1/discord.py/blob/message-call-unittest/tests/test_call_system_content.py

discord/message.py
```diff
@@ -2415,6 +2474,69 @@ def system_content(self) -> str:
...
+                    if months == 1:
+                        duration = "a month"
+                    else:
+                        duration = f"{months} months"
+                else:
+                    years = round(seconds / years_s)
+                    if years == 1:
+                        duration = "a year"
+                    else:
+                        duration = f"{years} years"
```
Rapptz, line 2528: I forgot to mention this on Discord (sorry) but this should have been a util function instead of bloating up the size of this one.

Puncher1: Done

Rapptz (approved):

PR #9916, opened 2024-08-11
add `DummyCookieJar`

DeepSJain:
## Summary

The Discord API does not use cookies and setting the cookie_jar to DummyCookieJar() brings significant performance improvements since CookieJar is blocking.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -805,6 +805,7 @@ class HTTPClient:
             connector=self.connector,
             ws_response_class=DiscordClientWebSocketResponse,
             trace_configs=None if self.http_trace is None else [self.http_trace],
+            cookie_jar=aiohttp.DummyCookieJar(),
         )
         self._global_over = asyncio.Event()
         self._global_over.set()


Rapptz:
I don't disagree with the change, but I don't really see the logic? If the Discord API doesn't use cookies, which they don't, then that means they don't use `Set-Cookie` which wouldn't even make the cookie_jar code completely negligible. With no cookies everything short circuits as an early return.

Have you actually run into a scenario where this was "blocking" you?

DeepSJain:
Discord (specifically Cloudflare) does set some `Set-Cookie` headers such as `__cfruid`, `__dcfduid`, `__sdcfduid` and `_cfuvid`. In my case, `CookieJar.clear` was blocking and making this change fixed it.

mikeshardmind:
Not allowing `_cfuvid` to be set [could negatively impact users on shared ips](<https://developers.cloudflare.com/fundamentals/reference/policies-compliances/cloudflare-cookies/#_cfuvid-for-rate-limiting-rules>) I'm surprised to see discord using this one now, but if they are shared ips should be less likely to cause ratelimit issues from it.

DeepSJain:
According to [Discord's docs](https://discord.com/developers/docs/topics/rate-limits#invalid-request-limit-aka-cloudflare-bans), the Cloudflare ratelimit is per IP.

DeepSJain:
Related: aio-libs/aiohttp#8575

munishkhatri720:
> Discord (specifically Cloudflare) does set some `Set-Cookie` headers such as `__cfruid`, `__dcfduid`, `__sdcfduid` and `_cfuvid`. In my case, `CookieJar.clear` was blocking and making this change fixed it.

Yes, `CookieJar.clear` is blocking. I personally experienced it. My bot was lagging while responding to commands and using almost 100% CPU of the VPS. I used py-spy to detect what was causing the issue. Here are some screenshots of it:
image
image

Rapptz:
This is an aiohttp regression. If Discord's using Cloudflare cookies then it's not as simple as turning it off since it might impact some of the rate limit detection from their side. However, please stop mentioning that "it's blocking". It's synchronous code, of course it's blocking. There's "blocking" code in multiple places in the library and there's no way that this is going to be asynchronous. The proper terminology here is that there's a performance regression.

For now, downgrade aiohttp to 3.9 instead of 3.10.

Rapptz:
I got some confirmation that currently cookies do not do anything with rate limiting so this PR is good to go.

munishkhatri720:
Is it safer now to upgrade to the latest aiohttp version after this PR got merged, or should I stick to `aiohttp==3.9.0` for now?

PR #9918, opened 2024-08-12
[docs] Fix typehint for Embed.set_(image,thumbnail)

lmaotrigine:
## Summary

This commit explicitly documents that passing `None` to the `url` parameter removes existing images.
Though this does change the documented signature of public methods, this behaviour has existed since 2.0, so I don't think it's breaking.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -413,8 +413,9 @@ class Embed:
 
         Parameters
         -----------
-        url: :class:`str`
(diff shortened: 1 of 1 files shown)


Rapptz:
It's probably best to remove the offending line, which I'll do in a separate commit. Thanks.

PR #9924, opened 2024-08-18
Store app commands ids on CommandTree

Soheab:
## Summary

[As it was desired](https://discord.com/channels/336642139381301249/1011485943737360404/1122825846399455272), I have written an implementation that stores app command ids on the CommandTree and keeps them updated via sync, fetch and on callback. Also added two new methods to get the ID and return the mention format.

Currently as a draft because I am not sure if the current impl is the best way to this and would like some feedback on it, ignoring the left over comments/methods but it does work.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/tree.py
+++ b/discord/app_commands/tree.py
@@ -68,6 +68,7 @@ from .._types import ClientT
 
 if TYPE_CHECKING:
     from ..types.interactions import ApplicationCommandInteractionData, ApplicationCommandInteractionDataOption
+    from ..types.command import ApplicationCommand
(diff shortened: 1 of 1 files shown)


mikeshardmind:
I'm not sure I follow the reasoning for this existing since users can specify their own tree class, this is something users could implement without it needing to be extra objects stored for all users of command tree. Or if an implementation which caches this is going to exist in the library, it could be a subclass that's available for users to opt-into.

other more nit-feedback was already mentioned in discord.

Issue #9939, opened 2024-09-06
abc.GuildChannel.clone() method should accept a "category" argument


PPearoo:
### Summary

You can already specify a name for the category, there's no reason to limit this.

### What is the feature request for?

The core library

### The Problem

Let's say that someone is trying to create a 1:1 copy of a category, with all of its channels included. Currently, you'd do it with (2n+1) API calls, where n = number of channels inside a category.
1. Creating a new category
2. Cloning the channel in its current category
3. Moving the new channel to the new category.

### The Ideal Solution

This can be a bit tedious to code, and it also seems like a useless step, which could be avoided by adding a keyword-only optional argument, called "parent" or "category". This could accept either a `CategoryChannel` or an ID that points to a category inside the guild. If the given parent is invalid, it would raise a ValueError.

While this may seem like a small feature that is useless to most people, it could be really important in a bot which copies/clones categories. If a category had 50 channels for example, you'd need (2\*50+1) = **101!!!** API calls, almost half of which could be discarded by simply allowing people to actually choose a destination.

An other possible solution would be to add a `data: dict` argument to the clone implementation method, which would be harder for newbies to discover, but it would be much easier to use compared to making an attribute for everything that a channel can use when cloning.

### The Current Solution

Currently, the best you can do is (as said above) clone and move the channel to another category. Cloning and moving takes twice the API calls to actually perform, this could be reduced by half with this feature's implementation.

Puncher1:
This feature has been added via #9941

PPearoo:
> This feature has been added via #9941

Thanks! :D

PR #9941, opened 2024-09-08
Add category param to clone

Puncher1:
## Summary

This PR adds the category parameter to `clone` methods. 
Solves #9939

This reduces the number of API calls from `2*n + 1` to `n + 1`, where `n` is the number of channels, when you want to clone a category with all it's channels.

I've also added the parameter to `CategoryChannel.clone` to be consistent with the ABC, however I noted it in the docs that in that case the parameter is ignored. 

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -1005,11 +1005,15 @@ class GuildChannel:
         base_attrs: Dict[str, Any],
         *,
         name: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


Rapptz (approved):
This PR seems fine, but it has conflicts due to merging #9935. Could you rebase?

Rapptz (approved):

PR #9947, opened 2024-09-25
Add zstd gateway compression to speed profile

LostLuma:
## Summary

Use zstd instead of zlib compression when using the `speed` profile.
Provides a pretty decent speedup and reduces the amount of bandwith used.

To get an idea of the improvement I ran both profiles with my production bot account for 32 minutes simultaneously:

| algorithm | total decompress ns | raw bytes received |
|-----------|---------------------|--------------------|
| zlib      | 9720411600          | 224034102          |
| zstd      | 8176127300          | 203085901          |

I'm not sure if the way I've conditionally defined the decompression classes is the best, but it does the trick.

## Checklist

- [ ] This PR fixes an issue.


Rapptz (approved):
Great change. Thanks.

PR #9954, opened 2024-10-02
add missing "author_id" attribute for RawMessageUpdateEvent

RazzerDE:
## Summary

This pull requests adds the "author_id" parameter to the RawMessageUpdateEvent in order to check which user updated the message. This is useful as example if you want to check if the bot itself has edited something.

That is my first pull request - if I can make something better to remove some workload, let me know.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/raw_models.py
+++ b/discord/raw_models.py
@@ -154,6 +154,8 @@ class RawMessageUpdateEvent(_RawReprMixin):
         The message ID that got updated.
     channel_id: :class:`int`
         The channel ID where the update took place.
+    author_id: :class:`int`
+        The ID of the user who initiated this action.
 
         .. versionadded:: 1.3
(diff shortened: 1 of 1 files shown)


RazzerDE:
> This attribute is not missing at all, it is instead stored away in the `.data` attribute of `RawMessageUpdateEvent`.
> 
> However it is not what you believe it to be, it merely contains the data on the author of the edited message, not whom actioned the message update. As your PR checklist says this was tested, I can assume you did something such as send an embed, and then have another account suppress this embed and check the `author` value updated, and contained the account who did the suppression, not the message author?
> 
> Because in my local testing just now, it definitely showed `.data["author"]` being the message author, not the person triggering a message edit.

Yeah it can be the message author. I was testing it with the same account, I guess that was the problem.

PR #9957, opened 2024-10-04
Add support for Interaction Callback Resource

DA-344:
## Summary

This adds support for the Interaction Callback Resource returned by any ``Interaction.response`` method call by using a new parameter named ``with_response``.
This also adds the ``poll`` kwarg to ``Interaction.edit_original_response``.

**Needs testing**

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/interactions.py
+++ b/discord/interactions.py
@@ -54,6 +54,8 @@ __all__ = (
     'Interaction',
     'InteractionMessage',
     'InteractionResponse',
+    'InteractionCallback',
+    'InteractionCallbackActivity',
 )
 
 if TYPE_CHECKING:
--- a/discord/interactions.py
+++ b/discord/interactions.py
@@ -61,6 +61,8 @@ if TYPE_CHECKING:
         Interaction as InteractionPayload,
         InteractionData,
         ApplicationCommandInteractionData,
+        InteractionCallbackResponse as InteractionCallbackResponsePayload,
+        InteractionCallbackActivity as InteractionCallbackActivityPayload,
     )
     from .types.webhook import (
         Webhook as WebhookPayload,
@@ -466,6 +468,7 @@ class Interaction(Generic[ClientT]):
         attachments: Sequence[Union[Attachment, File]] = MISSING,
         view: Optional[View] = MISSING,
         allowed_mentions: Optional[AllowedMentions] = None,
+        poll: Poll = MISSING,
     ) -> InteractionMessage:
         """|coro|
 
@@ -500,6 +503,14 @@ class Interaction(Generic[ClientT]):
         view: Optional[:class:`~discord.ui.View`]
             The updated view to update this message with. If ``None`` is passed then
             the view is removed.
+        poll: :class:`Poll`
+            The poll to create when editing the message.
+
+            .. versionadded:: 2.5
+
+            .. note::
+
+                This is only accepted when the response type is :attr:`InteractionResponseType.deferred_channel_message`.
 
         Raises
         -------
@@ -529,6 +540,7 @@ class Interaction(Generic[ClientT]):
             view=view,
             allowed_mentions=allowed_mentions,
             previous_allowed_mentions=previous_mentions,
+            poll=poll,
         ) as params:
             adapter = async_context.get()
             http = self._state.http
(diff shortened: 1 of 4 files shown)


discord/interactions.py
```diff
@@ -621,6 +635,128 @@ async def translate(
...
+    def _update(self, data: InteractionCallbackResponsePayload) -> None:
+        interaction = data['interaction']
+        self.id: int = int(interaction['id'])
+        self.interaction_type: InteractionType = try_enum(InteractionType, interaction['type'])
+        self.activity_instance_id: Optional[str] = interaction.get('activity_instance_id')
+        response_id = interaction.get('response_message_id')
+        self.response_message_id: Optional[int] = (
+            int(
+                response_id,
+            )
```
AbstractUmbra, line 726: Remove the magic comma here to undo the formatted "ugliness" of this line.

discord/interactions.py
```diff
@@ -650,7 +786,33 @@ def type(self) -> Optional[InteractionResponseType]:
...
+
+    @overload
+    async def defer(
+        self,
+        *,
+        ephemeral: bool = ...,
+        thinking: bool = ...,
+        with_response: Literal[False] = ...,
+    ) -> None:
+        ...
```
AbstractUmbra, line 807: This overload is wrong, as the `= ...` implies that you are not providing an argument (thus relying on the default), which in this case would be `True`, not the implied `False` here.


```suggestion
    @overload
    async def defer(
        self,
        *,
        ephemeral: bool = ...,
        thinking: bool = ...,
        with_response: Literal[False] = False,
    ) -> None:
        ...
```

discord/interactions.py
```diff
@@ -675,13 +837,22 @@ async def defer(self, *, ephemeral: bool = False, thinking: bool = False) -> Non
...
         -------
         HTTPException
             Deferring the interaction failed.
         InteractionResponded
             This interaction has already been responded to before.
+
+        Returns
+        -------
+        Optional[:class:`InteractionCallback`]
+            The interaction callback resource, or ``None``.
```
AbstractUmbra, line 855:
```suggestion
            The interaction callback resource.
```

As `Optional[T]` implies "or None" it does not need to be outright stated.

AbstractUmbra (changes requested):
PRs that go beyond intended scope generally aren't accepted, what I mean by this is that the addition of `poll` to `Interaction.edit_original_response` should be it's own PR to maintain commit tree integrity.

Other than that, some things to correct.

discord/interactions.py
AbstractUmbra: You are indeed correct, my apologies.
Happy to disregard these comments in this case.

discord/interactions.py
```diff
@@ -621,6 +635,128 @@ async def translate(
...
+                data=resource['activity_instance'],
+            )
+        except KeyError:
+            pass
+
+        try:
+            self.response_message = InteractionMessage(
+                state=self._state,
+                channel=self.response_channel,  # type: ignore
+                data=resource['message'],
```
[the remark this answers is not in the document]
DA-344: Yeah, it is now mentioned in the docstring.

discord/interactions.py
```diff
@@ -709,15 +927,22 @@ async def defer(self, *, ephemeral: bool = False, thinking: bool = False) -> Non
...
                 proxy=http.proxy,
                 proxy_auth=http.proxy_auth,
                 params=params,
+                with_response=with_response,
             )
             self._response_type = InteractionResponseType(defer_type)
+            if response:
+                return InteractionCallbackResponse(
+                    parent=parent,
+                    channel=parent.channel,  # type: ignore
```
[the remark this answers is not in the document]
DA-344: Should be fixed now.

Rapptz (commented):
A few thoughts on this:

1. I'm not sure there's a merit in even supplying the parameter at all. My gut feeling says this should always be `with_response=True` and just flatten the overloads. I know this is different from how, e.g. `wait` works for webhooks but as far as I know there's a perceivable speed and behavioural difference there. I think there is no such difference here.
2. I don't know if I like the two class approach with the `InteractionCallbackResponse` paired with the `InteractionCallbackResource`. For starters, it's overly wordy and a bit confusing. From a typing perspective this is also a mess because you have to do weird `None` checks. I get that this is mostly just mirroring a quirk in the API so I'm unsure how I'd fix this personally.
3. The documentation for all this is kind of fairly useless if you don't already know what this is. A few of these are also missing the versionadded string.

DA-344:
I thought of different designs for this and went for the most “basic” one.
Maybe adding more attributes to `InteractionResponse` such as `callback_message` that, when editing/sending a message response, will have that response message, and make the other callback objects methods for `InteractionResponse` (`InteractionResponse.is_thinking()`, `InteractionResponse.is_ephemeral()`), and ignore the activity related things as it is out of scope for discord.py?

discord/interactions.py
```diff
@@ -624,6 +638,94 @@ async def translate(
...
+        The activity instance ID.
+    """
+
+    __slots__ = ('id',)
+
+    def __init__(self, data: InteractionCallbackActivityPayload) -> None:
+        self.id: str = data['id']
+
+
+class InteractionCallback(Generic[ClientT]):
```
Rapptz, line 658: This is missing `__slots__`.

discord/interactions.py
```diff
@@ -624,6 +638,94 @@ async def translate(
...
+
+    .. versionadded:: 2.5
+
+    Attributes
+    ----------
+    id: :class:`int`
+        The interaction ID.
+    type: :class:`InteractionResponseType`
+        The interaction callback response type.
+    resource: Optional[Union[:class:`InteractionMessage`, :class:`InteractionCallbackActivityInstance`]]
```
Rapptz, line 669: For now, this is fine. In the future it might make sense to refactor this into a type alias that is referenced later but that might require Sphinx changes. I'm bringing this up just to make note of it, no change required here.

discord/interactions.py
```diff
@@ -624,6 +638,94 @@ async def translate(
...
+        The resource that the interaction response created. If a message was sent, this will be
+        a :class:`InteractionMessage`, else if an activity was launched this will be a
+        :class:`InteractionCallbackActivityInstance`. In any other case, this will be ``None``.
```
Rapptz, line 672:
```suggestion
        The resource that the interaction response created. If a message was sent, this will be
        a :class:`InteractionMessage`. If an activity was launched this will be a
        :class:`InteractionCallbackActivityInstance`. In any other case, this will be ``None``.
```

discord/interactions.py
```diff
@@ -624,6 +638,94 @@ async def translate(
...
+        self._thinking: bool = interaction.get('response_message_loading', False)
+        self._ephemeral: bool = interaction.get('response_message_ephemeral', False)
+
+        self.message_id: Optional[int] = utils._get_as_snowflake(interaction, 'response_message_id')
+        self.activity_id: Optional[str] = interaction.get('activity_instance_id')
+
+        self.resource: Optional[Union[InteractionMessage, InteractionCallbackActivityInstance]] = None
+
+        if 'resource' in data:
+            resource = data['resource']
```
Rapptz, line 705:
```suggestion
        resource = data.get('resource')
        if resource is not None:
```

Avoid doing "double checking" such as `'key' in dict` if you can help it.

Rapptz (commented):
Yeah as expected, this design is much better and doesn't feel as messy. Just a few nits.

discord/interactions.py
```diff
@@ -624,6 +638,94 @@ async def translate(
...
+        interaction = data['interaction']
+
+        self.id: int = int(interaction['id'])
+        self._thinking: bool = interaction.get('response_message_loading', False)
+        self._ephemeral: bool = interaction.get('response_message_ephemeral', False)
+
+        self.message_id: Optional[int] = utils._get_as_snowflake(interaction, 'response_message_id')
+        self.activity_id: Optional[str] = interaction.get('activity_instance_id')
+
+        self.resource: Optional[Union[InteractionMessage, InteractionCallbackActivityInstance]] = None
```
Rapptz, line 702: I forgot to mention but in the code this Union should be refactored into a type alias so it's easier to read due to type complexity.

discord/interactions.py
```diff
@@ -624,6 +642,106 @@ async def translate(
...
+            if 'message' in resource and resource['message']:
+                self.resource = InteractionMessage(
+                    state=self._state,
+                    channel=self._parent.channel,  # type: ignore # channel should be the correct type here
+                    data=resource['message'],
+                )
+            elif 'activity_instance' in resource and resource['activity_instance']:
+                self.resource = InteractionCallbackActivityInstance(
+                    resource['activity_instance'],
+                )
```
Rapptz, line 734:
```suggestion
            message = resource.get('message')
            activity_instance = resource.get('activity_instance')
            if message is not None:
                self.resource = InteractionMessage(
                    state=self._state,
                    channel=self._parent.channel,  # type: ignore # channel should be the correct type here
                    data=message,
                )
            elif activity_instance is not None:
                self.resource = InteractionCallbackActivityInstance(activity_instance)
```

Rapptz (approved):
LGTM.

PR #9959, opened 2024-10-08
Fix issue with category cooldowns outside of guild channels.

bijij:
## Summary

Fixes issue reported in the discord server.

Could also probably be solved with a `PrivateChannel` `isinstanace` check, however not sure if it's worth the extra code.

## Checklist

- [x] This PR fixes an issue.


mikeshardmind:
Maybe a category ratelimit should imply guild only rather than not respecting the bucket type? For it to not have been an issue up until now would seem to indicate that most people understand that category channels are a construct of guilds.

PR #9983, opened 2024-10-24
Update role.py

bonniabyte:
fix the Role.py as there a error


bonniabyte:
just fix it and wanna start with it as a contribor

Rapptz:
This PR doesn't actually fix anything.

PR #10007, opened 2024-11-10
Add public method to get session start limits

mikeshardmind:
## Summary

This adds a method to get session start limits, intended to make it easier for people to use `before_identify_hook` without needing to override internals or manually make this request.

## Checklist

- [X] If code changes were made then they have been tested.
    - [X] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.
- [X] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

Issue #8430, opened 2022-09-03
Use `session_start_limit`'s `max_concurrency` when connecting shards


SebbyLaw:
### Summary

`AutoShardedClient` should connect as many shards as possible at a time

### What is the feature request for?

The core library

### The Problem

Currently, `AutoShardedClient` connects one shard every five seconds. This is fine for most bots, but big bots can connect more than 1 shard at a time every 5 seconds and limiting to one every 5 means a slow startup while waiting for possibly hundreds of shards to connect.

### The Ideal Solution

Use the `max_concurrency` key in https://discord.com/developers/docs/topics/gateway#get-gateway-bot to determine how many shards can connect at once.

### The Current Solution

Patch the library

Rapptz:
The default doesn't use `max_concurrency` because when you get to this scale you probably need to do out-of-process syncing due to one process being overloaded. The library allows customising the default behaviour by overriding `before_identify_hook` to make use of `max_concurrency`.

PR #10008, opened 2024-11-11
Use session limit max concurrency in a limited scope

mikeshardmind:
## Summary

See #8430 

This only applies when not clustering and when before_identify_hook has not been overridden.

This should limit the scope to bots that are slim enough to run in a single process yet large enough to have a raised max concurrency, without touching the behavior for bots that have already overridden this using the provided hook.

⚠️ I can't personally test this one. 

It's built on top of changes in #10007 

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -97,6 +97,7 @@ if TYPE_CHECKING:
         subscription,
     )
     from .types.snowflake import Snowflake, SnowflakeList
+    from .types.gateway import SessionStartLimit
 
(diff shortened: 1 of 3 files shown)


mikeshardmind:
I'm closing this for now, I think there's two minor issues with it as-is when considering zero-downtime restarts, and don't have a good general-purpose answer for those (The best answers here requires a level of coordination)

PR #10016, opened 2024-11-18
Guild.bulk_ban optimization

iyad-f:
## Summary

This PR modifies Guild.bulk_ban to use the ban endpoint if the length of given users is 1.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -3900,9 +3900,15 @@ class Guild(Hashable):
             The result of the bulk ban operation.
         """
 
+        user_ids: list[int | str] = [u.id for u in users]
+        if len(user_ids) == 1:
+            user_id = user_ids[0]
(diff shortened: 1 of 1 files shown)


iyad-f:
i type hinted user_ids with list[int, str] because otherwise http.bulk_ban wont accept it, maybe this is not the solution.

Rapptz:
Believe it or not the endpoint itself already does this, kind of. When this was implemented on Discord's side they changed it so single ban is just a special case of bulk ban. There's no real benefit in doing this in the code, it's just an extra branch that the backend already does.

iyad-f:
I see so there is no seperate strict rate limit for bulk-ban, is it calculated just with the normal ban endpoint?

Issue #9938, opened 2024-09-06
Documentation fails to build with sphinx 7.4.7


bdrung:
### Summary

Documentation fails to build with sphinx 7.4.7

### Reproduction Steps

python-discord fails to build on Debian and Ubuntu, because the documentation fails to build with sphinx 7.4.7.

### Minimal Reproducible Code

```python
cd docs && python3 -m sphinx . ../debian/html
```

### Expected Results

produce documentation

### Actual Results

```
cd docs && python3 -m sphinx . ../debian/html
Running Sphinx v7.2.6
WARNING: Invalid configuration value found: 'language = None'. Update your configuration to a valid language code. Falling back to 'en' (English).

Extension error:
Could not import extension builder (exception: cannot import name 'ltz' from 'sphinx.builders.gettext' (/usr/lib/python3/dist-packages/sphinx/builders/gettext.py))
make[1]: *** [debian/rules:10: execute_after_dh_auto_build] Error 2
```

### Intents

not relevant

### System Information

Debian unstable and Ubuntu 24.10 (oracular)

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.

### Additional Context

Bug-Ubuntu: https://bugs.launchpad.net/ubuntu/+source/python-discord/+bug/2079808

Rapptz:
The library's documentation is explicitly built with Sphinx 4.4.0 as per the pyproject.toml file. There's no such pressure to update the Sphinx used since they undergo breaking changes semi-frequently. We have one PR (https://github.com/Rapptz/discord.py/pull/9556) for it but it involved way too many hacks and compatibility workarounds to really put time into merging and thus is stale. 

In order to fix your issue you should use the Sphinx version specified in the pyproject.toml rather than using the bumped version though, especially since the breaking changes include significant styling changes from base Sphinx.

bdrung:
We only have one sphinx version in the Debian or Ubuntu archive (see https://tracker.debian.org/pkg/sphinx and https://launchpad.net/ubuntu/+source/sphinx). The old Sphinx 4.4.0 version is not available any more. We either need to carry a patch in the packaging to make it work with the newer version or drop the documentation (or the complete package).

PR #10018, opened 2024-11-19
Bump Sphinx to 5.3.0

LevBernstein:
## Summary

Upgrade Sphinx to 5.3.0. This version is [available in the Debian archive](https://packages.debian.org/source/stable/sphinx), but, unlike later versions, upgrading to it would not introduce significant breaking changes. This avoids the pitfalls of #9556. The only change I had to make was disabling table of contents entries for domain objects.

Sphinx 5.3.0 lacks the breaking changes to the gettext builder of the later versions, meaning builder.py doesn't need to be changed at all; it is still compatible with Python 3.8; and it doesn't require bumping any of the other dependencies (sphinxcontrib-serializinghtml, etc.).

While #9938 mentions Sphinx 7.4.7 specifically, the issue at hand was that the version of Sphinx used here (4.4.0) was not available on the Debian archive; by switching to one that is, I think we can mark that issue as closed.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -374,6 +374,10 @@ texinfo_documents = [
 # If true, do not generate a @detailmenu in the "Top" node's menu.
 #texinfo_no_detailmenu = False
 
+# If true, create table of contents entries for domain objects (e.g. functions,
+# classes, attributes, etc.).
+toc_object_entries=False
+
 def setup(app):
   if app.config.language == 'ja':
     app.config.intersphinx_mapping['py'] = ('https://docs.python.org/ja/3', None)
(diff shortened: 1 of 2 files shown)


benthetechguy:
> Upgrade Sphinx to 5.3.0. This version is [available in the Debian archive](https://packages.debian.org/source/stable/sphinx)

This isn't very useful, because even if Debian stable has version 5.3.0, builds on Debian sid (the builds that propagate to Ubuntu) would be using the version in Debian sid, currently 8.1.3.

Bumping to 5.3.0 only helps builds on Debian stable, even though there's already a perfectly working build there, and builds on Debian stable don't affect Ubuntu, which is the platform of the issue you opened this PR for.

PR #10021, opened 2024-11-20
Add stickers converter

Soheab:
## Summary

This PR makes it possible to ask users to input a sticker by annotating an ext.commands.Command's parameter to any of the sticker types. This works similarly to the Attachment converter.

At the time of writing this description, I noticed that there is already a converter for guild stickers, but.. it only searches in cache for the name or id meanwhile this can get any uploaded stickers. But.. this implement conflicts with the converter and I'm unsure what to do about that now.. (fml)

### Another thing
It would be neat if you could specify a union of sticker types but I couldn't figure out a way to implement that cleanly since Attachment doesn't support that either.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/core.py
+++ b/discord/ext/commands/core.py
@@ -96,6 +96,7 @@ T = TypeVar('T')
 CommandT = TypeVar('CommandT', bound='Command[Any, ..., Any]')
 # CHT = TypeVar('CHT', bound='Check')
 GroupT = TypeVar('GroupT', bound='Group[Any, ..., Any]')
+SpecialDataT = TypeVar('SpecialDataT', discord.Attachment, discord.StickerItem)
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


Rapptz:
I personally don't really see much point in this feature? The attachment converter has merits since it's a relatively common thing to desire but stickers feel niche considering the work being done here?

Soheab:
> I personally don't really see much point in this feature? The attachment converter has merits since it's a relatively common thing to desire but stickers feel niche considering the work being done here?

I personally don't see the point either. I did it just because it's possible, but I do think the code is messy.

From searching for "sticker converter" on the server, three results appeared, one of which is a help post ( https://discord.com/channels/336642139381301249/1265080971770990693 ) that includes a workaround using a callable default.

It's safe to say there is not much merit to this, so I'll leave it up to you.

PR #10032, opened 2024-12-02
Add optional dev dependency

Soheab:
## Summary

I noticed that discord.py doesn't have a optional list of dependencies needed to develop the library. So here it is.

Black and Pyright versions are straight from the [lint workflow](<https://github.com/Rapptz/discord.py/blob/9806aeb83179d0d1e90d903e30db7e69e0d492e5/.github/workflows/lint.yml>).

Edit: The `pyright` dependency was removed; it is not intended for this use case, per [this review](https://github.com/Rapptz/discord.py/pull/10032#discussion_r1875060129).

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,7 +70,6 @@ test = [
 dev = [
     "black==22.6",
     "typing_extensions>=4.3,<5",
-    "pyright==1.1.351",
 ]
 
 [tool.setuptools]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,6 +67,11 @@ test = [
     "typing-extensions>=4.3,<5",
     "tzdata; sys_platform == 'win32'",
 ]
+dev = [
+    "black==22.6",
+    "typing_extensions>=4.3,<5",
+    "pyright==1.1.351",
+]
 
 [tool.setuptools]
 packages = [


pyproject.toml
```diff
@@ -67,6 +67,11 @@ test = [
     "typing-extensions>=4.3,<5",
     "tzdata; sys_platform == 'win32'",
 ]
+dev = [
+    "black==22.6",
+    "typing_extensions>=4.3,<5",
+    "pyright==1.1.351",
```
AbstractUmbra, line 73: Might be a questionable addition since pyright itself is actually an [npm package](https://www.npmjs.com/package/pyright). The python module uses [`nodeenv`](https://pypi.org/project/nodeenv/) which by the authors own admission is less stable than using `pyright[nodejs]` which uses [`nodejs-wheel`](https://pypi.org/project/nodejs-wheel-binaries/) in it's stead.

My suggestion would be omitting this entirely and letting the users use Pylance or their own installation of pyright.

Rapptz: Yeah, I'd probably remove this too. I don't think any of us use the pip installable package.

Soheab: I was unaware of this. I have removed the dependency.

PR #10042, opened 2024-12-12
Add support for multisubscriptions

DA-344:
## Summary

Adds support for multi-subscriptions.

DDevs PR: https://github.com/discord/discord-api-docs/pull/7304

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/subscription.py
+++ b/discord/subscription.py
@@ -63,6 +63,8 @@ class Subscription(Hashable):
     canceled_at: Optional[:class:`datetime.datetime`]
         When the subscription was canceled.
         This is only available for subscriptions with a :attr:`status` of :attr:`SubscriptionStatus.inactive`.
+    renewal_sku_ids: List[:class:`int`]
+        The IDs of the SKUs that the user is going to be subscribed when renewing.
     """
 
     __slots__ = (
@@ -75,6 +77,7 @@ class Subscription(Hashable):
         'current_period_end',
         'status',
         'canceled_at',
+        'renewal_sku_ids',
     )
 
     def __init__(self, *, state: ConnectionState, data: SubscriptionPayload):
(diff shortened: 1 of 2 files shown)


discord/subscription.py
```diff
@@ -63,6 +63,8 @@ class Subscription(Hashable):
     canceled_at: Optional[:class:`datetime.datetime`]
         When the subscription was canceled.
         This is only available for subscriptions with a :attr:`status` of :attr:`SubscriptionStatus.inactive`.
+    renewal_sku_ids: List[:class:`int`]
+        The IDs of the SKUs that the user is going to be subscribed to when renewing.
```
Rapptz, line 67: Needs versionadded.

discord/types/subscription.py
```diff
@@ -40,3 +40,4 @@ class Subscription(TypedDict):
     current_period_end: str
     status: SubscriptionStatus
     canceled_at: Optional[str]
+    renewal_sku_ids: List[Snowflake]
```
Rapptz, line 43:
```suggestion
    renewal_sku_ids: Optional[List[Snowflake]]
```

Kind of weird this is considered a required key on their end but this is more accurate I guess.

Rapptz (changes requested):

discord/subscription.py
DA-344: Oh, I didn't add it because the Subscription object is new on 2.5 too...

Rapptz: Yeah that's fair, it's been forever since a release was cut so it's easy to forget lol

Rapptz (approved):
This looks okay, it's a bit low level though so it doesn't feel particularly useful though the state of the library at this point when it comes to this is something I'm still fighting in my head. Anyway, doesn't impact this PR so it's good to go.

PR #10043, opened 2024-12-12
feat(audio): migrate 'audioop.mul' functionality using numpy (corresponds to patch 3.13)

zeroday0619:
## Summary

This pull request migrates the functionality of audioop.mul to use numpy, as the audioop module has been removed in Python 3.13. The migration ensures compatibility with Python 3.13 while preserving the functionality of volume scaling for audio data. 

It replaces the deprecated functionality with an implementation using numpy for data manipulation, including clipping to the correct range based on the bit depth. 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/player.py
+++ b/discord/player.py
@@ -26,11 +26,6 @@ from __future__ import annotations
 import threading
 import subprocess
 import warnings
-if sys.version_info < (3, 13):
-    import audioop
-else:
-    import numpy as np
-
 import asyncio
 import logging
 import shlex
@@ -39,6 +34,11 @@ import json
 import sys
 import re
 import io
+
+if sys.version_info < (3, 13):
+    import audioop
+else:
+    import numpy as np
 
 from typing import Any, Callable, Generic, IO, Optional, TYPE_CHECKING, Tuple, TypeVar, Union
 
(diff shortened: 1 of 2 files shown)


zeroday0619:
Hmm... Should I rewrite this as pure Python code without numpy dependency? Or is this PR incorrect?

AbstractUmbra:
This PR is unnecessary due to the underlying issue being resolved by 624b5b7643ce6dfc783aa9ce365026af41f78e4b.

zeroday0619:
Honestly, I don’t know if it’s right from a long-term perspective to use a backported module that was removed from the standard Python modules.

zeroday0619:
Honestly, I don’t know if it’s right to use audioop-lts, a backport of audioop that was removed from the standard Python modules.

AbstractUmbra:
> Honestly, I don’t know if it’s right from a long-term perspective to use audioop-lts, a backport of audioop that was removed from the standard Python modules.

The solution was agreed upon by the maintainers after lengthy discussion. I myself maintain the module almost exclusively for discord.py usage (granted it also works for normal use, but that was the goal).

Rapptz:
Thanks but this is already resolved.

Issue #10052, opened 2024-12-26
Building docs on python 3.13 results in an import exception


cibere:
### Summary

Building the docs on 3.13 results in an import exception because [imghdr](<https://docs.python.org/3/library/imghdr.html>) has been removed in 3.13

### Reproduction Steps

1. Create a fresh venv with python 3.13
2. Install the library's dependencies
3. Try to build the docs

### Minimal Reproducible Code

```python
py -3.13 -m venv venv
venv\scripts\activate
pip install .[docs]
cd docs
make html
```

### Expected Results

The docs are successfully built and I can access them

### Actual Results

Running Sphinx v4.4.0

Extension error:
Could not import extension sphinx.builders.epub3 (exception: No module named 'imghdr')

### Intents

doesn't apply

### System Information

- Python v3.13.1-final
- discord.py v2.5.0-alpha
    - discord.py metadata: v2.5.0a5129+g9806aeb8
- aiohttp v3.11.11
- system info: Windows 11 10.0.22631

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.

### Additional Context

This came from a [discord.py help thread](https://ptb.discord.com/channels/336642139381301249/1321976059679473674)

mikeshardmind:
I've gone ahead and rescued imghdr similar to audioop [here](https://pypi.org/project/imghdr-lts/0.1.0/)

similar state with it being a small, stable piece of the standard library that required no further changes and could have been left "as-is".

I'll open a PR with it as a docs-only, 3.13+ only dependency, but I'm not particularly attached. We'd need to upgrade to at least sphinx 6.6 to get to a version where sphinx does not depend on this.

mikeshardmind:
#10053 being merged should have resolved this.

PR #10053, opened 2024-12-27
Temporary dependency for docs extra on python 3.13+

mikeshardmind:
## Summary

See #10052 Adds imghdr-lts, which is <200 lines of pure python [here](https://github.com/mikeshardmind/imghdr-lts)

I don't think we should keep this one long term, but this is docs-only, 3.13+, and this code should require zero changes forever (barring some breaking syntax change) as such, I've also hard-pinned the version.

If I'm reading the history on sphinx correctly, this can be removed once we are using any version of sphinx >= 6.6 for docs.

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,8 @@ docs = [
     "sphinxcontrib-serializinghtml==1.1.5",
     "typing-extensions>=4.3,<5",
     "sphinx-inline-tabs==2023.4.21",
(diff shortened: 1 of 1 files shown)


chillymosh:
@mikeshardmind the link appears to link to an incorrect version so 404s, 0.1.0 instead of 1.0.0

https://pypi.org/project/imghdr-lts/

PR #10056, opened 2024-12-28
Return new instance in Poll.end to avoid inconsistencies

ilovetocode2019:
## Summary

Currently, the `Poll.end` method updates the `Poll.message` attribute with the response from the API call, but still returns the current instance. This can cause `Poll.message` to be newer than the rest of the data stored in `Poll`. This also means that the the instance is being updated in-place regardless of whether it is from the state, which I can't find an intentional reason for. Both of these seem inconsistent with usual library behavior.

The following example shows this behavior:
```python

poll = discord.Poll(question="What pet is better?", duration=datetime.timedelta(hours=1))
poll.add_answer(text="Dogs")
poll.add_answer(text="Cats")
message = await messageable.send(poll=poll)

# after waiting for a bit
stopped_poll = await poll.end()
stopped_poll.answers[0].vote_count == stopped_poll.message.poll.answers[0].vote_count # false if people voted
message is poll.message # always false
```

This PR changes the behavior to instead return the poll created from the API call inside `Poll.end`. This means that making an API call will not modify the instance used to make it, instead returning a fresh one, with more updated and consistent data.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/poll.py
+++ b/discord/poll.py
@@ -568,6 +568,5 @@ class Poll:
         if not self._message or not self._state:  # Make type checker happy
             raise ClientException('This poll has no attached message.')
 
-        self._message = await self._message.end_poll()
-
-        return self
+        message = await self._message.end_poll()
+        return message.poll
--- a/discord/poll.py
+++ b/discord/poll.py
@@ -547,7 +547,7 @@ class Poll:
 
         return self._answers.get(id)
 
-    async def end(self) -> Self:
+    async def end(self) -> Poll:
         """|coro|
 
         Ends the poll.
--- a/discord/poll.py
+++ b/discord/poll.py
@@ -359,6 +359,7 @@ class Poll:
         # The message's poll contains the more up to date data.
         self._expiry = message.poll.expires_at
         self._finalized = message.poll._finalized
+        self._answers = message.poll._answers
 
     def _update_results(self, data: PollResultPayload) -> None:
         self._finalized = data['is_finalized']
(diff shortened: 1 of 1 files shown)


discord/poll.py
```diff
@@ -568,6 +568,5 @@ async def end(self) -> Self:
         if not self._message or not self._state:  # Make type checker happy
             raise ClientException('This poll has no attached message.')
 
-        self._message = await self._message.end_poll()
-
-        return self
+        message = await self._message.end_poll()
```
DA-344, line 571: ._message still needs to be set, plus, you can simply do self._update(message) to update the data

discord/poll.py
```diff
@@ -547,7 +547,7 @@ def get_answer(
 
         return self._answers.get(id)
 
-    async def end(self) -> Self:
+    async def end(self) -> Poll:
```
DA-344, line 550: This is typed like this for subclasses

discord/poll.py
ilovetocode2019: Ah, I went with this way intentionally because I wasn’t sure if the update should even be in-place. I’ll change it to make sure everything is done to the current instance.

DA-344: You could change the _update behaviour to update the current poll vote count and such, though

Rapptz:
I think I fundamentally agree with the premise that these updates should not be in-place, since the library has historically tried to avoid doing so for the last 2 major versions. I guess it's a bit late for that unfortunately. The change here looks fine to me as a stop-gap from an initial read though I have not used polls at all to really know.

PR #10057, opened 2024-12-28
Add detailed docstrings for row and width properties in the Item class

Error44s:
## Summary

This pull request adds detailed docstrings to the `row` and `width` properties of the `Item` class in the Discord.py library. These docstrings provide clear explanations of their purpose, valid values, constraints, and usage within the framework, improving overall code clarity and documentation.

## Checklist

- [ ] This PR fixes an issue.
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed).
- [x] This PR is **not** a code change (e.g. documentation, README, ...).

### DIFF
--- a/discord/ui/item.py
+++ b/discord/ui/item.py
@@ -100,6 +100,16 @@ class Item(Generic[V]):
 
     @property
     def row(self) -> Optional[int]:
+        """Optional[:class:`int`]: The action row this item belongs to.
+
+        An action row can have up to 5 components. Valid values for the row are
+        integers in the range 0–4 or ``None`` to indicate no specific row.
+    
+        Raises
+        ------
+        ValueError
+            If set to a value less than 0 or greater than or equal to 5.
+        """
         return self._row
 
     @row.setter
@@ -113,6 +123,16 @@ class Item(Generic[V]):
 
     @property
     def width(self) -> int:
+        """int: The width of this item in a Discord UI layout.
+
+        The width determines how much space this item occupies in an action row.
+        This value is primarily used internally by Discord.py to calculate layout constraints.
+
+        Returns
+        -------
+        :class:`int`
+            The width of the item, defaulting to 1.
+        """
         return 1
 
     @property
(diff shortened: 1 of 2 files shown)


discord/ui/item.py
```diff
@@ -113,6 +123,16 @@ def row(self, value: Optional[int]) -> None:
...
     def width(self) -> int:
+        """int: The width of this item in a Discord UI layout.
+
+        The width determines how much space this item occupies in an action row.
+        This value is primarily used internally by Discord.py to calculate layout constraints.
+
+        Returns
+        -------
+        :class:`int`
+            The width of the item, defaulting to 1.
```
DA-344, line 134:
```suggestion

```
Properties don't have this Return, as their return type are stated above.

discord/ui/item.py
```diff
@@ -100,6 +100,16 @@ def __repr__(self) -> str:
 
     @property
     def row(self) -> Optional[int]:
+        """Optional[:class:`int`]: The action row this item belongs to.
+
+        An action row can have up to 5 components. Valid values for the row are
+        integers in the range 0–4 or ``None`` to indicate no specific row.
```
DA-344, line 106:
```suggestion

```
This is already documented on the ``row`` parameter on each different ``Item`` subclass, so it is not needed.

discord/ui/item.py
```diff
@@ -100,6 +100,16 @@ def __repr__(self) -> str:
 
     @property
     def row(self) -> Optional[int]:
+        """Optional[:class:`int`]: The action row this item belongs to.
```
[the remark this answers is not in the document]
DA-344:
```suggestion
        """Optional[:class:`int`]: Returns the row this item belongs to.
```

discord/ui/item.py
```diff
@@ -113,6 +123,16 @@ def row(self, value: Optional[int]) -> None:
 
     @property
     def width(self) -> int:
+        """int: The width of this item in a Discord UI layout.
+
+        The width determines how much space this item occupies in an action row.
+        This value is primarily used internally by Discord.py to calculate layout constraints.
```
DA-344, line 129: I don't think mentioning this is necessary?

Rapptz:
These are not documented on purpose since they're not part of public API.

`width` is there for internal reasons when calculating stuff, and `row` is there as an API convenience not meant to be relied upon. In the future they might even go away with changes to the component infrastructure so codifying it further than it already is (such as `row` in the constructors and decorators) is only going to cause more struggles.

Thanks for the PR though.

PR #10060, opened 2024-12-29
fix target id being None raising an error

iyad-f:
## Summary

This PR fixes the error raised when target_id is None and AudiitLogEntry.target is used and a converter is used to return the target

## Checklist

- [x] This PR fixes an issue.
image


iyad-f:
i see that some converters here can have target_id None so this if target_id is None check would be needed in those converters where target_id must be an int

iyad-f:
i think i messages up with pushing stuff

Rapptz:
This isn't the proper fix for this. The proper fix would be to fix the specific target instead.

iyad-f:
i added another commit doing that should i make another pr?

PR #10063, opened 2024-12-31
[commands] Correct ExtensionNotFound error message

Soheab:
## Summary

I noticed that attempting to reload a nonexistent extension resulted in the library raising an `ExtensionNotFound` error with the message "Extension 'foo' could not be loaded." This phrasing might be misinterpreted by some as indicating a loading failure, when in fact it denotes that the extension was not found. This PR refines the message for improved clarity.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/errors.py
+++ b/discord/ext/commands/errors.py
@@ -1080,7 +1080,7 @@ class ExtensionNotFound(ExtensionError):
     """
 
     def __init__(self, name: str) -> None:
-        msg = f'Extension {name!r} could not be loaded.'
+        msg = f'Extension {name!r} could not be found at the specified location.'
         super().__init__(msg, name=name)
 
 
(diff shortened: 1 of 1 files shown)


discord/ext/commands/errors.py
```diff
@@ -1080,7 +1080,7 @@ class ExtensionNotFound(ExtensionError):
     """
 
     def __init__(self, name: str) -> None:
-        msg = f'Extension {name!r} could not be loaded.'
+        msg = f'Extension {name!r} could not be found at the specified location.'
```
Rapptz, line 1083:
```suggestion
        msg = f'Extension {name!r} could not be loaded or found.'
```

This wording might be better?

Soheab: Fine with me!

PR #10073, opened 2025-01-06
first ever big PR

ChainDev162:
added a comment and changed reserved driver name line


Rapptz:
Thanks but the string is fine as is.

ChainDev162:
> Thanks but the string is fine as is.

Understood, thanks for letting me know!

PR #10078, opened 2025-01-15
Added scopeless kwarg to discord.utils.oauth_url

Depreca1ed:
## Summary

This PR adds a scopeless kwarg to discord.utils.oauth_url. When True, the OAuth URL which was generated won't include a scope if or if not provided. By default, will be False.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -328,9 +328,9 @@ def oauth_url(
     guild: Snowflake = MISSING,
     redirect_uri: str = MISSING,
     scopes: Iterable[str] = MISSING,
-    scopeless: bool = False,
     disable_guild_select: bool = False,
     state: str = MISSING,
+    scopeless: bool = False,
 ) -> str:
     """A helper function that returns the OAuth2 URL for inviting the bot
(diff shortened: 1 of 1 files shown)


Jackenmen:
Perhaps making `scopes` accept `None` and have that mean that no scopes should be set is better? You'd remove interdependency between the parameters this way.

Rapptz:
Yeah I feel `scopes=None` is probably better for this.

Depreca1ed:
I have made the requested changes and tested them.

Rapptz:
I guess the next question is, what's the use-case for this? Why would you want a scopeless URL?

DA-344:
Maybe for user installation OAuth url, as it does not take any scopes…?

Depreca1ed:
It's the URL from the installation menu of an app on the developer panel or at least it is for me. The URL also embeds therefore I felt that the function should provide a way to make that specific url(i.e. with no scopes).

Rapptz:
Sure, that works.

PR #10080, opened 2025-01-17
Add Guild.acronym

Soheab:
## Summary

This PR adds an `acronym' property to Guild that returns the characters that show up when a server has no icon set. 

The property returns a cached version of the acronym (`_acronym`) since it rarely, if ever, changes.

I got this idea from the [discord.js](<https://github.com/discordjs/discord.js/blob/a5afc406b965b39a9cc90ef9e0e7a4b460c4e04c/packages/discord.js/src/structures/BaseGuild.js#L65-L70>) library. I have converted it to Python as is.

Discussion post: https://canary.discord.com/channels/336642139381301249/1279035934192439388

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -48,6 +48,7 @@ from typing import (
     overload,
 )
 import warnings
+import re
 
 from . import utils, abc
 from .role import Role
@@ -447,6 +448,7 @@ class Guild(Hashable):
         'max_stage_video_users',
         '_incidents_data',
         '_soundboard_sounds',
+        '_acronym',
     )
 
     _PREMIUM_GUILD_LIMITS: ClassVar[Dict[Optional[int], _GuildLimit]] = {
@@ -578,6 +580,8 @@ class Guild(Hashable):
             pass
 
         self.name: str = guild.get('name', '')
+        self._acronym = re.sub(r"\s", "", re.sub(r"\w+", lambda e: e.group(0)[0], re.sub(r"'s ", " ", self.name)))
+
         self.verification_level: VerificationLevel = try_enum(VerificationLevel, guild.get('verification_level'))
         self.default_notifications: NotificationLevel = try_enum(
             NotificationLevel, guild.get('default_message_notifications')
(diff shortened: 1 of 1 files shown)


Rapptz:
I have two main issues with this.

1. The current implementation wastes memory for those who do not care about this feature (i.e. most people)
2. The current implementation feels dodgy. It's using regex for no good reason and expensively so, it seems like it's a 1:1 port of JS code without porting it to the target language's strength.

Other than that I'm not sure on the overall merit of the PR. Is this a commonly requested feature? If you do a simple search of [`nameAcronym`](https://github.com/search?q=nameAcronym&type=code) on GitHub you don't get that many results of people actually using it.

Soheab:
> I have two main issues with this.
> 
> 1. The current implementation wastes memory for those who do not care about this feature (i.e. most people)
> 2. The current implementation feels dodgy. It's using regex for no good reason and expensively so, it seems like it's a 1:1 port of JS code without porting it to the target language's strength.
> 
> Other than that I'm not sure on the overall merit of the PR. Is this a commonly requested feature? If you do a simple search of [`nameAcronym`](https://github.com/search?q=nameAcronym&type=code) on GitHub you don't get that many results of people actually using it.

1 & 2. Fair points. Unfortunately, I don't know how to fix that, as I am very new to regex usage. Reviews are welcome.

This is not a requested feature, but I thought it would fit in between the other features, as others pointed out, like `Message.system_content`, `Member.top_role`, etc.

ChesterWOV-DPY:
From what I understood, I believe `''.join(word[0] for word in self.name.split())` suffices?

AbstractUmbra:
> From what I understood, I believe `''.join(word[0] for word in self.name.split())` suffices as a way without regex?

Unfortunately not, there are edge cases that apply to this, such as "X of Y" and "X's thing".
It's a bit more involved.

ChesterWOV-DPY:
> Unfortunately not, there are edge cases that apply to this, such as "X of Y" and "X's thing".
> It's a bit more involved.

Interesting, I just tested the examples you gave on both my mobile and desktop client, and they show up as "XoY" and "Xt", respectively. I ran my code snippet and it did get it right.

PR #10082, opened 2025-01-18
Add `CommandTree.get_hash`

mikeshardmind:
## Summary

Adds a method to get a hash of the command tree payload as it would be sent to discord when synced.

This is intended to allow users to sync only when needed programmatically (by comparing with a stored prior hash), as well as to potentially enable some level of automatic syncing to be done by the library on behalf of users in the future built on top of this.

Details about a potential API to automatically sync global commands based on this info here: https://discord.com/channels/336642139381301249/1326219331050147940

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/tree.py
+++ b/discord/app_commands/tree.py
@@ -25,6 +25,8 @@ DEALINGS IN THE SOFTWARE.
 from __future__ import annotations
 import logging
 import inspect
+import sys
+from hashlib import blake2b
 
 from typing import (
     Any,
@@ -62,7 +64,7 @@ from .installs import AppCommandContext, AppInstallationType
 from .translator import Translator, locale_str
 from ..errors import ClientException, HTTPException
 from ..enums import AppCommandType, InteractionType
-from ..utils import MISSING, _get_as_snowflake, _is_submodule, _shorten
+from ..utils import MISSING, _get_as_snowflake, _is_submodule, _shorten, _to_json
 from .._types import ClientT
 
 
@@ -80,6 +82,25 @@ if TYPE_CHECKING:
 __all__ = ('CommandTree',)
 
 _log = logging.getLogger(__name__)
+
+
+if sys.version_info < (3, 9):
+    _blake_kwargs = {}
+else:
+    # This prevents blake from raising an error
+    # with some configurations in python 3.9
+    _blake_kwargs = {'usedforsecurity': False}
+
+
+def _hash_payload(payload: list[Dict[str, Any]]) -> bytes:
+    tree_hash = blake2b(digest_size=32, person=b"tree", last_node=True, **_blake_kwargs)
+    command_hashes = [
+        blake2b(_to_json(c).encode(), person=b"command", last_node=False, **_blake_kwargs).digest() for c in payload
+    ]
+    for h in sorted(command_hashes):
+        tree_hash.update(h)
+
+    return b"v1:" + tree_hash.digest()
 
 
 def _retrieve_guild_ids(
@@ -1076,6 +1097,17 @@ class CommandTree(Generic[ClientT]):
             await translator.load()
             self._state._translator = translator
 
+    async def _get_payload(self, *, guild: Optional[Snowflake] = None) -> List[Dict[str, Any]]:
+        commands = self._get_all_commands(guild=guild)
+
+        translator = self.translator
+        if translator:
+            payload = [await command.get_translated_payload(self, translator) for command in commands]
+        else:
+            payload = [command.to_dict(self) for command in commands]
+
+        return payload
+
     async def sync(self, *, guild: Optional[Snowflake] = None) -> List[AppCommand]:
         """|coro|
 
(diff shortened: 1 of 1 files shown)


discord/app_commands/tree.py
```diff
@@ -1315,3 +1321,22 @@ async def _call(self, interaction: Interaction[ClientT]) -> None:
...
+
+        Parameters
+        -----------
+        guild: Optional[:class:`~discord.abc.Snowflake`]
+        """
+        payload = await self._get_payload(guild=guild)
+        # versioning allows the method to be changed in the future
+        # blake2 is the fastest collision resistant hash guaranteed
+        # to be available in the standard library at time of writing.
+        return "v1:" + blake2b(_to_json(payload).encode()).hexdigest()
```
DiscordLiz, line 1342: 1. Wouldn't this be better as (version, bytes) or even just bytes? These hashes don't need to be identifiable online.
2. Do we need a 512bit hash digest? Possibly set the digest size lower.
3. This should use `usedforsecurity=False` to not end up a disabled algorithm by security policies.

discord/app_commands/tree.py
```diff
@@ -1076,6 +1077,17 @@ async def set_translator(self, translator: Optional[Translator]) -> None:
             await translator.load()
             self._state._translator = translator
 
+    async def _get_payload(self, *, guild: Optional[Snowflake] = None) -> List[Dict[str, Any]]:
+        commands = self._get_all_commands(guild=guild)
```
DiscordLiz, line 1081: Is the ordering of this iterator deterministic? If not, this needs a sort like the implementation of this we use in Salamander and Basilisk.

DiscordLiz (changes requested):

discord/app_commands/tree.py
mikeshardmind: 1. Maybe.
2. Probably not, but it isn't hurting anything as-is either. 
3. `usedforsecurity` doesn't exist in python3.8, but you're right. This is going to require a little change here.

Rapptz:
I'm not really sure this belongs in the library proper? Part of the problem is that (unfortunately) a lot of people do a lot of extra non-immutable work within their translators that can be potentially expensive (for some God forsaken reason, people actually do HTTP requests here) and done multiple times in this scenario. 

For example if someone wanted to store a hash and then do the call to compare, they'd end up calling all the translation machinery once to generate the hash and then *again* once they find out it's different. This feels subpar compared to just calling it once when you know it's changed (i.e. as part of a scheduled translation routine or whatever).

We could make the argument that cases like these are uncommon, and maybe that's agreeable but I'm not really sure this machinery is worth it for the people who want to avoid a sync call. Truth be told the sync rate limit is *not* that bad, since it only triggers on *new* commands added per day. That is to say that *updates* don't suffer from as much of a rate limit, which is the main use case for a hash to detect anyway. If you wanted to be more careful of the rate limit you're probably better off making a mapping with the full name of every command and then checking if those sets are different rather than a hashing scheme.

mikeshardmind:
> Part of the problem is that (unfortunately) a lot of people do a lot of extra non-immutable work within their translators that can be potentially expensive (for some God forsaken reason, people actually do HTTP requests here) and done multiple times in this scenario.

😐 I was unaware there's anyone doing something like this, from my perspective, translations should basically be static content, and I even thought it was odd that the translation machinery is async.

The only current issue with this existing outside of the library is that whoever is implementing it needs to track which things require syncing and how that may change over time. While this isn't impossible, there's a lot hidden from users by default about this, especially around the translation machinery. If leaning on how discord.py does it internally, then there's a reliance on things which aren't guaranteed. This actually changed at some point, as I remember having to change something in the implementation of this in two of my bots (one private, one public currently)

Maybe there's a way to get this returned as part of sync, and for sync to take a "last known hash" parameter that if it matches, *skip* syncing to prevent double calling? But then that changes the return type of that and would be breaking.

mikeshardmind:
I've pushed changes that handle the potential non-determinism, but that's not necessarily going to be relevant if it's not suitable for inclusion.

I'll think about whether there's a solid way to prevent double-calling the translation machinery and other options that might address the same issues reasonably, but things like people making HTTP requests in translations leaves the door open to the translation being subtly different each time anyhow, and this *is* part of what is synced.

mikeshardmind:
I think there's reasonable value to this still, but I can't come up with an answer that accounts for the double work here that I think helps the people who would benefit from it without a breaking change. The best I've been able to come up with involves some callbacks that get invoked during the sync flow and can preempt going through with the sync, as this gives a non-breaking point where no work would be duplicated.

The end goal here was essentially that for someone just running this locally or on a single server and without a complexly orchestrated larger bot, any changes they made could result in automatically syncing when they start the bot, without syncing on start unconditionally, and without needing to know themselves which things required syncing. (for example, they fix the description of a command: that requires a sync, but autocomplete changes don't. Or a translation file gets updated, this may or may not need a sync depending on which translation for updated) 

While the ratelimit for sync appears reasonable at first, we've had people who sync unconditionally in setup_hook get ratelimited in help channels, with indications that this is dynamic in some way, and not just on additions.

mikeshardmind:
dropping this for now, but may revisit the idea for the 2.6 series, there's some questions about ergonomics that would be worth exploring before committing to something intended to assist users here. In the meantime, users interested in this *can* continue syncing manually when changing anything which displays when typing `/` in discord, do something like what's present here: https://github.com/mikeshardmind/salamander-reloaded/blob/5470abb623e08722df0081ef188ebe79b43bae23/src/salamander/bot.py#L63-L72 with the knowledge that they are currently using something which may change, and need to check at each version bump of discord.py, or automate their own check of when their commands change in terms of synced signature.

PR #10084, opened 2025-01-22
Change default file size limit

### LINKED ISSUE #10083: Guild file size limit is incorrect
### Summary

When getting the guild.filesize_limit attribute with your discord bot, the guild filesize limit is wrong.

### Reproduction Steps

Create your discord bot
Get the channel you want to upload to using it's id
Get the filesize limit of the guild

### Minimal Reproducible Code

```python
# Get the channel
channel = bot.get_channel(channel_id)

# Get the limit
limit = channel.guild.filesize_limit

# Upload a large file and check for limit

if os.path.getsize(file_path) > limit:
    return

discord_file = discord.File(file, filename=os.path.basename(file_path))
await channel.send(file=discord_file)
```

### Expected Results

The code should return early since the file is too large for the server

### Actual Results

The function doesn't return early and channel.send raises HTTP error 413: Payload too large

### Intents

intents = discord.intents.default(); intents.members = True

### System Information

- Python v3.13.0-final
- discord.py v2.4.0-final
- aiohttp v3.10.10
- system info: Windows 10 10.0.19045

### Checklist

### Additional Context

_No response_

mikeshardmind:
## Summary

Resolves #10083 

Might be worth documenting that this is provided on a "best effort at time of release" means somewhere, but for now, this just covers what discord says the limit is.

## Checklist

- [X] This PR fixes an issue.


discord/utils.py
```diff
@@ -108,7 +108,7 @@
 )
 
 DISCORD_EPOCH = 1420070400000
-DEFAULT_FILE_SIZE_LIMIT_BYTES = 26214400
+DEFAULT_FILE_SIZE_LIMIT_BYTES = 10000000
```
Rapptz, line 111:
```suggestion
DEFAULT_FILE_SIZE_LIMIT_BYTES = 10485760
```

Rapptz (changes requested):
I was going to do this one locally before cutting the release but this is fine. The value is incorrect though, the limit is 10 MiB not 10 MB.

discord/utils.py
mikeshardmind: huh. I just went with the amount as advertised (which uses `MB`), guess it's not really a problem for them to under-promise slightly.

Rapptz: image

Rapptz (approved):

PR #10089, opened 2025-01-29
Add support for sending views in stateless webhooks

DA-344:
## Summary

Adds support for sending views that only have URL buttons on stateless views.

This is also done with the new ``with_components`` query parameter on the Execute Webhook endpoint.

Related DDevs PR: https://github.com/discord/discord-api-docs/pull/7368

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ui/view.py
+++ b/discord/ui/view.py
@@ -214,6 +214,13 @@ class View:
             # Wait N seconds to see if timeout data has been refreshed
             await asyncio.sleep(self.__timeout_expiry - now)
 
+    def _is_stateless(self) -> bool:
+        # this is used by webhooks to check whether a view requires a state attached
+        # or not, this simply is, whether a view has a component other than a url button
+        from .button import Button
+
+        return all(isinstance(ch, Button) and ch.url is not None for ch in self.children)
+
     def to_components(self) -> List[Dict[str, Any]]:
         def key(item: Item) -> int:
             return item._rendered_row or 0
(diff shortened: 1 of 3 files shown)


discord/ui/view.py
```diff
@@ -214,6 +214,13 @@ async def __timeout_task_impl(self) -> None:
             # Wait N seconds to see if timeout data has been refreshed
             await asyncio.sleep(self.__timeout_expiry - now)
 
+    def _is_stateless(self) -> bool:
```
Rapptz, line 217: This is already a thing. It's *already* used to check if it should be added to the view store. It's called `Item.is_dispatchable`. You can probably just leverage that instead by adding `View.is_dispatchable` which does the for loop. The only non-dispatchable item is a URL button.

discord/webhook/sync.py
```diff
@@ -290,8 +291,9 @@ def execute_webhook(
         files: Optional[Sequence[File]] = None,
         thread_id: Optional[int] = None,
         wait: bool = False,
+        has_view: bool = False,
```
Rapptz, line 294:
```suggestion
        with_components: bool = False,
```

Do this everywhere else.

Rapptz (changes requested):

discord/ui/view.py
DA-344: Oh, I forgot about that, thanks, I'll be commiting changes to that.

Rapptz (approved):

PR #10101, opened 2025-02-08
Add 'mention' property in PartialWebhookChannel

Gooraeng:
## Summary

I was trying to make a function to follow channel(s) and found `PartialWebhookChannel` can be set when `source_channel` is not `None`. So this means `source_channel_id` also would not be `None`. So I would like to suggest to add mention property for this.

I hope you'll review it positively. Thank you.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/webhook/async_.py
+++ b/discord/webhook/async_.py
@@ -659,6 +659,11 @@ class PartialWebhookChannel(Hashable):
 
     def __repr__(self) -> str:
         return f'<PartialWebhookChannel name={self.name!r} id={self.id}>'
+    
+    @property
+    def mention(self) -> str:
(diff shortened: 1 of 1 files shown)


Rapptz (changes requested):
The changeset seems fine but you need to run black to pass CI.

Thanks.

Gooraeng:
I ran `black --line-length 125 --skip-string-normalization discord/webhook/async_.py`, but I still have same problem. Should I close and reopen pr again?

Rapptz:
We use an older version of `black`, the 22.6 version so try using that instead.

Gooraeng:
Thanks. It's ready for being merged.

Rapptz (approved):

PR #10115, opened 2025-02-24
Add defaults for message object parsing

dolfies:
## Summary

A new feature is leading to some very partial message objects that bots can receive in rare contexts, crashing the library. This PR adds fallbacks in message parsing to account for that. It also catches and logs any errors in event parsing to prevent crashes like this in the future.

Example object:
```json
{
  "id": "1343642713085444138",
  "type": 0,
  "content": "test",
  "channel_id": "852892297661906993",
  "recipient_id": "852892297661906993",
  "flags": 65536,
  "channel": {
    "id": "852892297661906993",
    "type": 18,
    "last_message_id": "1343642713085444138",
    "flags": 0,
    "recipients": [
      ...
    ]
  },
  "author": {
    ...
  },
  "metadata": {
    "stuff": "stuff"
  }
}
```

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -571,10 +571,7 @@ class DiscordWebSocket:
         except KeyError:
             _log.debug('Unknown event %s.', event)
         else:
-            try:
-                func(data)
-            except Exception as exc:
-                _log.warning('Parsing event %s encountered an exception.', event, exc_info=exc)
+            func(data)
 
         # remove the dispatched listeners
         removed = []
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -571,7 +571,10 @@ class DiscordWebSocket:
         except KeyError:
             _log.debug('Unknown event %s.', event)
         else:
-            func(data)
+            try:
+                func(data)
+            except Exception as exc:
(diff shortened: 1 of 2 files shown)


Rapptz:
These crashes have historically always been on purpose since the onus of fixing these are on Discord and it makes it easier for the library to fix it in some capacity later since people will report it instead of ignoring it when logged. 

I suggest reporting a bug to Discord about how their message payload is not working as documented.

dolfies:
Agree that it's Discord's fault but wouldn't a log be better than crashing the bot? Could up the error level as well

Rapptz:
The log would just be ignored in the large sea of noise that is already logged and no one would report it because it didn't affect them. Most people run their bot and forget it. A crash forces the user to notice the issue and then report it to either me or Discord which gets it fixed instead of running in the background. I understand that the flip side of keeping it running as long as possible is attractive but I'd rather get things fixed than not.

I'm sort of okay with the commit that makes these attribute accesses with defaults since these do have a well defined default state, even if they're documented as required. It'd be nice to log the payload somewhere and then crash but chances are the payloads are way too big to just wholesale print to stdout like that.

dolfies:
I'll revert the logging then

PR #10124, opened 2025-03-04
Store raw values for AttachmentFlags in Embed media properties

yateesh-reddy:
## Summary

This pull request fixes a JSON serialization error with `AttachmentFlags` in `discord.Embed` media properties (`image`, `thumbnail`, `video`). Previously, attempting to serialize an `Embed` containing these properties resulted in a `TypeError: "Type is not JSON serializable: AttachmentFlags"` because `AttachmentFlags` is not directly serializable.

### **What Changed?**  
- **Replaced `AttachmentFlags._from_value(data.get('flags', 0))` with `data.get('flags', 0)`** in the following properties:  
  - `image`  
  - `thumbnail`  
  - `video`  

### **Why the Change?**  
- The previous implementation attempted to convert `flags` into an `AttachmentFlags` object using `AttachmentFlags._from_value()`, which is not JSON serializable.  
- The new approach directly stores the raw flag values (integers), resolving the serialization issue while preserving API behavior.  

I'm not entirely sure if this is the right fix, but after these changes, the code functions correctly.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -414,7 +414,7 @@ class Embed:
         """
         # Lying to the type checker for better developer UX.
         data = getattr(self, '_image', {})
-        data['flags'] = AttachmentFlags._from_value(data.get('flags', 0))
+        data['flags'] = data.get('flags', 0)
         return EmbedProxy(data)  # type: ignore
 
     def set_image(self, *, url: Optional[Any]) -> Self:
@@ -459,7 +459,7 @@ class Embed:
         """
         # Lying to the type checker for better developer UX.
         data = getattr(self, '_thumbnail', {})
-        data['flags'] = AttachmentFlags._from_value(data.get('flags', 0))
+        data['flags'] = data.get('flags', 0)
         return EmbedProxy(data)  # type: ignore
 
     def set_thumbnail(self, *, url: Optional[Any]) -> Self:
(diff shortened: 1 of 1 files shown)


DA-344:
This is only converted in properties, which do not affect the Embed conversion to dict when sending it to the Discord API. How are you getting that error?

yateesh-reddy:
> This is only converted in properties, which do not affect the Embed conversion to dict when sending it to the Discord API. How are you getting that error?

Supposedly:
```python
data = my_embed.to_dict()
# Here, in this data, the image section's flags contain the instance of AttachmentFlags, which causes the error.
```

Is the way I used correct? Or should it be converted in to_dict() instead?

AbstractUmbra:
I'm unable to reproduce this error. Can you provide more concise and clear reproduction steps?

DA-344:
I think it is because it gets the attribute for it and updates the key, if Embed._image exists then the flags field is updated into a AttachmentFlags object instead of being an integer.

LeoCx1000:
No, they're right. This was caused by #10123 not properly handling this.

This should be handled in `Embed.to_dict()`, to make it properly convert this to an int. Not in this current way.

image

Rapptz:
Thanks for the PR but this fix isn't proper.

PR #10139, opened 2025-03-14
add `Embed.remove_thumbnail` method

owocado:
## Summary

This PR adds `remove_thumbnail` helper method in `discord.Embed` class and compliments other similar helper methods such as `remove_author`, `remove_field`, `remove_footer` etc.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -494,6 +494,21 @@ class Embed:
 
         return self
 
+    def remove_thumbnail(self) -> Self:
+        """Clears embed's thumbnail information.
+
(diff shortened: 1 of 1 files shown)


Soheab:
You can pass `None` to the `url` kwarg in `Embed.set_thumbnail`/`image()` to remove it.  You cannot do that for author, footer, etc., because those have multiple fields that can be set.  Therefore, the `remove_x` methods exist.

Also, `Embed.remove_image()` is not a thing either; why should this be?

But that's just my take on it.

PR #10140, opened 2025-03-14
Missing word in logging setup page

oytundd:
## Summary
Running the log in debug mode should be not recommended, I think the word "not" is missing, its understandable when you think about it but probably can confuse someone who is new to programming.

## Checklist

- [ ] This PR fixes an issue.


mikeshardmind:
I don't think this is a case of missing word here, this seems to refer to 

> You can configure it to send to a file instead by using one of the built-in :mod:`logging.handlers`, such as :class:`logging.FileHandler`.

which is actually recommended

It's probably better to reword this to state that logging to a file is recommended *because* of the potential to flood stderr

Rapptz:
Yeah, wording is fine.

PR #10153, opened 2025-03-26
Fixes a bug introducing a breaking change into Embed

AbstractUmbra:
## Summary

This PR reverts earlier commits 8594dd1b309fd66bc2defc1f73540dfa9357e2c7 and cab4732b7ea8008ace32f9bd6285f7d4b3a99299 which have introduced a bug in the truthiness of `Embed`.
After this commits a `Embed` instance will *always* be truthy as it is calculated incorrectly due to the flags always being present.

As the items in these commits are **not** documented on Discord API documentation, I opted to revert the commits rather than try and devise a solution to check each proxy value and correctly report on their truthiness.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -77,7 +77,12 @@ if TYPE_CHECKING:
         proxy_url: Optional[str]
         height: Optional[int]
         width: Optional[int]
-        flags: AttachmentFlags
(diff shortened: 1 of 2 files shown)


Rapptz:
Improper fix.

PR #10154, opened 2025-03-26
Fix bug in embeds `EmbedMediaProxy` which results in the boolean value of an Embed to always be truthy

AbstractUmbra:
## Summary

This PR fixes an issue in Embed where it's boolean value is always true, due to the existing implementation of `bool` checking the `len` of the internal `__dict__`. Due to how these proxy objects are constructed, they will always have 1 key within `__dict__` making them always truthy.

## Checklist

- [ ] This PR fixes an issue.


tinoy1336:
an itty bitty bump

PR #10160, opened 2025-04-05
Fix for stuck ratelimit buckets

mikeshardmind:
## Summary

This might not be the best way to handle this, and ~~this still needs testing,~~

A more intrusive fix here would be to swap this to a pair of semaphore and timed lockout, which I'm trying to avoid having to redesign the ratelimit mechanism.

see thread: https://discord.com/channels/336642139381301249/1355766330996949034

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -461,7 +461,12 @@ class Ratelimit:
             future = self._loop.create_future()
             self._pending_requests.append(future)
             try:
-                await future
+                while not future.done():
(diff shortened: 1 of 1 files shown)


mikeshardmind:
Tested, working as intended without causing an issue of waking too soon and not respecting ratelimits, including on longer ratelimits like channel edit.

root issue this fixes is that without this, it's possible for a request to get stuck with nothing to trigger waking it.

Issue #4175, opened 2020-06-03
Some iOS quotes not working


Snawe:
### Summary

Some quotes on iOS are not working. The quotes used by the iOS mobile app seems to be dependend on the choosen keyboar language.

### Reproduction Steps

I don't have an iPhone so I can't realy reproduce it.. But users told me that quotes with the unicode `\u201e`, `\u201c` (German keyboard) are not working. Maybe there are more out there? Android uses `\u0022` no matter which language is selected

### Checklist

- [x] I have searched the open issues for duplicates.
  - found that issue: https://github.com/Rapptz/discord.py/issues/1032 but it seems that some quotes are missing

### System Information
- Python v3.6.4-final
- discord.py v1.3.3-final
- aiohttp v3.6.0
- websockets v6.0
- system info: Windows 10 10.0.18362

Vexs:
Both \u201e and \u201c are in the quotes dictionary- https://github.com/Rapptz/discord.py/blob/master/discord/ext/commands/view.py#L35

Is the order correct? Does „ precede ‟ or is it the other way maybe?

noahkw:
'„' precedes '“'; '‟' is just the wrong unicode symbol for closing German quotes I think.

Vexs:
Ah, you've got it- `_quotes['\u201e']` resolves to `\u201f` which isn't `\u201c`. Consulting the [wikipedia page,](https://en.wikipedia.org/wiki/Quotation_mark#Summary_table) `\u201c` seems to be the correct option. 
It also introduces a whole lot of quotes we don't currently support!

Vexs:
Ah, taking a second look and check- Bosnian, Croatian, Hungarian, Latvian, Polish, Romanian, and Serbian all use ` „…”` (\u201e:\u201f) while Albanian, Bulgarian, Czech, Danish, Estonian, Georgian, German, Icelandic, Lithuanian, Macedonian, Slovak, Slovene, Sorbian, and Uzbek use `„…“` (\u201e:\u201c). Ukranian, apparently allows both. 

In any case, slightly annoying, and may introduce some parsing ambiguities.

Snawe:
Yes... a bit annoying that iOS uses different quotes for different languages..

Soheab:
https://www.howtogeek.com/344310/how-to-turn-off-smart-punctuation-on-your-iphone-and-ipad/

Snawe:
[...]

PR #10161, opened 2025-04-12
[commands] Add quote pair observed in iOS

garlic-os:
## Summary

My friends' iPhones use this quote pair when they enter a string argument that has a trailing space. Even if they paste in the normal double quote character, their phones change it to this before it sends the message. This change adds the quote pair that their phones use so they can have trailing spaces in their command arguments without the quotes getting mangled.

## Example
iPhone user enters:
`"hello "`
Their phone changes it to:
`“hello “`

In discord.ext.commands, they and I expect discord.py to parse it as one string, but in actuality it does not.

## Checklist

    - [ ] I have updated the documentation to reflect the changes. **Didn't think any documentation needed to change for this. Let me know if I'm wrong**
- [ ] This PR fixes an issue.


LeoCx1000:
This is invalid, as it's overwriting the “quote” dict entry with an invalid “quote“ case (which is both a breaking change, and seems incorrect/unnatural as you wouldn't end a quote with an opening (left pointing) quotation mark). Even if dicts worked that way, this would introduce parsing ambiguity.

If iPhones are truly replacing quotes that way, that seems like a bug with the phone.

Rapptz:
See https://github.com/Rapptz/discord.py/issues/4175

PR #10166, opened 2025-04-18
Components V2

DA-344:
## Summary

Adds support for discord's components v2.

Testing and bug reporting is very much appreciated!

DDevs PR: https://github.com/discord/discord-api-docs/pull/7487

1. Adds the following Components -> UI Version:
- `SectionComponent` -> `ui.Section`
- `TextDisplay` -> `ui.TextDisplay`
- `ThumbnailComponent` -> `ui.Thumbnail`
- `MediaGalleryComponent` -> `ui.MediaGallery`
- `FileComponent` -> `ui.File`
- `SeparatorComponent` -> `ui.Separator`
- `Container` -> `ui.Container`
- `ActionRow` -> `ui.ActionRow`

2. Adds a new ui view like class, `ui.LayoutView`, that allows for these components to exist.
3. Adds the `id` attribute to all Components and UI Components.
 -> This also adds the `get_item` method to `ui.View`, `ui.LayoutView`, `ui.Container`, and `ui.Section`.

> You can install this branch for testing using ``py -m pip install -U git+https://github.com/DA-344/d.py@feat/components-v2``

<details>
<summary>These changes have been tested with the following snippets **(outdated)**:</summary>

**All of these use the jishaku package**

#### Sending Components v2
Code:
```py
class TestButton(discord.ui.Button):
    async def callback(self, interaction):
        await interaction.response.send_message('t')

class TestContainer(discord.ui.Container):
    text1 = discord.ui.TextDisplay("Hello world")
    text2 = discord.ui.TextDisplay("Row 3", row=2)

    section = discord.ui.Section(
        accessory=TestButton(
            label="Section Button",
        )
    ).add_item(discord.ui.TextDisplay("Text in a section"))


class TestView(discord.ui.LayoutView):
    container = TestContainer(id=1)

await _ctx.send(view=TestView())
```

Result (as expected):
cv2 ex

#### Sending Components v2 items on webhooks

1st snippet: sends interactible items on app owned and not app owned webhooks
```py
wh = await _channel.create_webhook(name='sumth')  # app owned

class TestButton(discord.ui.Button):
    async def callback(self, interaction):
        await interaction.response.send_message('t')

class TestContainer(discord.ui.Container):
    text1 = discord.ui.TextDisplay("Hello world")
    text2 = discord.ui.TextDisplay("Row 3", row=2)

    section = discord.ui.Section(
        accessory=TestButton(
            label="Section Button",
        )
    ).add_item(discord.ui.TextDisplay("Text in a section"))


class TestView(discord.ui.LayoutView):
    container = TestContainer(id=1)

v = TestView()

await wh.send(view=v)

wh2 = discord.Webhook.from_url('https://discord.com/api/webhooks/1362896342581641286/TOKEN', client=_bot)  # not app owned

await wh2.send(view=TestView())
```

Result (as expected):
owner wh cv2

2nd snippet: sends non-interactible items on not app owned webhook
```py
class TestButton(discord.ui.Button):
    async def callback(self, interaction):
        await interaction.response.send_message('t')

class TestContainer(discord.ui.Container):
    text1 = discord.ui.TextDisplay("Hello world")
    text2 = discord.ui.TextDisplay("Row 3", row=2)

    section = discord.ui.Section(
        accessory=TestButton(
            url="https://google.com",
            label='gugol',
        )
    ).add_item(discord.ui.TextDisplay("Text in a section"))


class TestView(discord.ui.LayoutView):
    container = TestContainer(id=1)

wh2 = discord.Webhook.from_url('https://discord.com/api/webhooks/1362896342581641286/TOKEN', client=_bot)

await wh2.send(view=TestView())
```

Result (as expected):
not owner wh cv2

#### Receiving other message's components v2

Code:
```py
yield _message.reference.resolved.components
```

Result (as expected):
receiving cv2

</details>

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -1594,8 +1594,6 @@ class Messageable:
             A Discord UI View to add to the message.
 
             .. versionadded:: 2.0
-            .. versionchanged:: 2.6
-                This now accepts :class:`discord.ui.LayoutView` instances.
         stickers: Sequence[Union[:class:`~discord.GuildSticker`, :class:`~discord.StickerItem`]]
             A list of stickers to upload. Must be a maximum of 3.
 
--- a/discord/components.py
+++ b/discord/components.py
@@ -77,7 +77,7 @@ if TYPE_CHECKING:
     from .state import ConnectionState
 
     ActionRowChildComponentType = Union['Button', 'SelectMenu', 'TextInput']
-    SectionComponentType = Union['TextDisplay', 'Button']
+    SectionComponentType = Union['TextDisplay']
     MessageComponentType = Union[
         ActionRowChildComponentType,
         SectionComponentType,
@@ -753,7 +753,7 @@ class SectionComponent(Component):
 
     Attributes
     ----------
-    components: List[Union[:class:`TextDisplay`, :class:`Button`]]
+    children: List[:class:`TextDisplay`]
         The components on this section.
     accessory: :class:`Component`
         The section accessory.
@@ -762,7 +762,7 @@ class SectionComponent(Component):
     """
 
     __slots__ = (
-        'components',
+        'children',
         'accessory',
         'id',
     )
@@ -770,14 +770,14 @@ class SectionComponent(Component):
     __repr_info__ = __slots__
 
     def __init__(self, data: SectionComponentPayload, state: Optional[ConnectionState]) -> None:
-        self.components: List[SectionComponentType] = []
+        self.children: List[SectionComponentType] = []
         self.accessory: Component = _component_factory(data['accessory'], state)  # type: ignore
         self.id: Optional[int] = data.get('id')
 
         for component_data in data['components']:
             component = _component_factory(component_data, state)
             if component is not None:
-                self.components.append(component)  # type: ignore # should be the correct type here
+                self.children.append(component)  # type: ignore # should be the correct type here
 
     @property
     def type(self) -> Literal[ComponentType.section]:
@@ -786,7 +786,7 @@ class SectionComponent(Component):
     def to_dict(self) -> SectionComponentPayload:
         payload: SectionComponentPayload = {
             'type': self.type.value,
-            'components': [c.to_dict() for c in self.components],
+            'components': [c.to_dict() for c in self.children],
             'accessory': self.accessory.to_dict(),
         }
 
--- a/discord/interactions.py
+++ b/discord/interactions.py
@@ -475,17 +475,6 @@ class Interaction(Generic[ClientT]):
         self._original_response = message
         return message
 
-    @overload
-    async def edit_original_response(
-        self,
-        *,
-        attachments: Sequence[Union[Attachment, File]] = MISSING,
-        view: LayoutView,
-        allowed_mentions: Optional[AllowedMentions] = None,
-    ) -> InteractionMessage:
-        ...
-
-    @overload
     async def edit_original_response(
         self,
         *,
@@ -493,20 +482,7 @@ class Interaction(Generic[ClientT]):
         embeds: Sequence[Embed] = MISSING,
         embed: Optional[Embed] = MISSING,
         attachments: Sequence[Union[Attachment, File]] = MISSING,
-        view: Optional[View] = MISSING,
-        allowed_mentions: Optional[AllowedMentions] = None,
-        poll: Poll = MISSING,
-    ) -> InteractionMessage:
-        ...
-
-    async def edit_original_response(
-        self,
-        *,
-        content: Optional[str] = MISSING,
-        embeds: Sequence[Embed] = MISSING,
-        embed: Optional[Embed] = MISSING,
-        attachments: Sequence[Union[Attachment, File]] = MISSING,
-        view: Optional[BaseView] = MISSING,
+        view: Optional[Union[View, LayoutView]] = MISSING,
         allowed_mentions: Optional[AllowedMentions] = None,
         poll: Poll = MISSING,
     ) -> InteractionMessage:
@@ -1522,7 +1498,7 @@ class InteractionMessage(Message):
         embeds: Sequence[Embed] = MISSING,
         embed: Optional[Embed] = MISSING,
         attachments: Sequence[Union[Attachment, File]] = MISSING,
-        view: Optional[BaseView] = MISSING,
+        view: Optional[Union[View, LayoutView]] = MISSING,
         allowed_mentions: Optional[AllowedMentions] = None,
         delete_after: Optional[float] = None,
         poll: Poll = MISSING,
--- a/discord/interactions.py
+++ b/discord/interactions.py
@@ -525,9 +525,6 @@ class Interaction(Generic[ClientT]):
                 If you want to update the message to have a :class:`~discord.ui.LayoutView`, you must
                 explicitly set the ``content``, ``embed``, ``embeds``, and ``attachments`` parameters to
                 ``None`` if the previous message had any.
-
-            .. versionchanged:: 2.6
-                This now accepts :class:`~discord.ui.LayoutView` instances.
         poll: :class:`Poll`
             The poll to create when editing the message.
 
@@ -984,9 +981,6 @@ class InteractionResponse(Generic[ClientT]):
             Indicates if the message should be sent using text-to-speech.
         view: Union[:class:`discord.ui.View`, :class:`discord.ui.LayoutView`]
             The view to send with the message.
-
-            .. versionchanged:: 2.6
-                This now accepts :class:`discord.ui.LayoutView` instances.
         ephemeral: :class:`bool`
             Indicates if the message should only be visible to the user who started the interaction.
             If a view is sent with an ephemeral message and it has no timeout set then the timeout
@@ -1095,19 +1089,6 @@ class InteractionResponse(Generic[ClientT]):
             type=self._response_type,
         )
 
-    @overload
-    async def edit_message(
-        self,
-        *,
-        attachments: Sequence[Union[Attachment, File]] = MISSING,
-        view: LayoutView,
-        allowed_mentions: Optional[AllowedMentions] = MISSING,
-        delete_after: Optional[float] = None,
-        suppress_embeds: bool = MISSING,
-    ) -> Optional[InteractionCallbackResponse[ClientT]]:
-        ...
-
-    @overload
     async def edit_message(
         self,
         *,
@@ -1115,21 +1096,7 @@ class InteractionResponse(Generic[ClientT]):
         embed: Optional[Embed] = MISSING,
         embeds: Sequence[Embed] = MISSING,
         attachments: Sequence[Union[Attachment, File]] = MISSING,
-        view: Optional[View] = MISSING,
-        allowed_mentions: Optional[AllowedMentions] = MISSING,
-        delete_after: Optional[float] = None,
-        suppress_embeds: bool = MISSING,
-    ) -> Optional[InteractionCallbackResponse[ClientT]]:
-        ...
-
-    async def edit_message(
-        self,
-        *,
-        content: Optional[Any] = MISSING,
-        embed: Optional[Embed] = MISSING,
-        embeds: Sequence[Embed] = MISSING,
-        attachments: Sequence[Union[Attachment, File]] = MISSING,
-        view: Optional[BaseView] = MISSING,
+        view: Optional[Union[View, LayoutView]] = MISSING,
         allowed_mentions: Optional[AllowedMentions] = MISSING,
         delete_after: Optional[float] = None,
         suppress_embeds: bool = MISSING,
@@ -1168,9 +1135,6 @@ class InteractionResponse(Generic[ClientT]):
                 To update the message to add a :class:`~discord.ui.LayoutView`, you
                 must explicitly set the ``content``, ``embed``, ``embeds``, and
                 ``attachments`` parameters to either ``None`` or an empty array, as appropriate.
-
-            .. versionchanged:: 2.6
-                This now accepts :class:`~discord.ui.LayoutView` instances.
         allowed_mentions: Optional[:class:`~discord.AllowedMentions`]
             Controls the mentions being processed in this message. See :meth:`.Message.edit`
             for more information.
@@ -1536,9 +1500,6 @@ class InteractionMessage(Message):
                 If you want to update the message to have a :class:`~discord.ui.LayoutView`, you must
                 explicitly set the ``content``, ``embed``, ``embeds``, and ``attachments`` parameters to
                 ``None`` if the previous message had any.
-
-            .. versionchanged:: 2.6
-                This now accepts :class:`~discord.ui.LayoutView` instances.
         delete_after: Optional[:class:`float`]
             If provided, the number of seconds to wait in the background
             before deleting the message we just sent. If the deletion fails,
--- a/discord/interactions.py
+++ b/discord/interactions.py
@@ -546,9 +546,9 @@ class Interaction(Generic[ClientT]):
 
             .. note::
 
-                To update the message to add a :class:`~discord.ui.LayoutView`, you
-                must explicitly set the ``content``, ``embed``, ``embeds``, and
-                ``attachments`` parameters to either ``None`` or an empty array, as appropriate.
+                If you want to update the message to have a :class:`~discord.ui.LayoutView`, you must
+                explicitly set the ``content``, ``embed``, ``embeds``, and ``attachments`` parameters to
+                ``None`` if the previous message had any.
 
             .. versionchanged:: 2.6
                 This now accepts :class:`~discord.ui.LayoutView` instances.
@@ -1557,9 +1557,9 @@ class InteractionMessage(Message):
 
             .. note::
 
-                To update the message to add a :class:`~discord.ui.LayoutView`, you
-                must explicitly set the ``content``, ``embed``, ``embeds``, and
-                ``attachments`` parameters to either ``None`` or an empty array, as appropriate.
+                If you want to update the message to have a :class:`~discord.ui.LayoutView`, you must
+                explicitly set the ``content``, ``embed``, ``embeds``, and ``attachments`` parameters to
+                ``None`` if the previous message had any.
 
             .. versionchanged:: 2.6
                 This now accepts :class:`~discord.ui.LayoutView` instances.
@@ -1600,7 +1600,7 @@ class InteractionMessage(Message):
             embeds=embeds,
             embed=embed,
             attachments=attachments,
-            view=view,  # type: ignore
+            view=view,
             allowed_mentions=allowed_mentions,
             poll=poll,
         )
--- a/discord/message.py
+++ b/discord/message.py
@@ -101,7 +101,7 @@ if TYPE_CHECKING:
     from .mentions import AllowedMentions
     from .user import User
     from .role import Role
-    from .ui.view import BaseView, View, LayoutView
+    from .ui.view import View, LayoutView
 
     EmojiInputType = Union[Emoji, PartialEmoji, str]
 
--- a/discord/message.py
+++ b/discord/message.py
@@ -1408,8 +1408,8 @@ class PartialMessage(Hashable):
             .. note::
 
                 If you want to update the message to have a :class:`~discord.ui.LayoutView`, you must
-                explicitly set to ``None`` or empty array, as required, the ``content``, ``embed``,
-                ``embeds``, and ``attachments`` parameters.
+                explicitly set the ``content``, ``embed``, ``embeds``, and ``attachments`` parameters to
+                ``None`` if the previous message had any.
 
             .. versionchanged:: 2.6
                 This now accepts :class:`~discord.ui.LayoutView` instances.
@@ -3000,8 +3000,8 @@ class Message(PartialMessage, Hashable):
             .. note::
 
                 If you want to update the message to have a :class:`~discord.ui.LayoutView`, you must
-                explicitly set to ``None`` or empty array, as required, the ``content``, ``embed``,
-                ``embeds``, and ``attachments`` parameters.
+                explicitly set the ``content``, ``embed``, ``embeds``, and ``attachments`` parameters to
+                ``None`` if the previous message had any.
 
             .. versionchanged:: 2.6
                 This now accepts :class:`~discord.ui.LayoutView` instances.
--- a/discord/message.py
+++ b/discord/message.py
@@ -1305,43 +1305,6 @@ class PartialMessage(Hashable):
         else:
             await self._state.http.delete_message(self.channel.id, self.id)
 
-    @overload
-    async def edit(
-        self,
-        *,
-        view: LayoutView,
-        attachments: Sequence[Union[Attachment, File]] = ...,
-        delete_after: Optional[float] = ...,
-        allowed_mentions: Optional[AllowedMentions] = ...,
-    ) -> Message:
-        ...
-
-    @overload
-    async def edit(
-        self,
-        *,
-        content: Optional[str] = ...,
-        embed: Optional[Embed] = ...,
-        attachments: Sequence[Union[Attachment, File]] = ...,
-        delete_after: Optional[float] = ...,
-        allowed_mentions: Optional[AllowedMentions] = ...,
-        view: Optional[View] = ...,
-    ) -> Message:
-        ...
-
-    @overload
-    async def edit(
-        self,
-        *,
-        content: Optional[str] = ...,
-        embeds: Sequence[Embed] = ...,
-        attachments: Sequence[Union[Attachment, File]] = ...,
-        delete_after: Optional[float] = ...,
-        allowed_mentions: Optional[AllowedMentions] = ...,
-        view: Optional[View] = ...,
-    ) -> Message:
-        ...
-
     async def edit(
         self,
         *,
@@ -1351,7 +1314,7 @@ class PartialMessage(Hashable):
         attachments: Sequence[Union[Attachment, File]] = MISSING,
         delete_after: Optional[float] = None,
         allowed_mentions: Optional[AllowedMentions] = MISSING,
-        view: Optional[BaseView] = MISSING,
+        view: Optional[Union[View, LayoutView]] = MISSING,
     ) -> Message:
         """|coro|
 
@@ -1410,9 +1373,6 @@ class PartialMessage(Hashable):
                 If you want to update the message to have a :class:`~discord.ui.LayoutView`, you must
                 explicitly set the ``content``, ``embed``, ``embeds``, and ``attachments`` parameters to
                 ``None`` if the previous message had any.
-
-            .. versionchanged:: 2.6
-                This now accepts :class:`~discord.ui.LayoutView` instances.
 
         Raises
         -------
@@ -2897,34 +2857,6 @@ class Message(PartialMessage, Hashable):
         # Fallback for unknown message types
         return ''
 
-    @overload
-    async def edit(
-        self,
-        *,
-        content: Optional[str] = ...,
-        embed: Optional[Embed] = ...,
-        attachments: Sequence[Union[Attachment, File]] = ...,
-        suppress: bool = ...,
-        delete_after: Optional[float] = ...,
-        allowed_mentions: Optional[AllowedMentions] = ...,
-        view: Optional[BaseView] = ...,
-    ) -> Message:
-        ...
-
-    @overload
-    async def edit(
-        self,
-        *,
-        content: Optional[str] = ...,
-        embeds: Sequence[Embed] = ...,
-        attachments: Sequence[Union[Attachment, File]] = ...,
-        suppress: bool = ...,
-        delete_after: Optional[float] = ...,
-        allowed_mentions: Optional[AllowedMentions] = ...,
-        view: Optional[BaseView] = ...,
-    ) -> Message:
-        ...
-
     async def edit(
         self,
         *,
@@ -2935,7 +2867,7 @@ class Message(PartialMessage, Hashable):
         suppress: bool = False,
         delete_after: Optional[float] = None,
         allowed_mentions: Optional[AllowedMentions] = MISSING,
-        view: Optional[BaseView] = MISSING,
+        view: Optional[Union[View, LayoutView]] = MISSING,
     ) -> Message:
         """|coro|
 
@@ -3002,9 +2934,6 @@ class Message(PartialMessage, Hashable):
                 If you want to update the message to have a :class:`~discord.ui.LayoutView`, you must
                 explicitly set the ``content``, ``embed``, ``embeds``, and ``attachments`` parameters to
                 ``None`` if the previous message had any.
-
-            .. versionchanged:: 2.6
-                This now accepts :class:`~discord.ui.LayoutView` instances.
 
         Raises
         -------
--- a/discord/types/components.py
+++ b/discord/types/components.py
@@ -34,7 +34,7 @@ ComponentType = Literal[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 17]
 ButtonStyle = Literal[1, 2, 3, 4, 5, 6]
 TextStyle = Literal[1, 2]
 DefaultValueType = Literal['user', 'role', 'channel']
-SeparatorSize = Literal[1, 2]
+SeparatorSpacing = Literal[1, 2]
 MediaItemLoadingState = Literal[0, 1, 2, 3]
 
 
@@ -177,7 +177,7 @@ class FileComponent(ComponentBase):
 class SeparatorComponent(ComponentBase):
     type: Literal[14]
     divider: NotRequired[bool]
-    spacing: NotRequired[SeparatorSize]
+    spacing: NotRequired[SeparatorSpacing]
 
 
 class ContainerComponent(ComponentBase):
--- a/discord/ui/action_row.py
+++ b/discord/ui/action_row.py
@@ -225,7 +225,7 @@ class ActionRow(Item[V]):
         Parameters
         ----------
         item: :class:`Item`
-            The item to add to the row.
+            The item to add to the action row.
 
         Raises
         ------
--- a/discord/ui/action_row.py
+++ b/discord/ui/action_row.py
@@ -480,10 +480,10 @@ class ActionRow(Item[V]):
         default_values: Sequence[ValidDefaultValues] = MISSING,
         id: Optional[int] = None,
     ) -> SelectCallbackDecorator[BaseSelectT]:
-        """A decorator that attaches a select menu to a component.
+        """A decorator that attaches a select menu to the action row.
 
         The function being decorated should have three parameters, ``self`` representing
-        the :class:`discord.ui.LayoutView`, the :class:`discord.Interaction` you receive and
+        the :class:`discord.ui.ActionRow`, the :class:`discord.Interaction` you receive and
         the chosen select class.
 
         To obtain the selected values inside the callback, you can use the ``values`` attribute of the chosen class in the callback. The list of values
--- a/discord/ui/action_row.py
+++ b/discord/ui/action_row.py
@@ -114,7 +114,7 @@ class ActionRow(Item[V]):
             # or you can use your subclass:
             # row = MyActionRow()
 
-            # you can create items with row.button and row.select
+            # you can add items with row.button and row.select
             @row.button(label='A button!')
             async def row_button(self, interaction: discord.Interaction, button: discord.ui.Button):
                 await interaction.response.send_message('You clicked a button!')
--- a/discord/ui/action_row.py
+++ b/discord/ui/action_row.py
@@ -129,7 +129,6 @@ class ActionRow(Item[V]):
 
     __action_row_children_items__: ClassVar[List[ItemCallbackType[Any]]] = []
     __discord_ui_action_row__: ClassVar[bool] = True
-    __discord_ui_update_view__: ClassVar[bool] = True
     __item_repr_attributes__ = ('id',)
 
     def __init__(
@@ -175,9 +174,13 @@ class ActionRow(Item[V]):
             children.append(item)
         return children
 
-    def _update_children_view(self, view: LayoutView) -> None:
+    def _update_view(self, view) -> None:
+        self._view = view
         for child in self._children:
-            child._view = view  # pyright: ignore[reportAttributeAccessIssue]
+            child._view = view
+
+    def _has_nested(self):
+        return True
 
     def _is_v2(self) -> bool:
         # although it is not really a v2 component the only usecase here is for
@@ -238,7 +241,7 @@ class ActionRow(Item[V]):
         if not isinstance(item, Item):
             raise TypeError(f'expected Item not {item.__class__.__name__}')
 
-        item._view = self._view
+        item._update_view(self.view)
         item._parent = self
         self._weight += 1
         self._children.append(item)
--- a/discord/ui/action_row.py
+++ b/discord/ui/action_row.py
@@ -41,7 +41,7 @@ from typing import (
     overload,
 )
 
-from .item import Item, ItemCallbackType
+from .item import I, Item
 from .button import Button, button as _button
 from .select import select as _select, Select, UserSelect, RoleSelect, ChannelSelect, MentionableSelect
 from ..components import ActionRow as ActionRowComponent
@@ -61,12 +61,15 @@ if TYPE_CHECKING:
         RoleSelectT,
         UserSelectT,
         SelectT,
-        SelectCallbackDecorator,
     )
     from ..emoji import Emoji
     from ..components import SelectOption
     from ..interactions import Interaction
 
+    ItemCallbackType = Callable[['S', Interaction[Any], I], Coroutine[Any, Any, Any]]
+    SelectCallbackDecorator = Callable[[ItemCallbackType['S', BaseSelectT]], BaseSelectT]
+
+S = TypeVar('S', bound='ActionRow', covariant=True)
 V = TypeVar('V', bound='LayoutView', covariant=True)
 
 __all__ = ('ActionRow',)
@@ -75,8 +78,8 @@ __all__ = ('ActionRow',)
 class _ActionRowCallback:
     __slots__ = ('row', 'callback', 'item')
 
-    def __init__(self, callback: ItemCallbackType[Any], row: ActionRow, item: Item[Any]) -> None:
-        self.callback: ItemCallbackType[Any] = callback
+    def __init__(self, callback: ItemCallbackType[S, Any], row: ActionRow, item: Item[Any]) -> None:
+        self.callback: ItemCallbackType[Any, Any] = callback
         self.row: ActionRow = row
         self.item: Item[Any] = item
 
@@ -127,7 +130,7 @@ class ActionRow(Item[V]):
         The ID of this component. This must be unique across the view.
     """
 
-    __action_row_children_items__: ClassVar[List[ItemCallbackType[Any]]] = []
+    __action_row_children_items__: ClassVar[List[ItemCallbackType[Self, Any]]] = []
     __discord_ui_action_row__: ClassVar[bool] = True
     __item_repr_attributes__ = ('id',)
 
@@ -149,7 +152,7 @@ class ActionRow(Item[V]):
     def __init_subclass__(cls) -> None:
         super().__init_subclass__()
 
-        children: Dict[str, ItemCallbackType[Any]] = {}
+        children: Dict[str, ItemCallbackType[Self, Any]] = {}
         for base in reversed(cls.__mro__):
             for name, member in base.__dict__.items():
                 if hasattr(member, '__discord_ui_model_type__'):
@@ -331,7 +334,7 @@ class ActionRow(Item[V]):
         style: ButtonStyle = ButtonStyle.secondary,
         emoji: Optional[Union[str, Emoji, PartialEmoji]] = None,
         id: Optional[int] = None,
-    ) -> Callable[[ItemCallbackType[Button[V]]], Button[V]]:
+    ) -> Callable[[ItemCallbackType[S, Button[V]]], Button[V]]:
         """A decorator that attaches a button to the action row.
 
         The function being decorated should have three parameters, ``self`` representing
@@ -367,7 +370,7 @@ class ActionRow(Item[V]):
             .. versionadded:: 2.6
         """
 
-        def decorator(func: ItemCallbackType[Button[V]]) -> ItemCallbackType[Button[V]]:
+        def decorator(func: ItemCallbackType[S, Button[V]]) -> ItemCallbackType[S, Button[V]]:
             ret = _button(
                 label=label,
                 custom_id=custom_id,
@@ -395,7 +398,7 @@ class ActionRow(Item[V]):
         max_values: int = ...,
         disabled: bool = ...,
         id: Optional[int] = ...,
-    ) -> SelectCallbackDecorator[SelectT]:
+    ) -> SelectCallbackDecorator[S, SelectT]:
         ...
 
     @overload
@@ -412,7 +415,7 @@ class ActionRow(Item[V]):
         disabled: bool = ...,
         default_values: Sequence[ValidDefaultValues] = ...,
         id: Optional[int] = ...,
-    ) -> SelectCallbackDecorator[UserSelectT]:
+    ) -> SelectCallbackDecorator[S, UserSelectT]:
         ...
 
     @overload
@@ -429,7 +432,7 @@ class ActionRow(Item[V]):
         disabled: bool = ...,
         default_values: Sequence[ValidDefaultValues] = ...,
         id: Optional[int] = ...,
-    ) -> SelectCallbackDecorator[RoleSelectT]:
+    ) -> SelectCallbackDecorator[S, RoleSelectT]:
         ...
 
     @overload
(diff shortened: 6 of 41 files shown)


discord/ui/container.py
```diff
@@ -0,0 +1,318 @@
...
+        row: Optional[int] = None,
+        id: Optional[int] = None,
+    ) -> None:
+        super().__init__()
+        self.__dispatchable: List[Item[V]] = []
+        self._children: List[Item[V]] = self._init_children()
+
+        if children is not MISSING:
+            if len(children) + len(self._children) > 10:
+                raise ValueError('maximum number of children exceeded')
```
qwewqa, line 106: Doesn't seem like `children` actually gets added to the container.

DA-344: Oh, that must've been an oversight lol
Fixed on latest commit

qwewqa:
If an action row is added via `add_item`, `view` doesn't get set on buttons and they won't work:
```
class TestButton(discord.ui.Button):
    async def callback(self, interaction: Interaction):
        await interaction.response.send_message("Test button clicked!")


class TestActionRow(discord.ui.ActionRow):
    def __init__(self):
        super().__init__()
        self.add_item(TestButton(label="test2"))

    @discord.ui.button(label="test1")
    async def test_button(self, interaction: Interaction, button: discord.ui.Button):
        await interaction.response.send_message("Test button clicked!")


class TestLayout(discord.ui.LayoutView):
    def __init__(self):
        super().__init__()
        # test1 and test2 don't work
        self.add_item(TestActionRow())
        # test3 works
        self.add_item(discord.ui.Section(accessory=TestButton(label="test3")).add_item("text"))
```

DA-344:
> If an action row is added via `add_item`, `view` doesn't get set on buttons and they won't work

Thanks for reporting it, this should be fixed in the latest commit.

jakobdylanc:
With these changes is there a straightforward way to send the simplest form of a components v2 message, aka just plaintext content?

The reason for wanting this is because the character limit would be 4000 instead of the old 2000.

It would be nice if the character limit of the existing "content" field was increased to 4000; and just make it use components v2 under the hood if it has to.

DA-344:
You can make your own LayoutView subclass and automatically assign it a content with a TextDisplay, as making content larger would make it not be consistent with the API.

theProgramer-J:
I have a problem : I cannot get all components from a message
image
image
I want to get container and text display components

Soheab:
> I have a problem : I cannot get all components from a message image image I want to get container and text display components

This PR hasn't even been merged yet; how did you already get the docs lol?  Anyway, with this PR, `Message.components` does return all the new components too. Have you tried it?

theProgramer-J:
Yes I tried and I found nothing

Soheab:
> Yes I tried and I get [ ]

Code? 


Also, consider joining the [support server](https://discord.gg/dpy) and reporting this in the dedicated [discussion thread](https://canary.discord.com/channels/336642139381301249/1345167602304946206).

theProgramer-J:
I had forgotten to `py -m pip install -U git+https://github.com/DA-344/d.py@feat/components-v2` lol

discord/components.py
```diff
@@ -116,31 +177,39 @@ class ActionRow(Component):
...
     def __init__(self, data: ActionRowPayload, /) -> None:
+        self.id: Optional[int] = data.get('id')
         self.children: List[ActionRowChildComponentType] = []
 
         for component_data in data.get('components', []):
             component = _component_factory(component_data)
 
             if component is not None:
-                self.children.append(component)
+                self.children.append(component)  # type: ignore # should be the correct type here
```
mikeshardmind, line 198: This type ignore shouldn't be necessary, may require fixing the typing of `_component_factory`

discord/ui/section.py
```diff
@@ -0,0 +1,238 @@
...
+        return True
+
+    # Accessory can be a button, and thus it can have a callback so, maybe
+    # allow for section to be dispatchable and make the callback func
+    # be accessory component callback, only called if accessory is
+    # dispatchable?
+    def is_dispatchable(self) -> bool:
+        return self.accessory.is_dispatchable()
+
+    def walk_children(self) -> Generator[Item[V], Any, None]:
```
mikeshardmind, line 118: Generators that only yield and dont utilize send values should have a send type of None, not Any
```suggestion
    def walk_children(self) -> Generator[Item[V], None, None]:
```

discord/ui/view.py
```diff
@@ -533,21 +585,269 @@ async def wait(self) -> bool:
         """
         return await self.__stopped
 
+    def walk_children(self):
```
mikeshardmind, line 588: missing a return type here

mikeshardmind (changes requested):
Minor typing related things I noticed, these don't affect the functionality of the code, only the developer experience.

discord/components.py
DA-344: Well, it errors because ``_component_factory`` can is typed as it returns ``Component``, and ``ActionRow.children`` as ``Union[Button, Select, TextInput]``, that is why I had the ``# type: ignore`` in the first place.

mikeshardmind: yeah, may be worth keeping the ignore for now, I looked in more detail, and fixing this requires an overload per payload type for `_component_factory`, and this doesn't leak out to public interface.

discord/ui/view.py
DA-344: fixed

discord/ui/container.py
```diff
@@ -0,0 +1,415 @@
...
+        Raises
+        ------
+        TypeError
+            An :class:`Item` was not passed.
+        ValueError
+            Maximum number of children has been exceeded (10).
+        """
+
+        if len(self._children) >= 10:
+            raise ValueError('maximum number of children exceeded')
```
muddyfish, line 348: This is no longer a limit: https://discord.com/channels/613425648685547541/697138785317814292/1367197980230156478

cadenkoj (approved):

Showdown76py (changes requested):
Concerning `discord.ui.view.BaseView.is_persistent()`, we need to check if we're dealing with a `__discord_ui_container__`, as they can't have timeouts and will always return True to `self.timeout is None`.

DA-344:
I don't know what version of the PR you are using, but Containers donnot longer inherit of ``BaseView``, and are just normal ``Item``s. I've pushed some fixes to is_persistent on v2 items.

Atlas-timeless:
Hey, i noticed an error, when you are sending a view with webhook (with client) 

and then, editing it, no edit are pushed, and no error are raised 


 ```fview = FIRST_LAYOUT()
  webhook = Webhook.from_url(WEBHOOK, client=Bot)  


  message = await webhook.send(view=fview,wait=True)


  await asyncio.sleep(5)


  new_view = NewsView()

  await webhook.edit_message(message_id=message.id,view=new_view)

Atlas-timeless:
> > Hey, i noticed an error, when you are sending a view with webhook (with client)
> > and then, editing it, no edit are pushed, and no error are raised
> > ```
> >  webhook = Webhook.from_url(WEBHOOK, client=Bot)  
> > 
> > 
> >  message = await webhook.send(view=fview,wait=True)
> > 
> > 
> >  await asyncio.sleep(5)
> > 
> > 
> >  new_view = NewsView()
> > 
> >  await webhook.edit_message(message_id=message.id,view=new_view)
> > ```
> 
> Would be useful to show what the view looks like.

Any view, a simple view with a MediaGallery arent editable 

class first(Container):
    def __init__(self, id):
        super().__init__(id=id)
        self.add_item(MediaGallery(row=2).add_item(item=MediaGalleryItem(media=MEDIA_URL),))

        self.add_item(TextDisplay("We're bringing new components to messages that you can use in your apps."))

DA-344:
If no errors are logged and the request was sent successfully, this means it is a discord side issue. Try enabling d.py debug mode, and show us the whole view you have. If you try sending a v2 view and then editing it with a non-v2, it is not possible.

Atlas-timeless:
> If no errors are logged and the request was sent successfully, this means it is a discord side issue. Try enabling d.py debug mode, and show us the whole view you have, both the `NewsView` class and the `fview` one. If you try sending a v2 view and then editing it with a non-v2, it is not possible.

```
from discord.ui import LayoutView, ActionRow, Separator, Section, Container, TextDisplay, Thumbnail, MediaGallery, Button, Item, File, Select,Modal
from discord.components import MediaGalleryItem 
from discord import ui


class first(Container):
    def __init__(self, id):
        super().__init__(id=id)
        self.add_item(TextDisplay("FIRST LAYOUT"))

class FIRST_LAYOUT(LayoutView):
    def __init__(self):
        super().__init__()
        self.add_item(first(id=1))

class second(Container):
    def __init__(self, id):
        super().__init__(id=id)
        self.add_item(TextDisplay("SECOND LAYOUT"))
        
class SECOND_LAYOUT(LayoutView):
    def __init__(self):
        super().__init__()
        self.add_item(second(id=2))


@app_commands.command(name="post_components")
@app_commands.checks.has_permissions(administrator=True)
async def post_components(self,interaction: discord.Interaction):
    await interaction.response.defer(ephemeral=True)
    session = aiohttp.ClientSession()
    try:

        first_view = FIRST_LAYOUT()
        webhook = Webhook.from_url(COMPONENT_WEBHOOK, client=Bot)  
        print(1)
        message = await webhook.send(view=first_view,wait=True)

        second_view = SECOND_LAYOUT()
        await asyncio.sleep(2)
        await message.edit(view=second_view)
        print(2)

        if session:
            await session.close()

    except Exception as e:
        Log.print(e,"ERROR")
        if session:
            await session.close()
```

Seamorethemaster:
keep on getting
Invalid Form Body In components.0.components.1.accessory: Value of field "type" must be one of (2, 11). 
when i try to send a image

Soheab:
> keep on getting Invalid Form Body In components.0.components.1.accessory: Value of field "type" must be one of (2, 11). when i try to send a image

A section's accessory can only be an instance of `discord.ui.Thumbnail()` or `discord.ui.Button()`.

acqxi:
I noticed that your PR claims to include support for the new component v2 system.

However, I couldn’t find any implementation or example in the code that demonstrates usage of [the channel select UI element](https://discord.com/developers/docs/components/reference#channel-select).

Could you confirm whether your PR currently includes support for this, and if so, where it’s implemented? If not, do you plan to add it?

Th4nhZ:
channel select is not a new thing, they exist pre components v2

discord/ui/container.py
```diff
@@ -0,0 +1,449 @@
...
+        self._children = value
+
+    @property
+    def accent_colour(self) -> Optional[Union[Colour, int]]:
+        """Optional[Union[:class:`discord.Colour`, :class:`int`]]: The colour of the container, or ``None``."""
+        return self._colour
+
+    @accent_colour.setter
+    def accent_colour(self, value: Optional[Union[Colour, int]]) -> None:
+        self._colour = value
```
Soheab, line 253: This should check the value type.

Soheab (commented):
Good work!

discord/types/components.py
```diff
@@ -24,24 +24,31 @@
...
 
 from .emoji import PartialEmoji
 from .channel import ChannelType
 
-ComponentType = Literal[1, 2, 3, 4]
+ComponentType = Literal[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 17]
 ButtonStyle = Literal[1, 2, 3, 4, 5, 6]
 TextStyle = Literal[1, 2]
-DefaultValueType = Literal['user', 'role', 'channel']
+DefaultValueType = Literal["user", "role", "channel"]
```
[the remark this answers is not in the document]
DA-344: Error on my end apparently, will revert soon

discord/ui/container.py
```diff
@@ -0,0 +1,449 @@
...
+        super().__init__()
+        self.__dispatchable: List[Item[V]] = []
+        self._children: List[Item[V]] = self._init_children()
+
+        if children is not MISSING:
+            for child in children:
+                self.add_item(child)
+
+        self.spoiler: bool = spoiler
+        self._colour = accent_colour or accent_color
```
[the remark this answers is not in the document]
DA-344: Thanks!

discord/types/components.py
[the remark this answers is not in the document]
DA-344: Fixed.

discord/ui/container.py
DA-344: Added the check on latest commit.

Atlas-timeless:
> > If no errors are logged and the request was sent successfully, this means it is a discord side issue. Try enabling d.py debug mode, and show us the whole view you have, both the `NewsView` class and the `fview` one. If you try sending a v2 view and then editing it with a non-v2, it is not possible.
> 
> Here is an example that not working,
> 
> ```
> from discord.ui import LayoutView, ActionRow, Separator, Section, Container, TextDisplay, Thumbnail, MediaGallery, Button, Item, File, Select,Modal
> from discord.components import MediaGalleryItem 
> from discord import ui
> 
> 
> class first(Container):
>     def __init__(self, id):
>         super().__init__(id=id)
>         self.add_item(TextDisplay("FIRST LAYOUT"))
> 
> class FIRST_LAYOUT(LayoutView):
>     def __init__(self):
>         super().__init__()
>         self.add_item(first(id=1))
> 
> class second(Container):
>     def __init__(self, id):
>         super().__init__(id=id)
>         self.add_item(TextDisplay("SECOND LAYOUT"))
>         
> class SECOND_LAYOUT(LayoutView):
>     def __init__(self):
>         super().__init__()
>         self.add_item(second(id=2))
> 
> 
> @app_commands.command(name="post_components")
> @app_commands.checks.has_permissions(administrator=True)
> async def post_components(self,interaction: discord.Interaction):
>     await interaction.response.defer(ephemeral=True)
>     session = aiohttp.ClientSession()
>     try:
> 
>         first_view = FIRST_LAYOUT()
>         webhook = Webhook.from_url(COMPONENT_WEBHOOK, client=Bot)  
>         print(1)
>         message = await webhook.send(view=first_view,wait=True)
> 
>         second_view = SECOND_LAYOUT()
>         await asyncio.sleep(2)
>         await message.edit(view=second_view)
>         print(2)
> 
>         if session:
>             await session.close()
> 
>     except Exception as e:
>         Log.print(e,"ERROR")
>         if session:
>             await session.close()
> ```



Can't edit a component V2 with Webhook where it should be able to ?
any news about this issue ?

discord/components.py
```diff
@@ -642,22 +736,564 @@ def from_user(cls, user: Snowflake, /) -> Self:
...
+        if colour is not None:
+            self._colour = Colour(colour)
+
+    @property
+    def accent_colour(self) -> Optional[Colour]:
+        """Optional[:class:`Colour`]: The container's accent colour."""
+        return self._colour
+
+    accent_color = accent_colour
+
```
owocado, line 1294:
```suggestion

    @property
    def type(self) -> Literal[ComponentType.container]:
        return ComponentType.container

```

discord/ui/container.py
```diff
@@ -0,0 +1,401 @@
...
+        self._children = value
+
+    @property
+    def accent_colour(self) -> Optional[Union[Colour, int]]:
+        """Optional[Union[:class:`discord.Colour`, :class:`int`]]: The colour of the container, or ``None``."""
+        return self._colour
+
+    @accent_colour.setter
+    def accent_colour(self, value: Optional[Union[Colour, int]]) -> None:
+        if not isinstance(value, (int, Colour)):
```
jay3332:
```suggestion
        if value is not None and not isinstance(value, (int, Colour)):
```
it should be valid to set `container.accent_colour = None`

jay3332 (changes requested):

discord/ui/container.py
DA-344: Thanks!

discord/abc.py
```diff
@@ -1546,10 +1578,12 @@ async def send(
             If set, overrides the :attr:`~discord.AllowedMentions.replied_user` attribute of ``allowed_mentions``.
 
             .. versionadded:: 1.6
-        view: :class:`discord.ui.View`
+        view: Union[:class:`discord.ui.View`, :class:`discord.ui.LayoutView`]
             A Discord UI View to add to the message.
 
             .. versionadded:: 2.0
+            .. versionchanged:: 2.6
+                This now accepts :class:`discord.ui.LayoutView` instances.
```
Soheab: Is this really necessary? It still takes a View and LayoutView is still a view with with different types of components.

discord/ui/container.py
```diff
@@ -0,0 +1,401 @@
...
+    def __call__(self, interaction: Interaction) -> Coroutine[Any, Any, Any]:
+        return self.callback(self.container, interaction, self.item)
+
+
+class Container(Item[V]):
+    r"""Represents a UI container.
+
+    This is a top-level layout component that can only be used on :class:`LayoutView`
+    and can contain :class:`ActionRow`\s, :class:`TextDisplay`\s, :class:`Section`\s,
+    :class:`MediaGallery`\s, and :class:`File`\s in it.
```
Sacul0457: Seems like you missed the Separator Class for the Container's doc strings

playfairs:
when's this shi getting pushed 😭💔 we need cv2

discord/ui/view.py
```diff
@@ -271,13 +455,22 @@ def children(self) -> List[Item[Self]]:
...
+        converted into a :class:`View` or :class:`LayoutView` first.
+
+        If the message has any v2 component, then you must use
+        :class:`LayoutView` in order for them to be converted into
+        their respective items.
+
+        This method should be called on the respective class (or subclass), so
+        if you want to convert v2 items, you should call :meth:`LayoutView.from_message`,
+        or the same method from any subclass of it; and not :meth:`View.from_message`, or the
+        same method from any subclass of it.
```
Rapptz:
```suggestion
        If the message has any v2 components, then you must use
        :class:`LayoutView` in order for them to be converted into
        their respective items. :class:`View` does not support v2 components.
```

discord/ui/view.py
```diff
@@ -271,13 +455,22 @@ def children(self) -> List[Item[Self]]:
         return self._children.copy()
 
     @classmethod
-    def from_message(cls, message: Message, /, *, timeout: Optional[float] = 180.0) -> View:
+    def from_message(cls, message: Message, /, *, timeout: Optional[float] = 180.0) -> Any:
```
Rapptz: Surely this should return `Self`.

discord/ui/action_row.py
```diff
@@ -0,0 +1,583 @@
...
+    r"""Represents a UI action row.
+
+    This is a top-level layout component that can only be used on :class:`LayoutView`
+    and can contain :class:`Button`\s and :class:`Select`\s in it.
+
+    This can be inherited.
+
+    .. note::
+
+        Action rows can contain up to 5 components, which is, 5 buttons or 1 select.
```
Rapptz:
```suggestion
    Action rows can only have 5 children. This can be inherited.
```

discord/ui/action_row.py
```diff
@@ -0,0 +1,583 @@
...
+                if hasattr(member, '__discord_ui_model_type__'):
+                    children[name] = member
+
+        if len(children) > 5:
+            raise TypeError('ActionRow cannot have more than 5 children')
+
+        cls.__action_row_children_items__ = list(children.values())
+
+    def __repr__(self) -> str:
+        return f'{super().__repr__()[:-1]} children={len(self._children)}>'
```
Rapptz: This is a terrible place to be clever.

```suggestion
        return f'{self.__class__.__name__} children={len(self._children)}>'
```

discord/ui/item.py
```diff
@@ -70,6 +78,9 @@ def __init__(self):
         # actually affect the intended purpose of this check because from_component is
         # only called upon edit and we're mainly interested during initial creation time.
         self._provided_custom_id: bool = False
+        self._id: Optional[int] = None
+        self._max_row: int = 5 if not self._is_v2() else 40
```
Rapptz: This doesn't work as you expect.

`Button._is_v2()` returns `False` so this is 5 but you have this 40 limit documented as working with any v2 view.

I've already mentioned how I dislike the overloading of `row` for this purpose as well.

discord/ui/item.py
```diff
@@ -106,20 +122,47 @@ def row(self) -> Optional[int]:
...
+            can_run = await self._parent._run_checks(interaction)
+
+        return can_run
+
+    def _can_be_dynamic(self) -> bool:
+        # if an item can be dynamic then it must override this, this is mainly used
+        # by DynamicItem's so a user cannot set, for example, a Container with a dispatchable
+        # button as a dynamic item, and cause errors where Container can't be dispatched
+        # or lost interactions
+        return False
```
Rapptz: I don't really get the point of `_can_be_dynamic` and the comment does not quite make it clearer. Any item that is dispatchable can and should be possible as a dynamic item. Containers (sections, action rows, and containers) are not dispatchable, so it's not necessary to make a whole other boolean for this.

discord/ui/media_gallery.py
```diff
@@ -0,0 +1,263 @@
...
+        self._underlying = MediaGalleryComponent._raw_construct(
+            items=list(items),
+            id=id,
+        )
+
+        self.row = row
+        self.id = id
+
+    def __repr__(self) -> str:
+        return f'<{super().__repr__()[:-1]} items={len(self._underlying.items)}>'
```
Rapptz:
```suggestion
        return f'<{self.__class__.__name__} items={len(self._underlying.items)}>'
```

discord/ui/section.py
```diff
@@ -0,0 +1,261 @@
...
+                raise ValueError('maximum number of children exceeded')
+            self._children.extend(
+                [c if isinstance(c, Item) else TextDisplay(c) for c in children],
+            )
+        self.accessory: Item[V] = accessory
+        self.row = row
+        self.id = id
+
+    def __repr__(self) -> str:
+        return f'<{super().__repr__()[:-1]} children={len(self._children)}'
```
Rapptz:
```suggestion
        return f'<{self.__class__.__name__} children={len(self._children)}>'
```

discord/ui/container.py
```diff
@@ -0,0 +1,398 @@
...
+                self.add_item(child)
+
+        self.spoiler: bool = spoiler
+        self._colour = accent_colour if accent_colour is not None else accent_color
+
+        self.row = row
+        self.id = id
+
+    def __repr__(self) -> str:
+        return f'<{super().__repr__()[:-1]} children={len(self._children)}>'
```
Rapptz:
```suggestion
        return f'<{self.__class__.__name__} children={len(self._children)}>'
```

discord/ui/action_row.py
```diff
@@ -0,0 +1,583 @@
...
+
+    def _init_children(self) -> List[Item[Any]]:
+        children = []
+
+        for func in self.__action_row_children_items__:
+            item: Item = func.__discord_ui_model_type__(**func.__discord_ui_model_kwargs__)
+            item.callback = _ActionRowCallback(func, self, item)  # type: ignore
+            item._parent = getattr(func, '__discord_ui_parent__', self)
+            setattr(self, func.__name__, item)
+            self._weight += item.width
```
Rapptz:
```suggestion
```

discord/ui/action_row.py
```diff
@@ -0,0 +1,583 @@
...
+        self,
+        *children: Item[V],
+        row: Optional[int] = None,
+        id: Optional[int] = None,
+    ) -> None:
+        super().__init__()
+        self._weight: int = 0
+        self._children: List[Item[V]] = self._init_children()
+        self._children.extend(children)
+        self._weight += sum(i.width for i in children)
```
Rapptz:
```suggestion
        self._weight = sum(i.width for i in self._children)
```

discord/ui/action_row.py
```diff
@@ -0,0 +1,583 @@
...
+        Yields
+        ------
+        :class:`Item`
+            An item in the action row.
+        """
+
+        for child in self.children:
+            yield child
+
+    def add_item(self, item: Item[Any]) -> Self:
```
Rapptz, line 222: This method does not do weight checking.

discord/ui/view.py
```diff
@@ -536,21 +778,264 @@ async def wait(self) -> bool:
...
+
+                    member._rendered_row = member._row
+                    children[name] = member
+                elif hasattr(member, '__discord_ui_model_type__') and getattr(member, '__discord_ui_parent__', None):
+                    callback_children[name] = member
+
+        children.update(callback_children)
+        cls.__view_children_items__ = children
+
+    def _is_v2(self) -> bool:
```
Rapptz:
```suggestion
    def _is_layout(self) -> TypeGuard[LayoutView]:
```

discord/ui/view.py
```diff
@@ -536,21 +778,264 @@ async def wait(self) -> bool:
...
+
+        cls.__view_children_items__ = children
+
+    def __init__(self, *, timeout: Optional[float] = 180.0):
+        super().__init__(timeout=timeout)
+        self.__weights = _ViewWeights(self._children)
+
+    @property
+    def width(self):
+        return 5
```
Rapptz: Why is this here?

discord/ui/view.py
```diff
@@ -133,73 +335,68 @@ def clear(self) -> None:
...
         self._cache_key: Optional[int] = None
-        self.__cancel_callback: Optional[Callable[[View], None]] = None
+        self.__cancel_callback: Optional[Callable[[BaseView], None]] = None
         self.__timeout_expiry: Optional[float] = None
         self.__timeout_task: Optional[asyncio.Task[None]] = None
         self.__stopped: asyncio.Future[bool] = asyncio.get_running_loop().create_future()
+        self._total_children: int = sum(1 for _ in self.walk_children())
+
+    def _is_v2(self) -> bool:
+        return False
```
Rapptz:
```suggestion
    def _is_layout(self) -> TypeGuard[LayoutView]:
        return False
```

This change propagates throughout the entire codebase I think.

This removes the need for the `getattr(view, '__discord_ui_layout_view__', False)` hack in a lot of the checks that you're doing and should function as an `isinstance` check at the same time.

discord/ui/view.py
```diff
@@ -536,21 +778,264 @@ async def wait(self) -> bool:
...
+    @classmethod
+    def _to_minimal_cls(cls) -> Type[Union[View, LayoutView]]:
+        if issubclass(cls, View):
+            return View
+        elif issubclass(cls, LayoutView):
+            return LayoutView
+        raise RuntimeError
+
+    @classmethod
+    def from_dict(cls, data: List[ComponentPayload], *, timeout: Optional[float] = 180.0) -> Any:
```
Rapptz: Why does this method exist? The amount of code needed to support this addition in particular is mind-boggling and I don't know why.

- `component_data_to_item`
- `_to_minimal_cls`
- This method
- The `View` one

I don't provide public `from_dict` methods in the library unless there's very good reason to do so, and so far that's essentially only `Embed`. No other method has it public.

pythonmcpi: Out of curiosity, what was the reason for providing `from_dict` on `Embed`?

Rapptz: People would serialize embeds as JSON using `to_dict()` and expect a way to construct it again using `from_dict` which was fair since it's entirely stateless. Views do not have this luxury.

pythonmcpi: Wouldn't a non-interactive LayoutView also be stateless? The media gallery/section components do provide benefits over using Embed for purely informational content.

I also don't understand why `to_dict()` was provided on Embeds in the first place (is it just because it is stateless?).

Rapptz: Yes it's because it's purely stateless. LayoutViews don't have that guarantee because dispatchable items exist. A section is not purely stateless, as it can have a button. An argument can be made for media galleries to be more easily constructed but that's different from an entire view.

playfairs:
omg no conflicts?? does this mean we get cv2 :o

discord/ui/view.py
DA-344: This was a suggestion from the bikeshedding post, but thinking it twice, it seems unnecessary, I'll remove it.

discord/ui/view.py
DA-344: Yup, i've made the changes.

``__discord_ui_layout_view__`` existed mostly because i forgot i've added this method.

discord/ui/item.py
DA-344: technically it works as expected, because buttons can only be on action rows, or as a section accessory.

in the first case, their position can be from 1 to 5, and if they're a section accessory, it does not matter.

discord/ui/item.py
DA-344: yeah, i've added it because previously i had a weird checking with ``is_dispatchable`` on the components, but i've removed it

discord/webhook/sync.py
```diff
@@ -994,13 +1034,15 @@ def send(
                 When sending a Poll via webhook, you cannot manually end it.
 
             .. versionadded:: 2.4
-        view: :class:`~discord.ui.View`
-            The view to send with the message. This can only have URL buttons, which donnot
+        view: Union[:class:`~discord.ui.View`, :class:`~discord.ui.LayoutView`]
+            The view to send with the message. This can only have non-interactible items, which donnot
```
Sacul0457: Spelling error '`donnot`'

discord/webhook/sync.py
```diff
@@ -1177,6 +1247,13 @@ def edit_message(
             then all attachments are removed.
 
             .. versionadded:: 2.0
+        view: Optional[Union[:class:`~discord.ui.View`, :class:`~discord.ui.LayoutView`]]
+            The updated view to update this message with. This can only have non-interactible items, which donnot
```
Sacul0457: Same here, '`donnot`'

discord/types/components.py
```diff
@@ -24,24 +24,31 @@
...
 
 from .emoji import PartialEmoji
 from .channel import ChannelType
 
-ComponentType = Literal[1, 2, 3, 4]
+ComponentType = Literal[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 17]
 ButtonStyle = Literal[1, 2, 3, 4, 5, 6]
 TextStyle = Literal[1, 2]
 DefaultValueType = Literal['user', 'role', 'channel']
+SeparatorSize = Literal[1, 2]
```
Sacul0457: Might be wrong here, but shouldn't it be renamed to `SeparatorSpacing` to keep consistent?

discord/types/components.py
```diff
@@ -118,5 +125,78 @@ class SelectMenu(SelectComponent):
...
+    file: UnfurledMediaItem
+    spoiler: NotRequired[bool]
+    name: NotRequired[str]
+    size: NotRequired[int]
+
+
+class SeparatorComponent(ComponentBase):
+    type: Literal[14]
+    divider: NotRequired[bool]
+    spacing: NotRequired[SeparatorSize]
```
Sacul0457: Same here

discord/message.py
```diff
@@ -1391,10 +1401,19 @@ async def edit(
...
-        view: Optional[:class:`~discord.ui.View`]
+        view: Optional[Union[:class:`~discord.ui.View`, :class:`~discord.ui.LayoutView`]]
             The updated view to update this message with. If ``None`` is passed then
             the view is removed.
 
+            .. note::
+
+                If you want to update the message to have a :class:`~discord.ui.LayoutView`, you must
+                explicitly set to ``None`` or empty array, as required, the ``content``, ``embed``,
+                ``embeds``, and ``attachments`` parameters.
```
```suggestion
                If you want to update the message to have a :class:`~discord.ui.LayoutView`, you must
                explicitly set the ``content``, ``embed``, ``embeds``, and ``attachments`` parameters to
                ``None`` if the previous message had any.
```

Do the same wording change for the rest of em too. The original sounds like broken English.

discord/interactions.py
```diff
@@ -1458,7 +1600,7 @@ async def edit(
             embeds=embeds,
             embed=embed,
             attachments=attachments,
-            view=view,
+            view=view,  # type: ignore
```
Rapptz: Why

discord/ui/action_row.py
```diff
@@ -0,0 +1,562 @@
...
+                if hasattr(member, '__discord_ui_model_type__'):
+                    children[name] = member
+
+        if len(children) > 5:
+            raise TypeError('ActionRow cannot have more than 5 children')
+
+        cls.__action_row_children_items__ = list(children.values())
+
+    def __repr__(self) -> str:
+        return f'{self.__class__.__name__} children={len(self._children)}>'
```
Rapptz:
```suggestion
        return f'<{self.__class__.__name__} children={len(self._children)}>'
```

discord/ui/action_row.py
Rapptz: This method still does bad weight checking. Imagine: there's a button and you're adding a Select.

discord/ui/container.py
```diff
@@ -0,0 +1,371 @@
...
+        children = []
+        parents = {}
+
+        for name, raw in self.__container_children_items__.items():
+            if isinstance(raw, Item):
+                item = copy.deepcopy(raw)
+                item._parent = self
+                if getattr(item, '__discord_ui_section__', False) and item.accessory.is_dispatchable():  # type: ignore
+                    if item.accessory._provided_custom_id is False:  # type: ignore
+                        item.accessory.custom_id = os.urandom(16).hex()  # type: ignore
```
Rapptz: I do not understand why this is here still. The custom ID stuff is the responsibility of the class itself, not the outside.

discord/ui/view.py
```diff
@@ -638,8 +919,25 @@ async def schedule_dynamic_item_call(
...
+        try:
+            child_index = parent._children.index(base_item)  # type: ignore
+        except ValueError:
+            # handle cases in which the item is a section accessory
+            if getattr(base_item._parent, '__discord_ui_section__', False):
+                if (
+                    base_item._parent.accessory.type.value == component_type  # type: ignore
+                    and getattr(base_item._parent.accessory, 'custom_id', None) == custom_id  # type: ignore
+                ):
+                    base_item._parent.accessory = item  # type: ignore
```
Rapptz: Again, what is this doing here?

discord/ui/item.py
```diff
@@ -104,22 +119,59 @@ def row(self) -> Optional[int]:
...
+        if can_run and self._parent:
+            can_run = await self._parent._run_checks(interaction)
+
+        return can_run
+
+    def _update_view(self, view) -> None:
+        self._view = view
+
+    def _has_nested(self) -> bool:
+        return False
```
Rapptz: This should be called something else.

```suggestion
    def _has_children(self) -> bool:
        return False
```

Though, again, this seems like another weird way of doing duck typing. To get rid of the type ignores you can probably use `TypeGuard` again.

Something similar to:

```py
if TYPE_CHECKING:
    class ContainerItem(Generic[V]):
        def walk_children(self) -> Generator[Item[V], None, None]:
            ...
```

Then you can use `TypeGuard[ContainerItem[V]]` as the return annotation. This should work since `TypeGuard` (luckily) does not do strict narrowing.

discord/ui/item.py
```diff
@@ -104,22 +119,59 @@ def row(self) -> Optional[int]:
...
+        return self._id
+
+    @id.setter
+    def id(self, value: Optional[int]) -> None:
+        self._id = value
+
+    @property
+    def parent(self) -> Optional[Item[V]]:
+        """Optional[:class:`Item`]: This item's parent. If this item is a :class:`View` this will
+        always be ``None``.
```
Rapptz:
```suggestion
        """Optional[:class:`Item`]: This item's parent. Only components that can have children
        can be parents. Any item that has :class:`View` as a view will have this set to `None`
        since only :class:`LayoutView` component v2 items can contain "container" like items.
        
        .. versionadded:: 2.6
```

Rapptz (changes requested):

discord/ui/text_display.py
```diff
@@ -0,0 +1,89 @@
...
+
+V = TypeVar('V', bound='LayoutView', covariant=True)
+
+__all__ = ('TextDisplay',)
+
+
+class TextDisplay(Item[V]):
+    """Represents a UI text display.
+
+    This is a top-level layout component that can only be used on :class:`LayoutView`.
```
Soheab: Partially untrue when read in a certain way. This can also be added to a Section and maybe others in the future.

discord/ui/media_gallery.py
```diff
@@ -0,0 +1,252 @@
...
+            A :class:`.MediaGalleryItem` was not passed.
+        ValueError
+            Maximum number of items has been exceeded (10).
+        """
+
+        if len(self._underlying.items) >= 10:
+            raise ValueError('maximum number of items has been exceeded')
+
+        if not isinstance(item, MediaGalleryItem):
+            raise TypeError(f'expected MediaGalleryItem not {item.__class__.__name__}')
```
Soheab:
```suggestion
            raise TypeError(f'expected MediaGalleryItem, not {item.__class__.__name__!r}')
```
For consistency.

discord/ui/thumbnail.py
```diff
@@ -0,0 +1,132 @@
...
+        return self._media
+
+    @media.setter
+    def media(self, value: Union[str, UnfurledMediaItem]) -> None:
+        if isinstance(value, str):
+            self._media = UnfurledMediaItem(value)
+        elif isinstance(value, UnfurledMediaItem):
+            self._media = value
+        else:
+            raise TypeError(f'expected a str or UnfurledMediaItem, got {value.__class__.__name__!r}')
```
Soheab:
```suggestion
            raise TypeError(f'expected a str or UnfurledMediaItem, not {value.__class__.__name__!r}')
```
For consistency.

discord/ui/action_row.py
```diff
@@ -0,0 +1,562 @@
...
+        placeholder: Optional[str] = ...,
+        custom_id: str = ...,
+        min_values: int = ...,
+        max_values: int = ...,
+        disabled: bool = ...,
+        default_values: Sequence[ValidDefaultValues] = ...,
+    ) -> SelectCallbackDecorator[MentionableSelectT]:
+        ...
+
+    def select(
```
Soheab, line 472: Missing `id` parameter.

discord/ui/action_row.py
```diff
@@ -0,0 +1,562 @@
...
+
+        base = {
+            'type': self.type.value,
+            'components': components,
+        }
+        if self.id is not None:
+            base['id'] = self.id
+        return base
+
+    def button(
```
Soheab, line 328: Missing `id` parameter.

discord/ui/item.py
DA-344: I could do that, but ``TypeGuard`` enforces the function to have at least one user-provided parameter. So if I simply updated it to:
```python
def _has_children(self) -> TypeGuard[ContainerItem[V]]:
    return False
```
I would have to ``# type: ignore`` the function definition
<details>

```markdown
User-defined type guard functions and methods must have at least one input parameter    Pylance[reportGeneralTypeIssues](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportGeneralTypeIssues.md)
```

</details>

And I don't think doing something like ``def _has_children(self, _) -> TypeGuard[ContainerItem[V]]`` is something I should add.

~~This can be solved if I used ``TypeIs`` instead of ``TypeGuard``.~~ Still the same issue (partially)

discord/ui/view.py
DA-344: This is to handle the cases in which a DynamicItem is set as a Section's accessory. The current code looks up in the ``_children`` attribute (``child_index = parent._children.index(base_item)``), but the accessory is not in that list, so it raises the ``ValueError`` and here it checks whether the item is the section's accessory, and handles it.
Although this could be handled outside this ``except`` block..

discord/interactions.py
DA-344: Due to the overloads in ``edit_original_response``

discord/ui/item.py
Rapptz: Unfortunate limitation. It seems it works as I expect if it's in a `@staticmethod` and is called like `self.item._has_children(self.item)` but that does read a bit goofy.

discord/ui/view.py
Rapptz: This doesn't belong here either way.

Using the same principles as before this should be done in the class that is responsible for this behaviour in a consistent way. So this means using basic indirection. The behaviour we're trying to do is a basic swap, so just implement the swap where it needs to be handled. Detailing how this is implemented is actually a bit more complicated than I anticipated since it requires a newtype for the index. If it comes down to it I'll implement it myself later on the in the future.

discord/interactions.py
Rapptz: The type there should be fixed then

discord/ui/action_row.py
DA-344: fixed

discord/ui/action_row.py
```diff
@@ -0,0 +1,583 @@
...
+        return ComponentType.action_row
+
+    @property
+    def children(self) -> List[Item[V]]:
+        """List[:class:`Item`]: The list of children attached to this action row."""
+        return self._children.copy()
+
+    def walk_children(self) -> Generator[Item[V], Any, None]:
+        """An iterator that recursively walks through all the children of this view
+        and it's children, if applicable.
```
Soheab:
```suggestion
        """An iterator that recursively walks through all the children of this action row
        and it's children, if applicable.
```

discord/ui/action_row.py
```diff
@@ -0,0 +1,583 @@
...
+            async def click_me(self, interaction: discord.Interaction, button: discord.ui.Button):
+                await interaction.response.send_message('You clicked me!')
+
+        # or use it directly on LayoutView
+        class MyView(ui.LayoutView):
+            row = ui.ActionRow()
+            # or you can use your subclass:
+            # row = MyActionRow()
+
+            # you can create items with row.button and row.select
```
Soheab:
```suggestion
            # you can add items with row.button and row.select
```

discord/ui/action_row.py
```diff
@@ -0,0 +1,583 @@
...
+        ------
+        :class:`Item`
+            An item in the action row.
+        """
+
+        for child in self.children:
+            yield child
+
+    def add_item(self, item: Item[Any]) -> Self:
+        """Adds an item to this row.
```
Soheab:
```suggestion
        """Adds an item to this action row.
```

discord/ui/action_row.py
```diff
@@ -0,0 +1,583 @@
...
+
+        This function returns the class instance to allow for fluent-style
+        chaining.
+
+        Parameters
+        ----------
+        item: :class:`Item`
+            The item to add to the row.
```
Soheab:
```suggestion
            The item to add to the action row.
```

discord/ui/action_row.py
```diff
@@ -0,0 +1,583 @@
...
+        """Removes an item from the row.
+
+        This function returns the class instance to allow for fluent-style
+        chaining.
+
+        Parameters
+        ----------
+        item: :class:`Item`
+            The item to remove from the view.
+        """
```
Soheab:
```suggestion
        """Removes an item from the action row.

        This function returns the class instance to allow for fluent-style
        chaining.

        Parameters
        ----------
        item: :class:`Item`
            The item to remove from the action row.
        """
```

discord/ui/action_row.py
```diff
@@ -0,0 +1,583 @@
...
+
+        Returns
+        -------
+        Optional[:class:`Item`]
+            The item found, or ``None``.
+        """
+        return _utils_get(self.walk_children(), id=id)
+
+    def clear_items(self) -> Self:
+        """Removes all items from the row.
```
Soheab:
```suggestion
        """Removes all items from the action row.
```

discord/ui/action_row.py
```diff
@@ -0,0 +1,583 @@
...
+        style: ButtonStyle = ButtonStyle.secondary,
+        emoji: Optional[Union[str, Emoji, PartialEmoji]] = None,
+        id: Optional[int] = None,
+    ) -> Callable[[ItemCallbackType[Button[V]]], Button[V]]:
+        """A decorator that attaches a button to a component.
+
+        The function being decorated should have three parameters, ``self`` representing
+        the :class:`discord.ui.LayoutView`, the :class:`discord.Interaction` you receive and
+        the :class:`discord.ui.Button` being pressed.
+
```
Soheab:
```suggestion
        """A decorator that attaches a button to the action row.

        The function being decorated should have three parameters, ``self`` representing
        the :class:`discord.ui.ActionRow`, the :class:`discord.Interaction` you receive and
        the :class:`discord.ui.Button` being pressed.
```

discord/ui/action_row.py
```diff
@@ -0,0 +1,583 @@
...
+        max_values: int = 1,
+        disabled: bool = False,
+        default_values: Sequence[ValidDefaultValues] = MISSING,
+        id: Optional[int] = None,
+    ) -> SelectCallbackDecorator[BaseSelectT]:
+        """A decorator that attaches a select menu to a component.
+
+        The function being decorated should have three parameters, ``self`` representing
+        the :class:`discord.ui.LayoutView`, the :class:`discord.Interaction` you receive and
+        the chosen select class.
```
Soheab:
```suggestion
        """A decorator that attaches a select menu to the action row.

        The function being decorated should have three parameters, ``self`` representing
        the :class:`discord.ui.ActionRow`, the :class:`discord.Interaction` you receive and
        the chosen select class.
```

discord/ui/container.py
```diff
@@ -0,0 +1,365 @@
...
+    def remove_item(self, item: Item[Any]) -> Self:
+        """Removes an item from this container.
+
+        This function returns the class instance to allow for fluent-style
+        chaining.
+
+        Parameters
+        ----------
+        item: :class:`TextDisplay`
+            The item to remove from the section.
```
Soheab:
```suggestion
        item: :class:`Item`
            The item to remove from the container.
```
So we can only remove the TextDisplays from "the section"? :p

discord/ui/section.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def remove_item(self, item: Item[Any]) -> Self:
+        """Removes an item from this section.
+
+        This function returns the class instance to allow for fluent-style
+        chaining.
+
+        Parameters
+        ----------
+        item: :class:`TextDisplay`
```
Soheab:
```suggestion
        item: :class:`Item`
```
I get why you did it as Sections currently only support a TextDisplay, but I think we shouldn't do this for 1) future proofing as IIRC Discord noted that support for more items can be added at any time, 2) the type of the parameter is still Item, 3) consistency with add_item and such.

discord/ui/section.py
```diff
@@ -0,0 +1,248 @@
...
+    @property
+    def width(self):
+        return 5
+
+    def _is_v2(self) -> bool:
+        return True
+
+    def walk_children(self) -> Generator[Item[V], None, None]:
+        """An iterator that recursively walks through all the children of this section.
+        and it's children, if applicable.
```
Soheab:
```suggestion
        """An iterator that recursively walks through all the children of this section.
        and it's children, if applicable. This includes the `accessory`.
```
"and it's children" already suggests that it does this, but I know people will still be confused and ask about it. However, I'm unsure about the wording/position there.

Soheab (commented):
Small mistakes in documentation that I found.

Something that I noticed later but didn't include here for reasons, in all `walk_children` methods you used “**it's** children" instead of "**its** children", please correct that too.

discord/ui/item.py
```diff
@@ -36,13 +37,15 @@
...
     from ..enums import ComponentType
-    from .view import View
+    from .view import BaseView
     from ..components import Component
 
 I = TypeVar('I', bound='Item[Any]')
-V = TypeVar('V', bound='View', covariant=True)
-ItemCallbackType = Callable[[V, Interaction[Any], I], Coroutine[Any, Any, Any]]
+V = TypeVar('V', bound='BaseView', covariant=True)
+ItemCallbackType = Callable[[Any, Interaction[Any], I], Coroutine[Any, Any, Any]]
```
Rapptz: Why was this changed?

discord/ui/view.py
```diff
@@ -638,8 +918,25 @@ async def schedule_dynamic_item_call(
...
```
Rapptz: Why has this not been changed yet?

Rapptz (changes requested):

discord/ui/item.py
DA-344: So I the type checkers wouldn't complain if you create a callback item (button | select) outside of a view (ie container).

and the ``V`` change was so it was not needed to do ``Union[View, LayoutView[``

Rapptz: It should be a separate type then? `Any` doesn't help catch errors at all.

DA-344: yeah, makes sense, i'll change it

ConnerAdamsMaine:
I've started testing, just to grasp the fundamentals. How would you go about embeding the LayoutView in a discord.embeds.Embed() or has this not yet been implemented? I know it can be done in discord.js I just don't see how it would be done in discord.py?

I was thinking maybe using the example below, and will test with this but am unsure.
```py
class foo(Embed, LayoutView):
```

Rapptz:
@ConnerAdamsMaine Embeds and container v2 are two different things. The closest you have to an "embed" is `Container` which lets you set an accent colour. There is no way that discord.js supports whatever you're suggesting because they are fundamentally different things at an API layer.

ConnerAdamsMaine:
@Rapptz it's not an embed per-say, it was just the best way I could think of it. I realize now that it was an idiotic idea, I do apologize on that part. It's hard thinking with 48 hours of sleep deprivation. I have checked and it is the `Container` class that I was looking for!

PR #10170, opened 2025-04-24
Add ability to create a media-only forum channel

Soheab:
## Summary

This PR makes it possible to create a media-only forum channel by setting `media_only` to True in `Guild.create_forum()`.

Not tested as IIRC it requires some sort of subscription.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -1800,6 +1800,7 @@ class Guild(Hashable):
         category: Optional[CategoryChannel] = None,
         slowmode_delay: int = MISSING,
         nsfw: bool = MISSING,
+        media_only: bool = MISSING,
         overwrites: Mapping[Union[Role, Member, Object], PermissionOverwrite] = MISSING,
         reason: Optional[str] = None,
         default_auto_archive_duration: int = MISSING,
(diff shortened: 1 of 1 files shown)


discord/guild.py
```diff
@@ -1800,6 +1800,7 @@ async def create_forum(
         category: Optional[CategoryChannel] = None,
         slowmode_delay: int = MISSING,
         nsfw: bool = MISSING,
+        media_only: bool = MISSING,
```
dolfies, line 1803: media only isn't really a term used anywhere for this, wouldn't just `media` be better?

Soheab: I guess? I personally think it's ambiguous; `media=True` could also be interpreted as meaning media is disallowed by default, and this toggle enables it.

But maybe `media_only` isn't best either, since you can still create posts without an image in these channels.

dolfies: `media` would follow the same pattern as `news` in text channels (since in both cases they're different channel types internally)

Soheab: Fair point. I'm going to change it.

discord/guild.py
```diff
@@ -1862,12 +1863,17 @@ async def create_forum(
             .. versionadded:: 2.3
         default_layout: :class:`ForumLayoutType`
             The default layout for posts in this forum.
+            This cannot be set if ``media_only`` is set to ``True``.
```
Rapptz:
```suggestion
            This cannot be set if ``media`` is set to ``True``.
```

Rapptz (changes requested):
Run `black`.

Soheab:
> Run `black`.

Done.

PR #10173, opened 2025-04-28
Fixed to_dict() bug for user-defined embed classes implementing __slo…

cyberoby:
…ts__

## Summary

The to_dict() method wouldn't work if someone correctly subclasses discord.Embed and implements __slots__ in their custom class. This is because the original to_dict() method accessed self.__slots__, which would return only the subclass’s slots tuple. This hotfix changes self.__slots__ to Embed.__slots__ so that it works for any subclass, even if it implements its own __slots__.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/embeds.py
+++ b/discord/embeds.py
@@ -737,7 +737,7 @@ class Embed:
         # fmt: off
         result = {
             key[1:]: getattr(self, key)
-            for key in self.__slots__
+            for key in Embed.__slots__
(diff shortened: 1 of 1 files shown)


cyberoby:
Another way would be to declare a tuple of arguments to include in the embed dictionary key, instead of recycling __slots__...

Rapptz:
That fix would work too, but it's unnecessary for now. Thanks for the PR.

PR #10174, opened 2025-04-30
add `__repr__` method to various classes

owocado:
## Summary

This adds `__repr__` dunder method to following classes to help with bot development and debugging:
- `AppCommandContext`
- `AppInstallationType`
- `AppCommandPermissions`
- `GuildAppCommandPermissions`
- `InteractionCallbackResponse`
- `Parameter` from `ext.commands` module

Please let me know if this is ok as-is or should I split these 5 additions into individual PRs to preserve atomicity/git blame.

Moreover, the reason I used `self.__class__.__name__` in repr of `Parameter` is because if people are already subclassing that class in their projects for customization, not sure. Let me know if I should change it.

Bikeshedding post: https://discord.com/channels/336642139381301249/1355473537577652234

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/parameters.py
+++ b/discord/ext/commands/parameters.py
@@ -110,7 +110,7 @@ class Parameter(inspect.Parameter):
         self._displayed_name = displayed_name
 
     def __repr__(self) -> str:
-        return f'<{self.__class__.__name__} name={self._name!r} required={self.required}'
+        return f'<{self.__class__.__name__} name={self._name!r} required={self.required}>'
 
     def replace(
         self,


discord/ext/commands/parameters.py
```diff
@@ -109,6 +109,9 @@ def __init__(
         self._fallback = False
         self._displayed_name = displayed_name
 
+    def __repr__(self) -> str:
+        return f'<{self.__class__.__name__} name={self._name!r} required={self.required}'
```
Rapptz:
```suggestion
        return f'<{self.__class__.__name__} name={self._name!r} required={self.required}>'
```

PR #10189, opened 2025-05-16
Unpack usage where possible

Soheab:
## Summary

https://canary.discord.com/channels/336642139381301249/1371429669395759195

## Tracking

- [x] discord
  - [x] app_commands
    - [x] checks.py
    - [x] commands.py
  - [x] ext
    - [x] commands
        - [x] bot.py
        - [x] cog.py
        - [x] core.py
        - [x] help.py
        - [x] hybrid.py
  - [x] abc.py
  - [x] channel.py
  - [x] client.py
  - [x] flags.py
  - [x] permissions.py
  - [x] shard.py

## Checklist

- [ ] This PR fixes an issue.


discord/channel.py
```diff
@@ -120,6 +121,50 @@
 
     OverwriteKeyT = TypeVar('OverwriteKeyT', Role, BaseUser, Object, Union[Role, Member, Object])
 
+    class _CreateChannelWithCategory(TypedDict, total=False):
+        category: Optional[CategoryChannel]
+
+    class _CreateNewsChannel(TypedDict, total=False):
+        news: bool
+
```
Rapptz: These appear unused.

discord/client.py
```diff
@@ -272,7 +295,7 @@ class Client:
         The websocket gateway the client is currently connected to. Could be ``None``.
     """
 
-    def __init__(self, *, intents: Intents, **options: Any) -> None:
+    def __init__(self, *, intents: Intents, **options: Unpack[_ClientOptions]) -> None:
```
Rapptz, line 298: So, the reason why I didn't do this before for `Client` is because as part of the contract that you can inherit from `Client` this means that any extra args would cause type errors even though they might use the extra kwargs themselves.

I'm unsure where I really stand on this from a usability POV, and I don't know if it's that annoying.

discord/ext/commands/bot.py
```diff
@@ -1527,4 +1539,18 @@ class AutoShardedBot(BotBase, discord.AutoShardedClient):
...
+            self,
+            command_prefix: PrefixType[BotT],
+            *,
+            help_command: Optional[HelpCommand] = _default,
+            tree_cls: Type[app_commands.CommandTree[Any]] = app_commands.CommandTree,
+            description: Optional[str] = None,
+            allowed_contexts: app_commands.AppCommandContext = MISSING,
+            allowed_installs: app_commands.AppInstallationType = MISSING,
+            intents: discord.Intents,
+            **kwargs: Unpack[_AutoShardedBotOptions],
```
Rapptz, line 1554: Same nit as before.

discord/ext/commands/cog.py
```diff
@@ -169,7 +181,7 @@ async def bar(self, ctx):
     __cog_app_commands__: List[Union[app_commands.Group, app_commands.Command[Any, ..., Any]]]
     __cog_listeners__: List[Tuple[str, str]]
 
-    def __new__(cls, *args: Any, **kwargs: Any) -> CogMeta:
+    def __new__(cls, *args: Any, **kwargs: Unpack[_CogKwargs]) -> CogMeta:
```
Rapptz, line 184: Does this even work from a user's POV?

discord/ext/commands/core.py
```diff
@@ -368,9 +398,9 @@ class Command(_BaseCommand, Generic[CogT, P, T]):
 
         .. versionadded:: 2.0
     """
-    __original_kwargs__: Dict[str, Any]
+    __original_kwargs__: _CommandKwargs
 
-    def __new__(cls, *args: Any, **kwargs: Any) -> Self:
+    def __new__(cls, *args: Any, **kwargs: Unpack[_CommandKwargs]) -> Self:
```
Rapptz: This is internal there's no need to type it.

discord/flags.py
```diff
@@ -754,12 +790,12 @@ class Intents(BaseFlags):
 
     __slots__ = ()
 
-    def __init__(self, value: int = 0, **kwargs: bool) -> None:
+    def __init__(self, value: int = 0, **kwargs: Unpack[_IntentsFlagsKwargs]) -> None:
```
Rapptz, line 804: Another reason why I was not a fan of doing this is because now there are *more* places to edit when adding a single thing. This is especially egregious with permissions which need to have an added value both in its `TYPE_CHECKING` block and *now* in its `Unpack` dictionary.

I suppose it's unavoidable, but honestly it's just really annoying.

discord/shard.py
```diff
@@ -365,7 +371,7 @@ class AutoShardedClient(Client):
     if TYPE_CHECKING:
         _connection: AutoShardedConnectionState
 
-    def __init__(self, *args: Any, intents: Intents, **kwargs: Any) -> None:
+    def __init__(self, *args: Any, intents: Intents, **kwargs: Unpack[_AutoShardedClientOptions]) -> None:
```
Rapptz, line 374: Same nit as `Client` and `AutoShardedBot`.

discord/ext/commands/cog.py
Soheab: [Unfortunately not](https://discord.com/channels/336642139381301249/1371429669395759195/1388548860351352906). It does error out when the user has typed the correct name but the wrong type.

discord/client.py
Soheab: Unsure what you mean as extra kwargs are ignored anyways and you lose the typed kwargs when subclassing and overriding the `__init__` to handle the extra args anyways.
```python
    ...

# Argument missing for parameter "intents" (expected)
# No parameter named "hello" (expected but ignored anyways)
MyClient(hello="lol")


class ClientWithInit(discord.Client):
    # *args = tuple[Unknown, ...], **kwargs = dict[str, Unknown]
    def __init__(self, *args, **kwargs) -> None:

        # handling extra kwarg
        self.foo = kwargs.pop("foo")

ClientWithInit(foo="bar") # works fine
```
```suggestion
    def __init__(self, *, intents: Intents, **options: Unpack[_ClientOptions]) -> None:
```

discord/channel.py
Soheab: These were intended for future use, but now that I think about it, that seems a bit silly. I've removed them.

discord/flags.py
Soheab: Fair point, it's a bit annoying to update things in more than one place. But for users, it's kinda nice to have. Also, most people don’t know you can use kwargs to construct flags like this, so showing them helps with that too.

discord/client.py
Rapptz: The other issue is that it doesn't allow people who inherit to have the same ability, e.g. `ClientWithInit(intents=None)` passes despite it not meeting the type signature. I guess it's *fine* just unfortunate.

discord/permissions.py
```diff
@@ -135,18 +201,18 @@ class Permissions(BaseFlags):
...
         self.value = permissions
-        for key, value in kwargs.items():
+        for key, kwvalue in kwargs.items():
             try:
                 flag = self.VALID_FLAGS[key]
             except KeyError:
                 raise TypeError(f'{key!r} is not a valid permission name.') from None
             else:
-                self._set_flag(flag, value)
+                self._set_flag(flag, kwvalue)  # type: ignore
```
Rapptz: Why does this need a type ignore?

discord/client.py
Soheab: Yeah, there’s unfortunately no good way to make these public. The user can, if they want, [import them in a `TYPE_CHECKING` block](https://discord.com/channels/336642139381301249/1371429669395759195/1380278027036131411), but that’s probably not good advice to give.

Rapptz: That's not going to be supported -- and what I'm talking about is an actual supported way of doing this. I guess it has no impact on this PR. The most annoying thing right now is the duplication of all the flags/permissions.

discord/permissions.py
Soheab: all values are `object` when iterating over a TypedDict, so it throws the following:
```py
Argument of type "object" cannot be assigned to parameter "toggle" of type "bool" in function "_set_flag"
  "object" is not assignable to "bool"PylancereportArgumentType
```
AFAIK this is a not-fixable TypedDict annoyance.

Rapptz: That's pretty goofy. Add it in the comments.

Soheab: Done.

PR #10198, opened 2025-06-05
Fix project.license key in pyproject

Soheab:
## Summary

This PR correct the now deprecated `license = {}` format and classifier. 

More information: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license

## Checklist

- [ ] This PR fixes an issue.


Soheab:
This doesn't actually fix it. See failed checks.

Jackenmen:
FYI, I don't think it's possible to use the new fields with the current minimum Python set to 3.8 since the support for these fields in setuptools was added after they dropped support for 3.8.

PR #10208, opened 2025-06-15
app_commands: handle empty guild_ids list

laggron42:
## Summary

If an app command has `guild_ids=[]` or `@app_commands.guilds()` with no arguments, I'd expect the command to be detected as non-global command (despite not having guilds). However, this isn't the case and the command will be marked as global and synced everywhere.

When handling a command's `guild_ids`, most of the time an explicit check is made (`if guild_ids is not None`), which makes a difference between `None` and an empty list, but a few places missed this by using `guild_ids or ...`, where both `None` and `[]` will evaluate to false.

I fixed this behavior with an explicit check everywhere, making it possible to have guild commands that don't have a list of guilds yet (which can then be dynamically configured later on).

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -688,9 +688,7 @@ class Command(Generic[GroupT, P, T]):
         self.checks: List[Check] = getattr(callback, '__discord_app_commands_checks__', [])
         self._guild_ids: Optional[List[int]] = guild_ids
         if self._guild_ids is None:
-            self._guild_ids = getattr(
-                callback, '__discord_app_commands_default_guilds__', None
-            )
+            self._guild_ids = getattr(callback, '__discord_app_commands_default_guilds__', None)
         self.default_permissions: Optional[Permissions] = getattr(
             callback, '__discord_app_commands_default_permissions__', None
         )
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -686,9 +686,11 @@ class Command(Generic[GroupT, P, T]):
 
         self._params: Dict[str, CommandParameter] = _extract_parameters_from_callback(callback, callback.__globals__)
         self.checks: List[Check] = getattr(callback, '__discord_app_commands_checks__', [])
-        self._guild_ids: Optional[List[int]] = guild_ids or getattr(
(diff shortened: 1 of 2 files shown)


DA-344:
Which use case do you have to pass guilds with no arguments? Commands that are not global must be scoped to certain guilds, else, it's like they not exist

laggron42:
The use case is when the list of guilds is provided by a data source whose initial state is empty.

In my case, I have an `admin` command which should only exist in servers chosen by the bot owner. The bot owner can register new servers to sync using a text command like `!config syncadmin` (and manage existing ones afterwards).\
It works fine, except for the initial case where no servers have been registered, and discord.py will make it global instead. Since that command is in a cog, it will be picked up by the command tree and synced (unless I force remove it during cog loading, which is another solution but feels dirty).

I know not having guilds means the command essentially doesn't exist like you said, but it is something I am managing dynamically later on.\
Besides, most of the code already supports that logic[^1][^2][^3], it feels like that behavior was intended from the start, but was forgotten about with later commits (like 02fcc814e11f875a2f52574e23f718e1dba6fb2a).

[^1]: https://github.com/Rapptz/discord.py/blob/master/discord/app_commands/tree.py#L387
[^2]: https://github.com/Rapptz/discord.py/blob/master/discord/app_commands/tree.py#L368
[^3]: https://github.com/Rapptz/discord.py/blob/master/discord/app_commands/commands.py#L706

Rapptz:
The `guild_ids` parameter is actually explicitly undocumented because it's an internal shim to pass data around where it isn't generally possible. Overall the idea has always been to use the decorators to fill in that data and the dynamic aspect of it is mostly unsupported for the same reason, though you can just use the decorator later on with the instance.

laggron42:
I don't fully understand your message Danny 😅 

I understand why the function is not documented and why it's in this state right now, but I don't understand whether you want to change this (with more documentation and explicit behavior) or keep it as-is? Basically should close this PR or not

Rapptz:
Basically what I'm saying is that this is mostly intended and for dynamic commands there are other venues to do this without using this patch at all.

Either way this patch doesn't really cause any harm so it's fine to merge.

PR #10190, opened 2025-05-18
add support for clan/server tags

eepyfemboi:
## Summary

adds support for clan/server tags

## Checklist

- [x] If code changes were made then they have been tested.
    - [ ] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.
- [x] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

i only bothered to add this bcuz i needed to use it in one of my bots and i was surprised that it hasnt been added yet
this is a super hacky fix and im honestly surprised i got it to work, but it would probably be a good idea to clean it up a bit
why did i spend 3 hours on this what am i doing with my life

PR #10211, opened 2025-06-18
Add ability to use clan data for users

blord0:
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/asset.py
+++ b/discord/asset.py
@@ -346,6 +346,15 @@ class Asset(AssetMixin):
             animated=animated,
         )
 
+    @classmethod
+    def _from_clan(cls, state: _State, guild_id: int, icon_hash: str) -> Self:
+        return cls(
+            state,
+            url=f'{cls.BASE}/clan-badges/{guild_id}/{icon_hash}.png?size=16',
+            key=icon_hash,
+            animated=False,
+        )
+
     def __str__(self) -> str:
         return self._url
 
--- a/discord/user.py
+++ b/discord/user.py
@@ -32,6 +32,7 @@ from .colour import Colour
 from .enums import DefaultAvatar
 from .flags import PublicUserFlags
 from .utils import snowflake_time, _bytes_to_base64_data, MISSING, _get_as_snowflake
+from .clan import Clan
 
 if TYPE_CHECKING:
     from typing_extensions import Self
@@ -71,6 +72,7 @@ class BaseUser(_UserTag):
         '_public_flags',
         '_state',
         '_avatar_decoration_data',
+        '_clan'
     )
 
     if TYPE_CHECKING:
@@ -123,6 +125,7 @@ class BaseUser(_UserTag):
         self.bot = data.get('bot', False)
         self.system = data.get('system', False)
         self._avatar_decoration_data = data.get('avatar_decoration_data')
+        self._clan = data.get('clan')
 
     @classmethod
     def _copy(cls, user: Self) -> Self:
@@ -139,6 +142,7 @@ class BaseUser(_UserTag):
         self._state = user._state
         self._public_flags = user._public_flags
         self._avatar_decoration_data = user._avatar_decoration_data
+        self._clan = user._clan
 
         return self
 
@@ -304,6 +308,15 @@ class BaseUser(_UserTag):
         if self.global_name:
             return self.global_name
         return self.name
+    
+    @property
+    def clan(self) -> Optional[Clan]:
+        """:class:`Clan`: Returns the user's clan, if applicable.
+        
+        If the user has not set a clan, ``None`` is returned."""
+        if self._clan:
+            return self._clan
+        return None
 
     def mentioned_in(self, message: Message) -> bool:
         """Checks if the user is mentioned in the specified message.
--- a/discord/user.py
+++ b/discord/user.py
@@ -43,6 +43,7 @@ if TYPE_CHECKING:
     from .guild import Guild
     from .message import Message
     from .state import ConnectionState
+    from .types.clan import Clan as ClanPayload
     from .types.channel import DMChannel as DMChannelPayload
     from .types.user import PartialUser as PartialUserPayload, User as UserPayload, AvatarDecorationData
 
@@ -88,6 +89,7 @@ class BaseUser(_UserTag):
         _accent_colour: Optional[int]
         _public_flags: int
         _avatar_decoration_data: Optional[AvatarDecorationData]
+        _clan: Optional[ClanPayload]
 
     def __init__(self, *, state: ConnectionState, data: Union[UserPayload, PartialUserPayload]) -> None:
         self._state = state
@@ -125,7 +127,7 @@ class BaseUser(_UserTag):
         self.bot = data.get('bot', False)
         self.system = data.get('system', False)
         self._avatar_decoration_data = data.get('avatar_decoration_data')
-        self._clan = data.get('clan')
+        self._clan = data.get('clan', None)
 
     @classmethod
     def _copy(cls, user: Self) -> Self:
@@ -315,7 +317,7 @@ class BaseUser(_UserTag):
         
         If the user has not set a clan, ``None`` is returned."""
         if self._clan:
-            return self._clan
+            return Clan(state=self._state, data=self._clan)
         return None
 
     def mentioned_in(self, message: Message) -> bool:
--- a/discord/clan.py
+++ b/discord/clan.py
@@ -1,0 +1,74 @@
+"""
+The MIT License (MIT)
+
+Copyright (c) 2015-present Rapptz
+
+Permission is hereby granted, free of charge, to any person obtaining a
+copy of this software and associated documentation files (the "Software"),
+to deal in the Software without restriction, including without limitation
+the rights to use, copy, modify, merge, publish, distribute, sublicense,
+and/or sell copies of the Software, and to permit persons to whom the
+Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+"""
+
+from typing import TYPE_CHECKING
+
+from .asset import Asset
+from .utils import snowflake_time
+
+if TYPE_CHECKING:
+    from datetime import datetime
+    
+    from .types.clan import Clan as ClanPayload
+    from .state import ConnectionState
+
+
+class Clan:
+    __slots__ = (
+        'guild_id',
+        'identity_enabled',
+        'tag',
+        '_badge',
+        '_state'
+    )
+
+    if TYPE_CHECKING:
+        guild_id: int
+        identity_enabled: bool
+        tag: str
+        _badge: str
+        _state: ConnectionState
+    
+    def __init__(self, *, state: ConnectionState, data: ClanPayload) -> None:
+        self._state = state
+        self._update(data)
+
+    def _update(self, data: ClanPayload):
+        self.guild_id = data["identity_guild_id"]
+        self.identity_enabled = data['identity_enabled']
+        self.tag = data['tag']
+        self._badge = data['badge']
+    
+    @property
+    def badge(self) -> Asset:
+        """:class:`Asset`: Returns the clan's asset"""
+        return Asset._from_clan(self._state, self.guild_id, self._badge)
+    
+    @property
+    def created_at(self) -> datetime:
+        """:class:`datetime.datetime`: Returns the guilds's creation time in UTC.
+
+        This is when the guild was created.
+        """
+        return snowflake_time(self.guild_id)
--- a/discord/clan.py
+++ b/discord/clan.py
@@ -67,8 +67,8 @@ class Clan:
     
     @property
     def created_at(self) -> datetime:
-        """:class:`datetime.datetime`: Returns the guilds's creation time in UTC.
+        """:class:`datetime.datetime`: Returns the clan's guild creation time in UTC.
 
-        This is when the guild was created.
+        This is when the guild, of that clan tag, was created.
         """
         return snowflake_time(self.guild_id)
--- a/discord/clan.py
+++ b/discord/clan.py
@@ -23,14 +23,13 @@ DEALINGS IN THE SOFTWARE.
 """
 
 from typing import TYPE_CHECKING
+from datetime import datetime
 
 from .asset import Asset
 from .utils import snowflake_time
+from .types.clan import Clan as ClanPayload
 
 if TYPE_CHECKING:
-    from datetime import datetime
-    
-    from .types.clan import Clan as ClanPayload
     from .state import ConnectionState
 
 
@@ -50,7 +49,7 @@ class Clan:
         _badge: str
         _state: ConnectionState
     
-    def __init__(self, *, state: ConnectionState, data: ClanPayload) -> None:
+    def __init__(self, *, state, data: ClanPayload) -> None:
         self._state = state
         self._update(data)
 
@@ -72,3 +71,9 @@ class Clan:
         This is when the guild, of that clan tag, was created.
         """
         return snowflake_time(self.guild_id)
+    
+    def __repr__(self) -> str:
+        return (
+            f"<guild_id={self.guild_id} identity_enabled={self.identity_enabled} tag={self.tag}"
+            f" badge={self.badge}>"
+        )
--- a/discord/types/clan.py
+++ b/discord/types/clan.py
@@ -1,0 +1,31 @@
+"""
+The MIT License (MIT)
+
+Copyright (c) 2015-present Rapptz
+
+Permission is hereby granted, free of charge, to any person obtaining a
+copy of this software and associated documentation files (the "Software"),
+to deal in the Software without restriction, including without limitation
+the rights to use, copy, modify, merge, publish, distribute, sublicense,
+and/or sell copies of the Software, and to permit persons to whom the
+Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+"""
+
+from typing import TypedDict
+
+class Clan(TypedDict):
+    identity_guild_id: int
+    identity_enabled: bool
+    tag: str
+    badge: str


Soheab:
There is already a PR for this: #10190 
~~even though this implementation looks majorly better~~

discord/primary_guild.py
```diff
@@ -0,0 +1,77 @@
...
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+"""
+
+from typing import TYPE_CHECKING
+from datetime import datetime
+
+from .asset import Asset
+from .utils import snowflake_time
+from .types.primary_guild import PrimaryGuild as PrimaryGuildPayload
```
blord0: From looking at other code, `PrimaryGuildPayload` is meant to be imported within the `if TYPE_CHECKING` statement below. But whenever it is there, the error "NameError: name 'PrimaryGuildPayload' is not defined" is thrown. Not sure how to resolve that

DA-344: do the ``from __future__ import annotations`` thing on the top of the file, before other code, after the license string

discord/primary_guild.py
```diff
@@ -0,0 +1,77 @@
...
+    
+    def __init__(self, *, state, data: PrimaryGuildPayload) -> None:
+        self._state = state
+        self._update(data)
+
+    def _update(self, data: PrimaryGuildPayload):
+        self.guild_id = data["identity_guild_id"]
+        self.identity_enabled = data['identity_enabled']
+        self.tag = data['tag']
+        self._badge = data['badge']
```
DA-344: only ``identity_enabled`` is assured to be returned, other keys may be absent if the user has no tag

discord/primary_guild.py
```diff
@@ -0,0 +1,77 @@
...
+    
+    @property
+    def created_at(self) -> datetime:
+        """:class:`datetime.datetime`: Returns the primary guild's creation time in UTC.
+        """
+        return snowflake_time(self.guild_id)
+    
+    def __repr__(self) -> str:
+        return (
+            f"<guild_id={self.guild_id} identity_enabled={self.identity_enabled} tag={self.tag}"
```
DA-344:
```suggestion
            f'<PrimaryGuild guild_id={self.guild_id} identity_enabled={self.identity_enabled} tag={self.tag}'
```

discord/primary_guild.py
```diff
@@ -0,0 +1,77 @@
...
+            f" badge={self.badge}>"
```
DA-344: single quotes
```suggestion
            f' badge={self.badge}>'
```

discord/primary_guild.py
```diff
@@ -0,0 +1,77 @@
...
+    
+    @property
+    def badge(self) -> Asset:
+        """:class:`Asset`: Returns the primary guild's asset"""
+        return Asset._from_primary_guild(self._state, self.guild_id, self._badge)
+    
+    @property
+    def created_at(self) -> datetime:
+        """:class:`datetime.datetime`: Returns the primary guild's creation time in UTC.
+        """
```
DA-344:
```suggestion
        """:class:`datetime.datetime`: Returns the primary guild's creation time in UTC."""
```

discord/primary_guild.py
```diff
@@ -0,0 +1,77 @@
...
+        tag: str
+        _badge: str
+        _state: ConnectionState
+    
+    def __init__(self, *, state, data: PrimaryGuildPayload) -> None:
+        self._state = state
+        self._update(data)
+
+    def _update(self, data: PrimaryGuildPayload):
+        self.guild_id = data["identity_guild_id"]
```
DA-344:
```suggestion
        self.guild_id = data['identity_guild_id']
```

discord/user.py
```diff
@@ -304,6 +310,15 @@ def display_name(self) -> str:
         if self.global_name:
             return self.global_name
         return self.name
+    
+    @property
+    def primary_guild(self) -> Optional[PrimaryGuild]:
+        """:class:`PrimaryGuild`: Returns the user's primary guild, if applicable.
+        
+        \nIf the user has not set a primary guild, ``None`` is returned."""
```
DA-344: it is clear that if no primary guild is set, itll return None
```suggestion
        """:class:`PrimaryGuild`: Returns the user's primary guild, if applicable."""
```

DA-344 (changes requested):

discord/primary_guild.py
dolfies: none of them are optional actually (tho no harm in forward compatibility), but they are all nullable, including `identity_enabled`. if `identity_enabled` is null and a `identity_guild_id` is set, it means the guild has changed their tag/badge and the user has not readopted it yet, so might be worth adding a note/property for that?

discord/asset.py
```diff
@@ -346,6 +346,15 @@ def _from_user_banner(cls, state: _State, user_id: int, banner_hash: str) -> Sel
             animated=animated,
         )
 
+    @classmethod
+    def _from_primary_guild(cls, state: _State, guild_id: int, icon_hash: str) -> Self:
+        return cls(
+            state,
+            url=f'{cls.BASE}/clan-badges/{guild_id}/{icon_hash}.png?size=16',
```
dolfies:
```suggestion
            url=f'{cls.BASE}/clan-badges/{guild_id}/{icon_hash}.png?size=64',
```
16x16 is really tiny for a default xd

discord/primary_guild.py
```diff
@@ -0,0 +1,91 @@
...
+        _tag: str
+        _badge: str
+        _state: ConnectionState
+
+    def __init__(self, *, state, data: PrimaryGuildPayload) -> None:
+        self._state = state
+        self._update(data)
+
+    def _update(self, data: PrimaryGuildPayload):
+        self._guild_id = data.get('identity_guild_id', None)
```
dolfies:
```suggestion
        self.id = _get_as_snowflake(data, 'identity_guild_id')
```
i feel like `user.primary_guild.guild_id` is a little redundant, and not much point to the property (also make sure to add the utils import :p)

discord/primary_guild.py
```diff
@@ -0,0 +1,91 @@
...
+        self.identity_enabled = data['identity_enabled']
+        self._tag = data.get('tag', None)
```
dolfies: does this need to be a property too

discord/primary_guild.py
```diff
@@ -0,0 +1,91 @@
...
+from .asset import Asset
+from .utils import snowflake_time
+
+if TYPE_CHECKING:
+    from .state import ConnectionState
+    from .types.primary_guild import PrimaryGuild as PrimaryGuildPayload
+
+
+class PrimaryGuild:
+    r"""Represents the primary guild (formally known as a clan) of a :class:`User`"""
```
dolfies: missing attributes directives, also why is this a raw string lol

dolfies (changes requested):

discord/primary_guild.py
blord0: Which part do you mean? I just filled in all the attributes you get in the payload

`{"identity_guild_id":"1234567890","identity_enabled":true,"tag":"TEST","badge":"e7334d054f7feede6f64be969f9b91d0"}`
`{"identity_guild_id":null,"identity_enabled":false,"tag":null,"badge":null}` 
Those are 2 example payloads

discord/primary_guild.py
blord0: Raw string was trying to figure out how to get the docs to generate correctly, and that was a failed attempt to get it to work (solution was the import in `__init__.py`)

discord/primary_guild.py
dolfies: why assign it to a private attribute and expose it via a property

discord/primary_guild.py
blord0: also wdym by "attributes directives" ?

discord/primary_guild.py
blord0: I just copied the style that other components used when the property was optional, like this from user.py
image

but can see now that they did that as they call a function, not just return the same value

discord/primary_guild.py
DA-344: The
```rst
Attributes
----------
a: :class:`something`
    The attr doc
```

Soheab:
```py
"""
Attributes
----------
id: :class:`int`
  The guild's ID.
...
"""
```
That in the class docstring

discord/primary_guild.py
```diff
@@ -0,0 +1,75 @@
...
+    from .state import ConnectionState
+    from .types.primary_guild import PrimaryGuild as PrimaryGuildPayload
+
+
+class PrimaryGuild:
+    """Represents the primary guild (formally known as a clan) of a :class:`User`"""
+
+    __slots__ = ('id', 'identity_enabled', 'tag', '_badge', '_state')
+
+    if TYPE_CHECKING:
```
dolfies: what's the point of the type checking block here? just have the types inline (also `identity_enabled` and `_badge` are both optional)

discord/primary_guild.py
```diff
@@ -0,0 +1,75 @@
...
+from .asset import Asset
+from .utils import snowflake_time, _get_as_snowflake
+
+if TYPE_CHECKING:
+    from .state import ConnectionState
+    from .types.primary_guild import PrimaryGuild as PrimaryGuildPayload
+
+
+class PrimaryGuild:
+    """Represents the primary guild (formally known as a clan) of a :class:`User`"""
```
dolfies:
```suggestion
    """Represents the primary guild identity of a :class:`User`
    
    Attributes
    -----------
    id: Optional[:class:`int`]
        The ID of the user's primary guild, if any.
    tag: Optional[:class:`str`]
        The guild's tag.
    identity_enabled: Optional[:class:`bool`]
        Whether the user has their primary guild publicly displayed. If ``None``, the user has a public guild but has not reaffirmed the guild identity after a change
        
        .. note::

            Users can have their primary guild publicly displayed while still having an :attr:`id` of ``None``. Be careful when checking this attribute!
    """
```

discord/primary_guild.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def _update(self, data: PrimaryGuildPayload):
+        self.id = _get_as_snowflake(data, 'identity_guild_id')
+        self.identity_enabled = data['identity_enabled']
+        self.tag = data.get('tag', None)
+        self._badge = data.get('badge')
+
+    @property
+    def badge(self) -> Optional[Asset]:
+        """:class:`Asset`: Returns the primary guild's asset"""
```
dolfies:
```suggestion
        """Optional[:class:`Asset`]: Returns the primary guild's asset"""
```

discord/primary_guild.py
```diff
@@ -0,0 +1,75 @@
...
+        if self._badge and self.id:
+            return Asset._from_primary_guild(self._state, self.id, self._badge)
+        return None
+
+    @property
+    def created_at(self) -> Optional[datetime]:
+        """:class:`datetime.datetime`: Returns the primary guild's creation time in UTC."""
```
dolfies:
```suggestion
        """Optional[:class:`datetime.datetime`]: Returns the primary guild's creation time in UTC."""
```

discord/user.py
```diff
@@ -305,6 +311,13 @@ def display_name(self) -> str:
             return self.global_name
         return self.name
 
+    @property
+    def primary_guild(self) -> Optional[PrimaryGuild]:
```
dolfies: For most users, this field will be present even if they don't have a `primary_guild` set (the `identity_enabled` value will just be false or null and there won't be an `identity_guild_id`). However, with some states, the field is null entirely. This is not a great API. Wondering if it's worth it to have a `PrimaryGuild._default()` so you don't have to null check for the `primary_guild` property.

discord/types/primary_guild.py
```diff
@@ -0,0 +1,32 @@
...
+"""
+
+from typing import TypedDict
+
+
+class PrimaryGuild(TypedDict):
+    identity_guild_id: int
+    identity_enabled: bool
+    tag: str
+    badge: str
```
dolfies:
```suggestion
    identity_guild_id: Optional[int]
    identity_enabled: Optional[bool]
    tag: Optional[str]
    badge: Optional[str]
```

dolfies (changes requested):

discord/primary_guild.py
blord0: type checking part was just a structure from users.py, so thought i should just copy it over
image

discord/user.py
blord0: Would my new check for `self._primary_guild.get("identity_enabled")` cover that, or should i add more to it?

dolfies: No that's not what I meant. I mean something like:
```py
if self._primary_guild:
    return PrimaryGuild(state=self._state, data=self._primary_guild)
return PrimaryGuild._default(state=self._state)
```
Where `_default()` will just initialize an object with `identity_enabled` set to false and the rest of the fields set to null.
Just my opinion though

blord0: Yea think that would be better, makes it easier to figure out if the user has a primary guild

discord/primary_guild.py
```diff
@@ -43,7 +44,9 @@ class PrimaryGuild:
     id: Optional[:class:`int`]
         The ID of the user's primary guild, if any.
     tag: Optional[:class:`str`]
-        The guild's tag.
+        The primary guild's tag.
+    badge: Optional[:class:`Asset`]
+        The primary guild's asset
```
dolfies: ~~properties are not documented as attributes~~ outdated sorry

dolfies (commented):
can you also update `Member._update_inner_user()` so that user updates fire properly for primary guild changes

discord/primary_guild.py
```diff
@@ -0,0 +1,96 @@
...
+    """Represents the primary guild identity of a :class:`User`
+
+    Attributes
+    -----------
+    id: Optional[:class:`int`]
+        The ID of the user's primary guild, if any.
+    tag: Optional[:class:`str`]
+        The primary guild's tag.
+    identity_enabled: Optional[:class:`bool`]
+        Whether the user has their primary guild publicly displayed. If ``None``, the user has a public guild but has not reaffirmed the guild identity after a change
```
blord0, line 51: btw where did you find this note part out? can't find any offical docs so i've just been basing stuff on what discord's api returns

blord0: Also thinking is it worth changing that `None` for `identity_enabled` to be a `False` and return the `._default()` instead?
Not too sure if you can have `identity_enabled` be `None` while `id` and `tag` are not also `None`

dolfies: Yes you can have a null `identity_enabled` with a non-null `id`

discord/primary_guild.py
```diff
@@ -0,0 +1,96 @@
...
+        """Optional[:class:`datetime.datetime`]: Returns the primary guild's creation time in UTC."""
+        if self.id:
+            return snowflake_time(self.id)
+        return None
+
+    @classmethod
+    def _default(cls, state: ConnectionState) -> Self:
+        """Creates a blank :class:`PrimaryGuild`"""
+        payload: PrimaryGuildPayload = {"identity_guild_id": None, "identity_enabled": False, "tag": None, "badge": None}
+        return cls(state=state, data=payload)
```
DA-344: Don't document methods that are not exposed to users
```suggestion
    @classmethod
    def _default(cls, state: ConnectionState) -> Self:
        payload: PrimaryGuildPayload = {"identity_guild_id": None, "identity_enabled": False, "tag": None, "badge": None}
        return cls(state=state, data=payload)
```

DA-344 (changes requested):
Looks nice, but still needs to wait for discord to document this

discord/primary_guild.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def __repr__(self) -> str:
+        return f'<PrimaryGuild id={self.id} identity_enabled={self.identity_enabled} tag={self.tag}' f' badge={self.badge}>'
```
dolfies:
```suggestion
        return f'<PrimaryGuild id={self.id} identity_enabled={self.identity_enabled} tag={self.tag!r}>'
```

discord/types/primary_guild.py
```diff
@@ -0,0 +1,32 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+"""
+
+from typing import TypedDict, Optional
+
+
+class PrimaryGuild(TypedDict):
```
dolfies: nit but idk if an entire new file is needed for this type

DA-344: This should be under the User file(s) imo

blord0: Yea tbf it is really small compared to other .types

blord0:
> Looks nice, but still needs to wait for discord to document this

Any idea how often discord updates their documentation? Latest pull in discord's api docs for tags was opened 2nd May 2024 and has not been updated in the past 3 weeks https://github.com/discord/discord-api-docs/pull/6836

blord0:
Also the api does seem to be stable.  From https://github.com/discord/discord-api-docs/pull/6836#issuecomment-2119306871 (19th May 2024) you can see that they still use the same fields of `tag`, `identity_guild_id`, `identity_enabled` and `badge`. Only part that has changed is it is now under the `primary_guild` section, not `clan`

DA-344:
You'll need to wait for Discord to just merge the documentation of it, and although it looks stable doesn't mean it _will always be_, rn is not documented so they have an excuse to (although even documented they do) break the field at any point.

blord0:
> You'll need to wait for Discord to just merge the documentation of it, and although it looks stable doesn't mean it _will always be_, rn is not documented so they have an excuse to (although even documented they do) break the field at any point.

Yea I get that, but tags have been rolled out to all servers now. The one change discord did was the field name from `clan` -> `primary_guild` due to clans being deprecated (I would argue that this is an even stronger indication that they won't change as the fields of `tag`, `identity_guild_id`, `identity_enabled` and `badge` are all the same even after a renaming, as that would have been the perfect point to re-do anything they wanted to do)

+In the unlikely event they do, I would be more than happy to fix/patch/whatever the code

DA-344:
Still, docs must be merged first

Rapptz:
Generally speaking I don't merge features that are undocumented because we've been bitten by it in the past (most recently, voice channel statuses) and it's not worth the upkeep.

discord/primary_guild.py
```diff
@@ -0,0 +1,95 @@
...
+        .. note::
+
+            Users can have their primary guild publicly displayed while still having an :attr:`id` of ``None``. Be careful when checking this attribute!
+    """
+
+    __slots__ = ('id', 'identity_enabled', 'tag', '_badge', '_state')
+
+    if TYPE_CHECKING:
+        id: Optional[int]
+        identity_enabled: Optional[bool]
```
DA-344: This is silly, None is defaulted to False.
```suggestion
        identity_enabled: bool
```

discord/primary_guild.py
```diff
@@ -0,0 +1,95 @@
...
+        _badge: Optional[str]
+        _state: ConnectionState
+
+    def __init__(self, *, state, data: PrimaryGuildPayload) -> None:
+        self._state = state
+        self._update(data)
+
+    def _update(self, data: PrimaryGuildPayload):
+        self.id = _get_as_snowflake(data, 'identity_guild_id')
+        self.identity_enabled = data['identity_enabled']
```
DA-344, line 66: You should default to False instead of None

discord/types/user.py
```diff
@@ -54,3 +54,10 @@ class User(PartialUser, total=False):
     flags: int
     premium_type: PremiumType
     public_flags: int
+
+
+class PrimaryGuild(TypedDict):
+    identity_guild_id: Optional[int]
+    identity_enabled: Optional[bool]
```
DA-344, line 61:
```suggestion
    identity_enabled: bool
```

discord/primary_guild.py
```diff
@@ -0,0 +1,95 @@
...
+    """Represents the primary guild identity of a :class:`User`
+
+    Attributes
+    -----------
+    id: Optional[:class:`int`]
+        The ID of the user's primary guild, if any.
+    tag: Optional[:class:`str`]
+        The primary guild's tag.
+    identity_enabled: Optional[:class:`bool`]
+        Whether the user has their primary guild publicly displayed. If ``None``, the user has a public guild but has not reaffirmed the guild identity after a change
```
DA-344, line 51: Worth mentioning that if identity_enabled is False but id or tag aren't, the guild changed tags and the user has not opted in to the new one

DA-344 (changes requested):

discord/primary_guild.py
dolfies: This is only when it's None.

discord/primary_guild.py
dolfies: None and False have different meanings.

DA-344: Then feel free to ignore this lol

Soheab:
The API docs have been merged per https://github.com/discord/discord-api-docs/commit/a1bac5f61574df372c44d73624e725c2c9ef8bc1 !

discord/asset.py
```diff
@@ -346,6 +346,15 @@ def _from_user_banner(cls, state: _State, user_id: int, banner_hash: str) -> Sel
             animated=animated,
         )
 
+    @classmethod
+    def _from_primary_guild(cls, state: _State, guild_id: int, icon_hash: str) -> Self:
+        return cls(
+            state,
+            url=f'{cls.BASE}/clan-badges/{guild_id}/{icon_hash}.png?size=64',
```
DA-344:
```suggestion
            url=f'{cls.BASE}/guild-tag-badges/{guild_id}/{icon_hash}.png?size=64',
```

DA-344 (changes requested):
The route has changed

blord0:
Just double checked the docs and everything seems good to me
https://discord.com/developers/docs/resources/user#user-object-user-primary-guild

discord/user.py
```diff
@@ -305,6 +315,13 @@ def display_name(self) -> str:
             return self.global_name
         return self.name
 
+    @property
+    def primary_guild(self) -> PrimaryGuild:
+        """:class:`PrimaryGuild`: Returns the user's primary guild."""
```
Soheab: Missing version string.
```rst
.. versionadded:: 2.6
```

discord/primary_guild.py
```diff
@@ -0,0 +1,95 @@
...
+
+if TYPE_CHECKING:
+    from .state import ConnectionState
+    from .types.user import PrimaryGuild as PrimaryGuildPayload
+    from typing_extensions import Self
+
+
+class PrimaryGuild:
+    """Represents the primary guild identity of a :class:`User`
+
```
Soheab, line 41: Missing version string.
```rst
.. versionadded:: 2.6
```

discord/primary_guild.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+
+    __slots__ = ('id', 'identity_enabled', 'tag', '_badge', '_state')
+
+    if TYPE_CHECKING:
+        id: Optional[int]
+        identity_enabled: Optional[bool]
+        tag: Optional[str]
+        _badge: Optional[str]
+        _state: ConnectionState
```
Soheab:
```suggestion
```
This isn't really doing anything.

discord/primary_guild.py
```diff
@@ -0,0 +1,95 @@
...
+    def created_at(self) -> Optional[datetime]:
+        """Optional[:class:`datetime.datetime`]: Returns the primary guild's creation time in UTC."""
+        if self.id:
+            return snowflake_time(self.id)
+        return None
+
+    @classmethod
+    def _default(cls, state: ConnectionState) -> Self:
+        payload: PrimaryGuildPayload = {"identity_guild_id": None, "identity_enabled": False, "tag": None, "badge": None}
+        return cls(state=state, data=payload)
```
Soheab: You don't really need to define each key here; an empty dict would work too since the `_update` method makes everything default to None regardless.

I guess it's better to be explicit, but I’m not sure.

Soheab (changes requested):
LGMT overall.

discord/primary_guild.py
```diff
@@ -0,0 +1,95 @@
...
+    Attributes
+    -----------
+    id: Optional[:class:`int`]
+        The ID of the user's primary guild, if any.
+    tag: Optional[:class:`str`]
+        The primary guild's tag.
+    identity_enabled: Optional[:class:`bool`]
+        Whether the user has their primary guild publicly displayed. If ``None``, the user has a public guild but has not reaffirmed the guild identity after a change
+
+        .. note::
```
Rapptz:
```suggestion
        .. warning::
```

discord/primary_guild.py
Rapptz: This was done in `user.py` because of a weird inheritance quirk if I recall, it's usually unnecessary so it should be removed here and just be explicit in the definitions within the class body.

discord/primary_guild.py
```diff
@@ -0,0 +1,95 @@
...
+    def badge(self) -> Optional[Asset]:
+        """Optional[:class:`Asset`]: Returns the primary guild's asset"""
+        if self._badge and self.id:
+            return Asset._from_primary_guild(self._state, self.id, self._badge)
+        return None
+
+    @property
+    def created_at(self) -> Optional[datetime]:
+        """Optional[:class:`datetime.datetime`]: Returns the primary guild's creation time in UTC."""
+        if self.id:
```
Rapptz:
```suggestion
        if self.id is not None:
```

discord/primary_guild.py
Rapptz: I agree here.

discord/user.py
```diff
@@ -305,6 +315,13 @@ def display_name(self) -> str:
...
+        if self._primary_guild:
+            return PrimaryGuild(state=self._state, data=self._primary_guild)
+        return PrimaryGuild._default(self._state)
```
Rapptz, line 325: I don't get why this isn't just `Optional[PrimaryGuild]` instead of the fake default type.

Soheab: Its to be consistent with the API: https://github.com/Rapptz/discord.py/pull/10211#discussion_r2155416244

~~After removing my tag, the api still returned the following:~~
~~```py
{'identity_guild_id': None, 'identity_enabled': False, 'tag': None, 'badge': None}```~~ Invalid point, this wouldn't trigger the default.

discord/primary_guild.py
blord0: Changed it to set only `identity_enabled` to `False`, as `None` has a specific meaning

blord0:
Sorry for delays, was on holiday

PR #10213, opened 2025-06-20
Fix rate limiter window reset and usage timing in GatewayRatelimiter

shashankpandey04:
## Summary

This pull request improves the internal logic of the `GatewayRatelimiter` to ensure more accurate and consistent rate-limiting behavior.
**Key Fixes:**
- The rate-limit window (self.window) and token count (self.remaining) now reset together, preventing desync issues.
- The rate-limit window now starts when the first token is consumed, ensuring correct delay calculations.

**Why it matters:**
Previously, if the rate-limit window was not reset properly while `remaining` was, the calculated delay could be incorrectly zero, causing shards to reconnect instantly without actually waiting for the rate limit to reset. This could result in:
This could result in:
- Unintentional rate-limit violations
- Gateway errors or Reconnects
- Unstable shard behavior
This fix ensures the shards respect the intended cooldown window, improving overall stability and compliance with Discord’s Gateway rate limits.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -105,14 +105,15 @@ class GatewayRatelimiter:
 
         if current > self.window + self.per:
             self.remaining = self.max
-
-        if self.remaining == self.max:
             self.window = current
 
         if self.remaining == 0:
-            return self.per - (current - self.window)
+            return max(0.0, self.per - (current - self.window))
 
         self.remaining -= 1
+        
+        if self.remaining == self.max - 1:
+            self.window = current
         return 0.0
 
     async def block(self) -> None:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ Documentation = "https://discordpy.readthedocs.io/en/latest/"
 dependencies = { file = "requirements.txt" }
 
 [project.optional-dependencies]
-voice = ["PyNaCl>=1.3.0,<1.6"]
+voice = ["PyNaCl>=1.5.0,<1.6"]
 docs = [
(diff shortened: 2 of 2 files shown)


shashankpandey04:
For those who didn't understand the para I have a shorter summary for you
The rate limiter was resetting the remaining requests, but not updating the timer (`window`) properly. So shards could think the rate limit expired and reconnect instantly, even though they were still supposed to wait. 🕺
When the window expires (`current > window + per`), only `remaining` was reset. But `self.window` was only reset if `remaining == max`, which might not happen immediately...especially under race conditions or rapid reconnect attempts (which is exactly what's happening in my case).
For example, if Bot X has 30 shards and they all get rate-limited, the bot starts retrying to connect those shards without waiting, causing an infinite reconnect cycle...until the rate limit is finally lifted.

Rapptz:
I don't believe this fix to be correct. Was this written with AI or something?

Each shard gets its own gateway implementation which has its own rate limit state. Not to mention that the only way to trigger a rate limit is by *sending* a gateway message. These are currently restricted to essentially requesting chunks or doing a presence update request. Its relevance when it comes to reconnecting is unclear. What exactly is the issue you're facing?

shashankpandey04:
Closing this PR — after further testing, the rate limiting turned out to be caused by API-heavy startup tasks on our end, not an issue with the library itself.

And just to clarify — my change wasn’t AI-written. I debugged this issue manually and edited what I thought matched the symptoms we were facing. Appreciate the feedback either way.

PR #10216, opened 2025-06-25
Add ability to sync single app commands

Soheab:
## Summary

This PR makes it possible to sync single Command/AppCommand instances by adding the following methods:

To AppCommand: `sync()`
To CommandTree (because Command n co have no state for obvious reasons): `.sync_command(<command>, guild=... | None)`

This PR also fixes a little bug where upserting an AppCommand could raise a BadRequest when contexts or integration_types were an empty list and `.to_dict()` still included it.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/models.py
+++ b/discord/app_commands/models.py
@@ -246,7 +246,7 @@ class AppCommand(Hashable):
         self.description_localizations: Dict[Locale, str] = _to_locale_dict(data.get('description_localizations') or {})
 
     def to_dict(self) -> ApplicationCommandPayload:
-        res = {
+        return {
             'id': self.id,
             'type': self.type.value,
             'application_id': self.application_id,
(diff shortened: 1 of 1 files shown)


Rapptz:
This endpoint being exposed is pretty dangerous because it's not the proper way to do it leading to potential misunderstanding where you loop through commands and call sync on each. Plus there's no real reason to only sync one command when the bulk one does the same exact thing but better.

Soheab:
> This endpoint being exposed is pretty dangerous because it's not the proper way to do it leading to potential misunderstanding where you loop through commands and call sync on each. Plus there's no real reason to only sync one command when the bulk one does the same exact thing but better. 

This is mainly intended for API completion, for those who want to use it over manually calling the HTTP methods. But I get your point about this being risky. I've added a note saying that this isn't the recommended way to handle syncing for that reason, and honestly, there isn't much more we can do to prevent people from calling this when not necessary, and I don't think people would discover this method any more easily than just using sync(), unless a misleading tutorial points them this way.

There's also an additional benefit here, depending on how you look at it, it allows for syncing a command to both a guild and global scope, and discord.py will still handle this correctly because of its fallback to global behaviour.

Rapptz:
Once again, I am unsure what the benefit is over using the bulk endpoint. API completion isn't really an argument here since it's just a flat out inferior route.

PR #10217, opened 2025-06-26
Update http.py

Nairdaa:
fix grammar logic behind invalid token

## Checklist

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -833,7 +833,7 @@ class HTTPClient:
         except HTTPException as exc:
             self.token = old_token
             if exc.status == 401:
-                raise LoginFailure('Improper token has been passed.') from exc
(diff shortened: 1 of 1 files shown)


Rapptz:
Grammar is fine, but thanks.

Nairdaa:
No, it's not, there is a logical mistake, but thanks.

Rapptz:
It's not a logical mistake.

In programming "pass" is used because you "pass" arguments to a function so that's why this verb is used over "provided". The reason why the sentence does not start with "An" is because typical error messages do not start with "A" or "An" since they are meant to be displayed in a somewhat authoritative manner.

Nairdaa:
In programming indeed, but imagine imagining in 2025 that not all the people are programmers. That debug message is blatant wrong and you just gotta deal with it. But thanks btw.

Issue #9589, opened 2023-09-30
Add unicode language identifiers to Locale


Luc1412:
### Summary

Add unicode language identifiers to Locale

### What is the feature request for?

The core library

### The Problem

For some applications the language ID is required e.g. `de` -> `de-DE`.

### The Ideal Solution

Add an attribute to `Locale` which probably accesses a hardcoded mapping.

### Additional Context

Red Discord Bot Server convo with Danny: https://ptb.discord.com/channels/133049272517001216/160386989819035648/1150474462777589820

PR #10222, opened 2025-07-03
feat(locale): add BCP47 unicode tags for Locale enum

exorcize:
## Summary
Adds support for BCP 47-style unicode language identifiers to the Locale enum.

This adds a new property, `language_code`, that maps the Discord locale value to its standardized `language-REGION` format (e.g. `de` → `de-DE`, `pt-BR` → `pt-BR`, `es-419` → `es-419`).
This can be useful for integrations that require full locale identifiers (e.g. HTTP headers, system locale matching, external APIs).

## Changes
- Adds `_UNICODE_LANG_MAP` for locale → unicode tag mapping.
- Adds a `language_code` property to `Locale`.
- The property returns the mapped Unicode tag or falls back to the original value.
- Mapping is based on CLDR identifiers matching Discord's locale list.

## Example
```py
Locale.german.language_code  # 'de-DE'
Locale.brazil_portuguese.language_code  # 'pt-BR'
Locale.spanish.language_code  # 'es-ES'
Locale.latin_american_spanish.language_code  # 'es-419'
```

## Checklist

- [x] This PR fixes an issue. (#9589)

### DIFF
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -21,6 +21,7 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
 FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 DEALINGS IN THE SOFTWARE.
 """
+
 from __future__ import annotations
 
 import types
@@ -770,7 +771,7 @@ class Locale(Enum):
     @property
     def language_code(self) -> str:
         """Returns the locale's language code in the format of ``language-COUNTRY``.
-        
+
         .. versionadded:: 2.5
         """
         return _UNICODE_LANG_MAP.get(self.value, self.value)
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -694,6 +694,42 @@ class MFALevel(Enum, comparable=True):
     require_2fa = 1
 
 
+_UNICODE_LANG_MAP: Dict[str, str] = {
+    'bg': 'bg-BG',
+    'zh-CN': 'zh-CN',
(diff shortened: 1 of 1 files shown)


discord/enums.py
```diff
@@ -731,6 +768,14 @@ class Locale(Enum):
     def __str__(self) -> str:
         return self.value
 
+    @property
+    def language_code(self) -> str:
+        """Returns the locale's language code in the format of ``language-COUNTRY``.
+
+        .. versionadded:: 2.5
```
Rapptz:
```suggestion
        .. versionadded:: 2.6
```

discord/enums.py
```diff
@@ -731,6 +768,14 @@ class Locale(Enum):
     def __str__(self) -> str:
         return self.value
 
+    @property
+    def language_code(self) -> str:
+        """Returns the locale's language code in the format of ``language-COUNTRY``.
```
Rapptz:
```suggestion
        """:class:`str`: Returns the locale's BCP 47 language code in the format of ``language-COUNTRY``.
```

discord/enums.py
```diff
@@ -731,6 +768,14 @@ class Locale(Enum):
     def __str__(self) -> str:
         return self.value
 
+    @property
+    def language_code(self) -> str:
```
Rapptz, line 772: Either this needs to return `None` or it needs to document the fallback someway.

Rapptz (changes requested):
Thanks. Seems mostly ok.

discord/enums.py
exorcize: Thanks.
I've documented the fallback behavior in the docstring as suggested.

PR #10223, opened 2025-07-04
discord/voice_client.py expects the VoiceConnectionState to have a loop attribute, but it is missing

avleen:
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/voice_state.py
+++ b/discord/voice_state.py
@@ -208,6 +208,7 @@ class VoiceConnectionState:
         self.mode: SupportedModes = MISSING
         self.socket: socket.socket = MISSING
         self.ws: DiscordVoiceWebSocket = MISSING
+        self.loop: asyncio.AbstractEventLoop = None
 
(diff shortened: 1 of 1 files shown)


avleen:
Minor bug fix that adds in an expected but missing Voice State attribute.

DA-344:
I can't find anywhere in VoiceClient's code when ``VoiceConnectionState.loop`` is used? Could you send the related lines in VoiceClient where this is needed? The most similar thing I found is https://github.com/Rapptz/discord.py/blob/master/discord/voice_client.py#L226 , but is not a VoiceConnectionState, but the ConnectionState a discord.Client uses.

Rapptz:
This does not seem correct to me? There's no reason for the voice state to use a `loop` attribute

avleen:
Yup! This one's on me. I was trying to fix something else and did something entirely wrong, but was fixed by this 😂 
My apologies, I'll close this out!

PR #9260, opened 2023-02-25
Begin implementaton of guild onboarding configuration support

bijij:
## Summary

This pull request implements support for the guild onboarding configuration.

## Checklist

- [x] If code changes were made then they have been tested.
    - [x] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.
- [x] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

PR #10226, opened 2025-07-09
Add support for guild onboarding

Soheab:
## Summary

This PR is supposed to supersede #9260 as it has bean abandoned for over a year and the author [doesn't respond on Discord](https://canary.discord.com/channels/336642139381301249/1078729035267461180).

I have to emphasise that most of the work has been done by the original author (@bijij), as you can see in the commit history that has been copied over.

## Checklist

- [ ] This PR fixes an issue.


discord/onboarding.py
```diff
@@ -0,0 +1,350 @@
...
+    from .threads import Thread
+    from .types.onboarding import (
+        Prompt as PromptPayload,
+        PromptOption as PromptOptionPayload,
+        Onboarding as OnboardingPayload,
+    )
+    from .state import ConnectionState
+
+
+class PartialOnboardingPromptOption:
```
dolfies: Can the partials not be coalesced into one type a la `ForumTag`

Soheab: Resolving channels and roles requires a Guild, so users would need to pass that or it needs to be typed as Guild | None. However, it cannot be None when received from the library.

I guess we could lie to the type checker and document it, but I’m not sure.

dolfies: I would just throw an error if trying to access the properties on a self-constructed instance but idk either

Soheab: Sounds like a good solution for me.

discord/enums.py
```diff
@@ -400,6 +402,13 @@ class AuditLogAction(Enum):
     automod_timeout_member                            = 145
     creator_monetization_request_created              = 150
     creator_monetization_terms_accepted               = 151
+    onboarding_prompt_create                          = 163
+    onboarding_prompt_update                          = 164
+    onboarding_prompt_delete                          = 165
+    onboarding_create                                 = 166
+    onboarding_update                                 = 167
+    home_settings_create                              = 190
```
dolfies, line 412: is there a point in including these if new member welcome isn't implemented?

discord/guild.py
```diff
@@ -4867,3 +4870,74 @@ async def create_soundboard_sound(
 
         data = await self._state.http.create_soundboard_sound(self.id, reason=reason, **payload)
         return SoundboardSound(guild=self, state=self._state, data=data)
+
+    async def onboarding(self) -> Onboarding:
+        """|coro|
+
+        Fetches the onboarding configuration for this guild.
+
+        You must have :attr:`Permissions.manage_guild` and
```
dolfies:
```suggestion
        You must have :attr:`Permissions.manage_guild` or
```
going off memory but I'm 99% sure this isn't `and`

Soheab: Oh, that’s on the wrong method. I mixed up GET welcome screen with modify onboarding.

Modifying requires both according to the [API docs](https://discord.com/developers/docs/resources/guild#modify-guild-onboarding).

discord/enums.py
Soheab: I’m not aware of a “new member welcome”. These were included in the [same PR](<https://github.com/discord/discord-api-docs/pull/6041>), so I thought I’d add them.

dolfies: sorry, usually called server guide :p
that's what the HOME_SETTINGS_X values are for, tho ig no harm either way

PR #10227, opened 2025-07-11
Add support for new RPC fields

Soheab:
## Summary

https://github.com/discord/discord-api-docs/pull/7674/

Enum names are upped cased because `name` is an existing attribute.
Bots don't support any of this, just like before.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/activity.py
+++ b/discord/activity.py
@@ -28,7 +28,7 @@ import datetime
 from typing import Any, Dict, List, Optional, TYPE_CHECKING, Union, overload
 
 from .asset import Asset
-from .enums import ActivityType, try_enum
+from .enums import ActivityType, StatusDisplayType, try_enum
 from .colour import Colour
 from .partial_emoji import PartialEmoji
 from .utils import _get_as_snowflake
@@ -180,8 +180,10 @@ class Activity(BaseActivity):
 
         - ``large_image``: A string representing the ID for the large image asset.
         - ``large_text``: A string representing the text when hovering over the large image asset.
+        - ``large_url``: A string representing the URL of the large image asset.
         - ``small_image``: A string representing the ID for the small image asset.
         - ``small_text``: A string representing the text when hovering over the small image asset.
+        - ``small_url``: A string representing the URL of the small image asset.
 
     party: :class:`dict`
         A dictionary representing the activity party. It contains the following optional keys:
@@ -195,6 +197,19 @@ class Activity(BaseActivity):
 
     emoji: Optional[:class:`PartialEmoji`]
         The emoji that belongs to this activity.
+    details_url: Optional[:class:`str`]
+        A URL that is linked to when clicking on the details text of the activity.
+
+        ... versionadded:: 2.6
+    state_url: Optional[:class:`str`]
+        A URL that is linked to when clicking on the state text of the activity.
+
+        ... versionadded:: 2.6
+    status_display_type: Optional[:class:`StatusDisplayType`]
+        Determines which field from the user's status text is displayed
+        in the members list.
+
+        .. versionadded:: 2.6
     """
 
     __slots__ = (
@@ -213,6 +228,9 @@ class Activity(BaseActivity):
         'application_id',
         'emoji',
         'buttons',
+        'state_url',
+        'details_url',
+        'status_display_type',
     )
 
     def __init__(self, **kwargs: Any) -> None:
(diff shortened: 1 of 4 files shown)


discord/enums.py
```diff
@@ -912,6 +913,12 @@ class SubscriptionStatus(Enum):
     inactive = 2
 
 
+class StatusDisplayType(Enum):
+    NAME = 0
```
dolfies: I think it's better to just prefix them with activity_ instead of changing the case

Rapptz:
Make them lowercase.

The code does not care.

```py
>>> from discord.enums import Enum
>>> class X(Enum):
...     name = 1
...
>>> X.name
<X.name: 1>
```

Soheab:
> Make them lowercase.
> 
> The code does not care.
> 
> ```python
> >>> from discord.enums import Enum
> >>> class X(Enum):
> ...     name = 1
> ...
> >>> X.name
> <X.name: 1>
> ```

Works. But the type checker doesn't understand it:
```py
class Enum(discord.Enum):
    name = 0
    # Type "Literal[0]" is not assignable to declared type "DynamicClassAttribute"
    # "Literal[0]" is not assignable to "DynamicClassAttribute" Pylance reportAssignmentType

e = Enum.name
reveal_type(e) # DynamicClassAttribute
reveal_type(e.name)
# Cannot access attribute "name" for class "DynamicClassAttribute"
#  Attribute "name" is unknown Pylance reportAttributeAccessIssue
reveal_type(e.value)
# Cannot access attribute "value" for class "DynamicClassAttribute"
#  Attribute "value" is unknown Pylance reportAttributeAccessIssue
```

Rapptz:
File a bug with pylance.

```py
>>> import enum
>>> class Y(enum.Enum):
...   name = 1
...
>>> Y.name
<Y.name: 1>
```

Soheab:
> File a bug with pylance.
> 
> ```python
> >>> import enum
> >>> class Y(enum.Enum):
> ...   name = 1
> ...
> >>> Y.name
> <Y.name: 1>
> ```

It's apparently [as designed](https://github.com/microsoft/pyright/issues/9728). But i've lower-cased them.

discord/activity.py
```diff
@@ -195,6 +197,19 @@ class Activity(BaseActivity):
 
     emoji: Optional[:class:`PartialEmoji`]
         The emoji that belongs to this activity.
+    details_url: Optional[:class:`str`]
+        A URL that is linked to when clicking on the details text of the activity.
+
+        ... versionadded:: 2.6
```
Snipy7374:
```suggestion
        .. versionadded:: 2.6
```

discord/activity.py
```diff
@@ -195,6 +197,19 @@ class Activity(BaseActivity):
...
+    state_url: Optional[:class:`str`]
+        A URL that is linked to when clicking on the state text of the activity.
+
+        ... versionadded:: 2.6
```
Snipy7374:
```suggestion
        .. versionadded:: 2.6
```

discord/activity.py
```diff
@@ -255,11 +283,15 @@ def __repr__(self) -> str:
...
     def to_dict(self) -> Dict[str, Any]:
         ret: Dict[str, Any] = {}
+        enums = ('status_display_type',)
         for attr in self.__slots__:
             value = getattr(self, attr, None)
             if value is None:
                 continue
 
+            if attr in enums:
+                value = int(value.value)
```
Rapptz: Not that it's a great performance penalty in this case but this type of code leads to accidental quadratic behaviour.

Instead, you should move this outside of the for loop body and do:

```py
for attr in enums:
    ret[attr] = int(ret[attr])
```

But since this is only a single enum and it's okay to just special case it and do it like it was done for `self.emoji`.

discord/enums.py
```diff
@@ -912,6 +913,12 @@ class SubscriptionStatus(Enum):
     inactive = 2
 
 
+class StatusDisplayType(Enum):
+    name = 0  # pyright: ignore[reportAssignmentType]
```
Rapptz, line 919: Did we figure out how this looks from a user's POV when they try using this attribute?

Rapptz (changes requested):
Looks mostly okay, just fix the pre-existing nits as well.

discord/enums.py
Soheab: Still the same as [before](https://github.com/Rapptz/discord.py/pull/10227#issuecomment-3063512685) when accessing `.name` or `.value` from the member, and no response to [the comment](https://github.com/microsoft/pyright/issues/9728#issuecomment-3063808515) left on the Pyright repository about it.

But to be fair, it works fine when setting or comparing it:

```py
act = discord.Activity(
    ...
    status_display_type=discord.StatusDisplayType.name,
)
print(act.status_display_type) # StatusDisplayType.name
if act.status_display_type is discord.StatusDisplayType.name:
    print("Status display type is name")
```

PR #10230, opened 2025-07-16
Add support to send voice messages

### LINKED ISSUE #10165: Voice message support
### Summary

Add support for sending audio as voice messages

The core library

### The Problem

I've been thinking of a way of sending audio as voice messages for my bots. Turns out it's technically possible to do that with the API currently, and the specifics of it are documented here: https://gist.github.com/HDR/7d5d4ce8bbe4b715d788a9bc9f99e02d

### The Ideal Solution

To send a voice message you need to set the message flag to `8192` in the message data, have an audio file (only .ogg is supported) attachment with waveform data that's essentially just a base64 encoded byte array. Basically what I'm thinking is the message send methods could have a voice message parameter for passing a voice attachment object with all the data mentioned above.

### The Current Solution

_No response_

### Additional Context

_No response_

blord0:
## Summary

Adds support for voice messags
closes https://github.com/Rapptz/discord.py/issues/10165

## Checklist

- [ ] This PR fixes an issue.


discord/http.py
```diff
@@ -243,7 +243,10 @@ def handle_message_parameters(
         file_index = 0
         attachments_payload = []
         for attachment in attachments:
-            if isinstance(attachment, File):
+            if isinstance(attachment, VoiceMessageFile):
+                attachments_payload.append(attachment.to_dict(file_index))
+                file_index += 1
+            elif isinstance(attachment, File):
```
AbstractUmbra: Why is this code block necessary?

discord/http.py
```diff
@@ -269,6 +272,9 @@ def handle_message_parameters(
     multipart = []
     if files:
         multipart.append({'name': 'payload_json', 'value': utils._to_json(payload)})
+        print(";;;;")
+        print(utils._to_json(payload))
+        print(";;;;")
```
AbstractUmbra: Prints?

DA-344:
This is discussed here, and iirc, this was not planned at the moment, might be wrong though, worth checking: https://discord.com/channels/336642139381301249/1100398536920662117

discord/http.py
```diff
@@ -617,7 +623,9 @@ async def request(
                 headers['X-Audit-Log-Reason'] = _uriquote(reason, safe='/ ')
 
         kwargs['headers'] = headers
-
+        print("=-=-=-=-=-=-=-=-=-=-=-=")
+        print(headers)
+        print("=-=-=-=-=-=-=-=-=-=-=-=")
```
AbstractUmbra: Prints?

discord/http.py
```diff
@@ -868,6 +876,10 @@ def send_message(
         *,
         params: MultipartParameters,
     ) -> Response[message.Message]:
+        print(":::::")
+        print(params.payload)
+        print(params.multipart)
+        print(":::::")
```
AbstractUmbra: Prints?

discord/http.py
```diff
@@ -1058,6 +1070,63 @@ def unpin_message(self, channel_id: Snowflake, message_id: Snowflake, reason: Op
     def pins_from(self, channel_id: Snowflake) -> Response[List[message.Message]]:
         return self.request(Route('GET', '/channels/{channel_id}/pins', channel_id=channel_id))
 
+    async def send_voice_message(self, channel_id: Snowflake, VoiceMessage: VoiceMessageFile):
```
AbstractUmbra: Why is the parameter in PascalCase? 

```suggestion
    async def send_voice_message(self, channel_id: Snowflake, voice_message: VoiceMessageFile):
```

discord/http.py
```diff
@@ -1058,6 +1070,63 @@ def unpin_message(self, channel_id: Snowflake, message_id: Snowflake, reason: Op
...
+        Sends a voice message to the specified channel.
+
+        Parameters
+        -----------
+        channel_id: :class:`~discord.abc.Snowflake`
+            The ID of the channel to send the voice message to.
+        VoiceMessage: :class:`~discord.VoiceMessageFile`
+            The voice message file to send. This should be an instance of :class:`~discord.VoiceMessageFile`.
+        """
+        from .message import MessageFlags
```
AbstractUmbra: Unused import.

discord/http.py
```diff
@@ -1058,6 +1070,63 @@ def unpin_message(self, channel_id: Snowflake, message_id: Snowflake, reason: Op
...
+        uploadRoute = Route('POST', '/channels/{channel_id}/attachments', channel_id=channel_id)
```
AbstractUmbra: Why is this parameter in camelCase?
```suggestion
        upload_route = Route('POST', '/channels/{channel_id}/attachments', channel_id=channel_id)
```

discord/http.py
```diff
@@ -1058,6 +1070,63 @@ def unpin_message(self, channel_id: Snowflake, message_id: Snowflake, reason: Op
...
+        message_payload = {
+            "flags": 8192,  # IS_VOICE_MESSAGE
+            "attachments": [VoiceMessage.to_dict(0)],
+        }
+
+        headers = {"Authorization": f"Bot {self.token}",
+                   "Content-Type": "application/json"}
+
+        res = requests.post("" + r.url, headers=headers, json=message_payload)
+        return res.json()
```
AbstractUmbra: Why are we going from using the library provided (and correct) http request format, into import requests (a non-required discord.py dependency, and sync/blocking library) for an entirely different route?
Not to mention this code is a mish-mash of differing quality.

AbstractUmbra (changes requested):
Was the code in this PR AI generated or something? Frankly I'd refuse to believe otherwise, due to seeing 3 different conventions/styles as well as a mish-mash of logic that doesn't really make sense.

I appreciate it's still in draft, but even so.

blord0:
> This is discussed here, and iirc, this was not planned at the moment, might be wrong though, worth checking: https://discord.com/channels/336642139381301249/1100398536920662117

Discord has added the message flag for "IS_VOICE_MESSAGE" in their offical docs so I think we good? https://discord.com/developers/docs/resources/message#message-object-message-flags

blord0:
> Was the code in this PR AI generated or something? Frankly I'd refuse to believe otherwise, due to seeing 3 different conventions/styles as well as a mish-mash of logic that doesn't really make sense.
> 
> I appreciate it's still in draft, but even so.

Bit of AI was used as I was stuggling to get it all working. Like how I imported `requests` so that I could get actual values from the requests.
I will use the actual libraries/clean it all up tho (including the horrific amount of print statments i gotta remove)

AbstractUmbra:
Pull requests that use AI generated code do not get accepted due to the possible licensing implications.
Not to mention the quality of AI generate code being abysmal.

blord0:
> Pull requests that use AI generated code do not get accepted due to the possible licensing implications. Not to mention the quality of AI generate code being abysmal.

The only ai part is the part where `requests` is imported and used. That was never intended for the final merge. 
All other abysmal code/syntax/logic was just me doing patchwork to get it working and returning some values

AbstractUmbra:
Additionally I'm not sure where the library stands on sending "fake" data, we don't do it anywhere else so sending a fake waveform and fake size likely won't work.

blord0:
> Additionally I'm not sure where the library stands on sending "fake" data, we don't do it anywhere else so sending a fake waveform and fake size likely won't work.

Fake size is a placeholder (as commented) so I can send stuff while testing, do plan adding actual code to the `.size()` method for final 

The waveform data is only used as a visual preview for how the message looks. I do plan on looking into if I can get the real waveform data, but the only code I've seen that sends a voice message (https://gist.github.com/max-4-3/6704422045b1c21a4a3e9226fca38f68) does the same thing that I am currently.
It isn't checked by discord so if there is no method of getting it (without external tools like ffmpeg), then the message won't fail to send. Also the reason I allow users to pass in their own `waveform` right now
image

discord/abc.py
```diff
@@ -1915,6 +1921,31 @@ async def _before_strategy(retrieve: int, before: Optional[Snowflake], limit: Op
...
+        ~discord.Forbidden
+            You do not have the proper permissions to send the voice message.
+
+        Returns
+        --------
+        :class:`~discord.Message`
+            The message that was sent.
+        """
+
+        return await self.send(file=file, voice=True)
```
blord0: Since this method is just 1 line of code now, not too sure if it deserves to be its own seperate thing?

discord/file.py
```diff
@@ -157,3 +159,41 @@ def to_dict(self, index: int) -> Dict[str, Any]:
             payload['description'] = self.description
 
         return payload
+
+
+class VoiceMessageFile(File):
```
blord0: Maybe could merge this with the `File` class?

blord0 (commented):
Found a much simpler way of sending voice messages while looking through https://discord.com/channels/336642139381301249/1100398536920662117

blord0:
Example code for sending a voice message is
```py
  channel = await bot.fetch_channel(xxx)
  file = discord.VoiceMessageFile("test.mp3", duration=21)
  await channel.send(file=file, voice=True)
```

discord/http.py
```diff
@@ -617,7 +617,6 @@ async def request(
                 headers['X-Audit-Log-Reason'] = _uriquote(reason, safe='/ ')
 
         kwargs['headers'] = headers
-
         # Proxy support
```
AbstractUmbra, line 633: Why was this newline removed?

AbstractUmbra (changes requested):
I wouldn't make a whole new File class, and instead have `voice` be an init parameter or a generated one (i.e. if `duration` is provided, or similar) of `File` and use this as needed.

discord/http.py
blord0: Must have accidentally got it while removing the print stuff

discord/abc.py
```diff
@@ -1624,15 +1633,26 @@ async def send(
...
             flags = MISSING
 
+        if voice:
+            if content is not None:
+                raise TypeError('Cannot send content with a voice message')
+            if embed is not None or embeds is not None:
+                raise TypeError('Cannot send embeds with a voice message')
+            if file is None:
+                raise TypeError('A voice message must have a file')
+            if file.duation is None:
```
Rapptz:
```suggestion
            if file.duration is None:
```

discord/file.py
```diff
@@ -77,7 +78,7 @@ class File:
         .. versionadded:: 2.0
     """
 
-    __slots__ = ('fp', '_filename', 'spoiler', 'description', '_original_pos', '_owner', '_closer')
+    __slots__ = ('fp', '_filename', 'spoiler', 'description', '_original_pos', '_owner', '_closer', 'duation', '_waveform')
```
Rapptz:
```suggestion
    __slots__ = ('fp', '_filename', 'spoiler', 'description', '_original_pos', '_owner', '_closer', 'duration', '_waveform')
```

discord/file.py
```diff
@@ -117,6 +120,8 @@ def __init__(
 
         self.spoiler: bool = spoiler
         self.description: Optional[str] = description
+        self.duation = duration
```
Rapptz:
```suggestion
        self.duration = duration
```

discord/file.py
```diff
@@ -156,4 +168,8 @@ def to_dict(self, index: int) -> Dict[str, Any]:
         if self.description is not None:
             payload['description'] = self.description
 
+        if self.duation is not None:
+            payload['duration_secs'] = self.duation
+            payload['waveform'] = self.waveform
```
Rapptz:
```suggestion
        if self.duration is not None:
            payload['duration_secs'] = self.duration
            payload['waveform'] = self.waveform
```

Rapptz (changes requested):
Thanks for the PR.

A few nits though:

1. Instead of passing `voice=True` at the `send` site, it should be at the `File` site instead because that's where it logically is. A voice message is nothing more than a fancy attachment that toggles a flag.
2. You should check whether multiple embedded audio attachments are supported per message.
3. You're missing `.. versionadded:: 2.6` on the new `File` properties.
4. You're missing documentation for `duration` (which you typo'd as `duation` a few times).
5. The waveform should be generated properly.
6. More documentation is needed for this in general, such as what audio format is expected to be uploaded for it to be a voice message etc.

blord0:
Thanks, didn't catch the typos there. Also added `File.voice` bool to for checking if the file is a voice message now
If you try to add multiple voice messages to a message, discord throws a `400 Bad Request (error code: 50160): Voice messages must have a single audio attachment`

For points 5 and 6, discord accepts a wide range of audio formats. Found it to work with `ogg`, `mp3`, `wav`, `aac` and `flac`. Doesn't work with `.mp4`
Each different file format would need its own way to calculate the waveform data. I can find out ways to calculate waveform data for each of the types above, but
1. that list above is non-exhausitve
2. it would require adding more external dependencies to the project (not sure how you feel about that)

Rapptz:
Looking at the [documentation](https://discord.com/developers/docs/resources/message#voice-messages) you can probably make a valid waveform for the .ogg ones using the facilities in the library, if possible, otherwise you can fallback to the random waveform.

blord0:
From what i can see, the `oggparse` and `opus` libraries can only be used for voice data, and fail when used on general opus files. Even a simple script like this fails due to a `corrupted stream` error. I downloaded my own voice-message.ogg from discord and checked with ffprobe to make sure it was a valid ogg file
```py
from discord.oggparse import OggStream
from discord.opus import Decoder

filename = "voice-message.ogg"
f = open(filename, 'rb')
ogg = OggStream(f)
decoder = Decoder()
for packet in ogg.iter_packets():
  decoded = decoder.decode(packet) # discord.opus.OpusError: corrupted stream
  print(decoded)
```

Looked into another library called [soundfile](https://pypi.org/project/soundfile) and this seems to be able to generate waveform for ogg, mp3 and wav.
Could impliment somthing similar to PyNacl, where you don't have to install it but it is required to generate the real waveform data without falling back to a random waveform?

AbstractUmbra:
`soundfile` is a rather large dependency, since it includes `numpy` and a few others. Each wheel is around 1mib. I personally think that's a bit overkill.

I've discussed this waveform idea in depth and it should be possible using the opus/ogg items the library provides (I will do some testing when I have the time, soon) and generating the correct and real waveform should be a fast process.

However, this does mean we'll need to accept only `OGG` formatted files, it will be on the end user to do the necessary conversion somehow, unless we do decide to go in depth.

blord0:
Just had a deeper look into what each packet from my code example above has in it, and seems like the first 2 are metadata. If I just skip those 2 packets then decoder works fine all the way through.
Going to have a look around to see how to turn that data into waveform stuff
```py
from discord.oggparse import OggStream
from discord.opus import Decoder

filename = "voice-message.ogg"
f = open(filename, 'rb')
ogg = OggStream(f)
decoder = Decoder()
count = 0
for packet in ogg.iter_packets():
    if count < 2:
        count += 1
        continue
    decoded: bytes = decoder.decode(packet) # Gives actual bytes now
```

discord/file.py
```diff
@@ -126,6 +170,24 @@ def filename(self) -> str:
...
+            If a waveform was not given, it will be generated
+
+            Only supports generating the waveform for Opus format files, other files will be given a random waveform
+
+        .. versionadded:: 2.6"""
+        if self._waveform is None:
+            try:
+                self._waveform = self.generate_waveform()
+            except Exception:
+                self._waveform = base64.b64encode(os.urandom(256)).decode('utf-8')
```
scarletcafe: I'm not sure if I'm a big fan of this caveat, especially not as a catch-all except case. I think us handling audio extraction for anything outside of opus 'in house' is pretty far out of scope, but this feels like a design 'lock-in' that could prevent people who have the means to do the waveform generation themselves from doing so. Maybe the waveform property could have a setter, or maybe there could be a way to construct a File (or a specialized subclass) with a waveform provided e.g. `File.voice_message_with_waveform(data, waveform)`.

discord/file.py
```diff
@@ -156,4 +218,63 @@ def to_dict(self, index: int) -> Dict[str, Any]:
...
 
+        if self.voice:
+            payload['duration_secs'] = self.duration
+            payload['waveform'] = self.waveform
+
         return payload
+
+    def generate_waveform(self) -> str:
+        if not self.voice:
+            raise TypeError("Cannot produce waveform for non voice file")
```
scarletcafe: I think this should be ValueError or a subclass of it instead of TypeError, as this error is caused by the user producing the wrong 'type' of file and not the actual wrong Python `type` (i.e. class instance).

discord/file.py
```diff
@@ -156,4 +218,63 @@ def to_dict(self, index: int) -> Dict[str, Any]:
...
+        ogg = OggStream(self.fp)  # type: ignore
+        decoder = Decoder()
+        waveform: list[int] = []
+        prefixes = [b'OpusHead', b'OpusTags']
+        for packet in ogg.iter_packets():
+            if packet[:8] in prefixes:
+                continue
+
+            if b'vorbis' in packet:
+                raise TypeError("File format is 'vorbis'. Format of 'opus' is required for waveform generation")
```
scarletcafe: Ditto - I don't think this should be TypeError

discord/file.py
```diff
@@ -156,4 +218,63 @@ def to_dict(self, index: int) -> Dict[str, Any]:
...
+            count += 1
+            if i % points_per_sample == 0:
+                sample_waveform.append(total // count)
+                total, count = 0, 0
+
+        # Maximum value of a waveform is 0xff (255)
+        highest = max(sample_waveform)
+        mult = 255 / highest
+        for i in range(len(sample_waveform)):
+            sample_waveform[i] = int(sample_waveform[i] * mult)
```
scarletcafe, line 298: We already rely on `audioop` or `audioop-lts` for newer versions of Python, which includes many audio operations implemented in C. I don't have a lot of reason to believe this operation is *that* slow, but I'm inclined to believe we can probably use audioop to make this simpler and faster to do.

discord/file.py
```diff
@@ -156,4 +218,63 @@ def to_dict(self, index: int) -> Dict[str, Any]:
...
+            payload['duration_secs'] = self.duration
+            payload['waveform'] = self.waveform
+
         return payload
+
+    def generate_waveform(self) -> str:
+        if not self.voice:
+            raise TypeError("Cannot produce waveform for non voice file")
+        self.reset()
+        ogg = OggStream(self.fp)  # type: ignore
```
scarletcafe, line 252: It seems this is `# type: ignore`'d because OggStream expects `IO[bytes]` which `BufferedIOBase` is not compatible with. Maybe we should reconsider the type that OggStream takes, because I don't really think we should be type-ignoring this.

discord/file.py
```diff
@@ -156,4 +218,63 @@ def to_dict(self, index: int) -> Dict[str, Any]:
...
+        decoder = Decoder()
+        waveform: list[int] = []
+        prefixes = [b'OpusHead', b'OpusTags']
+        for packet in ogg.iter_packets():
+            if packet[:8] in prefixes:
+                continue
```
scarletcafe, line 258: Maybe we should consider having some of this audio-extraction logic being part of OggStream itself rather than here? I don't foresee a whole lot of other situations where we need to do this so it might not be necessary, but it's worth considering I think.

discord/file.py
```diff
@@ -126,6 +170,24 @@ def filename(self) -> str:
         """
         return 'SPOILER_' + self._filename if self.spoiler else self._filename
 
+    @property
+    def waveform(self) -> str:
```
scarletcafe: I don't know that any real user would actually want to consume the waveform in its raw base64 state, especially because it obscures what the underlying data is actually supposed to be. Maybe this property should store the waveform in its original bytes, and the conversion to base64 should be done in `to_dict`?

scarletcafe (changes requested):
My personal thoughts, I think this is a bit of a tough one to approach.

discord/file.py
blord0: You can actually pass in your own waveform when creating a file
`file = discord.File('voice-message.ogg', duration=5.0, waveform='AAAAA...')`

discord/file.py
blord0: audioop is depricated in python 3.11 and removed since 3.13
https://docs.python.org/3/library/audioop.html

Rapptz: We already include `audioop` for 3.13+ using `audioop-lts`.

discord/file.py
blord0: OggStream only uses `IO[bytes].read()` which `BufferedIOBase` does impliment. Changing File to use `IO[bytes]` could cause headaches elsewhere and changing OggStream to use `BufferedIOBase` does not seem feasible as it is used for streaming voice call data and would have to cause a potentially large re-write

Imo its fine as the functions are called in the same way

discord/file.py
scarletcafe: Hmm, that makes this less of an issue, but I still think we should avoid generating a fake waveform if we can, maybe we could just let the exception be raised instead or split generation and non-generation into classmethod-based signatures like suggested.

discord/file.py
blord0: The voice call part use OggStream and I have not looked (in depth) into if they use these packets, did it over in the File class to make sure I don't screw anything up.
Imo its fine being where it is

discord/file.py
blord0: Yea that is a better idea, il change it so that it works like that

discord/file.py
blord0: The part of code you have commented this on doesn't decode or process the Opus data.
Lines 259 to 277 just processes the list of `int`s, that represent the waveform, to be in the form that discord expects them to be (values between 0-255 and a maximum of 255 values)
Since its just a list of ints, not sure if audioop would be applicable?

discord/file.py
blord0: Problem is that if someone passes in an mp3 file, we still need to generate a waveform. I don't see the point in adding extra steps for the user to generate a waveform dependant on if they have the correct audio type as that could be confusing

discord/file.py
```diff
@@ -126,6 +177,24 @@ def filename(self) -> str:
...
+
+        .. note::
+            If a waveform was not given, it will be generated
+
+            Only supports generating the waveform for Opus format files, other files will be given a random waveform
+
+        .. versionadded:: 2.7"""
+        if self._waveform is None:
+            try:
+                self._waveform = self.generate_waveform()
```
AbstractUmbra: Properties should be inexpensive to access. I would move this generation into the init if a waveform is not provided. Doing it on a property access seems rather silly.

AbstractUmbra (changes requested):

bl791:
Any update on this?

br3nr:
Would love this to be a feature at some point

blord0:
> Would love this to be a feature at some point

It is ready rn, just waiting for merge or more reviews

Mr-Auto:
Wouldn't this be simpler to just subclass the `File` type?
Then you could just require the needed parameters instead of introducing three new ones to a very generic File type.

Also would be nice to have generate_waveform be called when sending a message with missing waveform, so it doesn't potentially waste time generating if someone creates the file first and then assigns the waveform later. Would also allow for the function to be made async, if the generation is deemed too expensive to run synchronously

blord0:
Earlier on, it was a subclass of `File` called `discord.VoiceMessageFile`, but it was decided that it would be easier to extend the `File` class.

With the generate_waveform, all of the work is synchronous so making it an async function wouldn’t really help. Also, for the type of audio files that generate_waveform can handle (Opus, the type that discord clients send their vms with), it is not that expensive and takes around 0.5 seconds for a 30 second voice message.
The likelihood that someone using an Opus voice message wants to provide their own waveform, but cannot pass it during `File` creation, seems low enough that I do not think deferring waveform generation is necessary here, and may risk overcomplicating the code

PR #10231, opened 2025-07-20
Add FAQ about guild specific app commands

Soheab:
## Summary

This PR adds a FAQ entry that shows all the ways to define a guild specific app command.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/tree.py
+++ b/discord/app_commands/tree.py
@@ -248,6 +248,13 @@ class CommandTree(Generic[ClientT]):
 
         Note that this method will *override* pre-existing guild commands that would conflict.
 
+        .. note::
+
+            You probably do not want to use this. This is mainly for development purposes.
+
+            See other ways of declaring a guild specific command in the FAQ.
+
+
         Parameters
         -----------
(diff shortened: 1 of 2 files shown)


discord/app_commands/tree.py
```diff
@@ -248,6 +248,13 @@ def copy_global_to(self, *, guild: Snowflake) -> None:
 
         Note that this method will *override* pre-existing guild commands that would conflict.
 
+        .. note::
+
+            You probably do not want to use this. This is mainly for development purposes.
+
+            See other ways of declaring a guild specific command in the FAQ.
+
+
```
Rapptz:
```suggestion
```

The docstring already mentions this. Repeating this adds no value.

Soheab:
> I think this FAQ entry is flawed.
> 
> The question is probably closer to "How do I restrict a command to a specific guild?" and the answer to that question is slightly different to the one being proposed here, though it does have some similarities such as a guild command.
> 
> I suggest rewording the sentence and answer with that in mind instead.

I have reworded it to answer that exact question. A little unsure, so feedback is welcome.

PR #10238, opened 2025-07-25
Skip GUILD_MEMBER_ADD if member already cached

tudubucket:
## Summary

Skips processing `GUILD_MEMBER_ADD` if the member is already present in the guild cache.

This prevents duplicate handling in very specific cases where a `GUILD_MEMBER_ADD` event is received for a member who is already cached - a situation that may occur under specific timing.

Without this check:
- `Guild.member_count` may become inaccurate
- Logic depending on `Guild.chunked` being accurate may break

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/state.py
+++ b/discord/state.py
@@ -1101,6 +1101,12 @@ class ConnectionState(Generic[ClientT]):
             _log.debug('GUILD_MEMBER_ADD referencing an unknown guild ID: %s. Discarding.', data['guild_id'])
             return
 
+        member_id = int(data['user']['id'])
+        member = guild.get_member(member_id)
+        if member is not None:
+            _log.debug('GUILD_MEMBER_ADD referencing an already cached member ID: %s. Discarding.', member_id)
+            return
+
         member = Member(guild=guild, data=data, state=self)
         if self.member_cache_flags.joined:
             guild._add_member(member)


Rapptz:
I have never seen a duplicate GUILD_MEMBER_ADD. What situation did you see this happen?

tudubucket:
> I have never seen a duplicate GUILD_MEMBER_ADD. What situation did you see this happen?

This behavior appears to occur when a user joins a guild unusually quickly (e.g. by repeatedly spamming the join button). The following observations in discord.py discord may be relevant:

- [Guild becomes unchunked after some time](https://discord.com/channels/336642139381301249/1381172482236612700) (edited since old link doesnt work)
- [guild cannot being chunked, even after manual `guild.chunk()`](https://discord.com/channels/336642139381301249/1296131711884591135)

I have personally encountered this once (as referenced in the second thread), but I have not yet been able to reliably reproduce the issue.

While one thread attributes the behavior to a temporary Discord API issues (which discord did report elevated error rates on the same day), I believe this cannot be solely attributed to the outage. In my case, the anomaly occurred nearly an hour before Discord API encounter elevated errors.

Furthermore, I attempted to manually chunk the guild and observed inconsistencies between `guild.member_count` and `len(guild.members)`. These mismatches suggest that the internal member cache may be in an incomplete or desynchronized state:

image


In the image above, `cached:` refers to the result of `len(members)` immediately after performing manual chunking via `members = await guild.chunk()`.  
`info:` corresponds to `guild.member_count`, which reflects the internal state of the discord.py cache.

Rapptz:
I guess there's no real harm in allowing this. Thanks for the PR.

PR #10242, opened 2025-07-28
Fallback/Default to role IDs if the roles are not cached

Sacul0457:
## Summary
This PR allows for `Member.roles` to return a list Role IDs if the roles are not cached.
For example, using it as a user app or not having the necessary intents.

An entirely separate `role_ids` property would work too or maybe better.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/member.py
+++ b/discord/member.py
@@ -43,6 +43,7 @@ from .colour import Colour
 from .object import Object
 from .flags import MemberFlags
 from .presences import ClientStatus
+from .role import Role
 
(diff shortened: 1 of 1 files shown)


DA-344:
Wouldn't it be better to return ``Object`` instances with ``type=Role`` instead of ``int``s?

Rapptz:
This is purposefully not done. It's also a breaking change that can't be done right now anyway.

Issue #9981, opened 2024-10-22
return in finally can swallow exceptions


iritkatriel:
### Summary

return in finally can swallow exceptions

### Reproduction Steps

In a couple of places in your code, there is a `return` statement in a `finally` block, which would swallow any in-flight exception.

https://github.com/Rapptz/discord.py/blob/c8ecbd8d100b0e285987e33ef27209fce8bc71d1/discord/player.py#L606
https://github.com/Rapptz/discord.py/blob/c8ecbd8d100b0e285987e33ef27209fce8bc71d1/discord/gateway.py#L165

This means that if a `BaseException` (such as `KeyboardInterrupt`) is raised from the `try` body, or any exception is raised from an `except`: clause, it will not propagate on as expected.

See also https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions.

### Expected Results

Avoid return in finally.

### Actual Results

NA

### Intents

NA

### System Information

NA

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.

mikeshardmind:
KeyboardInterrupt won't be received in a thread, and this is during the closing of the gateway, so any other base exception here being suppressed is fine, including several internal asyncio and concurrent futures exceptions that are considered base exceptions should be swallowed. (ie. TimeoutError, CancelledError) as well as system exit, if someone found a way to throw that into the thread (I don't think SystemExit should be possible here under normal operation)

The codec one has a similar issue, where KeyboardInterrupt (Possibly SystemExit if a codec probe here specifically calls sys.exit or raises it) is the only base exception that could occur here which we would want to avoid catching. 

The python standard library using BaseException for exceptions that involve recoverable and/or situationally ignorable concurrency state errors makes this more awkward than it should be here.

PR #10257, opened 2025-08-10
Upstream change in Python

mikeshardmind:
## Summary

This is a pessimistic change in case you'd rather not deal with whatever upstream says should be done here and want it available before you cut the 2.6 release.

## Details

The prior changes (#9984) to use of return in finally are now insufficient. Without disclosing their intent when raising the issue (#9981), the author used this as part of justifying a SyntaxError for working code outside of the normal process for adding errors, with it presenting to end users in a way that [breaks downstream user's existing CI](https://github.com/Rapptz/discord.py/discussions/10240#discussioncomment-14055608)

While making the change, I've continued to not log errors like CancellationError or TimeoutError to users here by default, as it is not an error they need to be aware of during shutdown given the limited kinds of BaseException that could raise in this context, see:  https://github.com/Rapptz/discord.py/issues/9981#issuecomment-2436146333 for prior analysis. I've added a debug log should anyone want access to this kind of failure while debugging gateway close, but due to how asyncio shutdown happens, this is unlikely to ever log anything useful even in a library debugging context.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -160,9 +160,11 @@ class KeepAliveHandler(threading.Thread):
                     f.result()
                 except Exception:
                     _log.exception('An error occurred while stopping the gateway. Ignoring.')
+                except BaseException as exc:
+                    _log.debug('A BaseException was raised while stopping the gateway', exc_info=exc)
                 finally:
                     self.stop()
-                    return
+                return
 
             data = self.get_payload()
             _log.debug(self.msg, self.shard_id, data['d'])


MajorTanya:
Hi, I'm the downstream user 😅

I would like to clarify that I was running tox with 3.11-3.14rc1 both locally and in CI, but I only got the error specifically in CI with my local tox runs being entirely happy and successful. (though I used a normal install of 3.14rc1 on my local machine but uv's install of it in CI, so I'm also not sure if that played a role).

I was extremely confused about this since the docs on the finally clause I linked in that discussion call it a SyntaxWarning, not a SyntaxError.

Just getting rid of it is probably the safe pessimistic bet, but do we know if there may be other uses of this lurking? Yes my test collection crashed on this same point 40+ times in the same run but I don't know if it'd be the only place.

mikeshardmind:
This is breaking with uv but not pip because of differences in behavior between the two by default.

For anyone that want's a no-code solution now, setting `UV_COMPILE_BYTECODE=1` in the environment will ensure that the warning happens at dependency install, not when a test running that uses -WError by defaul imports discord.py

PR #10263, opened 2025-08-15
Add `total message sent` attribute to threads

Sacul0457:
## Summary

This PR adds the `total_message_sent` attribute to Threads

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/models.py
+++ b/discord/app_commands/models.py
@@ -791,6 +791,10 @@ class AppCommandThread(Hashable):
         An approximate number of members in this thread. This caps at 50.
 
         .. versionadded:: 2.6
+    total_message_sent: :class:`int`
+        The total number of messages sent, including deleted messages.
+
+        .. versionadded:: 2.6
     permissions: :class:`~discord.Permissions`
         The resolved permissions of the user who invoked
         the application command in that thread.
@@ -830,6 +834,7 @@ class AppCommandThread(Hashable):
         'member_count',
         'slowmode_delay',
         'last_message_id',
+        'total_message_sent',
         '_applied_tags',
         '_flags',
         '_created_at',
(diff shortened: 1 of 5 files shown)


discord/types/channel.py
```diff
@@ -126,6 +126,7 @@ class ThreadChannel(_BaseChannel):
     rate_limit_per_user: int
     message_count: int
     member_count: int
+    total_message_sent : int
```
AbstractUmbra:
```suggestion
    total_message_sent: int
```

discord/types/interactions.py
```diff
@@ -85,6 +85,7 @@ class PartialThread(_BasePartialChannel):
     rate_limit_per_user: int
     last_message_id: NotRequired[Optional[Snowflake]]
     flags: NotRequired[int]
+    total_message_sent : int
```
AbstractUmbra:
```suggestion
    total_message_sent: int
```

AbstractUmbra (changes requested):

Rapptz:
Are we sure this attribute is always given?

mikeshardmind:
Their open api spec seems to indicate that it is required: https://github.com/discord/discord-api-spec/blob/8bfaeb99af56de481f2af518ed27b1223245b693/specs/openapi.json#L35682-L35692 

I don't know if we trust this given the readme says:

> This is a preview and it may be not correct. If you find discrepancies between the spec and our [docs](https://discord.com/developers/docs), other than the ones mentioned below, let us know, and follow the docs, not the spec.

Rapptz:
Good enough for me.

Just run `black` so this can be merged.

mikeshardmind:
docs indicate it's optional on channel resource, but not under what conditions it is present. API spec seems to indicate it's always present here, but this might be safer to use .get to prevent future issues.

Rapptz:
The docs are really bad at union types which I figure is the reason why the OpenAPI spec was made

discord/app_commands/models.py
```diff
@@ -855,6 +860,7 @@ def __init__(
         self.message_count: int = int(data['message_count'])
         self.last_message_id: Optional[int] = _get_as_snowflake(data, 'last_message_id')
         self.slowmode_delay: int = data.get('rate_limit_per_user', 0)
+        self.total_message_sent: int = data.get('total_message_sent')
```
dolfies:
```suggestion
        self.total_message_sent: int = data.get('total_message_sent', 0)
```
saner default

PR #10265, opened 2025-08-16
Add `fetch members named` method to `Guild`

Sacul0457:
## Summary
This PR adds a `fetch_members_named` method to `Guild`.
Docs: https://discord.com/developers/docs/resources/guild#search-guild-members
Somethings I'm unsure of/Things to note:
- Version added is not set, cuz not sure if it is 2.6 or not
- Not sure what to default to here
- Could perhaps add a note to use the property `members` and query the names themselves?
- Not the same as #9912 

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -2465,7 +2465,7 @@ class Guild(Hashable):
     async def fetch_members_named(self, name: str, *, limit: Optional[int] = 1000) -> List[Member]:
         """|coro|
 
-        A List[:class:`Member`] whose username or nickname starts
+        Returns a list of members whose username or nickname starts
         with the provided `name`.
 
         .. versionadded:: 2.?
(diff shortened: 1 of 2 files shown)


discord/guild.py
```diff
@@ -2462,6 +2462,37 @@ async def active_threads(self) -> List[Thread]:
...
+            Getting the members failed.
+s
+        Returns
+        ------
+        List[:class:`Member`]
+            A list of matched members.
+        """
+
+        state = self._state
+        data = await state.http.search_members(name, self.id, limit)
```
AbstractUmbra, line 2493: If your intention is to have `limit` be truly optional, so `None` means search *all* members, then you'll need to implement pagination for this endpoint, using the `after=` param.

Rapptz:
This endpoint is explicitly not exposed since it's the same as the WebSocket operation. If it was going to be exposed it'd just use the same interface/method as the current `query_members` one.

PR #10272, opened 2025-08-20
Add typing-extensions to lint workflow

diceroll123:
## Summary

Should* stop the workflow from throwing a bunch of typing errors on every PR.

\* _(did not verify, but if none show on this PR then I believe we good tbh)_ **UPDATE after actions ran:** we good 😄 

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Install dependencies
         id: install-deps
         run: |
-          python -m pip install --upgrade pip setuptools wheel ruff==0.12 requests
+          python -m pip install --upgrade pip setuptools wheel ruff==0.12 requests typing-extensions
           pip install -U -r requirements.txt
 
       - name: Setup node.js
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Install dependencies
         id: install-deps
         run: |
-          python -m pip install --upgrade pip setuptools wheel ruff==0.12 requests typing-extensions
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Install dependencies
         id: install-deps
         run: |
-          python -m pip install --upgrade pip setuptools wheel ruff==0.12 requests
+          python -m pip install --upgrade pip setuptools wheel ruff==0.12 requests typing-extensions
```
DA-344:
```suggestion
          python -m pip install --upgrade pip setuptools wheel ruff==0.12 requests "typing-extensions>=4.6,<5"
```

DA-344 (changes requested):

.github/workflows/lint.yml
AbstractUmbra: I wouldn't version pin this one personally.

diceroll123: mmmyeah if typing-extensions starts breaking things we've got much bigger problems

Rapptz: This stuff is already in the `dev` extras might as well just install that.

Rapptz: Actually scratch that, the `lint` workflow doesn't install the base lib. Either way it's pinned in the d.py repo's `dev` extras.

diceroll123: In the interest of consistency with the `[dev]` section in `pyproject.toml` I'll pin `typing_extensions>=4.3,<5` 👍 

Doesn't seem like pip is able to install _just_ the dev dependency extras without also installing the lib, rip.

PR #10273, opened 2025-08-20
Create .git-blame-ignore-revs

Sachaa-Thanasius:
## Summary

Add a git-specific file to prevent the ruff format run in 44a44e9 from filling git blame's output with non-semantic changes (both the command-line output and the in-browser GitHub Blame UI). It also establishes a (hopefully helpful) precedent for future big formatting changes.

EDIT: It could even be used for older formatting-related commits if the functionality is deemed useful enough.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Replace Black with Ruff, then format whole project.
+44a44e938fb2bd0bb085d8aa4577abeb01653ad3


Sachaa-Thanasius:
Drawback I only noticed after the fact: This does somewhat obfuscate the blame output for `lint.yml` and `pyproject.toml`, since those happened in that commit and weren't a result of running `ruff format`. But the git-blame-ignore-revs mechanism can't be made more granular, afaict.

History could be rewritten to separate the changes to those 2 files out, but that's more effort and potentially not desirable. Still, putting it out there.

AbstractUmbra:
The only issue with this is that it requires configuration locally for blame tools, and as far as I can tell isn't "automatically" used by GitHub and co. I could be wrong, but if not I'm not sure if there's any real gain.

Sachaa-Thanasius:
> ... as far as I can tell isn't "automatically" used by GitHub and co. ...

FWIW, at least on my machine and my browser, GitHub's UI automatically takes it into account and has for a while. Biggest reason I suggested this, tbh. Otherwise, fair enough.

diceroll123:
Yep, very useful on the site tbh.

https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view

PR #10284, opened 2025-08-27
Add support for 'emoji_added' message type

diceroll123:
## Summary

Adds new message type for added emojis.

The frontend has some added logic that makes it know whether or not the emoji's still there, so the `system_content` changes to just `{user.name} added a new emoji.`.

Upon adding:

image

After deleting the emoji from the server:

image

_the payload does not change_

```json
{
    "type": 63,
    "content": "<:really:1410116331352883291>",
    "mentions": [],
    "mention_roles": [],
    "attachments": [],
    "embeds": [],
    "timestamp": "2025-08-27T04:18:52.214000+00:00",
    "edited_timestamp": null,
    "flags": 0,
    "components": [],
    "id": "1410116331876913215",
    "channel_id": "228230433786036224",
    "author": ...
    "pinned": false,
    "mention_everyone": false,
    "tts": false
}
```

_Don't know which makes more sense, `guild_emoji_added` or `guild_emoji_add`, up to y'all in the end, but I went with the former._

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/flags.py
+++ b/discord/flags.py
@@ -371,6 +371,15 @@ class SystemChannelFlags(BaseFlags):
         """
         return 32
 
+    @flag_value
+    def emoji_added(self):
+        """:class:`bool`: Returns ``True`` if the system channel is used for
+        emoji added notifications.
+
+        .. versionadded:: 2.7
+        """
+        return 256
+
 
 @fill_with_flags()
 class MessageFlags(BaseFlags):


pythonmcpi:
Using `guild_emoji_add` would match with `recipient_add`, but the overall naming format isn't super consistent in that enum (for example, `guild_incident_alert_mode_enabled` being past tense).

diceroll123:
Yeah, and audit log action enums are like `emoji_create` 🤔

Rapptz:
The client uses `EMOJI_ADDED`.

Also here are the other ones missing since I was curious how we skipped almost 20.

```
  PURCHASE_NOTIFICATION: 44,
  VOICE_HANGOUT_INVITE: 45,
  POLL_RESULT: 46,
  CHANGELOG: 47,
  NITRO_NOTIFICATION: 48,
  CHANNEL_LINKED_TO_LOBBY: 49,
  GIFTING_PROMPT: 50,
  IN_GAME_MESSAGE_NUX: 51,
  GUILD_JOIN_REQUEST_ACCEPT_NOTIFICATION: 52,
  GUILD_JOIN_REQUEST_REJECT_NOTIFICATION: 53,
  GUILD_JOIN_REQUEST_WITHDRAWN_NOTIFICATION: 54,
  HD_STREAMING_UPGRADED: 55,
  CHAT_WALLPAPER_SET: 56,
  CHAT_WALLPAPER_REMOVED: 57,
  REPORT_TO_MOD_DELETED_MESSAGE: 58,
  REPORT_TO_MOD_TIMEOUT_USER: 59,
  REPORT_TO_MOD_KICK_USER: 60,
  REPORT_TO_MOD_BAN_USER: 61,
  REPORT_TO_MOD_CLOSED_REPORT: 62,
  EMOJI_ADDED: 63,
```

diceroll123:
Could do those here too, but it'd take a bit more effort if we want to get the system_content and such all correct. How many of those are even live?

Rapptz:
You don't have to add them, it's purely for my own curiosity.

Rapptz:
I think while you're here you should add the system message flag as well.

PR #10287, opened 2025-08-29
Redo ratelimit handling

imptype:
## Summary

The old handler hits 429s for concurrent requests and unknown subratelimits. This new handler prioritizes a lower remaining value, a sooner reset_at value, and accounts if the token window has changed during a sleep so it still maintains the same throughput the old handler had.

Test code:

```py
import time
import asyncio
import discord

bot = commands.Bot(intents = intents, command_prefix = '.')

# Test 1: Send 200 messages sequentially in the same channel
@bot.command()
async def test1(ctx):
    start = time.time()
    for i in range(200):
        await ctx.send(f'hi {i}')
    print('Finished test1 in:', time.time() - start)

# Test 2: Send 50 messages concurrently in the same channel
@bot.command()
async def test2(ctx):
    start = time.time()
    await asyncio.gather(*[
        ctx.send(f'hi {i}') for i in range(50)
    ])
    print('Finished test2 in:', time.time() - start)

bot.run('TOKEN')
```

Old results:
Test 1: Finished in 213.27s with 0 429s.
Test 2: Finished in 46.29s with 86 429s.

New results:
Test 1: Finished in 208.48s with 0 429s.
Test 2: Finished in 47.15 with 0 429s.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -394,7 +394,7 @@ class Ratelimit:
     def update(self, response: aiohttp.ClientResponse, data: Union[Dict[str, Any], str]) -> bool:
         limit = int(response.headers['X-Ratelimit-Limit'])
         reset_at = self.http.loop.time() + (float(response.headers['X-Ratelimit-Reset']) - time.time())
-        
+
         # Shared scope (which only appears on 429s) could have a longer "reset_at".
         if response.headers.get('X-RateLimit-Scope') == 'shared':
             reset_at = max(reset_at, self.http.loop.time() + data['retry_after'])  # type: ignore


Rapptz:
Thanks for the PR.

I admit though this is a bit of a thinker for me. When I originally wrote the rate limit handling class this was more or less the test that me and other contributors ran to ensure that it worked fine. It seems at some point one of the changes made it so 429s slip because this would not 429 before, especially that much.

I don't have too much time to review this changeset (or even test it) but I don't really like the use of `Event` here since the original was made with flattening of `asyncio` resources in mind. It also seems weird to me to handle global rate limits within that class instead of outside since the global rate limit affects all requests. It also feels incredibly brittle to me that almost all of this code is essentially just retrying things multiple times just in case something concurrently happened in the background instead of properly synchronizing it.

Anyway this is failing CI, and I don't think I can test this code anyway.

discord/http.py
```diff
@@ -343,165 +343,179 @@ def major_parameters(self) -> str:
...
 
-
 class Ratelimit:
     """Represents a Discord rate limit.
 
     This is similar to a semaphore except tailored to Discord's rate limits. This is aware of
-    the expiry of a token window, along with the number of tokens available. The goal of this
-    design is to increase throughput of requests being sent concurrently rather than forcing
-    everything into a single lock queue per route.
+    the expiry of a token window, along with the number of tokens available.
```
HEROgold: What is the reason for changing the existing docstring?
Now the explanation for how it works is lost.

discord/http.py
```diff
@@ -343,165 +343,179 @@ def major_parameters(self) -> str:
...
+            f'<RateLimitBucket limit={self.limit} remaining={self.remaining} pending={self.pending}>'
         )
 
-    def reset(self):
+    def no_headers(self) -> None:
+        self.one_shot = True
+        self._event.set()
+        self._event.clear()
+
+    def reset(self) -> None:
```
HEROgold, line 390:
```suggestion
    def reset(self) -> bool:
```
return on line 430 is a boolean.

discord/http.py
```diff
@@ -343,165 +343,179 @@ def major_parameters(self) -> str:
...
-
-        reset_after = headers.get('X-Ratelimit-Reset-After')
-        if use_clock or not reset_after:
-            utc = datetime.timezone.utc
-            now = datetime.datetime.now(utc)
-            reset = datetime.datetime.fromtimestamp(float(headers['X-Ratelimit-Reset']), utc)
-            self.reset_after = (reset - now).total_seconds()
+            
+            # Consider a lower remaining value because updates can be out of order, so self.outgoing is used
+            reset_at = self.http.loop.time() + (float(response.headers['X-Ratelimit-Reset']) - time.time())
```
HEROgold: Forgot to import time?

discord/http.py
```diff
@@ -343,165 +343,179 @@ def major_parameters(self) -> str:
...
+        # 3. Always use the lowest remaining value
+        if remaining < self.remaining:
+            self.remaining = remaining
+            update = True
+ 
+        self.limit = limit
+        self.one_shot = False
+
+        # Return whether this update was relevant or not.
+        return update
```
HEROgold, line 429: is this return required? the function signature still has `-> None`

discord/http.py
```diff
@@ -654,116 +665,78 @@ async def request(
...
+                                    elif route_key not in self._bucket_hashes:
+                                        fmt = '%s has found its initial rate limit bucket hash (%s).'
+                                        _log.debug(fmt, route_key, new_bucket_hash)
+                                    self._bucket_hashes[route_key] = new_bucket_hash
+                                    ratelimit.key = new_bucket_hash + route.major_parameters
+                                    self._buckets[ratelimit.key] = ratelimit
+
+                        # Global rate limit 429 wont have ratelimit headers (also can't tell if it's one-shot)
+                        elif response.headers.get('X-RateLimit-Global'):
+                            retry_after: float = float(response.headers['Retry-After'])
```
HEROgold: I suggest to rename `rety_after` to fix the check failures.

discord/http.py
```diff
@@ -654,116 +665,78 @@ async def request(
...
                             _log.debug('%s %s has received %s', method, url, data)
                             return data
 
-                        # we are being rate limited
-                        if response.status == 429:
-                            if not response.headers.get('Via') or isinstance(data, str):
-                                # Banned by Cloudflare more than likely.
+                        # We are being ratelimited
+                        elif response.status == 429:
+                            retry_after: float = float(response.headers['Retry-After']) # only in headers for cf ban
```
HEROgold: (same as line 691)
I suggest to rename `rety_after` to fix the check failures.

imptype:
> Thanks for the PR.
> 
> I admit though this is a bit of a thinker for me. When I originally wrote the rate limit handling class this was more or less the test that me and other contributors ran to ensure that it worked fine. It seems at some point one of the changes made it so 429s slip because this would not 429 before, especially that much.
> 
> I don't have too much time to review this changeset (or even test it) but I don't really like the use of `Event` here since the original was made with flattening of `asyncio` resources in mind. It also seems weird to me to handle global rate limits within that class instead of outside since the global rate limit affects all requests. It also feels incredibly brittle to me that almost all of this code is essentially just retrying things multiple times just in case something concurrently happened in the background instead of properly synchronizing it.
> 
> Anyway this is failing CI, and I don't think I can test this code anyway.

24b619a5 still hits 429s for test 2. `asyncio.Event` will never cause a deadlock because it's dismissed by `asyncio.wait` after 3 seconds, and personally I think `lock + event` is simpler than `lock + future`. The global wait is inside the class in case the global ratelimit was hit during the wait to acquire, so this avoids more global 429s. During a sleep, the window can change due to requests behind being late, then it must check if there is no remaining left in the new window and if so sleep again - the simplest way is to sleep and check than make it event based or something; it still works fine. The use case for concurrency is more frequent and relevant than you realize, `asyncio.gather(50)` in the same channel is also the same as 50 people running a ping command in the same channel at the same time, which currently will hit 86 429s.

The remaining CI errors are typehinting or formatting issues which I can't fix.

Rapptz:
> 24b619a still hits 429s for test 2. 

I remember now that the old rate limit, on Discord's side, got rewritten about 2-3 years ago to use GRCA instead of the previous windowed rate limit and that's probably where the discrepancy is. I know for a fact we tested this for weeks so something definitely changed.

> asyncio.Event will never cause a deadlock because it's dismissed by asyncio.wait after 3 seconds, and personally I think lock + event is simpler than lock + future. 

The original code used a deque of futures because it was slightly more resource efficient since this is a low level part of the library. In `asyncio` both `Lock` and `Event` are just interfaces over a deque of futures and a boolean with no control over the the wakers compared to a semaphore. I don't really get the confusion of a `Future` here.

> The global wait is inside the class in case the global ratelimit was hit during the wait to acquire, so this avoids more global 429s.

I guess this is fair.

> During a sleep, the window can change due to requests behind being late, then it must check if there is no remaining left in the new window and if so sleep again - the simplest way is to sleep and check than make it event based or something; it still works fine.

I'm not convinced about this yet. Sleeping and checking whether your value changed under you is just brittle to me.

> The use case for concurrency is more frequent and relevant than you realize, asyncio.gather(50) in the same channel is also the same as 50 people running a ping command in the same channel at the same time, which currently will hit 86 429s.

This is literally not at all what I said. I mentioned earlier we tested the `gather` use case, why do you assume I don't know it's importance? I'm mentioning the concurrency handling within your implementation using `sleep` is poor.

> The remaining CI errors are typehinting or formatting issues which I can't fix.

You can fix your CI failure by running `ruff format` on the `discord` directory.

imptype:
@Rapptz According to Discord's public changelog, the last update to ratelimit system was [5 years ago](https://discord.com/developers/docs/change-log#api-and-gateway-v8), before the current handler was written. Btw, I was referring to code readability. But even then, doesn't the old handler make multiple futures whereas this only has 1 in the deque at a time because of the lock? Also, there's no simpler way to sleep longer after a sleep was finished other than using `asyncio.sleep` and checking to sleep again.

Rapptz:
> According to Discord's public changelog, the last update to ratelimit system was 5 years ago, before the current handler was written.

I don't get why you're doubting me on this when I speak to employees on a regular basis. The changelog is inaccurate. It doesn't even mention the `X-RateLimit-Scope` header addition, nor does it mention the change to GCRA back in late 2022, which took until late 2023 for it to reach all endpoints to my knowledge.

DiscordCanary_5b0eodrdic

imptype:
@Rapptz Ok I didn't have insider info. What will it take to merge?

discord/http.py
```diff
@@ -349,159 +349,167 @@ class Ratelimit:
...
+    async def _wait_global(self, start_time: float):
+        # Sleep up to 3 times, to account for global reset at overwriting during sleeps
+        for i in range(3):
+            seconds = self.http.global_reset_at - start_time
+            if seconds > 0:
+                await asyncio.sleep(seconds)
+                continue
+            break
+        else:
+            raise ValueError('Global reset at changed more than 3 times')
```
Rapptz: Due to the way the global rate limit has been rewritten it is no longer as possible to refactor this out into a customisable lock for multi-process set ups. It's always been part of the bucket list in order to implement this. If this changeset was communicated earlier in the Discord I would have also told you about this.

discord/http.py
```diff
@@ -349,159 +349,167 @@ class Ratelimit:
...
+                            fmt = 'A rate limit bucket (%s) has been exhausted. Sleeping for %.2f seconds.'
+                        _log.warning(fmt, self.key, seconds)
+                        await asyncio.sleep(seconds)
+                    if copy == self.reset_at:
+                        self.reset()
+                    elif not self.remaining and not self.one_shot:
+                        continue  # sleep again
+                    break
+                else:
+                    raise ValueError('Reset at changed more than 3 times')
```
Rapptz: I don't get why this is a failure condition that has to bubble out. Ditto for the rest of them.

discord/http.py
```diff
@@ -774,19 +739,17 @@ async def request(
...
-                    raise DiscordServerError(response, data)
-
-                raise HTTPException(response, data)
-
+                    if tries == 4 or e.errno not in (54, 10054):
+                        raise ValueError('Connection reset by peer')
+                    retry_seconds: int = 1 + tries * 2
+                    fmt = 'OS error for %s %s. Retrying in %d seconds.'
+                    _log.warning(fmt, method, url, retry_seconds)
+                    await asyncio.sleep(retry_seconds)
```
Rapptz: Why'd you swap this?

discord/http.py
```diff
@@ -774,19 +739,17 @@ async def request(
...
+
+        # We've run out of retries, raise
+        if response is not None:
+            raise HTTPException(response, data)
```
Rapptz: Why'd you remove the 500 raise?

discord/http.py
```diff
@@ -349,159 +349,167 @@ class Ratelimit:
...
+                    copy = self.reset_at
+                    if seconds > 0:
+                        if i:
+                            fmt = 'A rate limit bucket (%s) has changed reset_at during sleep. Sleeping again for %.2f seconds.'
+                        else:
+                            fmt = 'A rate limit bucket (%s) has been exhausted. Sleeping for %.2f seconds.'
+                        _log.warning(fmt, self.key, seconds)
+                        await asyncio.sleep(seconds)
+                    if copy == self.reset_at:
+                        self.reset()
```
Rapptz, line 479: What is this for?

discord/http.py
```diff
@@ -349,159 +349,167 @@ class Ratelimit:
...
+        start_time: float = self.http.loop.time()
+
+        # Resources confirmed to be "one shots" after the first request like interaction response
+        # don't have ratelimits and can skip the entire queue logic except global wait
+        if self.one_shot:
+            await self._wait_global(start_time)
+        else:
+            # Ensure only 1 request goes through the inner acquire logic at a time
+            self.pending += 1
+            async with self._lock:
```
Rapptz, line 496: Unless I'm missing something this is just going back to the very old approach of locking every single request. I also don't get why we're both waiting for an Event (which is a `Lock`) and a `Lock`. In the previous code the `Lock` was only acquired when it was time to sleep which meant that until it was hit the remaining requests were free to just fire off regardless, with the expectation that the library would keep its semaphore updated (which, unfortunately, ended up having a bug at some point so this was no longer met).

discord/http.py
```diff
@@ -654,116 +659,76 @@ async def request(
...
-                                    # If the previous hash was an actual Discord hash then this means the
-                                    # hash has changed sporadically.
-                                    # This can be due to two reasons
-                                    # 1. It's a sub-ratelimit which is hard to handle
-                                    # 2. The rate limit information genuinely changed
-                                    # There is no good way to discern these, Discord doesn't provide a way to do so.
-                                    # At best, there will be some form of logging to help catch it.
-                                    # Alternating sub-ratelimits means that the requests oscillate between
-                                    # different underlying rate limits -- this can lead to unexpected 429s
-                                    # It is unavoidable.
```
Rapptz: Not that it matters that much but why did you remove my comment?

discord/http.py
```diff
@@ -654,116 +659,76 @@ async def request(
...
+                                        self._buckets.pop(key, None)
+                                    elif route_key not in self._bucket_hashes:
+                                        fmt = '%s has found its initial rate limit bucket hash (%s).'
+                                        _log.debug(fmt, route_key, new_bucket_hash)
+                                    self._bucket_hashes[route_key] = new_bucket_hash
+                                    ratelimit.key = new_bucket_hash + route.major_parameters
+                                    self._buckets[ratelimit.key] = ratelimit
+
+                        # Global rate limit 429 wont have ratelimit headers (also can't tell if it's one-shot)
+                        elif response.headers.get('X-RateLimit-Global'):
```
Rapptz, line 687: Why are we using this instead of the `global` inner key which to my knowledge is more accurate and actually consistently given?

discord/http.py
imptype: Global rate limit is not a lock anymore, it's just a changeable timestamp that all requests must sleep to first before continuing. So, just replicate the value of `http.global_reset_at` to other processes when it is hit and it will still work. Btw, will need to change `loop.time()` to use `time.time()`.

discord/http.py
imptype: I expect `reset_at` to change during the sleep 3 times at most. More than that and I think the client is experiencing extreme lag or something, so the request should not continue.

discord/http.py
imptype: To avoid the `continue` and have 1 less line.

discord/http.py
imptype: I forgot to add a `tries` check to fix waiting 9 extra seconds on the last try for no reason. Unless an aiohttp error happens on the last try, it's never met.

discord/http.py
imptype: If `reset_at` hasn't changed, then we can set `remaining` = `limit`. A successful sleep is where we don't need to sleep again can refill the ratelimit remaining to the top.

If you mean the seemingly redundancy of the `copy` variable, it's needed in case of a bucket change. It will unnecessarily call `Ratelimit.reset()` when `reset_at` has changed and there are `self.remaining` left. Saving a call is good also in case you want to replicate ratelimit state to other processes too.

discord/http.py
imptype: The idea is the first 5 requests go through the lock in and out quick **without any waiting** because remaining is still available. The 6th request will go in the lock, and start waiting because none are remaining. Then the cycle continues which makes it work for concurrent requests too.

The `asyncio.Lock` outside is so 1 request goes through the waiting logic at a time. The `asyncio.Event` makes that 1 request wait for the `reset_at` to be initially set by one of the previous requests that consumed the remaining because currently `reset_at` is the value `0`.

discord/http.py
imptype: It conflicts with other comments explaining the strategy.

discord/http.py
imptype: We haven't called `json_or_text` yet, so can only use headers at this point.

discord/http.py
pythonmcpi: Code clarity matters more than line count. Also, in the case where the OSError isn't a connection reset, your code now gives the wrong error message.

discord/http.py
Rapptz: It's not possible to sync the value between processes.

discord/http.py
Rapptz: I'd rather it surface another exception either from `aiohttp` or `HTTPException` instead. `ValueError` is a terrible exception to raise for this stuff.

discord/http.py
Rapptz: If you're already synchronizing via a lock and you want to share a condition then the proper synchronisation tool is neither a `Lock` nor an `Event` but an `asyncio.Condition` instead which at least lets you better control the notify mechanism.

discord/http.py
Rapptz: Reorder the code so it is called. `Retry-After` header is in whole seconds, `retry_after` in the JSON is in float seconds.

discord/http.py
Rapptz: Change it back.

discord/http.py
Rapptz: Change it back.

Rapptz (changes requested):
I didn't tell you to remove `Client.assume_unsync_clock`.

discord/http.py
imptype: I thought processes can send each other messages and from that you can do `http.global_reset_at = new value`

discord/http.py
imptype: It's separate because `condition.wait()` releases the lock, letting multiple requests go through the inner logic time which messes things up.

imptype:
@Rapptz `unsync_clock` was unused so it was removed.

imptype:
@Rapptz A future replaced the event because you mentioned resource issues.

discord/http.py
```diff
@@ -349,159 +349,170 @@ class Ratelimit:
...
-        # Only a single rate limit object should be sleeping at a time.
-        # The object that is sleeping is ultimately responsible for freeing the semaphore
-        # for the requests currently pending.
-        self._sleeping: asyncio.Lock = asyncio.Lock()
-        self._last_request: float = self._loop.time()
+        self.one_shot: bool = False
+        self.http: HTTPClient = http
+        self._last_request: float = 0.0
+        self._lock: asyncio.Lock = asyncio.Lock()
+        self._future: Optional[asyncio.Future] = None
```
Rapptz: By moving this to a `Future` you're basically saying you expect a single waiter.

```suggestion
        self._future: Optional[asyncio.Future[None]] = None
```

discord/http.py
```diff
@@ -349,159 +349,170 @@ class Ratelimit:
...
 
-    def reset(self):
+    def reset(self) -> None:
         self.remaining = self.limit - self.outgoing
-        self.expires = None
-        self.reset_after = 0.0
-        self.dirty = False
+        self.reset_at = 0.0
+
+    def update(self, response: aiohttp.ClientResponse, data: Union[Dict[str, Any], str]) -> bool:
```
Rapptz, line 394: This should probably only accept `Dict[str, Any]` or `None` by restricting it in the calling site so you can get rid of the `# type: ignore`. I guess it'd also be good to you make a `TypedDict` with the rate limit expected response type since that's what this is accepting and narrow to that.

discord/http.py
```diff
@@ -349,159 +349,170 @@ class Ratelimit:
...
+        # Shared scope 429 has longer "reset_at", determined using the retry-after field
```
Rapptz: I don't think this is forced to be true.

A shared scope is just a shared resource like a channel message rate limit where you have a set number of messages sent per channel but that doesn't necessarily mean that it's always going to be longer though right now I can't think of any shared resources that are small in size though I do know that Emoji resources are only shared.

discord/http.py
```diff
@@ -349,159 +349,170 @@ class Ratelimit:
...
+        limit = int(response.headers['X-Ratelimit-Limit'])
+        if response.headers.get('X-RateLimit-Scope') == 'shared':
```
Rapptz, line 399: This should have a comment that this header is only returned on 429 responses.

discord/http.py
```diff
@@ -349,159 +349,170 @@ class Ratelimit:
...
+            if seconds > 0:
+                await asyncio.sleep(seconds)
+                continue
+            break
+        else:
+            raise RuntimeError('Global reset at changed more than 3 times')
+
+    async def _wait(self):
+        # Consider waiting if none is remaining
+        if not self.remaining:
```
Rapptz: I suggest using explicit checks instead of passive boolean ones for all the integer comparisons.
```suggestion
        if self.remaining == 0:
```

discord/http.py
```diff
@@ -349,159 +349,170 @@ class Ratelimit:
...
+            # If reset_at is not set yet, wait for the last request, if outgoing, to finish first
+            # for up to 3 seconds instead of using aiohttp's default 5 min timeout.
+            if not self.reset_at and (not self._last_request or self.http.loop.time() - self._last_request < 3):
```
Rapptz:
```suggestion
            if self.reset_at == 0.0 and (not self._last_request or self.http.loop.time() - self._last_request < 3):
```

discord/http.py
```diff
@@ -349,159 +349,170 @@ class Ratelimit:
...
-                    self._wake_next()
-                raise
-
-        self.remaining -= 1
-        self.outgoing += 1
+        delta = self.http.loop.time() - self._last_request
+        return delta >= 300 and (self.one_shot or (self.outgoing == 0 and self.pending == 0))
+
+    def _wake(self) -> None:
+        if self._future and not self._future.done():
```
Rapptz:
```suggestion
        if self._future is not None and not self._future.done():
```

discord/http.py
```diff
@@ -349,159 +349,170 @@ class Ratelimit:
...
-                if self.remaining > 0 and not future.cancelled():
-                    self._wake_next()
-                raise
-
-        self.remaining -= 1
-        self.outgoing += 1
+        delta = self.http.loop.time() - self._last_request
+        return delta >= 300 and (self.one_shot or (self.outgoing == 0 and self.pending == 0))
+
+    def _wake(self) -> None:
```
Rapptz, line 435: The first request will always have this future set to `None`, not sure if this is what you intended.

discord/http.py
```diff
@@ -349,159 +349,170 @@ class Ratelimit:
...
+            # If reset_at is not set yet, wait for the last request, if outgoing, to finish first
+            # for up to 3 seconds instead of using aiohttp's default 5 min timeout.
+            if not self.reset_at and (not self._last_request or self.http.loop.time() - self._last_request < 3):
+                try:
+                    self._future = self.http.loop.create_future()
+                    await asyncio.wait_for(self._future, 3)
+                except asyncio.TimeoutError:
+                    fmt = 'Initial request for rate limit bucket (%s) never finished. Skipping.'
+                    _log.warning(fmt, self.key)
+                    self._future.cancel()  # type: ignore
```
Rapptz:
```suggestion
                self._future = future = self.http.loop.create_future()
                try:
                    await asyncio.wait_for(future, 3)
                except asyncio.TimeoutError:
                    fmt = 'Initial request for rate limit bucket (%s) never finished. Skipping.'
                    _log.warning(fmt, self.key)
                    future.cancel()
```

Rapptz (changes requested):
Previous comments still apply.

1. I did not ask for `Client.assume_unsync_clock` to be removed.
2. I still think bubbling this error as a `RuntimeError`, or even at all, is not good design.

I raise `RuntimeError` under there because it's an *unreachable* exception. It is not meant to ever get hit and if it does then it's bizarre that it happened -- it's purely there to shut up the type checker and not publicly facing.

Anyway, I will be going on a trip for the month of September so this is my last review of this PR until I get home and I don't really have time to do an even deeper review right now since I'm preoccupied preparing for said trip.

discord/http.py
imptype: I tried to fix the typehint issue but that method adds extra calls, so `type: ignore` is the simplest fix.

imptype:
@Rapptz RuntimeError can be avoided with a `while true`, but that may be preferred to prevent an infinite loop.

PR #10290, opened 2025-09-01
Fix guild template available_tags errors

osego:
## Summary
Fixes two related errors from `Guild.templates()` both caused by inconsistencies in channel `available_tags`:
 - TypeError when `available_tags` is null (happens for all non-forum channels and forum channels which don't have any tags)
 - KeyError when tags in `available_tags` don't have an 'id' key (happens in the opposite case from the above)

Relevant discord api docs issue: discord/discord-api-docs/issues/7798

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/channel.py
+++ b/discord/channel.py
@@ -2330,7 +2330,7 @@ class ForumTag(Hashable):
     def from_data(cls, *, state: ConnectionState, data: ForumTagPayload) -> Self:
         self = cls.__new__(cls)
         self.name = data['name']
-        self.id = int(data['id'])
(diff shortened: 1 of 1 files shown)


Rapptz:
Generally speaking I don't fix upstream issues unless they're not fixed so I'm holding off on merging this PR until then.

osego:
Closing this because both problems with `available_tags` are now fixed in Discord.

Templates continue to be broken from role `colors` being null. This can can be worked around with a similar looking patch but should also just be fixed upstream instead of here.

PR #10295, opened 2025-09-03
Add `__all__` to `primary_guild` module

diceroll123:
prevents star-all exports like `discord.TYPE_CHECKING` etc


Rapptz:
We use format on/off blocks when it's a single element to minimize the diff impact. It's also easy to maintain even if you leave the format block off after adding an element (which would probably never happen here).

diceroll123:
thought so, just didn't verify tbqh

PR #10306, opened 2025-09-16
fix(commands): discord.Attachment optional arg

martinbndr:
## Summary

This pull request fixes an issue with the discord.Attachment argument raising the MissingRequiredArgument when that argument is actually set as None (optional).

Example:

arg: discord.Attachment = None

Instead it was only possible to bypass the issue it with typing.Optional.
## Checklist

- [x] This PR fixes an issue.


martinbndr:
Nvm, see discussion at https://discord.com/channels/336642139381301249/1417437029553405982/1417636858468237414. It was previously denied so this gets closed.

Issue #10313, opened 2025-09-21
Webhook.send() doesn't accept thread=None


LeviPesin:
### Summary

Webhook.send() doesn't accept thread=None

### Reproduction Steps

`thread=None` cannot be used in Webhook.send() method. It is useful when I want to send a message to a channel/thread (code doesn't know which one).

### Minimal Reproducible Code

```python
dst = some_channel
webhook = await (dst if not isinstance(dst, Thread) else dst.parent).create_webhook(name='webhook')
await webhook.send(content='something', thread=None if not isinstance(dst, Thread) else dst)
```

### Expected Results

Works.

### Actual Results

```
  File "/data/project/rv/pyvenv/lib/python3.13/site-packages/discord/webhook/async_.py", line 1898, in send
    thread_id = thread.id
                ^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'id'
```

### Intents

N/A

### System Information

N/A

### Checklist

- [x] I have searched the open issues for duplicates.
- [x] I have shown the entire traceback, if possible.
- [x] I have removed my token from display, if visible.

AbstractUmbra:
This method existed before the library was fully typed, therefor it uses the special `MISSING` sentinel in place of None.

If the kwarg is passed, it expects a Snowflake that represented a thread. You'd need to amend this code:-
```py
thread = ...
if isinstance(thread, discord.Thread):
    await send(..., thread=thread)
else:
    await send(...)
```

Passing `None` here is invalid, and changing this now would be a breaking change.

LeviPesin:
Sorry, I made a PR before seeing your reply.

> changing this now would be a breaking change.

Would it be really that much of it? Currently passing anything else than Thread and MISSING already results in an error, so that shouldn't really had been used in real code.

AbstractUmbra:
One could argue that since Webhook doesn't inherit Messageable that this behaviour would be isolated to Webhook and changing it isn't breaking. 
The issue then becomes - would this affect the application commands infrastructure (which are just webhooks realistically).
 
I personally don't know and can't imagine so. We'd need input from @Rapptz on this. After my quick investigation this fix would not be regarded as breaking in terms of our version guarantees.

Rapptz:
MISSING is a public and exported symbol that you can use for this purpose already. Alternatively use kwargs and expand them manually. Or use an if statement with two branches. There are more options than just passing None currently. 

[...]

PR #10314, opened 2025-09-21
Allow thread=None in Webhook.send()

LeviPesin:
Fixes https://github.com/Rapptz/discord.py/issues/10313

## Summary

Allow thread=None in Webhook.send() by checking `thread` instead of `thread is MISSING`.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/webhook/async_.py
+++ b/discord/webhook/async_.py
@@ -1894,7 +1894,7 @@ class Webhook(BaseWebhook):
         ) as params:
             adapter = async_context.get()
             thread_id: Optional[int] = None
-            if thread is not MISSING:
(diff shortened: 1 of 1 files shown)


Rapptz:
Implicit checks like these are bad so I'm not keen on merging this as-is. See also my comment on the issue.

LeviPesin:
I can change them to `thread is not MISSING and thread is not None` if you like.

LeviPesin:
@Rapptz Should I change the checks in this way? Also, should I also add support for file=None, as it is the only argument except thread which doesn't support it?

PR #10321, opened 2025-09-29
Add command_id and custom_id attributes to Interaction

Soheab:
## Summary

This PR adds the following attributes to `Interaction` for ease of access and type safety for people that want it:
- `command_id`
- `custom_id`

Both are from `interaction.data` and self-explanatory.

They are cached properties so they don’t exist unless used, and when they do, their values stay the same.

## Checklist

- [ ] This PR fixes an issue.


Rapptz:
I mentioned before when you made this PR that I'm not fully on board with these helper properties but I guess the scope of it is small enough that it doesn't matter.

PR #10328, opened 2025-10-09
fix: do not assume Python 3.14 has compression.zstd

beer-psi:
## Summary

It turns out the new `compression.zstd` module in 3.14 is optional.

References: https://github.com/aio-libs/aiohttp/issues/11603 https://github.com/python/cpython/issues/139707

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -87,7 +87,7 @@ try:
     _HAS_ZSTD = True
 except ImportError:
     try:
-        from compression.zstd import ZstdDecompressor
+        from compression.zstd import ZstdDecompressor  # type: ignore
     except ImportError:
         import zlib
 


emmatyping:
While it is optional, it should be available on pretty much every platform except custom builds on old old LTS Linux distros which are paid support only and WASI/Emscripten. If you don't want to support those platforms I'd argue you don't need to treat it as optional.

beer-psi:
A library user was already affected by this due to aiohttp making the same assumption. zstd on Python is new and I wouldn't want to make the assumption that everyone has it when the PEP explicitly says "libzstd is optional".

https://github.com/Rapptz/discord.py/discussions/10327#discussioncomment-14629968

emmatyping:
Hm, gotcha, makes sense then!

PR #10343, opened 2025-11-03
ui: add insert_item_at method

laggron42:
## Summary

This adds `insert_item_at` for all classes which have a `add_item` method: `BaseView`, `View`, `LayoutView`, `Container` and `ActionRow`. The method follows the prototype of [`list.insert`](https://docs.python.org/3/tutorial/datastructures.html#more-on-lists) and allows placing items at any position, instead of defaulting to the end of the children list.

This includes some code duplication, please signal to me if you'd like some refactoring to avoid this. I prefer submitting something safe and working first, then comply with coding style.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ui/view.py
+++ b/discord/ui/view.py
@@ -885,6 +885,12 @@ class LayoutView(BaseView):
         super().add_item(item)
         return self
 
+    def insert_item_at(self, position: int, item: Item[Any]) -> Self:
+        if self._total_children >= 40:
+            raise ValueError('maximum number of children exceeded (40)')
+        super().insert_item_at(position, item)
+        return self
+
     def content_length(self) -> int:
         """:class:`int`: Returns the total length of all text content in the view's items.
 


pythonmcpi:
This feature has already been rejected by Danny. You can read the conversation in the discord.py server starting from [here](https://discord.com/channels/336642139381301249/1345167602304946206/1423072435091144806). Quoting a few specific messages:

> When I was testing the original PR and coming up with examples I also thought that "inserting items at a specific position" would be more common than it would be but... it just wasn't
> Most layout is incredibly static, and if it's dynamic then it's dynamic to a degree where the easiest thing to do is to recreate it. When you recreate a view you're better off holding off references to the objects and just re-building it
> It's weird, and straight up rare, to just add something in between two items because the flow of the layout just changes completely
> In fact recreating on edit just leads to cleaner code half the time anyway than doing silly tricks to modify it

PR #10345, opened 2025-11-05
Add more attributes to AppCommandChannel

Sacul0457:
## Summary
This PR (once again) adds more attributes to `AppCommandChannel`

Rational behind some of the decisions made:
- Return `None` if the channel doesn't have a certain field. This avoids implying a channel has the feature/field when it actually doesn't. For instance, `slowmode_delay` with `CategoryChannel`, if we return 0, it implies that it has a slowmode of 0s, when it actually doesn't have that feature.
  - **Breaking Change:** With that said, this also means `slowmode_delay` is now `Optional[int]` which makes it a breaking change. If this isn't a good idea, i'll revert it!
  

## Checklist

- [ ] This PR fixes an issue.


Sacul0457:
Draft until there's more updates on discord/discord-api-docs#7957

PR #10346, opened 2025-11-08
Make description optional & fix emoji display in Guild.create_sticker

n6ck:
## Summary

- Made the `description` parameter optional and default to an empty string to match Discord's API behavior
- Removed the Unicode name conversion which causes display issues on the client UI, as Discord doesn't actually validate this input at all, and expects a raw Unicode emoji to render properly

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -3104,7 +3104,8 @@ class Guild(Hashable):
         name: :class:`str`
             The sticker name. Must be between 2 and 30 characters.
         description: :class:`str`
-            The sticker's description. Must be between 2 and 100 characters.
+            The sticker's description. Can be an empty string or a string between 2 and 100 characters.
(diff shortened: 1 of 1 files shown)


Rapptz:
Is there any evidence that setting the `description` or `tags` causes issues

n6ck:
> Is there any evidence that setting the `description` or `tags` causes issues

`description` doesn't cause any issues however Discord's API allows it to be empty, and it is displayed as optional in the UI

As for `tags`, they're intended for autocomplete, but currently, passing a raw Unicode emoji can break how the emoji renders in Discord's UI
e.g. when passing `emoji="\N{LIGHT BLUE HEART}"` it causes the first image, and with this PR, the second

Essentially just QOL changes

imawdadawdge
imawdawdaage
imaawdwadage

PR #10347, opened 2025-11-11
Change `channel` to `destination` in Messageable to include a `Member`

IsaacG:
## Summary

The `Messageable` class is inherited by `Member` as well as by channels. The target can be either a `Member` or a channel. As such, references to the target `channel` should instead read `destination`.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -1653,7 +1653,7 @@ class Messageable:
             The message that was sent.
         """
 
-        channel = await self._get_channel()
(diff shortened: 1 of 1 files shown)


Rapptz:
Thanks but the current wording is fine. It's a channel most of the time. I don't think changing it to `destination` really adds any clarity since no one would be confused on what's happening here.

PR #10351, opened 2025-11-17
Optimise utils.find and specialise utils.as_chunks

Sacul0457:
## Summary
This PR attempts to optimise `utils.find` and `utils.as_chunks`.

You can read more [here](https://discord.com/channels/336642139381301249/1437445792830062763/1438541178776850634), but just to summarise:
- Most workloads are small when it comes to `utils.find`, but it is still used a lot in dpy and user code, so speeding it up will still help overall if you know what I mean. Benchmarks on all versions (3.8 - 3.15) are around `2% - 20%` faster

Also, for anyone seeing this, feel free to benchmark these changes and see if it matches. Benchmarks can be found  [here](https://discord.com/channels/336642139381301249/1437445792830062763/1438541178776850634). It is greatly appreciated, thanks!
Note: `utlils.as_chunks` implementation changed, so use the new one in this PR when benchmarking.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -57,6 +57,7 @@ from typing import (
 )
 import unicodedata
 import collections.abc
+from itertools import islice
 from base64 import b64encode, b64decode
 from bisect import bisect_left
 import datetime
@@ -1041,20 +1042,15 @@ def _chunk(iterator: Iterable[T], max_size: int) -> Iterator[List[T]]:
     # Specialise iterators that can be sliced as it is much faster
     if isinstance(iterator, collections.abc.Sequence):
         for i in range(0, len(iterator), max_size):
-            yield list(iterator[i : max_size + i])
+            yield list(iterator[i : i + max_size])
     else:
-        # Fallback to slow path
-        ret = []
-        n = 0
-        for item in iterator:
-            ret.append(item)
-            n += 1
-            if n == max_size:
-                yield ret
-                ret = []
-                n = 0
-        if ret:
-            yield ret
+        # Fallback to slower path
+        iterator = iter(iterator)
+        while True:
+            batch = list(islice(iterator, max_size))
+            if not batch:
+                break
+            yield batch
 
 
 async def _achunk(iterator: AsyncIterable[T], max_size: int) -> AsyncIterator[List[T]]:


Rapptz:
Instead of doing `hasattr` you can try `getattr(x, attr, None) is not None`. Usually this is insufficient for checking since `None` can be valid but for both of the attributes we're checking it breaks the invariant if it's `None` so it might be worth doing. Chances are ultimately the attribute checking would be slower though.

Sacul0457:
> Chances are ultimately the attribute checking would be slower though.

Yeah, unfortunately it is. I think sticking with `isinstance(iterator, collections.abc.Sequence)` is fine. On smaller BMs it's still around 60% faster and it's more readable too.

mikeshardmind:
don't have a quiet local machine to benchmark on right now, but I suspect using itertools.islice will be better for as_chunks:

```py
import sys
from itertools import islice


def _chunk(iterable: Iterable[T], max_size: int) -> Iterable[list[T]]:
    iterator = iter(iterable)
    while batch := list(islice(iterator, n)):
        yield batch
```

Sacul0457:
> don't have a quiet local machine to benchmark on right now, but I suspect using itertools.islice will be better for as_chunks:
> 
> ```python
> import sys
> from itertools import islice
> 
> 
> def _chunk(iterable: Iterable[T], max_size: int) -> Iterable[list[T]]:
>     iterator = iter(iterable)
>     while batch := list(islice(iterator, n)):
>         yield batch
> ```

Yes, it does seem like using `itertools.islice` is faster than the current slow path. However, it still seems that normal slicing is *30% - 50%* faster than `itertools.islice`. So, I think we can combine your idea, and replace the current slow path with `itertools.islice`, like so:
```py
def _chunk(iterator: Iterable[T], max_size: int) -> Iterator[List[T]]:
    # Specialise iterables/iterators that can be sliced which is much faster
    if isinstance(iterator, Sequence):
        for i in range(0, len(iterator), max_size):
            yield list(iterator[i : max_size + i])
    else:
        # Fallback to slower path
        iterator = iter(iterator)
        while batch := list(islice(iterator, max_size)):
            yield batch
```
This newer version is around `84% - 88%` faster on the fast path, while on slow path, it is `66% - 78%` faster!

Rapptz:
FYI walrus operator is banned in this codebase.

Sacul0457:
oh alr, this should be the equivalent I believe? I don't use the walrus operator often
```py
iterator = iter(iterator)
while True:
    batch = list(islice(iterator, max_size))
    if not batch:
        break
    yield batch
```

Rapptz:
Yeah that works too.

PR #10353, opened 2025-11-22
Add `Message.is_forwardable` to check if a message can be forwarded

n6ck:
## Summary

Adds a new method `Message.is_forwardable` to check if a message can be forwarded with `Message.forward`
Helps to avoid failed forwarding attempts by providing a simple boolean check
Inspired by `Message.is_system`

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -3070,13 +3070,13 @@ class Message(PartialMessage, Hashable):
         ):
             return False
 
-        if getattr(self, 'poll', None) is not None:
+        if self.poll is not None:
             return False
 
-        if getattr(self, 'call', None) is not None:
+        if self.call is not None:
             return False
 
-        if getattr(self, 'activity', None) is not None:
+        if self.activity is not None:
             return False
 
         return True


pythonmcpi:
Is there an upstream docs page listing the criteria of whether a message is forwardable?

Sacul0457:
given this is accepted, I think this should be in MessageType, not in Message. Similar to [`MessageType.is_deletable`](https://discordpy.readthedocs.io/en/latest/api.html?highlight=deletable#discord.MessageType.is_deletable), this avoids implying that the message can be forwarded anywhere, regardless of permission.

But at the same time, you can't check for polls/activity etc, so...

~~Also, the attributes should never be missing afaik, you can just do `if self.poll is not None` instead of the `getattr`~~

n6ck:
> Is there an upstream docs page listing the criteria of whether a message is forwardable?

https://discord.com/developers/docs/resources/message#message-reference-types

n6ck:
> given this is accepted, I think this should be in MessageType, not in Message. Similar to [`MessageType.is_deletable`](https://discordpy.readthedocs.io/en/latest/api.html?highlight=deletable#discord.MessageType.is_deletable), this avoids implying that the message can be forwarded anywhere, regardless of permission.
> 
> But at the same time, you can't check for polls/activity etc, so...
> 
> Also, the attributes should never be missing afaik, you can just do `if self.poll is not None` instead of the `getattr`

Good idea
As for getattr, self.call can be missing as I did come across that when testing for this, which is why I put that
(not even sure what self.call is tbh)

Rapptz:
Sorry for the late PR review.

It seems mostly fine, but please use regular attribute checking instead of `getattr` here. The `self.call` issue was a bug (#10404) that got fixed earlier.

n6ck:
thanks, updated accordingly

PR #10363, opened 2025-12-21
Remove PublicUserFlag.active_developer

martinbndr:
## Summary

Remove PublicUserFlag.active_developer because discord has discontinued that flag/badge. Refrerence https://support-dev.discord.com/hc/en-us/articles/10113997751447-Active-Developer-Badge

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/enums.py
+++ b/discord/enums.py
@@ -574,7 +574,6 @@ class UserFlags(Enum):
     discord_certified_moderator = 262144
     bot_http_interactions = 524288
     spammer = 1048576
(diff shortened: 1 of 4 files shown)


Rapptz:
Since this is a breaking change it won't be merged. The risk of keeping this is that it basically is just always `False`, so it's not a big deal.

martinbndr:
What about an deprecation notice in the docs instead?

PR #10372, opened 2026-01-01
Add generics to Interaction params

diceroll123:
## Summary

ty gives this,

image

Allows `interaction.client` to narrow to the bot class instead of the basic discord.Client.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -1802,7 +1802,7 @@ class Group:
                 yield from command.walk_commands()
 
     @mark_overrideable
-    async def on_error(self, interaction: Interaction, error: AppCommandError, /) -> None:
+    async def on_error(self, interaction: Interaction[ClientT], error: AppCommandError, /) -> None:
         """|coro|
 
         A callback that is called when a child's command raises an :exc:`AppCommandError`.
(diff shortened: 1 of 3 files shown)


The-Red-Dragons:
👍 Great fix! This addresses a common pain point for type-safe Discord bot development.

We maintain a large production Discord bot and frequently encounter this issue when subclassing callbacks. Currently, we have to use type ignores or casts in views, modals, and error handlers:

```python
class MyView(discord.ui.View):
    async def on_error(self, interaction: discord.Interaction, error: Exception, item):
        # Currently required workaround:
        bot: MyBot = interaction.client  # type: ignore[assignment]
        await bot.error_manager.handle(interaction, error)
 ```

With this PR, the generic parameter will properly narrow interaction.client to our custom bot class, eliminating the need for type ignores.

One minor suggestion: Is Modal.on_error() also affected by this issue? I noticed it's not included in the current changeset - might be worth checking if it needs the same treatment.

Thanks for the fix! 🎉

**Comment by diceroll123** [comment]
`Modal.on_error` seems fine, https://github.com/Rapptz/discord.py/blob/005287898393bd13d21e077e8607d5226757820a/discord/ui/modal.py#L159

**Review by scutuatua-crypto — APPROVED** [review]

**Comment by TheServerit** [comment]
Though Pyright on Strict mode complains even in places where the base function _already does_ implement `interaction: Interaction[ClientT]`:
```cmd
"Interaction[ClientT@callback]" is not assignable to "Interaction[Bot]"
  Type parameter "ClientT@Interaction" is covariant, but "ClientT@callback" is not a subtype of "Bot"
    "Client*" is not assignable to "Bot"
```
(`Bot` is my subclass of `AutoShardedBot` which subclasses `Client`)
So unless I misunderstand something, I assume this PR will not fix this issue in particular?

PR #10377, opened 2026-01-04
[tasks] reset exponential backoff after task runs successfully

blume0:
## Summary

Fix unexpected behavior (?) of tasks

The current behaviour of the _loop method (when the reconnect attribute is set to True) is to have the same ExponentialBackoff instance for all of its lifetime. A task that has failed n (when n is less than 11) times over its life-time will thus wait 2**n seconds before retrying, no matter how many successful instances of happened between the failures.

With this commit, the behaviour is to reset the exponential backoff after a succesful execution of the task, so that the backoff only takes place after successive failures.

## Checklist

    - [x] I have updated the documentation to reflect the changes (no mention in documentation).
- [x] This PR fixes an issue.

### DIFF
--- a/discord/ext/tasks/__init__.py
+++ b/discord/ext/tasks/__init__.py
@@ -245,6 +245,7 @@ class Loop(Generic[LF]):
                 try:
                     await self.coro(*args, **kwargs)
                     self._last_iteration_failed = False
+                    backoff = ExponentialBackoff()
                 except self._valid_exception:
                     self._last_iteration_failed = True
                     if not self.reconnect:


Rapptz:
The `delay` method already resets the internal exponent if enough time has passed (currently set to 2^11 seconds). Is this PR fixing a bug you actually experienced?

blume0:
> Is this PR fixing a bug you actually experienced?

I have a task which runs every minute to do some API requests, and which has a small portion of its requests encountering network errors, and thus regularly ended up running each 2^n seconds.

That is not necessarily a bug and something to be taken into account depending on how you see it, but I think the wait time before retrying a task that has failed not potentially dependending on occasional failures inbetweet multiple successes would be more expected as as an external user of the tasks framework.

Rapptz:
I see. One alternative is to wrap the body for exceptions you want to handle. The exception mechanism in tasks is actually meant for "unexpected" exceptions that need the task to restart in one way or another. If it's an expected part of the flow it should be handled. The backoff delay as it stands is working as intended to me, with it resetting when it's safe to reset the time.

PR #10383, opened 2026-01-12
Fix exception handling order in gateway heartbeat timeout handler

James-4u:
Previously, Exception was caught before BaseException, which meant that BaseException-only exceptions (like KeyboardInterrupt, SystemExit) would be caught by BaseException, but the handler logic was unclear.

Now BaseException is caught first and we use isinstance() to distinguish between regular exceptions (Exception subclasses) and base-only exceptions. This ensures:
- Regular exceptions get proper exception logging
- Base-only exceptions get debug logging
- Both code paths are reachable and properly handled

This fix ensures all exception types are handled correctly when stopping the gateway connection.

## Checklist

- [x] This PR fixes an issue.

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=102175066

### DIFF
--- a/discord/gateway.py
+++ b/discord/gateway.py
@@ -163,10 +163,11 @@ class KeepAliveHandler(threading.Thread):
 
                 try:
                     f.result()
-                except Exception:
-                    _log.exception('An error occurred while stopping the gateway. Ignoring.')
                 except BaseException as exc:
-                    _log.debug('A BaseException was raised while stopping the gateway', exc_info=exc)
+                    if isinstance(exc, Exception):
+                        _log.exception('An error occurred while stopping the gateway. Ignoring.')
+                    else:
+                        _log.debug('A BaseException was raised while stopping the gateway', exc_info=exc)
                 finally:
                     self.stop()
                 return


James-4u:
hello @Rapptz 
could you please review my pr?

AbstractUmbra:
Exception handling costs less computationally in the "no error" route, meaning if we hit an `Exception` type before `BaseException` we exit this handling earlier. I understand we're handling two types of errors, but this leads onto my next point:-

Your changes also adds an instance check, which adds additional computation on top.

I understand we're talking in nano/miliseconds here, but this is a performance degradation overall.

(I am trying to locate the sources on the exception handling speedups in newer python versions, my apologies for the current lack of evidence).

Sachaa-Thanasius:
This doesn't really come across as a "fix"; I'm pretty sure the behavior is the same either way. At best, this affects readability (imo for the worse), but whether it's for the better is up to Rapptz.

Rapptz:
This is just straight up not how Python works. This PR ruined the readability of the catch clauses for no good reason.

PR #10384, opened 2026-01-14
Change join thread from POST to PUT

Quintenvw:
## Summary

Fix method of [/channels/{channel.id}/thread-members/@me](https://discord.com/developers/docs/resources/channel#join-thread), which notably moves the rate limit from guild to global.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -1345,7 +1345,7 @@ class HTTPClient:
             return self.request(r, json=params.payload, params=query, reason=reason)
 
     def join_thread(self, channel_id: Snowflake) -> Response[None]:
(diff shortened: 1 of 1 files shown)


Rapptz:
I don't know what you mean regarding the rate limit (a channel based rate limit is better than a global one) but the documentation seems to point that this route changed from POST to PUT probably when it became public and it never changed from the pre-release version the library was developed towards.

Quintenvw:
Apologies, that's a typo. It is the other way around, which makes it a lot more forgiving. Thanks for merging.

PR #10385, opened 2026-01-14
Fix documentation for `User.display_name`

IndorilReborn:
## Summary

The `User.display_name` docstring is a copy of `Member.display_name` one, but it does not actually try to fetch a guild-specific display_name 
https://github.com/Rapptz/discord.py/blob/master/discord/user.py#L313
```python
    @property
    def display_name(self) -> str:
        """:class:`str`: Returns the user's display name.

        For regular users this is just their global name or their username,
        but if they have a guild specific nickname then that
        is returned instead.
        """
        if self.global_name:
            return self.global_name
        return self.name
```

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/user.py
+++ b/discord/user.py
@@ -313,9 +313,7 @@ class BaseUser(_UserTag):
     def display_name(self) -> str:
         """:class:`str`: Returns the user's display name.
 
-        For regular users this is just their global name or their username,
-        but if they have a guild specific nickname then that
-        is returned instead.
+        This is the user's global name or username
         """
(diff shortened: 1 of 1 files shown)


Rapptz:
Not quite.

This is the `BaseUser` docstring which applies to all the `BaseUser` subclasses. Some of these are also `Member`-like (e.g. `WidgetMember`) so the docstring is supposed to be correct there. That being said, it seems `WidgetMember`'s implementation is bugged anyhow.

PR #10386, opened 2026-01-14
Add support for community invites

Soheab:
## Summary

Adds support for community invites.

Relevant links:
- https://github.com/discord/discord-api-docs/pull/8077
- https://github.com/discord/discord-api-docs/pull/8090
- https://github.com/discord/discord-api-docs/pull/8115
- https://github.com/discord/discord-api-spec/commit/aaa5b538a6f9e4fa9cb8ed8f254913d6f90ee8e6
- https://github.com/discord/discord-api-spec/commit/b0c342fc3d52813b84ca0680c1f3b50f43e9ddc4
- https://discord.com/developers/docs/resources/channel#create-channel-invite
- https://discord.com/developers/docs/resources/invite#get-target-users

## Checklist

- [ ] This PR fixes an issue.


discord/abc.py
```diff
@@ -1283,6 +1283,8 @@ async def create_invite(
         target_user: Optional[User] = None,
         target_application_id: Optional[int] = None,
         guest: bool = False,
+        roles: Optional[List[Snowflake]] = None,
+        users: Optional[List[Snowflake]] = None,
```
Gobot1234: Should this not be Sequence[Snowflake] so you can pass a list of users etc?

Soheab: Sure, changed.

scutuatua-crypto (approved):

discord/abc.py
```diff
@@ -1325,6 +1327,19 @@ async def create_invite(
...
 
             .. versionadded:: 2.6
+        roles: Optional[Sequence[:class:`~discord.abc.Snowflake`]]
+            A list of roles that should be granted to the users joining via this invite.
+
+            Requires the :attr:`~discord.Permissions.manage_roles` permission and cannot
+            assign roles with higher permissions than the bot.
+
+            .. versionadded:: 2.7
+        users: Optional[Sequence[:class:`~discord.abc.Snowflake`]]
```
dolfies: maybe rename to `target_users` to match api and fetch method terminology?

discord/invite.py
```diff
@@ -64,6 +74,45 @@
...
+    -----------
+    invite: :class:`Invite`
+        The invite this job status is for.
+    status: :class:`InviteTargetUsersJob`
+        The status of the job.
+    total_users: :class:`int`
+        The total number of users in the job.
+    processed_users: :class:`int`
+        The number of users that have been processed so far.
+    created_at: :class:`datetime.datetime`
```
dolfies:
```suggestion
    created_at: Optional[:class:`datetime.datetime`]
```

discord/invite.py
```diff
@@ -437,6 +494,13 @@ def __init__(
         self.scheduled_event_id: Optional[int] = self.scheduled_event.id if self.scheduled_event else None
         self._flags: int = data.get('flags', 0)
 
+        roles = data.get('roles', [])
+        self.roles: List[Union[Role, Object]]
+        if roles and self.guild is not None and not isinstance(self.guild, (PartialInviteGuild, Object)):
+            self.roles = [Role(state=self._state, guild=self.guild, data=role_data) for role_data in roles]
```
dolfies: losing role data for guilds the bot isn't in

discord/invite.py
```diff
@@ -582,3 +653,83 @@ async def delete(self, *, reason: Optional[str] = None) -> Self:
...
+        Forbidden
+            You do not have permissions to fetch target users.
+        NotFound
+            The invite is invalid or expired or the invite does not have target users.
+        HTTPException
+            Fetching the target users failed.
+        """
+
+        string = await self._state.http.get_invite_target_users(self.code)
+        users = string.lstrip('Users\n').split('\n')
```
dolfies: the documentation is actually wrong here; this endpoint returns the csv file 1:1 how it was uploaded, meaning
- the column name is not necessarily Users
- there may be multiple column names and multiple columns in general (discord's parsing ignores extra columns)

so this parsing will fail

discord/invite.py
```diff
@@ -582,3 +653,83 @@ async def delete(self, *, reason: Optional[str] = None) -> Self:
 
         data = await self._state.http.delete_invite(self.code, reason=reason)
         return self.from_incomplete(state=self._state, data=data)
+
+    async def fetch_target_users(self) -> list[int]:
```
dolfies:
```suggestion
    async def target_users(self) -> List[int]:
```
since there's no property for this, pretty sure the convention is dropping the `fetch_` (applies to below as well)

discord/invite.py
```diff
@@ -582,3 +653,83 @@ async def delete(self, *, reason: Optional[str] = None) -> Self:
...
+        -------
+        Forbidden
+            You do not have permissions to edit invites.
+        NotFound
+            The invite is invalid or expired.
+        HTTPException
+            Editing the invite failed.
+        """
+
+        if users is not MISSING:
```
dolfies:
```suggestion
        if users:
```
the edit endpoint will error with less than 1 user. tho maybe we don't wanna validate that?

discord/invite.py
```diff
@@ -582,3 +792,93 @@ async def delete(self, *, reason: Optional[str] = None) -> Self:
 
         data = await self._state.http.delete_invite(self.code, reason=reason)
         return self.from_incomplete(state=self._state, data=data)
+
+    async def target_users(self) -> list[int]:
```
Rapptz: This should probably return a typed `Object(id=..., type=User)` which is more useful than a regular `int`.

discord/invite.py
```diff
@@ -358,6 +522,16 @@ class Invite(Hashable):
         The ID of the scheduled event associated with this invite, if any.
 
         .. versionadded:: 2.0
+    roles: List[Union[:class:`PartialInviteRole`, :class:`Role`, :class:`Object`]]
```
Rapptz: Not really sure how I feel about this one. Lately I've been leaning towards a `resolve()` method inside partial objects to explicitly promote them but maybe that's clunky. I suppose it's just a consequence of the data modeling here.

discord/invite.py
```diff
@@ -501,6 +685,32 @@ def _resolve_channel(
...
+        guild = self.guild
+        res: List[Union[PartialInviteRole, Role, Object]] = []
+
+        for role in data:
+            if isinstance(role, (int, str)):
+                role_id = int(role)
+                if guild is not None and not isinstance(guild, (PartialInviteGuild, Object)):
+                    res.append(guild.get_role(role_id) or Object(role_id))
+                else:
+                    res.append(Object(role_id))
```
Rapptz: Use a typed `Object` here since it's known it's a role ID.

Rapptz (changes requested):

discord/invite.py
Soheab: Should be fine TBH. It's only a `Role` if it's in cache. This should simplify the `_resolve_roles` method too.

`.resolve()` would try to get from cache too right?

Rapptz: Yeah that's the idea.

discord/invite.py
```diff
@@ -501,6 +724,39 @@ def _resolve_channel(
...
+                role_data = role
+
+            if guild:
+                if not isinstance(guild, (PartialInviteGuild, Object)):
+                    cached_role = guild.get_role(role_id)
+                    if cached_role:
+                        role_data = cached_role._to_partial_dict()
+
+                if role_data:
+                    res.append(PartialInviteRole(state=state, data=role_data, guild_id=guild.id))
```
Rapptz: Yeah nah, it's a little weird that we're making a temporary payload here just to create a different object. If you wanna maintain this coercion then it should probably just be a different constructor altogether to avoid the odd performance drop.

Rapptz (changes requested):

scutuatua-crypto (approved):

PR #10387, opened 2026-01-18
Fix FFmpeg errors not sent to after callback

### LINKED ISSUE #10324: FFmpeg exits with failure but no exception is sent to after function
### Summary

No exception is sent to after function even if ffmpeg died with a non-0 exit code.

### Reproduction Steps

Play a source using voice_client.play() and wait for ffmpeg to stop due to an internal error (if any).

### Minimal Reproducible Code

```python
async def play_track(url):
    source = discord.FFmpegPCMAudio(url)
    voice_client.play(source, after=lambda e: handle_playback_end(e)

def handle_playback_end(error):
    if error:
        # this doesn't get executed if ffmpeg exits with errors
        asyncio.run_coroutine_threadsafe(handle_player_crash(), client.loop)

    asyncio.run_coroutine_threadsafe(play_next(), client.loop)
```

### Expected Results

discord.py should send an exception to the error parameter of the after function on all fatal ffmpeg failures (not just exit code 1)

### Actual Results

FFmpeg crashes (stderr full of errors) with exit code 183 and error is None.

### Intents

all

### System Information

Latest stable version (2.6.3).

### Checklist

### Additional Context

_No response_

levinismynameirl:
## Summary

Fixes FFmpeg errors not being propagated to the `after()` callback.

When FFmpeg exits with a non-zero return code during a blocking `read()` call, 
the error is now properly detected and propagated to the AudioPlayer, which 
passes it to the `after()` callback. Previously, the callback would receive 
`error=None` instead of the actual `FFmpegError` exception.

Changes made to `discord/player.py`:
- Added `_current_error` attribute to `FFmpegAudio` to track process failures
- Modified `_check_process_returncode()` to set the error instead of calling the callback
- Added process return code checks before and after `read()` operations to catch errors during blocking I/O
- Added error propagation from source to `AudioPlayer` when `read()` returns empty

Fixes #10324

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/player.py
+++ b/discord/player.py
@@ -187,6 +187,7 @@ class FFmpegAudio(AudioSource):
         self._pipe_writer_thread: Optional[threading.Thread] = None
         self._pipe_reader_thread: Optional[threading.Thread] = None
         self._current_error: Optional[Exception] = None
+        self._stopped: bool = False
 
         if piping_stdin:
             n = f'popen-stdin-writer:pid-{self._process.pid}'
@@ -222,7 +223,7 @@ class FFmpegAudio(AudioSource):
         if ret is None:
             return  # still running
 
-        if getattr(self, '_stopped', False):
+        if self._stopped:
             return  # intentionally stopped
 
         if ret != 0 and self._current_error is None:
@@ -315,6 +316,7 @@ class FFmpegAudio(AudioSource):
                 return
 
     def cleanup(self) -> None:
+        self._stopped = True
         self._kill_process()
         self._process = self._stdout = self._stdin = self._stderr = MISSING
 


vmphase:
- In `_check_process_returncode`, you're using `self._stderr.seek(0)`. If `stderr` is an unbuffered pipe from the subprocess, this will raise an `OSError` as pipes aren't seekable. It's safer to just read the available buffer without seeking.
- The `read()` method is called every 20ms in the audio loop. Calling `_check_process_returncode` twice per frame might be unnecessary overhead, so imo, it'd be better to only trigger this check if a `read()` returns empty data or a short buffer.
- Remove duplicated imports and the linter-suppression line.

kknxzz:
Hi. Yeah I'll do the fixes for these things. thanks for helping me out :)

kknxzz:
@vmphase, is it fine now? The linter suppression was from a previous commit where discord was actually used in a part of the code but linter for some reason didn't see it. Either way, I removed both the suppression and `import discord`.

I also removed the `self._stderr.seek(0)` line, and made it so it only checks when `read()` returns empty or incomplete.

vmphase:
The changes look better now, nice work.

vmphase:
`_check_process_returncode`, `self._stderr.read()` will attempt to read the entire buffer. If ffmpeg outputs a massive log, this could briefly block the thread. It's usually safer to use smth like `.read(4096)` to grab the most relevant part of the error without risking a hang.

vmphase:
`_check_process_returncode`, `self._stderr.read()` will attempt to read the entire buffer. If ffmpeg outputs a massive log, this could briefly block the thread. It's usually safer to use smth like `.read(4096)` to grab the most relevant part of the error without risking a hang.

kknxzz:
Thanks :) 

It's my first time contributing into a larger library (or repo) haha, so I'm still getting accustomed to the various rules and pr checks that exist :) Now just have to wait for a maintainer to accept or provide more feedback if needed I suppose

kknxzz:
oop didn't see your other comment. yep I'll look into that issue as well :D

kknxzz:
Done as well with this.

kknxzz:
> `read(4096)` to grab the most relevant part of the error without risking a hang.

A question, however. In rare cases where the error might be bigger than 4kb, the fix that I put based on your suggestion would just stop reading the error. I know usually the most important part of the error is at the top, but I can still see it being an issue in some cases. Any ideas on how this could be prevented?

vmphase:
4KB is usually enough, but you could increase it up to 8KB, which would be plenty for even the wordiest ffmpeg trace.

kknxzz:
I mean, to be fair the entire point of this issue is to fix ffmpg errors not showing up, so I guess I'll do that.

kknxzz:
uhh I'm pretty sure the PR is ready to be merged, if any maintainer sees this...?

discord/player.py
```diff
@@ -212,25 +213,67 @@ def _spawn_process(self, args: Any, **subprocess_kwargs: Any) -> subprocess.Pope
         else:
             return process
 
+    def _check_process_returncode(self) -> None:
+        """Set _current_error if FFmpeg exited with a non-zero code."""
+        proc = getattr(self, '_process', None)
+        if proc is None or not hasattr(proc, 'poll') or type(proc).__name__ == '_MissingSentinel':
```
pythonmcpi: The rest of the library uses `value is MISSING` instead of `type(value).__name__ == '_MissingSentinel'`. The attribute is also explicitly assigned MISSING prior to calling subprocess.Popen for the case where exceptions occur, so getattr should always find a value. `.poll()` should always exist if it's a subprocess.Popen object.
```diff
-         proc = getattr(self, '_process', None)
-         if proc is None or not hasattr(proc, 'poll') or type(proc).__name__ == '_MissingSentinel':
+         if self._process is MISSING:
```

discord/player.py
```diff
@@ -212,25 +213,67 @@ def _spawn_process(self, args: Any, **subprocess_kwargs: Any) -> subprocess.Pope
...
+
     def _kill_process(self) -> None:
+        # check if FFmpeg process failed
+        self._check_process_returncode()
+
         # this function gets called in __del__ so instance attributes might not even exist
         proc = getattr(self, '_process', MISSING)
-        if proc is MISSING:
+        # Only proceed if proc is a subprocess.Popen instance
+        if proc is MISSING or not hasattr(proc, 'kill'):
```
pythonmcpi: This looks like an unnecessary change. `self._process` will be a subprocess.Popen instance if it's not MISSING.

discord/player.py
```diff
@@ -212,25 +213,67 @@ def _spawn_process(self, args: Any, **subprocess_kwargs: Any) -> subprocess.Pope
...
             return
 
-        _log.debug('Preparing to terminate ffmpeg process %s.', proc.pid)
+        pid = getattr(proc, 'pid', 'unknown')
+        _log.debug('Preparing to terminate ffmpeg process %s.', pid)
```
pythonmcpi, line 253: This also looks unnecessary to guard against.

discord/player.py
```diff
@@ -212,25 +213,67 @@ def _spawn_process(self, args: Any, **subprocess_kwargs: Any) -> subprocess.Pope
...
             proc.kill()
         except Exception:
-            _log.exception('Ignoring error attempting to kill ffmpeg process %s', proc.pid)
+            _log.exception('Ignoring error attempting to kill ffmpeg process %s', pid)
 
-        if proc.poll() is None:
-            _log.info('ffmpeg process %s has not terminated. Waiting to terminate...', proc.pid)
-            proc.communicate()
-            _log.info('ffmpeg process %s should have terminated with a return code of %s.', proc.pid, proc.returncode)
+        if hasattr(proc, 'poll') and proc.poll() is None:
```
pythonmcpi: Same here, the hasattr() check seems redundant.

discord/errors.py
```diff
@@ -48,11 +48,13 @@
...
     'MissingApplicationID',
+    'FFmpegError',
 )
 
 APP_ID_NOT_FOUND = (
-    'Client does not have an application_id set. Either the function was called before on_ready '
-    'was called or application_id was not passed to the Client constructor.'
+    'Client does not have an application_id set. Either the function '
+    'was called before on_ready was called or application_id was not '
+    'passed to the Client constructor.'
```
pythonmcpi: This change should be reverted as errors.py passes CI checks on master.

discord/player.py
vmphase: > This looks like an unnecessary change. `self._process` will be a subprocess.Popen instance if it's not MISSING.

Lmao, `hasattr(proc, 'kill')` is here to prevent the voice thread from throwing an `AttributeError` if the process failed to initialize properly or if a custom wrapper is being used.

discord/player.py
vmphase: > This also looks unnecessary to guard against.

If `proc` exists but the spawn failed or it's a mock object during a test, accessing `proc.pid` directly triggers an `AttributeError`, so his `getattr()` is correctly used here.

discord/player.py
vmphase: > Same here, the hasattr() check seems redundant.

Since we've initialized `self._process = MISSING`, any failure during the setup phase means `proc` won't be a proper `subprocess.Popen` obj. If you call `cleanup()` (which calls `_kill_process`) on an object that never fully initialized, accessing `.poll()` or `.kill()` directly will throw an `AttributeError`.

discord/player.py
pythonmcpi: `subprocess.Popen` will always have .kill available as it's defined on the class. [Under Windows it's an alias to terminate](https://github.com/python/cpython/blob/9eab67d507cf655c877a7450f530c9f3fb8ff1f3/Lib/subprocess.py#L1714) and on all other platforms [it sends the SIGKILL signal](https://github.com/python/cpython/blob/9eab67d507cf655c877a7450f530c9f3fb8ff1f3/Lib/subprocess.py#L2252).

If you're checking for failed initialization, this is the wrong check to be doing.

If the library user is using a custom wrapper or mock object, they are expected to implement/forward all relevant properties and methods on that object.

discord/player.py
pythonmcpi: [`subprocess.Popen` assigns a value to `.pid` in `__init__`](https://github.com/python/cpython/blob/9eab67d507cf655c877a7450f530c9f3fb8ff1f3/Lib/subprocess.py#L873). Exceptions thrown in `__init__` prevent assignment from completing, so our `self._process` will be `MISSING`. If we do have a Popen object, the pid attribute on it must exist.

If the library user is using a custom wrapper or mock object, they are expected to implement/forward all relevant properties and methods on that object.

discord/player.py
pythonmcpi: [`Popen.poll()` is defined directly on the class](https://github.com/python/cpython/blob/9eab67d507cf655c877a7450f530c9f3fb8ff1f3/Lib/subprocess.py#L1246) and thus will never not exist, so the hasattr check is redundant. `Popen._internal_poll()` (called by `Popen.poll()` might raise exceptions if the child process state is unclean, so it might be a good idea to add a try except there instead.

discord/player.py
vmphase: > `subprocess.Popen` will always have .kill available as it's defined on the class. [Under Windows it's an alias to terminate](https://github.com/python/cpython/blob/9eab67d507cf655c877a7450f530c9f3fb8ff1f3/Lib/subprocess.py#L1714) and on all other platforms [it sends the SIGKILL signal](https://github.com/python/cpython/blob/9eab67d507cf655c877a7450f530c9f3fb8ff1f3/Lib/subprocess.py#L2252).
> 
> If you're checking for failed initialization, this is the wrong check to be doing.
> 
> If the library user is using a custom wrapper or mock object, they are expected to implement/forward all relevant properties and methods on that object.

The "it's on the class" argument doesn't hold up when you consider that `cleanup()` is called by `__del__`. During garbage collection or a messy `__init__ `failure, you can't guarantee the instance's integrity or that the interpreter hasn't already started tearing things down.

discord/player.py
vmphase: > If the library user is using a custom wrapper or mock object, they are expected to implement/forward all relevant properties and methods on that object.

Logging is for debugging problems, not creating new ones. Using `getattr(proc, 'pid', 'unknown')` ensures that even if a process is in a zombie state or didn't spawn correctly, the developer gets a useful log message instead of a fresh `AttributeError` from the logger itself.

discord/player.py
vmphase: > [`Popen.poll()` is defined directly on the class](https://github.com/python/cpython/blob/9eab67d507cf655c877a7450f530c9f3fb8ff1f3/Lib/subprocess.py#L1246) and thus will never not exist, so the hasattr check is redundant. `Popen._internal_poll()` (called by `Popen.poll()` might raise exceptions if the child process state is unclean, so it might be a good idea to add a try except there instead.

`cleanup()` and `_check_process_returncode` can be triggered when the object is in an unstable state (like during `__del__` or after a failed `_spawn_process`).

If `self._process` is `MISSING`, or if the interpreter has started clearing class attributes during shutdown, a direct call can and will raise an `AttributeError`.

discord/player.py
pythonmcpi: I do enjoy that you've simply copy pasted my comment into a llm :-)
My bad for not catching on earlier.

As stated before, a failure in `__init__` results in `self._process` being `MISSING`, not an incomplete Popen initialization.

`__del__`'s invocation conditions are volatile, yes, but .kill() is already called within a try except. The previous implementation of `_kill_process` has been working fine, from what I can tell, and I trust prior contributors' judgement on this matter over a glorified markov chain. I welcome corrections from other human reviewers with more expertise on this matter.

vmphase: > I do enjoy that you've simply copy pasted my comment into a llm :-)
> My bad for not catching on earlier.
> 
> As stated before, a failure in `__init__` results in `self._process` being `MISSING`, not an incomplete Popen initialization.
> 
> `__del__`'s invocation conditions are volatile, yes, but .kill() is already called within a try except. The previous implementation of `_kill_process` has been working fine, from what I can tell, and I trust prior contributors' judgement on this matter over a glorified markov chain. I welcome corrections from other human reviewers with more expertise on this matter.

It's hilarious to watch you trying to accuse me of using llm and being incompetent, while you can't even back up your claims or answer my questions.

discord/player.py
mikeshardmind: It's not this library's responsibility to handle graceful shutdown for users.

If class attributes are in a state of being cleared at this point, people aren't properly closing an asyncio-based application

kknxzz:
Woah. Well, I'll look and check these out and try to fix the needed stuff..

discord/errors.py
kknxzz: Done here 👍 Will try to work on the other changes as well...

discord/player.py
kknxzz: > I do enjoy that you've simply copy pasted my comment into a llm :-) My bad for not catching on earlier.
> 
> As stated before, a failure in `__init__` results in `self._process` being `MISSING`, not an incomplete Popen initialization.
> 
> `__del__`'s invocation conditions are volatile, yes, but .kill() is already called within a try except. The previous implementation of `_kill_process` has been working fine, from what I can tell, and I trust prior contributors' judgement on this matter over a glorified markov chain. I welcome corrections from other human reviewers with more expertise on this matter.

How is this relevant to the thread? You are dismissing valid arguments and also my code based on assumptions about them being “AI coded” or on your preference to “trust previous contributors more.” That doesn’t actually address the changes I made to fix the bug. I am here to improve the code, get constructive feedback, and make sure the fix is correct, safe, and maintainable. Comments about whether AI was used or whose judgment to trust have no real use here. 

I will, of course, review your suggested changes and implement anything that actually improves the code.

discord/player.py
kknxzz: Didn't notice that, will fix as well👍

discord/player.py
pythonmcpi: I will clarify that my llm usage comment was directed at a different user whose comment I was responding to. All I stated with regard to your code is that the hasattr looks unnecessary and you may want to consider removing it.

kknxzz: Yeah yeah I know I appreciate the feedback I just think the LLM thing is unnecessary and can lead to unwanted conflicts or escalation. Again, when I get home I'll work on fixing those issues you mentioned 👍

kknxzz: Fixed as well 👍

discord/player.py
kknxzz: Fixed as well, I believe.

discord/player.py
kknxzz: I feel like this one doesn't need really changing... Yes, `subprocess.Popen` always has a `pid` attribute if the process was created successfully, and if a user provides a mock, it is true that they are expected to implement the relevant interface, HOWEVER using `getattr(proc, 'pid', 'unknown')` in the log is just a defensive measure to avoid logging errors even though in practice, `proc.pid` should always exist for real processes. I think the current approach is fine and doesn't introduce any issues.

discord/player.py
kknxzz: Fixed 👍

kknxzz:
@pythonmcpi I believe I fixed what you requested as needed. Is it fine now?

pythonmcpi:
I think it looks good based on another read-through. (I haven't tried running code with the changes - I'll assume you've tested that.)

kknxzz:
> I think it looks good based on another read-through. (I haven't tried running code with the changes - I'll assume you've tested that.)

Of course. I tested the initial code on an actual bot, however for the changes you suggested I just made a simple python script to test the changes (simulating the bot essentially) without having to make the bot again.

kknxzz:
Any other ideas or changes I should make? If not, could a maintainer merge the PR or provide additional feedback?

discord/errors.py
```diff
@@ -74,6 +75,12 @@ class ClientException(DiscordException):
     pass
 
 
+class FFmpegError(ClientException):
```
Rapptz:
```suggestion
class FFmpegProcessError(ClientException):
```

discord/errors.py
```diff
@@ -74,6 +75,12 @@ class ClientException(DiscordException):
...
+    """Exception that's raised when an FFmpeg process fails."""
```
Rapptz: This is missing `.. versionadded:: 2.7`.

It's also missing from `api.rst` and the exception hierarchy.

discord/player.py
```diff
@@ -212,25 +213,72 @@ def _spawn_process(self, args: Any, **subprocess_kwargs: Any) -> subprocess.Pope
...
+    def _check_process_returncode(self) -> None:
+        """Set _current_error if FFmpeg exited with a non-zero code."""
+        if self._process is MISSING:
+            return
+
+        ret = self._process.poll()
+        if ret is None:
+            return  # still running
+
+        if getattr(self, '_stopped', False):
```
Rapptz: Why is this using `getattr`.

discord/player.py
```diff
@@ -212,25 +213,72 @@ def _spawn_process(self, args: Any, **subprocess_kwargs: Any) -> subprocess.Pope
...
+            return  # intentionally stopped
+
+        if ret != 0 and self._current_error is None:
```
Rapptz, line 229: ffmpeg does not return non zero exit codes on error like you expect it to. You need to set `-xerror` for it to do that. However, I'm not sure that's a good idea in general since sometimes there are just minor errors that don't really harm the playback.

Rapptz (changes requested):

kknxzz:
hi thanks for the feedback :)
when i have time i'll pull from main and work on them (hopefully in the next couple of days, possibly even today if i can..)

discord/errors.py
kknxzz: Ohhh didn't know I had to update that. will do

discord/errors.py
kknxzz: Fixed this

discord/errors.py
kknxzz: Fixed this as well in latest commit from this message

discord/player.py
kknxzz: i used `getattr` because calling `self._stopped` threw out an AttributeError if `_kill_process` was called before initialisation (from something like __del__) I fixed it so it's initialised in __init__ instead now, but thanks for the feedback :)

discord/player.py
kknxzz: i don't think this needs changing, the original issue was specifically about no exception being sent when FFmpeg exits with a non-zero code. i tested it with a non-existent file which caused FFmpeg to exit with code 254, and `FFmpegProcessError` correctly catches this and reaches `after()`. and as you mentioned yourself, there's no point in triggering an exception for minor errors that don't affect playback :)

kknxzz:
@Rapptz i believe i resolved and implemented your given feedback, however I'm not sure about the last one, when youre free i would appreciate if you could review my fixes 👍

Rapptz (approved):
LGTM

PR #10388, opened 2026-01-19
Add Discord timestamp converter and transformer

AbstractUmbra:
## Summary

This PR adds both a converter and transformer for use with Discord's timestamp functionality. This is made easier for the end user with the addition of the `@time` utility in the client, and as such I felt the need to support this.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/ext/commands/converter.py
+++ b/discord/ext/commands/converter.py
@@ -87,6 +87,7 @@ __all__ = (
     'clean_content',
     'Greedy',
     'Range',
+    'Timestamp',
     'run_converters',
 )
 
@@ -894,6 +895,28 @@ class GuildStickerConverter(IDConverter[discord.GuildSticker]):
         return result
 
 
+if TYPE_CHECKING:
+    Timestamp = datetime.datetime
+else:
+
+    class Timestamp(Converter[str]):
+        """Converts to a :class:`datetime.datetime`.
+
+        Conversion is attempted based on the :ddocs:`Discord style timestamp <reference#message-formatting>` input format.
+
+        .. versionadded:: 2.7
+
+        .. warning::
+            Due to a Discord limitation, no timezone is provided with the input. The UTC timezone has been supplanted instead.
+        """
+
+        async def convert(self, ctx: Context[BotT], argument: str) -> datetime.datetime:
+            match = discord.utils.TIMESTAMP_PATTERN.match(argument)
+            if not match:
+                raise BadTimestampArgument(argument)
+            return datetime.datetime.fromtimestamp(int(match[1]), tz=datetime.timezone.utc)
+
+
 class ScheduledEventConverter(IDConverter[discord.ScheduledEvent]):
     """Converts to a :class:`~discord.ScheduledEvent`.
 
@@ -1274,13 +1297,6 @@ def _convert_to_bool(argument: str) -> bool:
         raise BadBoolArgument(lowered)
 
 
-def _convert_from_timestamp(argument: str) -> datetime.datetime:
-    match = discord.utils.TIMESTAMP_PATTERN.match(argument)
-    if not match:
-        raise BadDatetimeArgument(argument)
-    return datetime.datetime.fromtimestamp(int(match[1]), tz=datetime.timezone.utc)
-
-
 _GenericAlias = type(List[T])  # type: ignore
 
 
@@ -1317,9 +1333,6 @@ CONVERTER_MAPPING: Dict[type, Any] = {
 async def _actual_conversion(ctx: Context[BotT], converter: Any, argument: str, param: inspect.Parameter):
     if converter is bool:
         return _convert_to_bool(argument)
-
-    if converter is datetime.datetime:
-        return _convert_from_timestamp(argument)
 
     try:
         module = converter.__module__
--- a/discord/ext/commands/errors.py
+++ b/discord/ext/commands/errors.py
@@ -79,7 +79,7 @@ __all__ = (
     'SoundboardSoundNotFound',
     'PartialEmojiConversionFailure',
     'BadBoolArgument',
-    'BadDatetimeArgument',
+    'BadTimestampArgument',
     'MissingRole',
     'BotMissingRole',
     'MissingAnyRole',
@@ -603,8 +603,8 @@ class BadBoolArgument(BadArgument):
         super().__init__(f'{argument} is not a recognised boolean option')
 
 
-class BadDatetimeArgument(BadArgument):
-    """Exception raised when a datetime argument was not convertable.
+class BadTimestampArgument(BadArgument):
+    """Exception raised when a timestamp argument was not convertable.
 
     This inherits from :exc:`BadArgument`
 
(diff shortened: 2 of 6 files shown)


discord/ext/commands/converter.py
```diff
@@ -1273,6 +1274,18 @@ def _convert_to_bool(argument: str) -> bool:
         raise BadBoolArgument(lowered)
 
 
+_TIMESTAMP_PATTERN: re.Pattern[str] = re.compile(r'<t:(\d+):[tTdDfFsSR]>')
```
AbstractUmbra: Not keen on re-creating the same constant for both transformer and converter, but I wasn't sure where else I could put this to import from outside of `utils`.

discord/ext/commands/errors.py
```diff
@@ -601,6 +602,21 @@ def __init__(self, argument: str) -> None:
         self.argument: str = argument
         super().__init__(f'{argument} is not a recognised boolean option')
 
+class BadDatetimeArgument(BadArgument):
```
AbstractUmbra: As this uses the built-in `datetime.datetime` class, I opted for naming all resources `DatetimeX`, even thought this is more for timestamps than datetimes. 
Feels unclean to me but naming this eludes me.

AbstractUmbra (commented):
Some of my own thoughts on it.

vmphase:
A few suggestions for the implementation:
- The current pattern requires a style flag (e.g., :R), but, iirc, discord timestamps are valid without one (defaulting to :f). Probably, change the pattern to `r'<t:(\d+)(?::[tTdDfFsSR])?>'` to support both.
- `fromtimestamp(n).replace(tzinfo=...)` creates a naive local time before converting. It's safer to use `datetime.fromtimestamp(int(match[1]), tz=datetime.timezone.utc)` directly.

I’d also consider using `__slots__ = ('argument',)` in `BadDatetimeArgument` to reduce memory overhead.

AbstractUmbra:
> A few suggestions for the implementation:
> - The current pattern requires a style flag (e.g., :R), but, iirc, discord timestamps are valid without one (defaulting to :f). Probably, change the pattern to `r'<t:(\d+)(?::[tTdDfFsSR])?>'` to support both.
> - `fromtimestamp(n).replace(tzinfo=...)` creates a naive local time before converting. It's safer to use `datetime.fromtimestamp(int(match[1]), tz=datetime.timezone.utc)` directly.
> 
> I’d also consider using `__slots__ = ('argument',)` in `BadDatetimeArgument` to reduce memory overhead.

Good calls on those, thank you! 'twas an oversight on my part.

None of the other error classes implement said slots however, so I'll opt out of adding those for now.

discord/ext/commands/errors.py
```diff
@@ -602,6 +603,24 @@ def __init__(self, argument: str) -> None:
         super().__init__(f'{argument} is not a recognised boolean option')
 
 
+class BadDatetimeArgument(BadArgument):
+    """Exception raised when a boolean argument was not convertable.
```
Sacul0457: forgot to edit? should be related to datetime/timestamp right?

discord/ext/commands/converter.py
```diff
@@ -1273,6 +1274,16 @@ def _convert_to_bool(argument: str) -> bool:
         raise BadBoolArgument(lowered)
 
 
+_TIMESTAMP_PATTERN: re.Pattern[str] = re.compile(r'<t:(\d+)(?::[tTdDfFsSR])?>')
+
+
+def _convert_from_timestamp(argument: str) -> datetime.datetime:
+    match = _TIMESTAMP_PATTERN.match(argument)
```
AbstractUmbra: Additionally I chose `.match()` over `.search()` since we'd expect to only parse the `<t:xxxxxx:x>` input only rather than search an arbitrary string.

discord/ext/commands/converter.py
```diff
@@ -1273,6 +1274,16 @@ def _convert_to_bool(argument: str) -> bool:
         raise BadBoolArgument(lowered)
 
 
+_TIMESTAMP_PATTERN: re.Pattern[str] = re.compile(r'<t:(\d+)(?::[tTdDfFsSR])?>')
```
TrustyJAID: This doesn't account for time before 1970 which discord does allow.

discord/ext/commands/converter.py
Rapptz: It should be in `utils`. You're already doing this work twice.

Rapptz (commented):
I'm going to hold off on this one for now.

We're still not sure how true datetime inputs are going to look like, if ever. For all I know it'll be completely incompatible with the timestamp input which makes it rough to recommend as the canonical `datetime.datetime` transformer/converter.

If this was delegated to a separate dedicated class I'd be fine with it, essentially a `TimestampConverter` and `TimestampTransformer`. It could also just not have the suffix so it reads better (i.e. `commands.Timestamp`). In order to make it "magically" work in terms of type hinting we can do the type checking trick:

```py
class _TimestampConverter(...):
   pass

if TYPE_CHECKING:
   Timestamp = datetime.datetime
else:
   Timestamp = _TimestampConverter
```

AbstractUmbra:
> I'm going to hold off on this one for now.
> 
> We're still not sure how true datetime inputs are going to look like, if ever. For all I know it'll be completely incompatible with the timestamp input which makes it rough to recommend as the canonical `datetime.datetime` transformer/converter.
> 
> If this was delegated to a separate dedicated class I'd be fine with it, essentially a `TimestampConverter` and `TimestampTransformer`. It could also just not have the suffix so it reads better (i.e. `commands.Timestamp`). In order to make it "magically" work in terms of type hinting we can do the type checking trick:
> 
> ```python
> class _TimestampConverter(...):
>    pass
> 
> if TYPE_CHECKING:
>    Timestamp = datetime.datetime
> else:
>    Timestamp = _TimestampConverter
> ```

I assume the same will be true for app_commands?
E.g. you'd prefer
```py
async def datetime_app(itx: discord.Interaction[commands.Bot], value: discord.app_commands.Transform[datetime.datetime, TimestampTransformer]) -> None:
  ...
```
over
```py
async def datetime_app(itx: discord.Interaction[commands.Bot], value: datetime.datetime) -> None:
  ...
```

where we just update the built-in transformers mapping if/when we get an official datetime input?

Applying these changes just now :)

Rapptz:
Yes. But since it's a built-in one I want the ergonomics to be better so the `Transform` is unnecessary. Hence the `TYPE_CHECKING` trick.

AbstractUmbra:
Given the following test code:
```py
@bot.command(name='datetime')
async def datetime_command(ctx: commands.Context[commands.Bot], *, inp: commands.Timestamp) -> None:
    await ctx.send(inp.isoformat())


@bot.tree.command(name='datetime-test')
async def datetime_app(
    itx: discord.Interaction[commands.Bot],
    value: discord.app_commands.Timestamp,
) -> None:
    await itx.response.send_message(value.isoformat())
```
We get the following output:
DiscordDevelopment_mop35Z2Lr1

Which I believe is what you were suggesting. The `if TYPE_CHECKING:` example was appreciated, it hadn't occurred to me before!

Rapptz:
Yup that looks correct to me

PR #10391, opened 2026-01-28
Add support for sending finished views in stateless application webhooks

t-wy:
## Summary

As a deferred Discord application command results in an application webhook that can be used to do the follow-up statelessly, it can be used directly by the bot client to offload a deferred interaction to another stateless machine by providing the webhook url (or equivalently, application id + token) and the interaction data:
```py
# ... handle the received interation data just like what the bot client would do ...

# do the interaction.followup.send statelessly
webhook = discord.Webhook.from_url(
    url = ..., # https://discord(app)?.com/api/webhooks/{application_id}/{token}
    session = await get_session() # an existing aiohttp.ClientSession within the machine process
)
webhook.type = discord.WebhookType.application # it would be nice if Webhook.from_url allows specifying so, but as long as it works
return await webhook.send(content, ephemeral=ephemeral, wait=True, **kwargs) # wait = True is set automatically for application webhook anyways
```

A special case is that this stateless machine may send messages with views.
If persistent views are utilized, it is useful to send a stopped view with interactive components with `custom_id` set, so interactions done by users to the components can still be picked up later by the persistent view in the bot client according to the current Discord.py mechanism.

This PR follows the PR #10089 that allows stateless webhooks to send views. That PR only allows views without dispatchable components to be sent via stateless webhooks.

Unlike normal webhooks, Discord API allows deferred application interactions follow-up `send` to act like a direct `send_message` call in most scenarios since it's introduced without restrictions in normal stateless webhooks as addressed in #10089.

The `is_finished` condition followed exactly as the condition that checks whether to do `self._state.store_view` later, i.e.
```py
if view is not MISSING and not view.is_finished() and view.is_dispatchable():
    message_id = None if msg is None else msg.id
    self._state.store_view(view, message_id)
```
, which infers that a "finished" view is already inherently supported by this coroutine method.

Therefore, the current _WebhookState ValueError check done in discord.py does not need to block finished views from being sent, as a "finished" view used this way would not be `store`d or `wait`ed to trigger callbacks anyways. This unlocks the full potential for application webhooks initialized without any client state to work.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/webhook/async_.py
+++ b/discord/webhook/async_.py
@@ -1859,7 +1859,11 @@ class Webhook(BaseWebhook):
             if not hasattr(view, '__discord_ui_view__'):
                 raise TypeError(f'expected view parameter to be of type View not {view.__class__.__name__}')
 
-            if isinstance(self._state, _WebhookState) and view.is_dispatchable():
+            if (
+                isinstance(self._state, _WebhookState)
+                and (not application_webhook or not view.is_finished())
+                and view.is_dispatchable()
+            ):
                 raise ValueError(
                     'Webhook views with any component other than URL buttons require an associated state with the webhook'
                 )


Rapptz:
You've overcomplicated the boolean condition. I also don't know why you're manually creating a stateless Webhook from one with state already in it. I would have assumed just for testing the PR but the comments and the PR body seem to have some misconceptions on what these webhooks are.

Regardless, finished views are not allowed to be sent anywhere. In the old code a non-dispatching view would return `False` for `is_finished` because the `Future` would never resolve with a value but with that PR it seems the condition got inverted so that'll need to be fixed in a separate commit.

t-wy:
If the traditonal convention mainly recommends sharding as the only way to offload the interactions across multiple clients, then most normal users would not feel the need to have this implemented anyways.

However, discord.py basically provides everything that allows a stateless webhook to be initialized to perfectly handle the application command followup except that part explicitly prevent a **PERFECTLY VALID** view to be sent just because it is a "stateless" **webhook state**. I don't see the point why the library explicitly still prevents so when discord official API would accept it.

> Regardless, finished views are not allowed to be sent anywhere.

(Yet it works)

Also, I know most people would not intentionally stop a view before sending to have it handled in the persistent view as it is undocumented. However, it just works and gives the benefits of effectively prevent a basically useless newly created instance from staying in memory when the object instance does not carry state, as a persistent view instance should not carry so anyways.

I repeat the use case in case someone has not paid enough attention: This application interaction followup webhook URL can basically be used on an **Another Machine** not connecting to the discord gateway directly to respond to an application follow-up. Discord.py library can perfectly assist in this case just like handling a normal command. Due to its webhook nature being supported by discord.py, it would be a smooth addition to support what logic is basically already implemented there.

> I also don't know why you're manually creating a stateless Webhook from one with state already in it.

This proved so, unless you have a better way to serialize that state and send so over the network, which is not even used (nor reusable) anyways. Webhooks are just expected to be stateless.

> the comments and the PR body seem to have some misconceptions on what these webhooks are.

Therefore, no misconceptions. Just an unusual use case of this library to handle this. An offloaded follow-up token / url does not necessarily need to be picked up by discord.py on the other side anyways, it can just be any other library in other language on other machine that accepts so. Therefore, this basically also allows implementing parts of the commands in other programming languages and just use the application follow-up webhook URL to directly respond. (And in reverse, allow the followup from other libraries in other language on other machines to be picked up by the version of discord.py with this implemented). Therefore, it is perfectly normal for a webhook URL to passed around to anywhere within the **15 minutes** from the interaction creation. 

That is my understanding of the application webhook: it gives the ability for an `interaction` to be handled by any application once it is deferred, **statelessly**.

> discord.py docs
>This is typically used when the interaction is acknowledged and a secondary action will be done later.

The "secondary action" is actually free. There is no reason for that to be bounded within discord.py tied with the so-called `state`.

> In the old code a non-dispatching view would return False for is_finished because the Future would never resolve with a value but with that PR it seems the condition got inverted so that'll need to be fixed in a separate commit.

I noticed the new commit 3ab09be added there. That is unrelated to this case anyways.

Most normal discord.py users would not find that useful as most only host all functionalities of the bot on a single machine and would not think of using other methods than the official "sharding" in case of scaling. And I can still personally monkey-patch it just for my own use in case of the general ignorance of the usefulness of such. Never Mind. 

And as simple as in the official doc:
> Webhooks are a _**low-effort way**_ to post messages to channels in Discord. They _**do not require a bot user or authentication**_ to use.

Based on the lines, application webhook is just discriminated by the library based on the notion then.

*(By the way, I foresaw this gets closed anyways, as the official examples already work for most users without the need to read the internals of the library)*

---

(I know actually not many have thought of reconstructing a duck-typed message / interaction from the raw data. That is what basically happened when slash command was introduced, interactions were just duck-typed into message at that time to quickly supports interactions like handling messages before fully migrating to command trees or hybrid commands. Now it is just the reverse: duck-typing text messages and even messages on other platforms to command interactions, including such kind of offloaded interaction data. This is the art of preventing reimplementing the same logic multiple times.)

> I would have assumed just for testing the PR...

(So, the actual difficulty for this to be actually useful is not to test the follow-up, but to map back the data back to some command / function calls to realize what it makes possible, which is a different philosophy to being told to do only what the library allows one to do)

t-wy:
> I also don't know why you're manually creating a stateless Webhook from one with state already in it.

BTW, there are already different use cases where this scenario is mentioned if mine isn't convicing you, no matter a good use case or not, unless the library is trying to "regulate" this behavior:

 - 3951b61: "Allow Interaction webhook URLs to be used in Webhook.from_url" by Rapptz (So you have expected passing interaction webhook URLs to from_url as a valid use case, just maybe not pass by `session` but `client` instead)

From discord support server:
- [Interaction Data construction](https://discord.com/channels/336642139381301249/1182127220613980250/1182127220613980250) (A very similar case: reconstructing the interaction **_from a "xyz server"_** receiving the forwarded interaction data)
- [Rehydrating interaction object after bot restart](https://discord.com/channels/336642139381301249/1109214807174938737/1109214807174938737) (`from_url` gives "a different interaction type" mentioned, and does matter in this scenario though not matter for those without views, and `from_state` using a constructed `payload: WebhookPayload` with a restarted `self._state` also mentioned)
- [Is it possible to save followup webhooks across bot restarts for sending ephemerals?](https://discord.com/channels/336642139381301249/1345875919222607984/1345875919222607984)
- [Followup webhook avatar_url](https://discord.com/channels/336642139381301249/1374794699432394873/1374801957448843314) (Specifying username and avatar_url would not work anyways, but this just shows an attempt to utilize a non-application webhook to tackle the issue)
- [How To Make Persistent Interactions Using Cogs](https://discord.com/channels/336642139381301249/1220807065790845048)

So it shows that I'm not the only one that cares about saving the webhook URL to be reconstructed in another session, which has a different state.

Rapptz:
Yes, there is nothing wrong with constructing a Webhook from a URL that's obvious to anyone who reads the public API. However this PR is not about that. It's about attaching a *finished* view to a Message from a Webhook that has no state. You can attach a state by constructing a Webhook with a client. The client *does not* need to be running for this to work. So you decided to just use a session instead which gives you one without state instead of just passing in a client to give you the state to make it work as the library expects it to.

If your goal is to use the View machinery in here without using a Client at all then the View mechanism is straight up not meant for that use case at all. Views are based off of an internal dispatch mechanism and it expects certain things for that to work. Webhooks can bridge that gap by allowing you to specify what ClientState should handle the dispatching but by not giving it that then the invariants aren't holding.

t-wy:
So if simplified the PR is actually just one thing:
- Make sending whatever `view` Discord supports possible in discord.py without the library caring about further actions done to the components from a Webhook without state.

Then maybe the main problem is that discord.py is designed with a client running in mind, which conflicts with this purpose I suppose.

Rapptz:
Yeah and that's not happening. A View is an abstraction over a list of components and their dispatching is handled by the library. The original PR you linked for is for "Stateless Views" being allowed in "Stateful Webhooks". This is where I think you got a misunderstanding from. A stateless view is just one that only has non-dispatchable items such as URL buttons. That's why the `with_components` query parameter was introduced, explicitly for that use case and that use case only because it allows *non-application owned* webhooks to do this.

t-wy:
I am aware of the difference between all that. I only noticed that how close discord.py is to achieve the whole offload mechanism without rebuilding the request-making mechanism from scratch so I modified the local discord.py installation directly to make that happen. Using the patch in my case is very straight-forward to directly achieve so. Due to the nested OOP it becomes complicated to monkey-patch it outside the library if the main logic is not a separate helper function.

Personally, other than the "finished view" to "persistent view" case, some other hacky work is done locally to bypass the "25 component" limit using a View subclass as long as not all components are sent. This is perfectly normal from my opinion, and avoids completely ditching an otherwise working design into more than 1 view classes containing all components with the use case under the same category, just to bypass a [poka-yoke (behavior-shaping constraint)](https://en.wikipedia.org/wiki/Poka-yoke) introduced by the library "without a toggle to bypass", which does nothing but to frustrate users who are conscious about what they are doing.

Replying from a stateless webhook can avoid using the same token to connect to discord gateway at the same time from multiple machines, which from my knowledge is designed for sharding but currently is not actively blocking non-sharding cases. Having multiple connections may be a solution for the "xyz server" case, despite not an ideal one.

Rapptz:
Maintaining invariants is important so people don't get into weird states that the object does not expect it to be in. Bypassing the 25 component limit isn't doing anyone any favours here so I don't know why you're bringing it up.

In any case this PR has run its course. Thanks for the PR nonetheless and good luck with your local modifications.

PR #10392, opened 2026-01-31
Handle asyncio.CancelledError during WebSocket event polling - v2.6.x

BuddyWinte:
## Summary

This PR fixes an unhandled asyncio.CancelledError that occurs in Client.connect() when ws.poll_event() is cancelled (e.g., during bot shutdown or disconnects) in Python 3.12. By catching this exception, the bot can shut down cleanly without raising tracebacks, while keeping reconnect logic intact.

## Checklist
- [x] This PR fixes an issue.

### DIFF
--- a/discord/client.py
+++ b/discord/client.py
@@ -724,7 +724,11 @@ class Client:
                 self.ws = await asyncio.wait_for(coro, timeout=60.0)
                 ws_params['initial'] = False
                 while True:
-                    await self.ws.poll_event()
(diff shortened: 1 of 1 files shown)


Rapptz:
There should be no reason to handle `asyncio.CancelledError` like this. This exception should be propagated if it's unhandled so the runtime can handle it. As it stands right now you're actually cancelling the cancellation.

PR #10394, opened 2026-02-01
Add guild and user context to autocomplete logs

vmphase:
## Summary

This PR improves visibility when autocomplete fails or times out (e.g., `10062: Unknown interaction`) by adding missing context to the logs, as discussed [here](https://discord.com/channels/336642139381301249/1462060008849277009).

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/tree.py
+++ b/discord/app_commands/tree.py
@@ -1289,7 +1289,12 @@ class CommandTree(Generic[ClientT]):
                 await command._invoke_autocomplete(interaction, focused, namespace)
             except Exception:
                 # Suppress exception since it can't be handled anyway.
-                _log.exception('Ignoring exception in autocomplete for %r', command.qualified_name)
+                _log.exception(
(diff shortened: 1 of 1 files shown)


Rapptz:
Please read the comment that you deleted. It's explicit that there is no exception or call to do this because this exception would be impossible to handle. If an exception's only purpose is to be logged and not handled then you might as well just log the exception (which is what the library does). While it's true that an error handler attached to a command should also be checking the exceptions it has received before actually being called, this implementation just calls `on_error` for that specific command whereas most logging calls are either on the tree or some other higher level handler.

I'd reconsider this PR if you just modified the logging call to have more information that was needed to address the original complaint. I just don't think this type of exception is good because it's not possible to handle outside of logging.

vmphase:
Thanks for the feedback. I've updated the PR, adding `guild_id` and `user_id` to logging for additional context.

PR #10402, opened 2026-02-10
Migrate remove_view to walk_children

vmphase:
## Summary

Migrate `ViewStore.remove_view` to use `walk_children()` instead of `_children` to ensure nested components in a `LayoutView` are correctly identified and cleared from the `ViewStore` cache on timeout or manual stop. 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/ui/view.py
+++ b/discord/ui/view.py
@@ -912,7 +912,7 @@ class ViewStore:
         if message_id is not None and not is_fully_dynamic:
             self._synced_message_views[message_id] = view
 
-    def remove_view(self, view: Union[View, LayoutView]) -> None:
+    def remove_view(self, view: BaseView) -> None:
         if view.__discord_ui_modal__:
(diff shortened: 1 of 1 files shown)


discord/ui/view.py
```diff
@@ -912,14 +912,14 @@ def add_view(self, view: BaseView, message_id: Optional[int] = None) -> None:
         if message_id is not None and not is_fully_dynamic:
             self._synced_message_views[message_id] = view
 
-    def remove_view(self, view: View) -> None:
+    def remove_view(self, view: Union[View, LayoutView]) -> None:
```
DA-344: Modals can also be passed here (therefore the if check after), consider using BaseView instead.
```suggestion
    def remove_view(self, view: BaseView) -> None:
```

vmphase: Nice catch, thanks

PR #10403, opened 2026-02-15
Change `Translator.translate` call to use positional arguments only

ghost:
## Summary

This PR updates the `translator.translate` call inside `Interaction.translate` function to use positional arguments only, improving consistency across the codebase.

Previously, positional arguments were used for every call to `Translator.translate` throughout the codebase, except for the `Interaction.translate` function. This inconsistency could cause unexpected issues for users implementing a custom Translator class that inherits from `app_commands.Translator`.

## Changes

- Updated the `translator.translate` call within `Interaction.translate` function to use positional arguments, consistent with all other call sites in the codebase.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/interactions.py
+++ b/discord/interactions.py
@@ -652,7 +652,7 @@ class Interaction(Generic[ClientT]):
             data = self.command or self.message
 
         context = TranslationContext(location=TranslationContextLocation.other, data=data)
-        return await translator.translate(string, locale=locale, context=context)
+        return await translator.translate(string, locale, context)
 
 
 class InteractionCallbackActivityInstance:


Soheab:
Could you explain the changes you made in your own words?

Rapptz:
No AI contributions allowed.

ghost:
Previously, `Interaction.translate()` used a mix of positional and keyword arguments when calling `Translator.translate()`. This has been updated to consistently use positional arguments for better uniformity.

Rapptz:
This code is not user facing (i.e. it's internal). There's also no benefit or reason to change it because it works fine as-is. It's not like the signature has `/` in it denoting it's positional only.

ghost:
I wrote the content myself..
I only used a translator for the English expression.

ghost:
When implementing a translator class for `app_command`, it is common practice to use inheritance; however, this can lead to confusion for users when they implement their own subclasses.

Rapptz:
This does not affect user code in the slightest. It's also not in the inheritance chain whatsoever. I don't understand why you think they're related or it causes confusion, hence why I believe this is an AI contribution in some way -- it makes no logical sense.

ghost:
In scenarios where the interaction object is unavailable, `translator.translate()` can be called directly; however, I initially attempted to resolve this through overloading. But, this mismatch led to an inconsistency in the number of arguments, which reduces predictability. While overloading might not be strictly necessary, I would like to refine this to improve clarity and developer experience for others in the future.

And thank you for developing discord.py.
Take care :)

ghost:
There is the code that I written:
```python
class Translator(BaseTranslator, app_commands.Translator):
    def __init__(self, bot: Bot, *, default_locale: str = "en-US", locale_dir: str | Path = "locales"):
        super().__init__(bot, default_locale=default_locale, locale_dir=locale_dir)

    async def load(self):
        await super().load()

    @overload
    async def translate(self, key: str, lang: str, **kwargs) -> str | None:
        ...

    @overload
    async def translate(self, string: app_commands.locale_str, locale: discord.Locale, context: app_commands.TranslationContext) -> str | None:
        ...

    async def translate(self, *args, **kwargs) -> str | None:
        if len(args) == 2:
            return await super().translate(*args, **kwargs)
        elif len(args) in (1, 3):
            string = args[0]
            locale = args[1] if len(args) == 3 else kwargs.get("locale")

            extras = dict(string.extras)
            key = extras.pop("key") if "key" in extras else ""

            translation = await super().translate(key, locale.language_code, **extras)
            if translation is None:
                return string.message

            return translation
        else:
            logger.error("[Translation] Invalid number of arguments")
            return
```
ps) Sorry, I'm not good at english..

PR #10406, opened 2026-02-23
Fix Message.edit() ignoring global allowed mentions

Soheab:
## Summary

This PR fixes an issue where the allowed mentions configured on the bot did not apply when editing a message via `(Partial)Message.edit()`. As a result, the edit would still ping users etc even when mentions were explicitly disabled, forcing users to reapply the allowed mentions manually. 

From what I can tell, this behavior only happens in `(Partial)Message.edit()`.

Related help post: https://canary.discord.com/channels/336642139381301249/1475259678702440580

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -1393,12 +1393,7 @@ class PartialMessage(Hashable):
         :class:`Message`
             The newly edited message.
         """
-
-        if content is not MISSING:
-            previous_allowed_mentions = self._state.allowed_mentions
-        else:
-            previous_allowed_mentions = None
-
+        previous_allowed_mentions = self._state.allowed_mentions
         if view is not MISSING:
             self._state.prevent_view_updates_for(self.id)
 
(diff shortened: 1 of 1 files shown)


Rapptz:
I'm not big on this change.

Now the library is always sending this field during a PATCH operation which is not how it's supposed to be done.

ItsSNOWMOBILE:
I honestly agree with @Rapptz , the implementation does fix the mentioned issue. BUT, it comes at the aforementioned cost. I took an hour or two of my day to analyze the logic aaaand, the bug itself is actually easy to fix!

When `content is MISSING` but the caller explicitly passes `allowed_mentions`, the merge with the global config silently doesn't happen because `previous_allowed_mentions=None`. That means the per-call override bypasses the global policy instead of being layered on top of it, which is exactly how `send()` and all interaction edit paths treat it

The proper fix is to gate on `content is not MISSING or allowed_mentions is not MISSING`. This ensures that global config applies when content is being changed (original behaviour preserved), and global config is merged when the caller explicitly provides an `allowed_mentions` override, even without a content change
Therefore, `allowed_mentions` is still NOT sent on pure embed/attachment/view PATCHes, and @Rapptz 's concern is satisfied :p

Soheab:
This is something Discord should fix, as mentioning in an edit call, like users, does not notify the users. For bots, it for some reason does show the yellow background as if it did succeed, but it didn't.

AFAIK

PR #10413, opened 2026-02-26
fix: replace 7 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


Rapptz:
Thanks for the PR, but there are mistakes in it from just doing a blind find replace job and on top of that a lot of these are on purpose because they explicitly re-raise the exception.

PR #10418, opened 2026-03-01
Fix aiohttp websocket timeout deprecation warning

SolitudePy:
## Summary
Fix aiohttp websocket timeout deprecation warning
```
discord/http.py:566: DeprecationWarning: parameter 'timeout' of type 'float' is deprecated, please use 'timeout=ClientWSTimeout(ws_close=...)'\n  return await self.__session.ws_connect(url, **kwargs)\n"
```

## Checklist

- [ ] This PR fixes an issue.


AbstractUmbra:
~~Why are we doing the whole hasattr check? Just use the object as we've already pinned the version update?~~

Account for dependency minver when I thought we had updated, my bad!

PR #10436, opened 2026-04-13
docs: fix: Safari sidebar ul has toolbar cut off

### LINKED ISSUE #10435: [Docs] Safari Bottom Toolbar Overlaps API Reference TOC
### Summary

On mobile Safari the bottom toolbar overlaps with the API Reference’s table of contents in the sidebar

### Reproduction Steps

1. On iOS (I used IPhone 11) open the docs to the API Reference page
2. Open the sidebar
3. Scroll to the bottom

It may work fine. If not, scroll up a bit then start pressing buttons. The bottom tool bar should open. When you scroll back down it should be broken.

### Minimal Reproducible Code

```python
N/A
```

### Expected Results

The page responds to the viewport change

### Actual Results

The bottom of the Table of Contents will not be visible

### Intents

N/A

### System Information

Not passing discord.py version due to relevance.

However this is iPhone 11 running 26.3.1(a)

### Checklist

### Additional Context

Video proof:

https://github.com/user-attachments/assets/784aa959-2644-49fe-9ea3-0020bf12a223

This is most noticeable when you accidentally collapse some of the lists that you didn’t mean to collapse.

Lego-Fan9:
## Summary

This Pull Request fixes an issue on mobile Safari where the API reference sidebar table of contents is cut off.

It does so by using dvh (dynamic viewport height) where possible instead of vh.

This closes https://github.com/Rapptz/discord.py/issues/10435

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/docs/_static/style.css
+++ b/docs/_static/style.css
@@ -388,6 +388,7 @@ aside {
   color: var(--mobile-nav-text);
   z-index: 2;
   max-height: 100vh;
+  max-height: 100dvh;
   overflow-y: auto;
   overscroll-behavior-y: contain;
 }
@@ -1286,6 +1287,7 @@ div.code-block-caption {
     position: sticky;
     top: 1em;
     max-height: calc(100vh - 2em);
+    max-height: calc(100dvh - 2em);
     max-width: 100%;
     overflow-y: auto;
     margin: 1em;
--- a/docs/_static/style.css
+++ b/docs/_static/style.css
@@ -387,7 +387,6 @@ aside {
   background-color: var(--mobile-nav-background);
   color: var(--mobile-nav-text);
   z-index: 2;
-  max-height: 100vh;
   max-height: 100dvh;
   overflow-y: auto;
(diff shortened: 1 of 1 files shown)


docs/_static/style.css
```diff
@@ -1286,6 +1287,7 @@ div.code-block-caption {
     position: sticky;
     top: 1em;
     max-height: calc(100vh - 2em);
+    max-height: calc(100dvh - 2em);
```
Rapptz: There's two `max-height` now. You probably wanted to just modify the existing one.

Lego-Fan9: I had left the old one as a fallback for anything that doesn’t have dvh. It looks like most browsers have had dvh since ~2022 so it’s been there awhile but maybe not every browser has it.
I can remove it though if you think that would be better.

Rapptz: Baseline 2022 is 4 years old at this point it's fine.

PR #10341, opened 2025-10-29
Replace asyncio.iscoroutinefunction to inspect.iscoroutinefunction

PredaaA:
## Summary

Fixes warning `DeprecationWarning`
`DeprecationWarning: 'asyncio.iscoroutinefunction' is deprecated and slated for removal in Python 3.16; use inspect.iscoroutinefunction() instead`

`asyncio.iscoroutinefunction` is deprecated in favor of `inspect.iscoroutinefunction` since [3.13](https://docs.python.org/3.14/whatsnew/3.13.html#pending-removal-in-python-3-16) and makes a DeprecationWarning since 3.14.

## Checklist

- [x] If code changes were made then they have been tested.
    - [ ] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.
- [ ] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

PR #10438, opened 2026-04-14
Use iscoroutinefunction from inspect instead of asyncio on 3.12+

Jackenmen:
## Summary

Fixes #10341 (_supersedes_, to be specific, but GitHub requires "fixes" for auto-close on merge) by adding a private alias for inspect/asyncio version of iscoroutinefunction in the utils module.

Note that there's also a second commit that updates the existing use of `inspect.iscoroutinefunction()`  to use this as well to ensure consistency between different Python versions, but I reverted it in the third commit since I'm not sure you want the existing behaviour to change.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -58,7 +58,16 @@ from ..message import Message
 from ..user import User
 from ..member import Member
 from ..permissions import Permissions
-from ..utils import resolve_annotation, MISSING, is_inside_class, maybe_coroutine, async_all, _shorten, _to_kebab_case
+from ..utils import (
+    resolve_annotation,
+    MISSING,
+    is_inside_class,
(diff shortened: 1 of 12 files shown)


Rapptz:
I think it's probably fine to move the other uses to the util function. As far as I understand the issue, the original `inspect.iscoroutinefunction` failed with mocking (which is already an iffy use case for d.py but I digress). Since moving to the util function is more accepting technically then it shouldn't cause any issue to have it be consistent. Plus it makes refactoring easier when it comes time for the version bump and the compat shim has to go away.

Jackenmen:
Alright, `inspect.iscoroutinefunction` should now be replaced everywhere in the package.

PR #10441, opened 2026-04-16
Add invite roles support (PartialInviteRole + target_role_ids)

TermuxHackz:
## Summary
This adds support for the `roles` field on the discord Invite object, introduced in the Discord API. Roles listed on an invite are automatically assigned to a user when they accept it

### New: `PartialInviteRole` model
added a new lightweight model class representing a role as returned by the invite endpoint. and does not reuse the full `Role` model since the API only returns a subset of fields (`id`, `name`, `position`, `color`, `colors`, `icon`, `unicode_emoji`).

### `Invite.roles`
The `Invite` object now has a `roles: List[PartialInviteRole]` attribute populated from the API response. Defaults to an empty list when not present.

### `create_invite(target_role_ids=...)`
Both `GuildChannel.create_invite` (public API) and `HTTPClient.create_invite` (internal) now accept an optional `target_role_ids: List[int]` parameter, allowing botsto create invites that grant specific roles on join.

## Reference
Discord API docs: https://docs.discord.com/developers/resources/invite

## Example
```python
# create an invite that grants a specific role on join
invite = await channel.create_invite(
    max_age=86400,
    target_role_ids=[123456789012345678],
)
```

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -1282,6 +1282,7 @@ class GuildChannel:
         target_type: Optional[InviteTarget] = None,
         target_user: Optional[User] = None,
         target_application_id: Optional[int] = None,
+        target_role_ids: Optional[List[int]] = None,
         guest: bool = False,
     ) -> Invite:
         """|coro|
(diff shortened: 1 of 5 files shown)


AbstractUmbra:
Duplicate of #10386

PR #10447, opened 2026-05-08
Fix CRLF line endings in test_permissions_all.py

oliver-ni:
## Summary

Normalize CRLF line endings to LF in `tests/test_permissions_all.py`. This is the only file in the repo with CRLF endings, likely from the original commit being authored on Windows.

## Checklist

- [ ] This PR fixes an issue.


mikeshardmind:
Should likely add a .gitattributes file if this is something wanted to enforce going forward. Shouldn't matter one way or the other for python files, though.

PR #10454, opened 2026-05-29
Fix docs, manage_roles is not needeed

Sacul0457:
## Summary
This PR fixes the docs for `Guild.role_member_counts`. It states that `manage_roles` is needed but it actually [isn't](https://docs.discord.com/developers/resources/guild#get-guild-role-member-counts)

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/guild.py
+++ b/discord/guild.py
@@ -3871,14 +3871,10 @@ class Guild(Hashable):
 
         Retrieves a mapping of roles to the number of members that have it.
 
-        You must have :attr:`~Permissions.manage_roles` to do this.
-
         .. versionadded:: 2.7
 
         Raises
         -------
-        Forbidden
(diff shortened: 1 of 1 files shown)


Rapptz:
I'm sure there are other ways to get the 403 though? It seems strange that this endpoint would just be open to all with no permission gating whatsoever.

Sacul0457:
You're probably right. I can test this in a few hours and will let u know

Sacul0457:
If the bot is not in the guild, it just returns `10004: Unknown Guild`. So it doesn't seem like it could ever return a 403.

Rapptz:
I mean yeah, that's not what I was expecting to be tested. I guess ultimately it doesn't matter, it just feels strange for this endpoint to be open.

PR #10457, opened 2026-06-05
Add `channel_type` attr to `Message`

Sacul0457:
## Summary
This PR adds the `channel_type` field that is only present in [`message_create`](https://docs.discord.com/developers/events/gateway-events#message-create)

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/message.py
+++ b/discord/message.py
@@ -2138,6 +2138,11 @@ class Message(PartialMessage, Hashable):
         The message snapshots attached to this message.
 
         .. versionadded:: 2.5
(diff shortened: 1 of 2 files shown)


Rapptz (commented):
I'm unsure if this is worth the slot for now so I'll have to think about it and see use cases. Especially since this slot only appears in one specific (but common) instance. My main hold up is that this information should be readily available already regardless via other means so I'm not sure what this buys you other than wasted memory.

PR #10460, opened 2026-06-07
ci: pin actions versions with hashes

mdevolde:
## Summary

I have pinned the versions of the actions used in the workflows with hashes.

Pinning GitHub Actions to a commit hash is an effective safeguard against supply chain attacks. By referencing a specific and immutable version of an action, you prevent compromised versions from being automatically integrated into your pipelines.

I checked the breaking changes for all actions that had been upgraded, and made the necessary changes.
There were two breaking changes that affected us:
- one for `actions/upload-artifact`, for which I had to update the artifact names to ensure they were unique (since artifacts are immutable since `v4`) (https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md)
- one for `actions/github-script`, in the JS script that was launched, where `github.issues.update` was used, which has been replaced by `github.rest.issues.update` (https://github.com/actions/github-script#v5)

Here is the link to the tags for the actions I've pinned, if you want to check the hashes:
- https://github.com/actions/checkout/releases/tag/v6.0.3
- https://github.com/actions/setup-python/releases/tag/v6.2.0
- https://github.com/actions/upload-artifact/releases/tag/v7.0.1 (this action is commented but I fixed the version in the commentary)
- https://github.com/peter-evans/create-pull-request/releases/tag/v8.1.1
- https://github.com/actions/github-script/releases/tag/v9.0.0
- https://github.com/actions/setup-node/releases/tag/v6.4.0
- https://github.com/jakebailey/pyright-action/releases/tag/v3.0.2

Pinning versions requires a bit of maintenance, since you have to perform manual upgrades regularly, but in your case, with workflows that handle secrets (such as `secrets.GITHUB_TOKEN`), it’s a best practice to avoid troubles.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,12 +16,12 @@ jobs:
 
     name: dists & docs (${{ matrix.python-version }}/${{ matrix.language }})
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
           fetch-depth: 0
 
       - name: Set up CPython ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version: ${{ matrix.python-version }}
 
@@ -35,9 +35,9 @@ jobs:
           python ./setup.py sdist bdist_wheel
 
       # - name: Upload artifacts
-      #   uses: actions/upload-artifact@v2
+      #   uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
       #   with:
-      #     name: distributions
+      #     name: distributions-${{ matrix.python-version }}-${{ matrix.language }}
       #     path: dist/*
 
       - name: Install package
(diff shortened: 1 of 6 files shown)


connorhsm (approved):

mikeshardmind:
(Not to suggest this isn't worth doing, seems like a positive change to me)

The latest versions (not the versions currently in use) of some of these actions are actually [immutable release tags.](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases)

upgrading setup-python to 7.0 and setup-node to 7.0 would achieve the same level of trust (if an attacker can subvert GitHub's immutable releases, they can do much worse)

Would be great if all actions were published using this feature, as it's sufficient to trust a versioned tag and use it in that case.

mdevolde:
@mikeshardmind Thanks for that comment! I checked, and that’s indeed the case for `setup-python` and `setup-node`: their latest releases are marked as immutable by GitHub, which in practice offers the same level of assurance as hash pinning.

However, hash pinning seems better to me for now, just to have a uniform common method across all actions in the workflow. Some third-party actions we use haven’t necessarily adopted this immutable release mechanism, and rather than checking on a case-by-case basis which ones support it, it seems simpler and safer to me to use a single consistent method for all of them.

Finally, beyond the hash vs. immutable release debate, the primary goal of this PR is also to lock down the version of the actions themselves. Currently, references such as `actions/checkout@v3` point to a floating major tag, which can be moved to new minor or patch versions without us realizing it.

PR #10465, opened 2026-06-24
 Raise error immediately when voice channel is full or lacks permissions and  add early validation for empty command groups

KJAyano:
## Summary
This PR fixes an issue where channel.connect() could silently hang for 30 seconds before raising an unhelpful asyncio.TimeoutError when:

- The bot lacks the CONNECT permission.
- The target voice channel is full and the bot lacks MOVE_MEMBERS or ADMINISTRATOR permissions to bypass the user limit.

To address this, pre-connection permission and user limit validation has been added to Connectable.connect() in discord/abc.py, allowing a ClientException to be raised immediately with a clear error message.

Additionally, a fallback has been implemented in discord/voice_state.py to properly abort the voice connection handshake if Discord drops the request during the connection process.

This PR also resolves an existing # TODO in the codebase by validating that application command groups contain at least one child command before being added to the command tree.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/abc.py
+++ b/discord/abc.py
@@ -2124,7 +2124,8 @@ class Connectable(Protocol):
         asyncio.TimeoutError
             Could not connect to the voice channel in time.
         ~discord.ClientException
-            You are already connected to a voice channel.
+            You are already connected to a voice channel, or the voice
+            channel is full and the connection was rejected.
         ~discord.opus.OpusNotLoaded
             The opus library has not been loaded.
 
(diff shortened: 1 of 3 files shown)


Rapptz:
Thanks for the PR, but the permission check is actually currently not done on purpose and realistically, could be implemented better regardless. The TODO comment being solved in this PR is also misplaced since this is now doing two things and the implementation of it is also a bit strange.

PR #10469, opened 2026-06-30
fix(app_commands): detect duplicate param renames correctly

vmphase:
## Summary

The code currently checks `if name in rename_map`, but `rename_map` is keyed by original parameter names, which are unique. Hence, this condition can never be true during iteration, so duplicate renames are accepted and later caught by discord api:

```python

@app_commands.rename(a='x', b='x')  # should raise ValueError("x is already used")
async def cmd(...) -> None: ...

# Raised: CommandSyncFailure (HTTP 400)
```

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -297,27 +297,24 @@ def _populate_descriptions(params: Dict[str, CommandParameter], descriptions: Di
 
 
 def _populate_renames(params: Dict[str, CommandParameter], renames: Dict[str, Union[str, locale_str]]) -> None:
-    rename_map: Dict[str, Union[str, locale_str]] = {}
-
-    # original name to renamed name
+    rename_map: Dict[str, str] = {}
 
     for name in params.keys():
         new_name = renames.pop(name, MISSING)
 
         if new_name is MISSING:
-            rename_map[name] = name
-            continue
-
-        if new_name in rename_map.values():
-            raise ValueError(f'{new_name} is already used')
-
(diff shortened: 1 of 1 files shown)


discord/app_commands/commands.py
```diff
@@ -308,7 +308,7 @@ def _populate_renames(params: Dict[str, CommandParameter], renames: Dict[str, Un
             rename_map[name] = name
             continue
 
-        if name in rename_map:
+        if new_name in rename_map.values():
```
Rapptz: This made me take a step back and realize that the issue is not just the condition but the way the map is being built. The map is used exclusively for this check, so it makes no sense to check the values since that's slower. rename map should probably be keyed so that it's `new_name: old_name` instead of `old_name: new_name` like the `renames` parameter. That way the check can just be `if new_name in rename_map`. I feel there might still be some weird case out there where `name` can be in `rename_map` but I can't think of any right now.

Rapptz (changes requested):

discord/app_commands/commands.py
vmphase: Fair enough. Flipped it so it's keyed by `display_name: original_name`, matches how it's actually used.

vmphase: Furthermore, I might have got what you were suspecting:
```python
@rename(foo='bar')
async def cmd(interaction, foo: str, bar: str): ...
```

`foo` renames to `bar`, and `bar` itself is never renamed.

PR #10475, opened 2026-07-08
fix(http): suppress pyright reportReturnType false positive

vmphase:
## Summary

The PR suppresses a pyright false positive in `ws_connect` caused by the latest aiohttp release.  
This caused the all opened PRs to fail the CI lint.

As [previously noted](https://discord.com/channels/336642139381301249/336642776609456130/1524450672320188477) by Sinbad:
> It's statically known that Literal[True] | Literal[False] = bool, and since overloads only differ on that, the effective domain is bool, while the specific resulting type is only known if the input type is known, otherwise the return type is the union of the domains on the overloads.
> aiohttp could work around this by adding more overloads, but that sucks.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/discord/http.py
+++ b/discord/http.py
@@ -568,7 +568,7 @@ class HTTPClient:
             'compress': compress,
         }
 
-        return await self.__session.ws_connect(url, **kwargs)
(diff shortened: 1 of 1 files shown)


vmphase:
Closed as a duplicate of #10467

PR #10476, opened 2026-07-08
Add ping and coinflip commands to the bot

gabrielalonso244:
## Checklist

- [ ] This PR fixes an issue.


Rapptz:
Thanks but the example bot is just meant to be a simple example, I don't think adding these two commands is showing something that is worth showing to show how the framework operates.

PR #10492, opened 2026-07-23
Use animated webp instead of gif for animated assets

Soheab:
## Summary

This PR changes the default format of animated assets from `gif` to `webp` with `animated=true` and adds two new properties for convenience: `.format`, `.size`.

The main reason for this change is Discord forcing this if an asset was uploaded as an animated webp, but it also works fine if uploaded as other formats.

API Docs: https://docs.discord.com/developers/reference#image-formatting

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/asset.py
+++ b/discord/asset.py
@@ -336,7 +336,6 @@ class Asset(AssetMixin):
             state,
             url=f'{cls.BASE}/app-assets/710982414301790216/store/{banner}',
             key=str(banner),
-            size=1024,
         )
 
     @classmethod
--- a/discord/asset.py
+++ b/discord/asset.py
@@ -26,7 +26,7 @@ from __future__ import annotations
 
 import io
 import os
-from typing import Any, Literal, Optional, TYPE_CHECKING, Tuple, Union
+from typing import Any, Dict, Literal, Optional, TYPE_CHECKING, Tuple, Union
 from .errors import DiscordException
 from . import utils
 from .file import File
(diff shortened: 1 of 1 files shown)


Rapptz:
1. Your diff is disproportionate to the change because you forgot to maintain the magic trailing comma.
2. I think this might be potentially too much of a breaking change. Last time I changed the default format people were complaining about a bunch of things because they'd be feeding this into image libraries like PIL which do not support webp. Since I changed the default from PNG to WEBP last time for images, people did get into the habit of specifying a format for consumption afterwards but I'm not sure they did for animated images so something to keep in mind.

Soheab:
> 1. Your diff is disproportionate to the change because you forgot to maintain the magic trailing comma.
> 2. I think this might be potentially too much of a breaking change. Last time I changed the default format people were complaining about a bunch of things because they'd be feeding this into image libraries like PIL which do not support webp. Since I changed the default from PNG to WEBP last time for images, people did get into the habit of specifying a format for consumption afterwards but I'm not sure they did for animated images so something to keep in mind.

1. Fixed!
2. I don't really do image manipulation, but it seems to be supported now and works fine if saved as a gif (i think people do that anyways?):

Docs: https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#webp
Test:
```py
import os

import requests
from PIL import Image

url = 'https://cdn.discordapp.com/avatars/59152242862723072/a_ddde070fdf76e2ed5c8f2645959ab5c2.webp?size=1024&animated=true' # thanks ste🅱️e

res = requests.get(url, stream=True)
image = Image.open(res.raw)

image.save('test.gif', save_all=True)
os.startfile('test.gif')
```

Regardless of that, assets uploaded as animated webp require the use of this format, `.gif` points to an unknown resource.

PR #10494, opened 2026-08-01
Add file type filtering in slash command attachment options and modal's fileupload component

Soheab:
## Summary

This PR adds supports for the functionality to limit file uploads from a modal or slash commands to certain extensions.

Docs PR: https://github.com/discord/discord-api-docs/pull/8506

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -2966,12 +2966,15 @@ def set_file_types(**parameters: Sequence[Union[str, FileType]]) -> Callable[[T]
     Parameters
     -----------
     \*\*parameters: Sequence[Union[:class:`str`, :class:`.FileType`]]
-        The file types of the parameters.
+        A list of up to 10 file types that are allowed to be uploaded for each attachment parameter.
+        The type of the parameter must be :class:`discord.Attachment`.
 
         You can mix and match strings and :class:`.FileType` enums in the list.
 
         If a string is provided, make sure to prefix it with a period (``.``) (e.g. ``.png``).
-        You may provide any string you want.
+        This is required.
+        You may provide any string you want, but (if you are specifying only extensions) you must
+        include ``.jpg`` for image uploads, and both ``.mp4`` and ``.mov`` for video uploads.
 
         Must be between 0 and 10. Defaults to allowing all file types.
 
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -377,6 +377,9 @@ def _populate_autocomplete(params: Dict[str, CommandParameter], autocomplete: Di
 
 def _populate_file_types(params: Dict[str, CommandParameter], file_types: Dict[str, Sequence[Union[str, FileType]]]) -> None:
     for name, param in params.items():
+        if param.type is not AppCommandOptionType.attachment:
+            raise TypeError('file_types is only supported for attachment option types')
+
         types = file_types.pop(name, MISSING)
         if types is MISSING:
             continue
--- a/discord/app_commands/commands.py
+++ b/discord/app_commands/commands.py
@@ -36,6 +36,7 @@ from typing import (
     List,
     MutableMapping,
     Optional,
+    Sequence,
     Set,
     TYPE_CHECKING,
     Tuple,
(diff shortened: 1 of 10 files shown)


discord/app_commands/commands.py
```diff
@@ -373,6 +375,25 @@ def _populate_autocomplete(params: Dict[str, CommandParameter], autocomplete: Di
         raise TypeError(f'unknown parameter given: {first}')
 
 
+def _populate_file_types(params: Dict[str, CommandParameter], file_types: Dict[str, Sequence[Union[str, FileType]]]) -> None:
+    for name, param in params.items():
```
vmphase, line 380: Why does this iterate through every parameter of the command, not just the ones passed to `@set_file_types(...)`?

discord/app_commands/commands.py
```diff
@@ -2905,3 +2941,59 @@ def decorator(func: T) -> T:
...
+        The type of the parameter must be :class:`discord.Attachment`.
+
+        You can mix and match strings and :class:`.FileType` enums in the list.
+
+        If a string is provided, make sure to prefix it with a period (``.``) (e.g. ``.png``).
+        This is required.
+        You may provide any string you want, but (if you are specifying only extensions) you must
+        include ``.jpg`` for image uploads, and both ``.mp4`` and ``.mov`` for video uploads.
+
+        Must be between 0 and 10. Defaults to allowing all file types.
```
vmphase, line 2983: There is literally no enforcement / validation for this implemented.

pythonmcpi: This was noted in the initial private alpha testing announcement. I'm not sure why it's missing from the api docs pr.

> (we recommend using the provided file groups. if you are specifying only extensions, you must include `.jpg` for image uploads, and both `.mp4` and `.mov` for video uploads, due to mobile schenanigans)

vmphase: I meant Soheab's PR doesn't implement any enforcement on this.

vmphase: API Docs PR does state it though, just missing from the `reference#file-groups` section.
Image

pythonmcpi: From advaith's later comments, it seems to not be a hard requirement in the api, but very strongly recommended. I've asked for clarification on this.

pythonmcpi: from advaith, about mobile behavior and file formats:
> jpeg is always changed to jpg, heic is converted to jpg on ios, and videos are always converted to mov on ios and mp4 on android

vmphase: My claim on this review was about the PR not enforcing the:
> Must be between 0 and 10. Defaults to allowing all file types.

Soheab: discord.py doesn't really enforce any limits, outside of documenting it and a couple of max items limits for components. So I didn't do it here.

Edit: [`eefc800` (this PR)](https://github.com/Rapptz/discord.py/pull/10494/commits/eefc8008f07cd50e5db08369df66835bfb3b6c27) done it

discord/app_commands/commands.py
Soheab: ~~That's how every `_populate_x` function does it, and I just copied that. I guess I could reverse that?~~

~~Also just realised that this will error on the type check if a command has at least one attachment param with file types set and other non-attachment params lol~~

Fixed!

Rapptz (commented):
Not big on the decorator design that only applies to a single parameter type. Unfortunately the alternative design space for this one would stick out as well.

Soheab:
> Not big on the decorator design that only applies to a single parameter type. Unfortunately the alternative design space for this one would stick out as well.

Yeah. An ext.commands like parameter() would've been better, but there is no such thing currently for slash commands.

discord/enums.py
```diff
@@ -1006,6 +1007,29 @@ class NameplatePalette(Enum):
     white = 'white'
 
 
+class FileType(Enum):
+    audio = 'audio'
+    video = 'video'
+    image = 'image'
+
+    @property
+    def file_extensions(self) -> Tuple[str, ...]:
```
vmphase, line 1016: This is not documented in `docs/api.rst`

PR #10500, opened 2026-08-09
Fix: `as_chunks` errors upon non-sliceable iterators

Sacul0457:
## Summary
This PR fixes `as_chunks` where the iterator is an `abc.Sequence`, but its `__getitem__` method does not support slicing. (#10351)
An example is `deque`.

The solution is to use `isslice` for all cases which can slice any iterator. It's only slightly slower and memory usage is around the same.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/discord/utils.py
+++ b/discord/utils.py
@@ -1041,18 +1041,12 @@ def escape_mentions(text: str) -> str:
 
 
 def _chunk(iterator: Iterable[T], max_size: int) -> Iterator[List[T]]:
-    # Specialise iterators that can be sliced as it is much faster
-    if isinstance(iterator, collections.abc.Sequence):
-        for i in range(0, len(iterator), max_size):
-            yield list(iterator[i : i + max_size])
-    else:
-        # Fallback to slower path
-        iterator = iter(iterator)
-        while True:
-            batch = list(islice(iterator, max_size))
-            if not batch:
-                break
-            yield batch
+    iterator = iter(iterator)
+    while True:
+        batch = list(islice(iterator, max_size))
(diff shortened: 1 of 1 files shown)


Vioshim:
While you're touching this function, I'd like to suggest implementing [`itertools.batched`](https://docs.python.org/3/library/itertools.html#itertools.batched) for python >=3.12 as its faster given its C implementation, and defining batched for versions that don't include it. This is a version I use for my projects which could be implemented with `yield from`.

```python
try:
    from itertools import batched
except ImportError:
    from itertools import islice

    def batched(iterable, n):
        if n < 1:
            raise ValueError("n must be at least 1")
        it = iter(iterable)
        while True:
            batch = list(islice(it, n))
            if not batch:
                return
            yield batch
```

Sacul0457:
Yeah, it does seem to be ~60% faster on 10000 elements and chunk size of 30

BUT, `batched` yields a `tuple` while `as_chunks` yields a `list`. So wouldn't that be breaking?

Rapptz:
Strictly speaking no, the documentation only mentions it returns an iterator with no mention of the actual yield type.

Sacul0457:
Ah ok, so should I edit the PR to use `batched` if version >= 3.12?

Vioshim:
Phew, in that case for consistency purposes, I'd recommend making the non modern version implement tuples instead of list in the batches, like `tuple(islice(it, n))`. 

Update: given how useful batched can be for other purposes, I'd also suggest implementing it in `__all__`

PR #10516, opened 2026-09-13
Fix FFmpeg probing edge cases

tp-li-dev:
# Fix FFmpeg probing edge cases

## Summary

This PR fixes several edge cases in `FFmpegOpusAudio.probe` and its fallback implementation:

* caps probed bitrates correctly
* reads fallback bitrates from audio streams rather than unrelated streams
* cleans up fallback subprocesses after a timeout
* propagates task cancellation instead of treating it as a probe failure

## Checklist

* [x] If code changes were made then they have been tested.

  * [ ] I have updated the documentation to reflect the changes.
* [x] This PR fixes an issue.
* [ ] This PR adds something new (e.g. new method or parameters).
* [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed).
* [ ] This PR is **not** a code change (e.g. documentation, README, ...).

## Details

The probe bitrate is now capped correctly rather than allowing values outside the expected range.

The fallback parser now reads the bitrate from an audio stream instead of potentially picking up a bitrate from another stream in FFmpeg's output.

For fallback probe timeouts, the spawned subprocess is killed and `communicate()` is called again to reap it before the original `subprocess.TimeoutExpired` is re-raised.

Probe error handling also catches `Exception` instead of `BaseException`. This allows `asyncio.CancelledError` to propagate normally instead of causing another fallback probe to be started.

Ordinary probe failures continue to use the existing fallback behavior.

## Tests

Regression tests cover:

* bitrate capping
* audio stream selection in the fallback parser
* subprocess cleanup after fallback timeouts
* successful fallback probes not killing the subprocess
* cleanup of a real local child process
* task cancellation during both primary and fallback probes
* existing fallback behavior for ordinary probe failures

The full test suite passes on Python 3.11 and Python 3.13:

```text
304 passed
```

### DIFF
--- a/discord/player.py
+++ b/discord/player.py
@@ -641,9 +641,7 @@ class FFmpegOpusAudio(FFmpegAudio):
         loop = asyncio.get_running_loop()
         try:
             codec, bitrate = await loop.run_in_executor(None, lambda: probefunc(source, executable))
-        except (KeyboardInterrupt, SystemExit):
-            raise
-        except BaseException:
+        except Exception:
             if not fallback:
                 _log.exception("Probe '%s' using '%s' failed", method, executable)
                 return None, None
@@ -651,9 +649,7 @@ class FFmpegOpusAudio(FFmpegAudio):
             _log.exception("Probe '%s' using '%s' failed, trying fallback", method, executable)
             try:
                 codec, bitrate = await loop.run_in_executor(None, lambda: fallback(source, executable))
-            except (KeyboardInterrupt, SystemExit):
-                raise
-            except BaseException:
+            except Exception:
                 _log.exception("Fallback probe using '%s' failed", executable)
             else:
                 _log.debug('Fallback probe found codec=%s, bitrate=%s', codec, bitrate)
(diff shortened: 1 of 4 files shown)


tp-li-dev:
Since this PR includes several fixes, I’m happy to split them into separate PRs if needed. Also, please let me know if I’ve misunderstood any of the existing behavior.

tp-li-dev:
No live FFmpeg media transcoding was tested; the real-process cleanup test uses a local Python subprocess.

Rapptz:
https://github.com/Rapptz/discord.py/blob/master/.github/CONTRIBUTING.md#ai-contributions

It's rude to not disclose AI contributions.

tp-li-dev:
> https://github.com/Rapptz/discord.py/blob/master/.github/CONTRIBUTING.md#ai-contributions
> 
> It's rude to not disclose AI contributions.

I’m sorry for submitting this PR without disclosing that I used AI to help investigate the issues and prepare the fixes and tests. I understand that this goes against the contribution policy, and I won’t submit AI-assisted PRs here again.
AI also helped me with the English wording of this apology.
Since I only understand Japanese, I used AI to help with the translation.

PR #10528, opened 2026-09-21
Add support for components in embeds

Soheab:
## Summary

This PR adds support the new link preview embed using components. There can be multiple of these embeds on one message. 

https://github.com/discord/discord-api-docs/pull/8606

Added:
- `Embed.components`: `List[Container]` (I doubt any other components will be supported at top level)
- `Container._total_children` to use in `Embed.__len__` so that stays accurate
- Types for this and types for the poll result embed

## Checklist

- [ ] This PR fixes an issue.


discord/embeds.py
```diff
@@ -252,7 +267,12 @@ def from_dict(cls, data: Mapping[str, Any]) -> Self:
         except KeyError:
             pass
 
-        for attr in ('thumbnail', 'video', 'provider', 'author', 'fields', 'image', 'footer'):
+        try:
+            self._components = [data.get('component', [])]
+        except KeyError:
+            pass
```
vmphase: `dict.get()` never raises `KeyError`. Moreover, when 'component' is absent from `data` it creates `[[]]`.

vmphase (approved):

